PR #1, opened 2016-08-31
Changes needed for MSVC compilation

borrrden:
Please review for any changes that don't seem appropriate

---
This change is [Reviewable](https://reviewable.io/reviews/couchbase/couchbase-lite-core/1)

### DIFF
--- a/C/c4Document1.cc
+++ b/C/c4Document1.cc
@@ -24,6 +24,7 @@
 #include "SecureRandomize.hh"
 #include "SecureDigest.hh"
 #include "varint.hh"
+#include "fdb_errors.h"
 #include <ctime>
 #include <algorithm>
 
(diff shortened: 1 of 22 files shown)


C/c4Document1.cc
```diff
@@ -338,7 +339,7 @@ static revidBuffer generateDocRevID(C4Slice body, C4Slice parentRevID, bool dele
     }
     return revidBuffer(generation, digest, kDigestType);
 #else
-    error::_throw(FDB_RESULT_CRYPTO_ERROR);
+    error::_throw(error::ForestDB, FDB_RESULT_CRYPTO_ERROR);
```
snej: Hm, this would be better as `error::Unimplemented`. Could you change it to that? (And then remove the `#include "fdb_errors.h"`)

borrrden: I almost changed it to that too, but then I left it as it was :p.  I'll change this and the other location.

snej:
I merged this and then made the above changes (and a few more.)

I hadn't noticed that you added a PR to SQLiteCpp — I would rather not make our own changes in that repo because it's a fork of a 3rd party project. Do you need to create an MSVC project in that repo? I just added its source files directly to the LiteCore Xcode project.

borrrden:
Sure I don't mind leaving that repo untouched, but I think it should be a separate project shouldn't it?  Otherwise you are effectively baking in SQLite support and eliminating the pluggable nature of the storage engines unless there is another mechanism I am unaware of?

PR #2, opened 2016-09-09
Round 2: Get the tests compiling and running on MSVC

borrrden:
Not sure what to replace `srandomdev()` with, if anything.


snej:
I've added a header MSVC_Compat.hh in the Fleece repo where I'm collecting all of the things that need to be defined differently for MSVC. I'd rather put new stuff in there than sprinkle it into the individual source files. Unless it's something larger that's only needed in one place, like the code you've added to FilePath.cc. Could you make those changes?

borrrden:
Ok, done!  I think that's everything

PR #17, opened 2016-11-17
Cmake windows

borrrden:
Changes for CMake on Windows:

Note:  Blocked on a PR on forestdb
Also requires some downstream PRs

Also appears to be missing functions (retain and release on alloc_slice?)


borrrden:
@snej I'd like to merge this before it turns into a god PR that starts including even more things that are not directly related to cmake.  Let me know if you see any issues.

PR #32, opened 2016-12-06
Made QueryParser class not copy/move/assign-able

### LINKED ISSUE #31: Latest QueryParser breaks GCC 4.x compatibility
It has `std::stringstream` member variables, which are not move constructable until libstdc++5 and thus a QueryParser cannot be returned from a function (as in QueryParserTests::parserWith).  There are two choices here:

1. Require libstdc++5 or libc++
2. Remove those member variables

Considering Android, the second seems to be the only viable option...

snej:
Works around limitation in GCC 4.x STL
Fixes #31

### DIFF
--- a/LiteCore/Storage/QueryParser.hh
+++ b/LiteCore/Storage/QueryParser.hh
@@ -43,6 +43,9 @@ namespace litecore {
         static void writeSQLString(std::ostream &out, slice str);
 
     private:
+        QueryParser(const QueryParser&) =delete;
+        QueryParser& operator=(const QueryParser&) =delete;
+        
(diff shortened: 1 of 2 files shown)


borrrden:
Note: 4.8.5 fails, 4.9(.2?)+ required!

To accomplish this for CentOS:

```
#One time
sudo yum install centos-release-scl
sudo yum install devtoolset-3-gcc* #devtoolset-3 == 4.9.2, devtoolset-4 == 5.2.1

#After each login
scl enable devtoolset-3 bash
```

snej:
Could you add those instructions to the README?

PR #68, opened 2017-03-25
Implement N1QL regexp_like function

borrrden:


snej:
I had no idea the C++ library had regexes!

PR #77, opened 2017-03-31
Fixes for Windows (and possibly other non-Apple)

borrrden:
Dependent on https://github.com/couchbaselabs/BLIP-Cpp/pull/1 and https://github.com/couchbaselabs/fleece/pull/16

Most of the commit here is C# stuff, but there are a few things in C++ (notably, Windows complains about the ambiguity between fleece::slice and litecore::blip::slice despite them being the same thing)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -167,7 +167,7 @@ endif()
 
 add_subdirectory(vendor/BLIP-Cpp   EXCLUDE_FROM_ALL)
 
-#add_subdirectory(Replicator)
+add_subdirectory(Replicator)
 
 ### JNI:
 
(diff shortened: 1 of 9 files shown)


snej:
The slice ambiguity is strange. Could you show me what the compile errors look like?

borrrden:
That's literally the message. It says slice is ambiguous between those two types. I will paste the exact message when I get back to my machine later

borrrden:
> D:\Development\couchbase-lite-net-2\vendor\couchbase-lite-core\Replicator\c4Replicator.cc(141): error C2872: 'slice': ambiguous symbol
> D:\Development\couchbase-lite-net-2\vendor\couchbase-lite-core\vendor\fleece\Fleece\slice.hh(147): note: could be 'fleece::slice'
> d:\development\couchbase-lite-net-2\vendor\couchbase-lite-core\vendor\blip-cpp\include\blip_cpp\Message.hh(29): note: or       'litecore::blip::slice'

snej:
I'll try to get around the so-called ambiguity so we don't have to spell out the namespaces here. I think the `using` declaration in Message.hh is the culprit; I thought its effect would only be local to the enclosing `namespace{...}` block, but I guess not.

PR #86, opened 2017-04-06
.NET Replication

borrrden:
Stripped out the bits of c4socket.cc that were needed into c4websocket.cc and set up the CMake project to choose between them based on an option.

This implementation uses the first byte of bytes to send to insert the opcode for the web socket.  
opcode + payload

This allows the same path to signal a close as well as a message still.  A large portion of this is copy and paste though, so it feels evil.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,6 +11,7 @@ set(CMAKE_CXX_STANDARD 11)
 set(CMAKE_C_STANDARD_REQUIRED ON)
 set(CMAKE_C_STANDARD 11)
 OPTION(LITECORE_BUILD_SQLITE "If ON, builds libsqlite as part of the build process")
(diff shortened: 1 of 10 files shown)


snej:
A platform that has a WebSocket library shouldn't need any of the code you brought in from WebSocketImpl. That library will already take care of things like sending close messages. In fact I don't think a WebSocket library will let you send a close frame, only text or binary.

borrrden:
You are right, it won't.  I have special logic that takes care of sending a close message if a close payload is received but I want to make sure that I sent the correct close status and message that LiteCore generated (i.e. int status = 1000, fleece::slice message = fleece::nullslice), and that the correct information gets sent to the delegate when the close comes from the .NET side (i.e. delegate().onWebSocketClose(status)).  I'm having trouble separating out which is just for web socket behavior and which is used for LiteCore logic.

PR #126, opened 2017-05-24
Blob garbage collection with in-memory set

### LINKED ISSUE #50: Blob garbage collection
Need to implement deletion of blob files that have no references in the database (as in CBL 1.x.)

1. Index all the blob digests used in the database
   1. Scan database for documents with attachments (by checking metadata flag)
   2. Scan those docs for references to blobs (dicts with a `"_cbltype":"blob"` property)
   3. Collect the `digest:` property values of all those dicts
2. Delete blob files whose digests don't appear in that index

In 1.x we collected the digests in memory as a Set data structure. This is fast and easy but has limited scalability, of course. I don't believe anyone has run into problems with the memory use, though. Two ways around that:

* Store the digest index in the database. (Interestingly, that turns step 1 into a pretty typical map/reduce process.)
* Store the index as a Bloom filter, which uses vastly less space but will allow a small fraction of unused attachments to remain.

borrrden:
Fixes #50

### DIFF
--- a/LiteCore/Database/Database.cc
+++ b/LiteCore/Database/Database.cc
@@ -234,10 +234,37 @@ namespace c4Internal {
         }
     }
 
+    unordered_set<string> Database::collectBlobs() {
+        RecordEnumerator::Options options;
+        options.contentOptions = kHasAttachmentsOnly;
+        RecordEnumerator e(defaultKeyStore(), 0, INT64_MAX, options);
+        unordered_set<string> usedDigests;
+        for(const Record& rec : e) {
+            auto doc = documentFactory().newDocumentInstance(rec);
+            doc->selectCurrentRevision();
+            do {
+                if(!doc->loadSelectedRevBodyIfAvailable()) {
+                    continue;
+                }
+                
+                const Dict* body = Value::fromTrustedData(doc->selectedRev.body)->asDict();
+                auto keys = _db->documentKeys();
+                Document::findBlobReferences(body, keys, [&usedDigests](const blobKey& key, uint64_t size) {
+                    usedDigests.insert(key.base64String());
+                });
+            } while(doc->selectNextRevision());
+            
+            delete doc;
+        }
+        
+        return usedDigests;
+    }
 
     void Database::compact() {
         mustNotBeInTransaction();
         dataFile()->compact();
+        unordered_set<string> digestsInUse = collectBlobs();
+        blobStore()->deleteAllExcept(digestsInUse);
     }
 
 
--- a/LiteCore/Database/Database.cc
+++ b/LiteCore/Database/Database.cc
@@ -236,7 +236,7 @@ namespace c4Internal {
 
     unordered_set<string> Database::collectBlobs() {
         RecordEnumerator::Options options;
-        options.contentOptions = kHasAttachmentsOnly;
+        options.onlyBlobs = true;
         RecordEnumerator e(defaultKeyStore(), 0, INT64_MAX, options);
         unordered_set<string> usedDigests;
         for(const Record& rec : e) {
@@ -250,7 +250,7 @@ namespace c4Internal {
                 const Dict* body = Value::fromTrustedData(doc->selectedRev.body)->asDict();
                 auto keys = _db->documentKeys();
                 Document::findBlobReferences(body, keys, [&usedDigests](const blobKey& key, uint64_t size) {
-                    usedDigests.insert(key.base64String());
+                    usedDigests.insert(key.filename());
                 });
             } while(doc->selectNextRevision());
             
--- a/LiteCore/Storage/RecordEnumerator.hh
+++ b/LiteCore/Storage/RecordEnumerator.hh
@@ -25,7 +25,8 @@ namespace litecore {
 
     enum ContentOptions {
         kDefaultContent = 0,
-        kMetaOnly = 0x01
+        kMetaOnly = 0x01,
+        kHasAttachmentsOnly = 0x02
     };
 
     /** KeyStore enumerator/iterator that returns a range of Records.
--- a/LiteCore/Storage/RecordEnumerator.hh
+++ b/LiteCore/Storage/RecordEnumerator.hh
@@ -25,8 +25,7 @@ namespace litecore {
 
     enum ContentOptions {
         kDefaultContent = 0,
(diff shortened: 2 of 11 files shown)


LiteCore/Storage/SQLiteDataFile.cc
```diff
@@ -499,11 +541,45 @@ path.path().c_str());
...
+        while(_collectBlobStmt->executeStep()) {
+            slice sl = SQLiteKeyStore::columnAsSlice(_collectBlobStmt->getColumn(0));
+            slice body = RawRevision::getCurrentRevBody(sl);
+            const Dict* root = Value::fromTrustedData(body)->asDict();
+            findBlobs(root, digests, documentKeys());
+        }
+    }
+    
+    void SQLiteDataFile::removeUnusedBlobs(const unordered_set<string> &used) {
+        FilePath attsPath(filePath().dirName() + "Attachments", "");
```
snej: There's a BlobStore class that manages access to blobs. You should go through that instead of groping into the directory here.

LiteCore/Storage/SQLiteDataFile.cc
```diff
@@ -94,6 +100,42 @@ namespace litecore {
             Warn("SQLite error (code %d): %s", errCode, msg);
         }
     }
+    
+    static void findBlobs(const Array* array, unordered_set<string> &digests, SharedKeys* sharedKeys) {
```
snej: There's already an implementation of this in IncomingRev.cc — if it's being done in two places we should move it into a separate file so they can both call it.

LiteCore/Storage/SQLiteDataFile.cc
```diff
@@ -499,11 +541,45 @@ path.path().c_str());
...
+        s << "SELECT body FROM kv_" << storeName << " WHERE (flags & " <<
+            (int)DocumentFlags::kHasAttachments << ") != 0 AND (flags & " <<
+            (int)DocumentFlags::kDeleted << ") == 0";
+        
+        string sql = s.str();
+        compile(_collectBlobStmt, sql.c_str());
+        UsingStatement u(_collectBlobStmt);
+        while(_collectBlobStmt->executeStep()) {
+            slice sl = SQLiteKeyStore::columnAsSlice(_collectBlobStmt->getColumn(0));
+            slice body = RawRevision::getCurrentRevBody(sl);
```
snej: I _think_ we should preserve blobs from all available revisions of a document, not just the current one.

LiteCore/Storage/SQLiteDataFile.cc
```diff
@@ -499,11 +541,45 @@ path.path().c_str());
...
+        });
+    }
 
     void SQLiteDataFile::compact() {
         checkOpen();
         beganCompacting();
         maybeVacuum();
+        auto storeNames = allKeyStoreNames();
+        unordered_set<string> digestsInUse;
+        for_each(storeNames.cbegin(), storeNames.cend(), [&](const string &storeName) {
```
snej: Only the default KeyStore contains documents. The others just have key/value stuff like checkpoints and other db metadata. DataFile itself doesn't know about which ones contain what, so this logic probably belongs up at the Database class.

snej:
Please use a commit message that describes the change directly, without having to look up issue numbers. Thanks!

borrrden:
Sorry about the messages!

I tried a bit to refactor up the the Database level instead of DataFile but my head started swimming because of the different ways it seems that there are to get at things.  Is `RecordEnumerator` the correct class to use to query all the documents?  Is this logic correct?

1. Setup a record enumerator on "this"
2. Create a document for the next entry
3. Load the revision body and run the findBlobReference function on it after decoding it to a `Dict *`
4. Repeat 3 until selectNextRevision returns false

That would give a definitive list of all the in-use digests?

snej:
That logic sounds right. What's missing in RecordEnumerator is a way to select only the records with the hasAttachments flag. You could add that as another flag in the Options (`onlyAttachments`?). Or we could ignore that optimization for now and just file an issue to do it later.

borrrden:
Ok I made those changes.  One note:  The previous revisions are automatically gobbled up by the library unless told otherwise.  Is there a current or planned way to set the flag to not do this from CBL?  If so another test is needed to make sure older revisions' blobs are preserved.

LiteCore/Storage/RecordEnumerator.hh
```diff
@@ -25,7 +25,8 @@ namespace litecore {
 
     enum ContentOptions {
         kDefaultContent = 0,
-        kMetaOnly = 0x01
+        kMetaOnly = 0x01,
+        kHasAttachmentsOnly = 0x02
```
snej: ContentOptions specify which content to retrieve from a document, not which documents to return; so this should be a field of RecordEnumerator::Options instead, like `onlyBlobs`.

LiteCore/Storage/SQLiteDataFile.cc
```diff
@@ -40,6 +40,7 @@ extern "C" {
 #endif
 
 using namespace std;
+using namespace fleece;
```
snej: It looks like there are no real changes in this file anymore, so best to just reset it.

LiteCore/Database/Database.cc
```diff
@@ -234,10 +234,37 @@ namespace c4Internal {
...
+            doc->selectCurrentRevision();
+            do {
+                if(!doc->loadSelectedRevBodyIfAvailable()) {
+                    continue;
+                }
+                
+                const Dict* body = Value::fromTrustedData(doc->selectedRev.body)->asDict();
+                auto keys = _db->documentKeys();
+                Document::findBlobReferences(body, keys, [&usedDigests](const blobKey& key, uint64_t size) {
+                    usedDigests.insert(key.base64String());
```
snej: If you changed this to `key.filename()`, then BlobStore::deleteAllExcept wouldn't have to transform its filenames to look them up in the set, which would be a lot faster.

borrrden:
@snej Updated based on feedback!  Thanks!

snej:
Code looks good, but there should be a unit test. (sorry I didn't call this out before!)

borrrden:
I thought to myself "but there is one" and then I realized it is at the CBL level. I'll make a lite core version too.

PR #150, opened 2017-06-07
Updated JNI for the replicator.

hideki:


snej:
Please feel free to just commit/push changes. If you're not asking someone else to review, there is no need to create a PR.

Issue #154, opened 2017-06-11
Possible SQLite issue with replicating and saving at the same time


borrrden:
Original issue reference: https://github.com/couchbase/couchbase-lite-net/issues/872

Reproduced on .NET Core with both macOS and Windows.  This could be a .NET implementation bug but I'm not sure what could be going wrong in this case.  It seems like the replicator's enumeration is causing the save operations to return SQLITE_BUSY.  Here is a verbose log of the run:

https://gist.github.com/borrrden/532c84e2ab791e23a19b29dbe6e0896a

To summarize the repro process:  

1. Start a continuous push and pull (or just push) replication
2. Add a number of documents (in the case of the above, 1000)
3. Start updating them sequentially at 300 ms intervals (the first update causes the locked issue)

Adding a 200 ms delay between 2 and 3 causes the issue to go away.

borrrden:
Note: the test thread changes from 8 to 15 in the middle due to awaiting async stuff.  I redid the test without async stuff in the test and the thread remained 8 with the same results.

snej:
It would help if you could find out where the error is originating from. Maybe add a line of code to the `SQLite::Exception` constructors, to log and dump a backtrace?

borrrden:
Here is a stack trace from where the error is thrown.

> at SQLite::printStackTrace in d:\development\couchbase-lite-net-2\vendor\couchbase-lite-core\vendor\sqlitecpp\src\statement.cpp: line: 260: address: 0xE98E44A0
        at SQLite::Statement::exec in d:\development\couchbase-lite-net-2\vendor\couchbase-lite-core\vendor\sqlitecpp\src\statement.cpp: line: 333: address: 0xE98E3B50
        at litecore::SQLiteKeyStore::set in d:\development\couchbase-lite-net-2\vendor\couchbase-lite-core\litecore\storage\sqlitekeystore.cc: line: 263: address: 0xE98EB040
        at litecore::VersionedDocument::save in d:\development\couchbase-lite-net-2\vendor\couchbase-lite-core\litecore\revtrees\versioneddocument.cc: line: 88: address: 0xE98A7AD0
        at c4Internal::TreeDocument::save in d:\development\couchbase-lite-net-2\vendor\couchbase-lite-core\litecore\database\treedocument.cc: line: 210: address: 0xE986CE80
[...]

PR #162, opened 2017-06-22
Fix for #154 (Possible SQLite issue with replicating and saving at th…

borrrden:
…e same time)

Plugs a place where the mutex protection to SQLiteDataFile was bypassed and adds an assert to attempt to ensure that it doesn't happen again

### DIFF
--- a/LiteCore/Storage/SQLiteDataFile.cc
+++ b/LiteCore/Storage/SQLiteDataFile.cc
@@ -220,7 +220,7 @@ path.path().c_str());
         // Deliberately make unordered queries unpredictable, to expose any LiteCore code that
         // unintentionally relies on ordering:
         if (arc4random() % 1)
-            _sqlDb->_exec("PRAGMA reverse_unordered_selects=1");
+            _sqlDb->exec("PRAGMA reverse_unordered_selects=1");
 #endif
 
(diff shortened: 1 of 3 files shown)


LiteCore/Storage/SQLiteDataFile.cc
```diff
@@ -382,25 +382,30 @@ path.path().c_str());
...
 
 
-    int SQLiteDataFile::exec(const string &sql) {
+    int SQLiteDataFile::_exec(const string &sql) {
         LogVerbose(SQL, "%s", sql.c_str());
         return _sqlDb->exec(sql);
     }
 
+    int SQLiteDataFile::exec(const string &sql) {
+        assert(inTransaction());
```
snej: Use Assert not assert

LiteCore/Storage/SQLiteDataFile.cc
```diff
@@ -366,7 +366,7 @@ path.path().c_str());
     void SQLiteDataFile::_beginTransaction(Transaction*) {
         checkOpen();
         LogVerbose(SQL, "BEGIN");
```
snej: While you're making changes, could just delete this log message? I just noticed it's redundant, since `_exec` will log. Thx!

PR #192, opened 2017-07-24
Compile LiteCore with sqlcipher

borrrden:


snej:
Is this going to affect the iOS Obj-C build? I don't know if we want to add SQLCipher to that, as it'll increase the binary size by about 1MB. If not, looks good to me.

borrrden:
Without it lite core will simply crash on iOS 9 though.

borrrden:
That being said it doesn't look like I switched the xcodeproject.

snej:
> Without it lite core will simply crash on iOS 9 though.

Oh right, I forgot that iOS 9's SQLite is too old.

snej:
Do you know why Github shows the entire contents of the CMakeLists.txt files as changed? Maybe you changed the line-break character?

borrrden:
I probably did.  Git loves messing with me on that.  If you are curious then I bet I can rebase this.

snej:
It would be good to get this merged. It looks fine to me, so you can push it after rebasing without a review.

Issue #183, opened 2017-07-18
Support SELECT *


pasin:
From the discussion last week, we would like SELECT * to be able to return the content of the document. 

1. The content will be extracted to the root level for the following Query:
   - SELECT * FROM
   - SELECT * with other projections such as SELECT *, count(*) FROM
2. The content could be grouped to each data source alias name for JOINs query with multiple data sources.

snej:
I don't really understand what you're saying in item 1. Can you explain?

I think it makes sense for "*" to map to a single column in the result. That's the way it's going to come back from the underlying query — as a blob containing a Fleece dictionary.

snej:
Implemented by Jim and merged. The special property name `.*`, used in a `WHAT` column list, represents the entire document.

PR #201, opened 2017-07-28
Implement functionality for SELECT *

borrrden:
Implements #183

### DIFF
--- a/LiteCore/Query/QueryParser.cc
+++ b/LiteCore/Query/QueryParser.cc
@@ -283,7 +283,7 @@ namespace litecore {
         // TODO: Add 'WHERE' clause for use with SQLite 3.15+
     }
 
-
+    // Unused?
     void QueryParser::writeResultColumn(const Value *val) {
         switch (val->type()) {
             case kArray:
@@ -320,10 +320,14 @@ namespace litecore {
 
 
     void QueryParser::writeStringLiteralAsProperty(slice str) {
-        require(str.size > 0 && str[0] == '.',
-                "Invalid property name '%.*s'; must start with '.'", SPLAT(str));
+        require(str.size > 0 && (str[0] == '.' || str[0] == '*'),
+                "Invalid property name '%.*s'; must start with '.' or be '*'", SPLAT(str));
         str.moveStart(1);
-        writePropertyGetter("fl_value", str.asString());
+        if(str[0] == '*') {
+            writePropertyGetter("fl_value", string());
+        } else {
+            writePropertyGetter("fl_value", str.asString());
+        }
     }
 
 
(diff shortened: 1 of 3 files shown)


LiteCore/Query/SQLiteFleeceFunctions.cc
```diff
@@ -28,12 +28,18 @@ namespace litecore {
     // Core SQLite functions for accessing values inside Fleece blobs.
 
 
-    // fl_value(fleeceData, propertyPath) -> propertyValue
+    // fl_value(fleeceData, [optional]propertyPath) -> propertyValue
```
borrrden: Oops, that was a leftover from the previous iteration!

LiteCore/Query/QueryParser.cc
```diff
@@ -320,9 +320,12 @@ namespace litecore {
 
 
     void QueryParser::writeStringLiteralAsProperty(slice str) {
-        require(str.size > 0 && str[0] == '.',
-                "Invalid property name '%.*s'; must start with '.'", SPLAT(str));
-        str.moveStart(1);
+        require(str.size > 0 && (str[0] == '.' || str[0] == '*'),
```
borrrden: Yeah I knew that when I wrote it, I was just wondering how often that would happen.  Well, actually I think I can change it to check that the size is also 1.

snej: `str == "*"_sl`

borrrden: Yeah I considered that...will that avoid allocating memory by virtue of the const char * being in the const section of the binary and having the slice just point to it?

borrrden:
@jens Updated!

snej:
I think you missed my comment above? "we should make a fl_root function that just extracts the rev body and returns it as a blob. So the query parser would respond to a * property-path by just generating a call to fl_root(body)."

I'm being picky about this, because this is performance-sensitive code that has a big effect on query speed. Thanks.

borrrden:
Indeed I did miss that...actually that was similar to the first way I approached this so if I roll back this shouldn't take long.

LiteCore/tests/QueryTest.cc
```diff
@@ -100,6 +100,25 @@ TEST_CASE_METHOD(DataFileTestFixture, "Query SELECT WHAT", "[Query]") {
     REQUIRE(num == 101);
 }
 
+TEST_CASE_METHOD(DataFileTestFixture, "Query SELECT All", "[Query]") {
+    addNumberedDocs(store);
+    Retained<Query> query{ store->compileQuery(json5("{WHAT: ['*', ['*', ['.num'], ['.num']]], WHERE: ['>', ['.num'], 10]}")) };
```
snej: The multiple asterisks here confused me at first! That highlights an ambiguity where `*` is the multiplication operator, except in the `WHAT` column list. That will prevent us from using the `*` property in expressions in the future. I think we should rename it.

`.*` would work well, since the prefix unambiguously identifies it as a property. What do you think?

borrrden:
Actually I was worried that it wouldn't work but it did.  It's not actually two stars, but a star, and a star inside of an array.  Regardless it is kind of weird looking.  `.*` would be fine, but if that's the case I'd like to refactor fl_value to defer to fl_root when it receives a star instead of checking if it starts with `.` AND then has a star.  How would that be?

snej:
I'd prefer to keep the checking at compile-time. I know it doesn't take long to compare the string to `"*"`, but just on principle :)

PR #207, opened 2017-08-08
Feature/android icu

borrrden:
For review next week (8/14).  Implements the ICU collator on Android (CC @snej)


borrrden:
Note:  Most of these files are straight ICU headers.  The only interesting changes are in CMakeLists.txt and ndk_icu.c

hideki:
Hi @borrrden,

I don't remember all history of Collator and ICU discussion.

I have one question:
- Do languages, that LiteCore supports, NOT require the dictionary?

With CBL 1.x experience, ICU collator requires dictionary which could be a separated file or a part of binary data in the library.

borrrden:
The dictionaries are present on the OS. The android team says that we are allowed to use system ICU as an exception to the no system library rule and eventually they will provide an official API for it. For now I used a wrapper that they wrote.

hideki:
@borrrden Thanks!

borrrden:
Yes, you may notice that there are almost no source files in this PR. The only one there is for dynamically looking up the symbols in the system lib. But before you merge it i think we might need a few more for the toupper and tolower functions so it's best to rebase and test first.

hideki:
@borrrden can you point the affected test cases? I will run it.

borrrden:
See Unicode string functions in SQLiteFunctionsTest.cc and other tests surrounded by `#if` guards like this:  `#if __APPLE__ || defined(_MSC_VER) || LITECORE_USES_ICU`

hideki:
Hi @borrrden,

Could you please review latest my changes? ndk_icu.c and UnicodeCollator_ICU.cc

Thanks!
Hideki

hideki:
Note to Self:

- [Android API version and ICU version matrix](https://developer.android.com/guide/topics/resources/internationalization.html#prenougat)
- [ICU Collator API](http://icu-project.org/apiref/icu4c/ucol_8h.html#a3abc6779e6452106415918199308fab4)
- [NDK with ICU commit](https://android-review.googlesource.com/#/c/platform/development/+/121180/)
- [How to use ICU from native](https://android-review.googlesource.com/#/c/platform/development/+/153001/)
- [NDK ICU discussion](https://github.com/couchbase/couchbase-lite-core/pull/207)

borrrden:
If you have run the tests and they pass on Android then this looks alright to merge

hideki:
I tested from only Java, not from native test code. I will run the test from CTest and CppTest.

hideki:
Note to self:
- To run only Query test:` ./C4Tests "[Query]"`
  - Blob test causes the segmentation fault.
- `ICU_DATA` environment variable should be set for running C4Test
- CollatedQueryTest fails with `ucol_strcoll(...)`. Need to use `ucol_strcollUTF8(...)`

TODO:
- `ucol_strcollUTF8(...)` is not supported with Android API 16 and 17. Figure out the solution.
- Port CollatedQueryTest of C4Test to Java unit tests

hideki:
Hi @borrrden 

I could solve the problem with using `ucol_strcollIter(...)` which is another method to compare UTF-8 encoding string.

Following is commit. Can you review it? If you are fine with this change, we could merge this PR.
https://github.com/couchbase/couchbase-lite-core/pull/207/commits/17f22be1e5f7e941168337df74722ad6b1cfc78c

PR #210, opened 2017-08-14
Introduce a way to copy prebuilt DBs in from an application

borrrden:
I'm on the fence about the last refactor to put stuff into a C++ file (similar to Upgrader.cc) because there is nothing actually C++-ish about it, but I did anyway to be consistent and to allow a signature that uses `FilePath`.

### DIFF
--- a/C/c4Database.cc
+++ b/C/c4Database.cc
@@ -108,10 +108,12 @@ C4Database* c4db_openAgain(C4Database* db,
 }
 
 bool c4db_copy(C4String sourcePath, C4String destinationPath, const C4DatabaseConfig* config,
-               C4Error *error) {
+               C4Error *error) noexcept {
     FilePath from(slice(sourcePath).asString());
     FilePath to(slice(destinationPath).asString());
-    return CopyPrebuiltDB(from, to, config, error);
+    return tryCatch(error, [=] {
+        return CopyPrebuiltDB(from, to, config);
+    });
 }
 
 
--- a/C/c4Database.cc
+++ b/C/c4Database.cc
@@ -28,6 +28,7 @@
 #include "FilePath.hh"
 #include "Error.hh"
 #include "StringUtil.hh"
+#include "PrebuiltCopier.hh"
 #include <thread>
 
 using namespace fleece;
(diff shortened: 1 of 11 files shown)


LiteCore/Database/PrebuiltCopier.cc
```diff
@@ -0,0 +1,84 @@
...
+            if(db) {
+                Log("Resetting UUIDs...");
+                db->resetUUIDs();
+                c4db_free(db);
+                
+                if(needBackup) {
+                    Log("Backing up destination DB...");
+                    
+                    // TODO: Have a way to restore these temp backups if a crash happens?
+                    string p = to.path();
```
snej: You can use `FilePath::appendingToName` to add `_TEMP` more easily.

LiteCore/Database/PrebuiltCopier.cc
```diff
@@ -0,0 +1,84 @@
...
+                    backupPath = FilePath(p + "_TEMP/");
+                    backupPath.delRecursive();
+                    to.moveTo(backupPath);
+                }
+                
+                Log("Moving source DB to destination DB...");
+                temp.moveTo(to);
+                Log("Finished, asynchronously deleting backup");
+                thread( [=]{
+                    backupPath.delRecursive();
```
snej: What happens if an exception is thrown from an anonymous thread like this? Does the C++ runtime catch it or does it abort the process?

LiteCore/Database/PrebuiltCopier.cc
```diff
@@ -0,0 +1,84 @@
...
+        FilePath backupPath;
+        Log("Copying prebuilt database from %s to %s", from.path().data(), to.path().data());
+        bool needBackup = to.exists();
+
+        try {
+            FilePath temp = FilePath::tempDirectory().mkTempDir();
+            temp.delRecursive();
+            from.copyTo(temp);
+            
+            auto db = c4db_open(c4str(temp.path().data()), config, outError);
```
snej: Use `c_str()`, not `data()`

LiteCore/Database/PrebuiltCopier.cc
```diff
@@ -0,0 +1,84 @@
...
+                    Log("Copier finished async delete of backup db at %s", backupPath.path().c_str());
+                } ).detach();
+                
+                return true;
+            }
+        } catchError(outError)
+        
+        try {
+            Warn("Failed to finish copying database (%s)",
+                 error::_what((error::Domain)outError->domain, outError->code).data());
```
snej: Use `c_str`, not `data`

LiteCore/Database/PrebuiltCopier.cc
```diff
@@ -0,0 +1,84 @@
...
+            Database::deleteDatabaseAtPath(to.path(), config);
```
snej: Yikes! This assumes the database at path `to` is the new one, but that doesn't happen till line 52. And the old database isn't moved aside until line 48. So if the exception/error occurred before that, you're deleting the old database here :(

LiteCore/Database/PrebuiltCopier.hh
```diff
@@ -0,0 +1,16 @@
+//
+//  PrebuiltCopier.hpp
+//  LiteCore
+//
+//  Created by Jim Borden on 2017/08/15.
+//  Copyright © 2017 Couchbase. All rights reserved.
+//
+
+#pragma once
+#include "FilePath.hh"
```
snej: You don't need to include these headers; just declare them with `class FilePath;` etc. inside the `namespace` block below.

LiteCore/Database/PrebuiltCopier.cc
borrrden: Whoa, I didn't know there was a difference!

LiteCore/Database/PrebuiltCopier.cc
borrrden: I'm not sure, but FYI this same thing happens at the end of the upgrade process.

borrrden: Relevant stack answer -> https://stackoverflow.com/a/7273925/1155387 so it looks like it would be safer to use `std::async`

LiteCore/Database/PrebuiltCopier.cc
borrrden: `appendingToName` doesn't work here because it is a directory, and that function will append to the entire path, making `FOLDER/_TEMP` instead of `FOLDER_TEMP`

LiteCore/Database/PrebuiltCopier.cc
snej: `data()` just points to the characters; it doesn't say anything about what comes after the last character. `c_str()` explicitly promises there is a 0 byte after the last character.

In practice, they return the same pointer, because the string object's internal buffer is nul-terminated. I think this has been made explicit in C++14 or C++17. But even so I think it's best to call `c_str()` if you require a nul-terminated string.

LiteCore/Database/PrebuiltCopier.cc
snej: Dang, you're right. IIRC it's not supposed to do that; the method name implies it appends to the filename. The only place it's used is on files not directories, so it would be safe to fix this.

C/c4Database.cc
```diff
@@ -103,6 +107,15 @@ C4Database* c4db_openAgain(C4Database* db,
     return c4db_open({path.data(), path.size()}, c4db_getConfig(db), outError);
 }
 
+bool c4db_copy(C4String sourcePath, C4String destinationPath, const C4DatabaseConfig* config,
+               C4Error *error) noexcept {
+    FilePath from(slice(sourcePath).asString());
```
snej: I'm nervous about leaving these lines outside the `try` block. There's nothing in their current implementation that explicitly throws exceptions, but they're allowed to, and `std::string` might conceivably throw `bad_alloc`.

LiteCore/Database/PrebuiltCopier.cc
```diff
@@ -0,0 +1,78 @@
...
+            temp.moveTo(to);
+        } catch(exception &) {
+            Warn("Failed to finish copying database");
+            to.delRecursive();
+            if(needBackup) {
+                if(backupPath.exists()) {
+                    backupPath.moveTo(to);
+                } else {
+                    WarnError("The backup of the database has vanished. \
+                              This should be reported immediately");
```
snej: This string will have a bunch of spaces in it before "This". It's better to use a separate string literal on each line (the parser will concatenate them with no spaces between.)

LiteCore/Database/PrebuiltCopier.cc
```diff
@@ -0,0 +1,78 @@
...
+            chomp(p, '\\');
+            backupPath = FilePath(p + "_TEMP/");
+            backupPath.delRecursive();
+            to.moveTo(backupPath);
+        }
+        
+        try {
+            Log("Moving source DB to destination DB...");
+            temp.moveTo(to);
+        } catch(exception &) {
```
snej: It's safest to use `catch(...)` if you want to catch _everything_ (and don't care what it is).

Issue #211, opened 2017-08-14
Named index support


borrrden:
This is needed for a new API item (indices that support naming so that they can be overwritten and not leave old indices dangling around pulling down performance).  This would involve altering the signatures for two existing methods:

`c4db_createIndex(C4Database* database, C4Slice name, C4Slice propertyPath, C4IndexType indexType, const C4IndexOptions* indexOptions, C4Error* outError) noexcept`
(+ name parameter)

`c4db_deleteIndex(C4Database* database, C4Slice name, C4Error* outError)`
(- propertyPath parameter, - indexType parameter, + name parameter)

Behind the scenes this doesn't involve much for regular indices.  Basically, the name generated for the property path needs to be swapped for the index name and things will work as before.  The challenging part comes from FTS which doesn't use an index, but a virtual table.  That means that instead of automatically being used based on property index, as a normal index would, it needs to be explicitly joined during the query creation.  That means there is a need for some form of mapping between index name and FTS expression on the index.  The current top proposal is to write both the alias and the expression being mapped into the table name of the virtual table and query the sqlite metadata table to find the correct table name for joining (and also during the deletion).  It makes sense to use a simple encoding scheme such as:

`<expression>::<alias>`

Then at query time `LIKE <expression>::%` could be used, and at deletion time `LIKE %::<alias>` could be used.  The order here is important because if I recall correctly, `LIKE` lookups are faster when the wildcard is at the end vs the beginning and deletion will probably not happen as often as queries.  

At creation time, a check for an existing alias will need to be performed and the previous virtual table dropped.

borrrden:
@snej How is the above proposal?

snej:
Sounds good. the query to look up an FTS index by name would be like
```sql
SELECT name FROM sqlite_master WHERE type='table' AND name LIKE '<tablename>::<indexname>::%'
```
where `<tablename>` is the name of the SQLite table for the KeyStore (always `kv_default` in our usage) and `<expression>` is the JSON expression, usually a property, being FTS-indexed.

We'll need to be careful with separators and quoting; for instance, with the above syntax we'd have to quote occurrences of `::` in the index name.

>if I recall correctly, LIKE lookups are faster when the wildcard is at the end vs the beginning and deletion will probably not happen as often as queries.

Performance shouldn't be a concern here, because creating queries doesn't happen that often.

snej:
[...]

PR #217, opened 2017-08-18
Refactor index creation and deletion to use named aliases

borrrden:
Implements #211

### DIFF
--- a/LiteCore/Query/QueryParser.cc
+++ b/LiteCore/Query/QueryParser.cc
@@ -288,7 +288,7 @@ namespace litecore {
 
     void QueryParser::writeCreateIndex(const string &name, const Array *expressions) {
         reset();
-        _sql << "CREATE INDEX IF NOT EXISTS \"" << name << "\" ON " << _tableName << " ";
+        _sql << "CREATE INDEX \"" << name << "\" ON " << _tableName << " ";
         Array::iterator iter(expressions);
         writeColumnList(iter);
         // TODO: Add 'WHERE' clause for use with SQLite 3.15+
--- a/LiteCore/Query/QueryParser.cc
+++ b/LiteCore/Query/QueryParser.cc
@@ -286,9 +286,9 @@ namespace litecore {
     }
 
 
-    void QueryParser::writeCreateIndex(const Array *expressions) {
+    void QueryParser::writeCreateIndex(const string &name, const Array *expressions) {
         reset();
-        _sql << "CREATE INDEX IF NOT EXISTS \"" << indexName(expressions) << "\" ON " << _tableName << " ";
+        _sql << "CREATE INDEX IF NOT EXISTS \"" << name << "\" ON " << _tableName << " ";
         Array::iterator iter(expressions);
         writeColumnList(iter);
         // TODO: Add 'WHERE' clause for use with SQLite 3.15+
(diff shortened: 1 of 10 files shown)


borrrden:
Actually hold off on this, I realized that this won't properly drop aliases from the mapping table

borrrden:
Ok, it is ready for review now.

LiteCore/Query/QueryParser.cc
```diff
@@ -286,9 +286,9 @@ namespace litecore {
     }
 
 
-    void QueryParser::writeCreateIndex(const Array *expressions) {
+    void QueryParser::writeCreateIndex(const string &name, const Array *expressions) {
         reset();
-        _sql << "CREATE INDEX IF NOT EXISTS \"" << indexName(expressions) << "\" ON " << _tableName << " ";
+        _sql << "CREATE INDEX IF NOT EXISTS \"" << name << "\" ON " << _tableName << " ";
```
snej: Should take out `IF NOT EXISTS` because if there's already an index with that name, it's no longer a no-op but an error (since the index expression might be different.)

LiteCore/Storage/SQLiteKeyStore.cc
```diff
@@ -390,17 +409,18 @@ namespace litecore {
                 }
                 sql << ")";
                 db().exec(sql.str(), LogLevel::Info);
-
+                db().exec("INSERT OR REPLACE INTO kv_fts_map (alias, expression) VALUES (\"" + alias +
```
snej: Take out `OR REPLACE` — by this point, there shouldn't already be an entry with this alias (since we were supposed to have detected that up above.)

LiteCore/Storage/SQLiteKeyStore.cc
```diff
@@ -353,7 +353,8 @@ namespace litecore {
     }
 
 
-    void SQLiteKeyStore::createIndex(slice expression,
+    void SQLiteKeyStore::createIndex(slice indexName,
```
snej: This is a good point to sanity-check the indexName. At a minimum, it shouldn't be empty and shouldn't contain a `"`.

LiteCore/tests/QueryTest.cc
```diff
@@ -46,6 +46,32 @@ static void addNumberedDocs(KeyStore *store) {
     t.commit();
 }
 
+TEST_CASE_METHOD(DataFileTestFixture, "Create/Delete Index", "[Query]") {
+    KeyStore::IndexOptions options { "en", true };
+    store->createIndex("num"_sl, "[[\".num\"]]"_sl, KeyStore::kFullTextIndex, &options);
+    bool gotError = false;
+    try {
```
snej: Please use the function `ExpectException` (from LiteCoreTest.hh) — it's simpler, and it sets some state so that the exception can be ignored by the debugger (the Xcode project does this.)

LiteCore/Storage/SQLiteKeyStore.cc
```diff
@@ -409,26 +429,38 @@ namespace litecore {
...
 
-
-    void SQLiteKeyStore::deleteIndex(slice expression, IndexType type) {
-        alloc_slice expressionFleece;
-        const Array *params;
-        tie(expressionFleece, params) = parseIndexExpr(expression, type);
-        string indexName = SQLIndexName(params, type, true);
-
-        Transaction t(db());
+    void SQLiteKeyStore::_deleteIndex(slice name, IndexType type) {
```
snej: This method should sanity-check the `name` parameter too (probably use a common `validateIndexName` function.)

LiteCore/Query/QueryParser.cc
borrrden: In that case I think I should mimic the full text behavior and drop the first index in favor of the new one.  Is there a way to tell what the existing index is on without causing headaches?

borrrden:
@snej Updated

PR #226, opened 2017-08-23
Fix c4db_getIndexes’s result is not being retained

pasin:
Called sliceResult() method to create a retained C4SliceResult. Without doing this, I couldn't get the correct result when converting it to array.


borrrden:
Oops that totally makes sense!

Issue #227, opened 2017-08-23
Database table locked when creating a new index with an existing name


pasin:
The error happened when deleting the existing index.

Error:
```
2017-08-22 23:22:08.515353-0700 xctest[64534:42293846] CouchbaseLite  WARNING: SQLite error (code 6): statement aborts at 14: [DROP INDEX IF EXISTS myindex] database table is locked
```

Failing Test: 
https://github.com/couchbase/couchbase-lite-ios/blob/feature/2.0-new-index/Objective-C/Tests/DatabaseTest.m#L1371-L1372

Trace:
```
* thread #1, queue = 'com.apple.main-thread', stop reason = breakpoint 1.1
    frame #0: 0x0000000112750b86 libc++abi.dylib`__cxa_throw
    frame #1: 0x0000000125fa892a CouchbaseLite`SQLite::Database::check(this=0x00007f92b8e21590, aRet=6) const at Database.h:385
    frame #2: 0x0000000125fa8996 CouchbaseLite`SQLite::Database::exec(this=0x00007f92b8e21590, apQueries="DROP INDEX IF EXISTS myindex") at Database.cpp:149
    frame #3: 0x0000000125e940a2 CouchbaseLite`SQLite::Database::exec(this=0x00007f92b8e21590, aQueries="DROP INDEX IF EXISTS myindex") at Database.h:182
    frame #4: 0x0000000125e91353 CouchbaseLite`litecore::SQLiteDataFile::_exec(this=0x00007f92b8e16610, sql="DROP INDEX IF EXISTS myindex", logLevel=Info) at SQLiteDataFile.cc:412
    frame #5: 0x0000000125e93dae CouchbaseLite`litecore::SQLiteDataFile::exec(this=0x00007f92b8e16610, sql="DROP INDEX IF EXISTS myindex", logLevel=Info) at SQLiteDataFile.cc:417
  * frame #6: 0x0000000125f59c39 CouchbaseLite`litecore::SQLiteKeyStore::_deleteIndex(this=0x00007f92b8c30fa0, name=slice @ 0x00007fff542d5d28) at SQLiteKeyStore.cc:458
[...]

PR #231, opened 2017-08-23
Fix database table locked when replacing an existing index

pasin:
Reset getExistingSQL statement after using it.

#227


pasin:
I don't know if this is a proper fix or not, but that fixed the issue.

borrrden:
This is what I usually do.  I'm also not sure if it is correct but I don't know another way.

PR #237, opened 2017-08-25
Fixes #236 (Select results not getting populated in full text queries)

### LINKED ISSUE #236: Select results do not get populated during full text queries
https://forums.couchbase.com/t/fts-resulting-documents-are-empty-all-properties-are-null/13917/6

The results have the correct number of rows, but every column is null.

borrrden:
Fixes #236 

Normally I would just commit this but I wasn't sure if doing this step unconditionally would cause any problems.  The result columns are only populated currently if the result is ruled to be non-FTS.  That means that every select result in couchbase lite is null instead of what it should be.


snej:
I'm surprised this wasn't caught by unit tests. Could you add a unit test to QueryTest.cc to check for this, please?

borrrden:
Now the existing full text unit test will check for this

snej:
Thanks!

PR #277, opened 2017-10-22
build_unix.sh corrections to fix building on Linux

mz2:
- Adds `-p` to `mkdir` command (script will fail on subsequent script calls after the first one if folder `unix` already exists without this).
- Adds workaround to inability of clang (3.8.0) on Ubuntu 16.04 like Linux to find `__cxxabi_config.h` even when `libc++-dev` & `libc++abi-dev` are installed (same issue as noted here: http://lists.alioth.debian.org/pipermail/pkg-llvm-team/2015-September/005208.html with the same workaround working).
- Tweaks CMakeLists.txt and some supporting header locations to clearer separate Android from other non-Apple UNIX systems, in the process fixing a linking / compilation error with the REST target on Linux.

The addition of `-I/usr/include/libcxxabi` may not be the most elegant workaround but at least it doesn't cause failures on flavours of unix where that folder does not exist.

**NOTE:** Requires https://github.com/couchbaselabs/fleece/pull/24 in order to fully fix build on Linux.


mz2:
This PR now, with the latest additions I made to it just now as well as the fleece changes in PR https://github.com/couchbaselabs/fleece/pull/24, fixes building on Linux. Or at least it Works On My Machine ™️.

snej:
Cool! I can't easily review this myself since I don't have a Linux machine/VM available, but I think @borrrden does.

borrrden:
The beauty is that docker makes that kind of thing wicked easy.  I tested the OS they mentioned (it had a Dockerfile available) and the errors happened, and the fixes fix them.  I will make some notes.

StephenBoyd:
@mz2 's fix worked for me on Ubuntu 17.04 with clang 3.9.1

CMakeLists.txt
```diff
@@ -109,8 +109,10 @@ include_directories("vendor/fleece/Fleece"
 if(WIN32)
   set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} /ignore:4099")
   include_directories("MSVC")
-elseif(UNIX AND NOT APPLE)
+elseif(ANDROID)
   include_directories("LiteCore/Android")
```
borrrden: This directory is now missing arc4random.h, so Android compilation fails (LiteCore/Unix can be safely included on Android, but not vice versa)

CMakeLists.txt
```diff
@@ -203,13 +205,15 @@ if(MSVC)
...
     MSVC/strptime.cc)
+elseif(ANDROID)
+    set(ALL_SRC_FILES ${ALL_SRC_FILES} 
+    LiteCore/Unix/strlcat.c
+    LiteCore/Unix/arc4random.cc
+    LiteCore/Android/unicode/ndk_icu.c)
 elseif(UNIX AND NOT APPLE)
-  set(ALL_SRC_FILES ${ALL_SRC_FILES} 
-    LiteCore/Android/strlcat.c
-    LiteCore/Android/arc4random.cc)
```
borrrden: Please make this change (LiteCore/Android/arc4random.cc -> LiteCore/Unix/arc4random.cc) in LiteCore/tests/CMakeLists.txt as well or CppTests will fail to compile

CMakeLists.txt
mz2: Oopsies, of course.

build_cmake/scripts/build_unix.sh
```diff
@@ -3,10 +3,10 @@
 SCRIPT_DIR=`dirname $0`
 pushd $SCRIPT_DIR/..
 
-mkdir unix
+mkdir -p unix
 pushd unix
 core_count=`getconf _NPROCESSORS_ONLN`
-CC=clang CXX=clang++ cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo -DLITECORE_BUILD_SQLITE=1 ../..
+CC=clang CXX=clang++ CXXFLAGS="-I/usr/include/libcxxabi" cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo -DLITECORE_BUILD_SQLITE=1 ../..
```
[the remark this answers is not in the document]
mz2: Is that going to behave identically in the case that /usr/include/libcxxabi does not exist? This will not fail in that scenario.

CMakeLists.txt
mz2: Done.

build_cmake/scripts/build_unix.sh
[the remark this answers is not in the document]
mz2: (Done.)

CMakeLists.txt
mz2: Done.

mz2:
The latest commit here should hopefully address your review comments.

Issue #286, opened 2017-11-03
get_repo_version.sh fails when building CBL with Carthage


pasin:
This is from https://github.com/couchbase/couchbase-lite-ios/issues/1947. The script fails and exits when running this command `git rev-parse HEAD`. The main reason that the command fails is that the project has no git repo when checkout with Carthage.

PR #287, opened 2017-11-03
Fix get_repo_version.sh fails when project is not in a git repo

pasin:
When getting GIT_COMMIT, ignore (result to an empty value) if the command fails.

#286


pasin:
@borrrden I'm merging this in myself to unblock my DB19 release. Please review and refix if you want to do it differently.

PR #291, opened 2017-11-05
Fix to Ubuntu 17.10 xlocale.h associated build failure

mz2:
I'm going to be a killjoy again with building on Linux: building on a relatively fresh Ubuntu 17.10 would fail because some `libc++` headers would search for `<xlocale.h>` which doesn't get found. This fixes, but I'm not 100% sure it's the cleanest plausible solution.

I also added `libicu` into the list of library dependencies as I noticed following failing with the provided build instructions on this fresh new system with relatively few common development dependencies included that it was missing at least on Ubuntu.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -82,6 +82,7 @@ if("${CMAKE_SYSTEM_NAME}" STREQUAL "Linux")
     endif()
     message("Using libc++ header files in ${LIBCXX_INCLUDE}")
     include_directories("${LIBCXX_INCLUDE}/c++/v1")
+    include_directories("${LIBCXX_INCLUDE}/c++/v1/support/xlocale") # this fixed path is here to avoid compilation on Ubuntu 17.10 where xlocale.h is searched by some header(s) in libc++ as <xinclude.h> but not found from search path without this modification.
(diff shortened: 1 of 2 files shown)


borrrden:
Actually I already ran into this when I was messing around in Docker...I solved it in a different way (by making a symlink to be able to find xlocale.h) but this is probably better.  I don't see a problem with this PR.

borrrden:
This broke the build on our servers because xlocale.h is a zero-length file.  It is also a zero-length file on Ubuntu 17.10 so I don't know what it is there for?

borrrden:
Turns out that the problem is that GLIBC 2.26 (which is used in Ubuntu 17.10) removed /usr/include/xlocale.h.  Since we build our official builds on CentOS to use an older version of GLIBC to maintain compatibility, that file is still there at /usr/include/xlocale.h.  CMake has an odd quirk that any include directories get put seemingly *before* /usr/include and so an empty xlocale.h from the c++ Support directory (probably put there specifically to maintain compatibility despite this change) got included instead of the real one and many locale types vanished.  So I added a condition that the original file does not exist before adding this include directory.

mz2:
Oh joy... Sorry about that.

PR #337, opened 2017-12-21
Cross platform implementation of ‘cblite’ tool

borrrden:
Almost entirely changes needed for running in interactive mode only.  Summary:

Replaced libedit with linenoise (fork with Windows support and UTF-8 input support)
Added a few headers and defines to make some previously undefined functions available (isatty, fnmatch)

### DIFF
--- a/tools/Tool.cc
+++ b/tools/Tool.cc
@@ -9,6 +9,7 @@
 #include "Tool.hh"
 #include "Logging.hh"
 #include "linenoise.h"
+#include "argumentTokenizer.hh"
 #include <stdio.h>
 #include <regex>
 
@@ -58,53 +59,6 @@ static bool outputIsColor() {
     return sColor;
 }
 
-static bool tokenize(const char* input, deque<string> &args)
-{
-    if(input == nullptr) {
-        return false;
-    }
-    
-    // " (Matches a " character (ASCII 34)
-    // ( (First capture group)
-    //   (?: (Non capturing group)
-    //     \\"|[^"] (Either an escaped '"', or any character other than '"')
-    //   )
-    // * (Zero or more of the previous group)
-    // )
-    // " (Matches a " character (ASCII 34)
-    //
-    // | (OR, the following)
-    //
-    // ( (Second capture group)
-    //   \S+ One or more non-whitespace characters
-    // )
-    //
-    // This regular expression will split a string into tokens of either quoted items
-    // (that can contain escaped quotation marks) or whitespace delimited entries
-    regex pattern("\"((?:\\\\\"|[^\"])*)\"|(\\S+)");
-    cregex_iterator iter(input, input + strlen(input), pattern);
-    cregex_iterator end;
-    
-    for(; iter != end; iter++) {
-        cmatch const &what = *iter;
-        if(what.str(1).length() > 0) {
-            // Remove the escaped quotes
-            string match = what.str(1);
-            size_t start_pos = 0;
-            while((start_pos = match.find("\\\"", start_pos)) != std::string::npos) {
-                match.replace(start_pos, 2, "\"");
-                start_pos += 1;
-            }
-            
-            args.push_back(match);
-        } else {
-            args.push_back(what.str(0));
-        }
-    }
-    
-    return true;
-}
-
 // See <https://en.wikipedia.org/wiki/ANSI_escape_code#CSI_codes>
 #define ANSI_COLOR_ESC(STR) "\033[" STR "m"
 #define ANSI_COLOR_PROMPT   ANSI_COLOR_ESC("1")     // Bold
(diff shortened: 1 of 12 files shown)


tools/Tool.cc
```diff
@@ -51,6 +57,61 @@ static bool outputIsColor() {
...
 }
 
+static bool tokenize(const char* input, deque<string> &args)
+{
+    string str(input);
+    int start = 0;
+    for( size_t i=0; i<str.length(); i++){
+        char c = str[i];
+        if( c == ' ' ) {
+            if(i - start > 1) {
```
snej: This test seems wrong. If the first argument is 1 byte long, like `x y`, then this test will be false and the first argument will be ignored. Initializing `start` to -1 might fix this.

tools/Tool.cc
```diff
@@ -51,6 +57,61 @@ static bool outputIsColor() {
...
+                args.push_back(str.substr(start, i - start));
+            }
+            
+            while(str[i] == ' ') {
+                i++;
+            }
+            
+            start = i;
+            i--;
+        } else if(c == '\"' ) {
```
snej: There's no way to escape a quote.

tools/Tool.cc
```diff
@@ -51,6 +57,61 @@ static bool outputIsColor() {
     return sColor;
 }
 
+static bool tokenize(const char* input, deque<string> &args)
+{
+    string str(input);
```
snej: Turning the input into a C++ string doesn't seem worth it. The nice thing with parsing a C string is you can treat the trailing NUL byte as a special character, which simplifies parsing. Here you have to do some special checks to handle finishing the last argument, mismatched quotes, etc.

tools/Tool.cc
```diff
@@ -51,6 +57,61 @@ static bool outputIsColor() {
...
+            while( i < str.size() - 1 && str[i] != '\"' ){ i++; }
+            if(str[i] != '\"') {
+                return false;
+            }
+            
+            args.push_back(str.substr(start + 1, i - start - 1));
+            start = i;
+        } else if(c == '\r' || c == '\n') {
+            break;
+        } else if(!isascii(c)) {
```
snej: This seems like the wrong way to treat non-ASCII bytes. For example, `foo¿bar` will produce the argument `bar` because `start` moves forward past the `foo`. (No wait, actually it'll produce invalid UTF-8 because the subsequent byte(s) of the UTF-8 sequence for `¿` will be emitted but not the first.

tools/Tool.cc
```diff
@@ -51,6 +57,61 @@ static bool outputIsColor() {
...
+            start = i+1;
+        }
+    }
+    
+    int offset = 0;
+    int carriageReturn = str.find("\r");
+    int lineFeed = str.find("\n");
+    if(carriageReturn > 0) {
```
snej: You're checking for CR or LF anywhere in the string, but the use of `offset` seems to assume they're at the end.

tools/Tool.cc
```diff
@@ -51,6 +57,61 @@ static bool outputIsColor() {
...
+    if(lineFeed > 0) {
+        offset++;
+    }
+    
+    if(str[str.length() - offset - 1] == '\"') {
+        offset++;
+    }
+    
+    if(str.length() - start - offset > 1) {
+        args.push_back(str.substr(start, str.length() - start - offset));
```
snej: It would be a lot easier to handle this last argument as part of normal processing, if you scanned `input` until you hit a `\0`.

tools/Tool.cc
```diff
@@ -51,6 +58,53 @@ static bool outputIsColor() {
     return sColor;
 }
 
+static bool tokenize(const char* input, deque<string> &args)
+{
+    if(input == nullptr) {
+        return false;
+    }
+    
+    // " (Matches a " character (ASCII 34)
```
snej: This will still match a double-quote preceded by a backslash, won't it? Meaning you can't escape quotes.

tools/cblite/cbliteTool.hh
```diff
@@ -12,14 +12,17 @@
...
-#include <fnmatch.h>        // POSIX (?)
 #include <fstream>
 #include <iomanip>
 #include <map>
 #include <set>
 #include <sstream>
 #include <vector>
 
+#ifndef _MSC_VER
+#include <fnmatch.h>        // POSIX (?)
```
snej: This include could be moved to cbliteTool+ls.cc.

tools/Tool.cc
borrrden: You *can* escape quotes, just not unless they are also inside quotes (which in my mindset at the time was the only place that they would occur, oops).  This could probably be solved by adding a "not backslash" to each of the outer quotes

PR #339, opened 2017-12-24
Implement Mutable Fleece in Java

pasin:
Note: MValue.registerDelegate(delegate) is required to be called first thing before using the MFleece API to set delegate implementation for MValue to Native value conversion.

https://github.com/couchbase/couchbase-lite-android/issues/1447


Java/src/com/couchbase/litecore/fleece/MArray.java
```diff
@@ -13,98 +13,189 @@
...
-        free();
-        super.finalize();
+    public boolean insert(long index, Object value) {
+        if (!isMutable())
+            throw new IllegalStateException("Cannot insert items in a non-mutable MArray");
+
+        if (index < 0 || index > count())
+            return false;
+
+        if (index < count())
```
hideki: Is calling `populateValues()` once enough? I think it is not necessary to call the method always if `index < count()`.

Java/src/com/couchbase/litecore/fleece/MArray.java
```diff
@@ -13,98 +13,189 @@
...
+        if (end <= start)
+            return end == start;
 
-    private static native void initInSlot(long handle, long mv, long parent);
+        long count = count();
+        if (end > count)
+            return false;
 
-    private static native void initAsCopyOf(long handle, long mDict, boolean isMutable);
+        if (end < count)
```
hideki: Is calling `populateValues()` once enough? I think it is not necessary to call the method always if `end < count()`.

Java/src/com/couchbase/litecore/fleece/MDict.java
```diff
@@ -13,92 +13,244 @@
...
+            _count = 0;
+        }
     }
 
-    public boolean contains(String key) {
-        return contains(_handle, key);
+    public void initAsCopyOf(MDict d, boolean isMutable) {
+        super.initAsCopyOf(d, isMutable);
+        _dict = d._dict;
+        _map = new HashMap<>(d._map);
```
hideki: question: Is it enough to do a shallow copy instead of deep copy?

Java/src/com/couchbase/litecore/fleece/MDict.java
```diff
@@ -13,92 +13,244 @@
...
         }
+
+        if (_dict != null) {
+            FLDictIterator itr = new FLDictIterator();
+            try {
+                itr.begin(_dict);
+                String key;
+                SharedKeys sk = new SharedKeys(getSharedKeys());
+                while ((key = SharedKeys.getKey(itr, sk)) != null) {
+                    if (!_map.containsKey(key))
```
hideki: question: Does `getKeys()` return the key for its value is `MValue.EMPTY`?

Java/src/com/couchbase/litecore/fleece/MDict.java
```diff
@@ -13,92 +13,244 @@
...
-    private static native boolean remove(long handle, String key);
+            if (_dict != null) {
+                FLDictIterator itr = new FLDictIterator();
+                try {
+                    itr.begin(_dict);
+                    String key;
+                    SharedKeys sk = new SharedKeys(getSharedKeys());
+                    while ((key = SharedKeys.getKey(itr, sk)) != null) {
+                        if (!_map.containsKey(key)) {
+                            enc.writeKey(key);
```
hideki: Should not serialize if a `value` in the `_map` is `EMPTY`?

Java/src/com/couchbase/litecore/fleece/MArray.java
pasin: I followed the original MArray's implementation. Can you create a ticket to revise the implementation?

https://github.com/couchbaselabs/fleece/blob/260479eac5b274fae9642135978ab7b7d6edd3c5/Fleece/Mutable/MArray.hh#L93-L94

Java/src/com/couchbase/litecore/fleece/MArray.java
pasin: I believe it should be done once, but I followed the original MArray's implementation. Can you create a ticket to revise the implementation?

https://github.com/couchbaselabs/fleece/blob/260479eac5b274fae9642135978ab7b7d6edd3c5/Fleece/Mutable/MArray.hh#L114-L115

Java/src/com/couchbase/litecore/fleece/MDict.java
pasin: This part is not clear. I need your input. Do you think what the following line is doing?

https://github.com/couchbaselabs/fleece/blob/260479eac5b274fae9642135978ab7b7d6edd3c5/Fleece/Mutable/MDict.hh#L55

Note .NET doesn't do copy.

Java/src/com/couchbase/litecore/fleece/MDict.java
pasin: It shouldn't. Why do you think it will return the empty value keys?

Java/src/com/couchbase/litecore/fleece/MDict.java
pasin: It shouldn't. Why do you think it will serialize the empty value keys?

Java/src/com/couchbase/litecore/fleece/MArray.java
hideki: https://github.com/couchbaselabs/fleece/issues/27

Java/src/com/couchbase/litecore/fleece/MArray.java
hideki: https://github.com/couchbaselabs/fleece/issues/27

Java/src/com/couchbase/litecore/fleece/MDict.java
hideki: This code works as expected. Originally I misunderstood the behavior if the value is EMPTY. no issue here.

Java/src/com/couchbase/litecore/fleece/MDict.java
hideki: same here. no issue.

Java/src/com/couchbase/litecore/fleece/MDict.java
hideki: http://www.cplusplus.com/reference/unordered_map/unordered_map/operator=/

I believe `_map = d._map;` is case 1. But not sure if it is shallow copy or deep copy.

PR #340, opened 2017-12-27
Add in some initial implementation for failure testing

borrrden:
Let me know if this is something that seems worthwhile.  It can increase our code coverage fairly easily while not changing anything in release builds.  A runtime toggleable item (similar to expect exceptions) can control whether or not a method, or a call to an outside method, fails so that we can see what happens.

### DIFF
--- a/C/c4Private.h
+++ b/C/c4Private.h
@@ -27,9 +27,28 @@ extern "C" {
         ignore the exception. */
     CBL_CORE_API extern std::atomic_int gC4ExpectExceptions;
 
+#if DEBUG
+    /** If > 0, the library will force a failure for every function in the C API that uses a try 
+        catch section */
+    CBL_CORE_API extern std::atomic_int gC4ForceFailure;
+#endif
+
 #else
     CBL_CORE_API extern atomic_int gC4InstanceCount;
     CBL_CORE_API extern atomic_int gC4ExpectExceptions;
+#if DEBUG
+    CBL_CORE_API extern atomic_int gC4ForceFailure;
+#endif
+#endif
+
+#if DEBUG
+#define MAYBE_FAIL if(gC4ForceFailure) litecore::error::_throw(litecore::error::Domain::LiteCore, litecore::error::LiteCoreError::AssertionFailed);
+#define MAYBE_RECORD_ERROR(outErr, retVal) if(gC4ForceFailure) { recordError(LiteCoreDomain, kC4ErrorAssertionFailed, "Test failure", outErr); return retVal; }
+#define MAYBE_FAIL_CALL(call) (gC4ForceFailure && !call) || call
+#else
+#define MAYBE_FAIL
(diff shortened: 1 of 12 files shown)


C/c4Private.h
```diff
@@ -27,9 +27,28 @@ extern "C" {
         ignore the exception. */
     CBL_CORE_API extern std::atomic_int gC4ExpectExceptions;
 
+#if DEBUG
+    /** If > 0, the library will force a failure for every function in the C API that uses a try
```
snej: I think you mean "...for every function in the C API that outputs a C4Error"?

C/c4Private.h
```diff
@@ -27,9 +27,28 @@ extern "C" {
...
 #else
     CBL_CORE_API extern atomic_int gC4InstanceCount;
     CBL_CORE_API extern atomic_int gC4ExpectExceptions;
+#if DEBUG
+    CBL_CORE_API extern atomic_int gC4ForceFailure;
+#endif
+#endif
+
+#if DEBUG
+#define MAYBE_FAIL if(gC4ForceFailure) litecore::error::_throw(litecore::error::Domain::LiteCore, litecore::error::LiteCoreError::AssertionFailed);
```
snej: Could you wrap these macro definitions to 100-char lines, please?

C/c4Private.h
```diff
@@ -27,9 +27,28 @@ extern "C" {
...
+#define MAYBE_RECORD_ERROR(outErr, retVal) if(gC4ForceFailure) { recordError(LiteCoreDomain, kC4ErrorAssertionFailed, "Test failure", outErr); return retVal; }
+#define MAYBE_FAIL_CALL(call) (gC4ForceFailure && !call) || call
```
snej: The logic here seems weird. If `gForceFailure` is true and `call` returns false, then `call` ends up being evaluated again and its result used. I think what you're trying to do is to evaluate `call` but force the result to be false? Then how about `(call) && !gC4ForceFailure`?

snej: Also, be careful to parenthesize the macro parameter (`call`) and the body of the macro, to avoid unintended results.

borrrden: Good call on that (no pun intended...)

borrrden:
> I’m not sure this kind of testing is useful, though

Nor am I, which is why I only took it so far (this PR is not complete, I just wanted to give an idea).  It is very useful in the area of code coverage obviously, where I struggle to gain single percents at a time, but I'm not sure if increasing like this can be considered too self-serving or not.  The `MAYBE_FAIL_CALL` method probably has the most practical application since it tests what happens if *outside* things fail.  Second place is the `MAYBE_FAIL` which would be useful probably once to make sure that C++ exceptions are properly getting stopped and wrapped at the C boundary.  `MAYBE_RECORD_ERROR` is probably the most self serving of the three.

snej:
This is ancient...

PR #347, opened 2018-01-06
Fixes #329 (Specifying negative limit/offset values fetches the entire DB)

### LINKED ISSUE #329: Specifying negative limit/offset values fetches the entire DB
CBL Query:
```
    let searchQuery = Query
        .select(SelectResult.all())
        .from(DataSource.database(db))
        .limit(limit,offset: offset)
```
Run the above query with limit = -5 and offset = -5
The query returned 31591 results against the travel sample bucket.

Equivalent n1ql query:
```
select * from `travel-sample` where meta().id not like "_sync%" limit -5 offset -5
```
returned:
```
{
  "results": []
}
```

borrrden:
Made this a PR because the changes are awfully messy but I can't see a cleaner way to do it.  The N1QL example is apples and oranges (fixing *that* case is trivial.  Fixing the case of query params with negative values is messy)

### DIFF
--- a/LiteCore/Query/QueryParser.cc
+++ b/LiteCore/Query/QueryParser.cc
@@ -142,6 +142,7 @@ namespace litecore {
         _context.clear();
         _context.push_back(&kOuterOperation);
         _parameters.clear();
+        _limitOrOffsetParameters.clear();
         _variables.clear();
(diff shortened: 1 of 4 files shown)


borrrden:
Not going to go with this implementation

PR #366, opened 2018-01-23
VS2017 improvements

d122y:
This fixes compilation errors in VS2017

### DIFF
--- a/LiteCore/Database/SequenceTracker.cc
+++ b/LiteCore/Database/SequenceTracker.cc
@@ -217,7 +217,8 @@ namespace litecore {
                     external = i->external;
                 else if (i->external != external)
                     break;
-                changes[n++] = {i->docID, i->revID, i->sequence, i->bodySize};
+                Change change = {i->docID, i->revID, i->sequence, i->bodySize};
+                changes[n++] = change;
             }
             ++i;
         }
--- a/LiteCore/Support/PlatformIO.hh
+++ b/LiteCore/Support/PlatformIO.hh
@@ -22,6 +22,8 @@
     #define strcasecmp  _stricmp
     #define fdclose ::_close
 
+    struct stat;
+
     namespace litecore {
         int mkdir_u8(const char* const path, int mode);
         int stat_u8(const char* const filename, struct stat* const s);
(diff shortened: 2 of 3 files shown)


LiteCore/Database/SequenceTracker.cc
```diff
@@ -217,7 +217,8 @@ namespace litecore {
                     external = i->external;
                 else if (i->external != external)
                     break;
-                changes[n++] = {i->docID, i->revID, i->sequence, i->bodySize};
```
snej: I'd rather avoid the extra copying. Does `changes[n++] = Change{i->docID, i->revID, i->sequence, i->bodySize}` work?

LiteCore/Support/PlatformIO.hh
```diff
@@ -22,6 +22,7 @@
     #define strcasecmp  _stricmp
     #define fdclose ::_close
 
+struct stat;
```
snej: Please indent 4 spaces and put a blank line after.

LiteCore/Database/SequenceTracker.cc
borrrden: It does not.  This seems like a compiler error, unless something in C++14 prohibits this.  The issue comes from the mismatch of types between i->docID (alloc_slice) and docID on change (slice).  A copy constructor exists between alloc_slice and slice but it seems it is not getting hit with the initializer list.  For example, this works:

```c++
slice docID = i->docID;
slice revID = i->revID;
changes[n++] = { docID, revID, i->sequence, i->bodySize };
```

LiteCore/Support/PlatformIO.hh
```diff
@@ -22,6 +22,8 @@
     #define strcasecmp  _stricmp
     #define fdclose ::_close
 
+    struct stat;
```
d122y: @snej , updated

borrrden:
I just realized that this PR is on the db021 branch.  It should be on dev instead.  Is there a way to modify this PR to do so?

d122y:
@borrrden, created another PR #374 to dev. Feel free to close this one if it's easier

Issue #348, opened 2018-01-08
Need LiteCore To Compile With SEE or SQLite


djpongh:

borrrden:
Initial work for this is on the `feature/sqlite_switch` branch.  CMake compiles with both SQLite and SQLCipher as controlled by the `BUILD_ENTERPRISE` CMake flag.  All tests pass in both versions.

snej:
CMake can now build EE, but Xcode can't. I believe we need this for building couchbase-lite-ios. But I'm not sure how to make it possible to build with either of two SQLite source files depending on a configuration.

borrrden:
Yes we do need this for iOS.  We could do some ifdef incantations if that would make things easier. Or switch to another way of doing things. It doesn’t have to be source switching.

PR #370, opened 2018-01-24
Remove sqlcipher and encryption functionality

borrrden:
This is half of the logic needed for #348 (Need LiteCore To Compile With SEE or SQLite).  LiteCore CE will now compile here only with vanilla sqlite.c

@snej Xcode project updates needed, and feedback about the process of switching.  CMake makes it easy to sneak in a substitute sqlite3.c file with encryption features, but I'm not sure how easy that is for Xcode.


CMakeLists.txt
```diff
@@ -20,7 +21,28 @@ add_definitions(-DSQLITE_OMIT_LOAD_EXTENSION        # This one's for SQLiteCpp
                 -DUSE_IPV6                          # Enable IPv6 support in CivetWeb
                 -DCMAKE                             # Let the source know this is a CMAKE build
                 -D__STDC_LIMIT_MACROS               # Ensure the presence of UINT64_MAX, etc
-                -DNOMINMAX)                         # Get rid of pesky Windows macros for min and max
+                -DNOMINMAX                          # Get rid of pesky Windows macros for min and max
+                -DSQLITE_DEFAULT_WAL_SYNCHRONOUS=1
```
snej: These flags were removed here because SQLite is now built by SQLiteCpp's nested CMakeLists, right?
Did you modify that other file to include the same flags as here?

borrrden:
I think this is ready to merge now.  Until it is, the enterprise builds will be invalid.

thamht4190:
As I read from [here ](https://github.com/couchbase/couchbase-lite-core/commit/897b2bc79e87724369e9101052836f1985324502):

> Database encryption will be an Enterprise Edition (paid) feature in Couchbase Lite 2.0.

Can you please provide me with some information about paid plan (like lifetime or subscription, how much, etc...)? Thanks!

borrrden:
@ssmotra Please see above

PR #374, opened 2018-01-24
VS2017 improvements

d122y:


borrrden:
Hmmm this didn't go quite as planned in terms of GitHub keeping track of what went in, but the Fleece submodule update was committed also as part of this PR.

Issue #381, opened 2018-01-29
Getting a "disk I/O error" when running "test_no_conflicts_enabled_with_revs_limit" repeatedly


hrajput89:
We are testing "no conflict" enabled with revision_limit set test, for multiple options, On running the test with different parameters individually, all test passes. However, when running them in succession to each other, test starts to fail after first successful run. Below are the details of the test - 

>         1. Enable allow_conflicts = false in SG config and specified revs_limit
>         2. Add a few docs through CBL.
>         3. Update the doc few times in CBL.
>         4. Do push replication to Sg with continous replication
>         5. Update docs in CBL again few more times which can cross the revs_limit
>         6. Check the revision list for the doc

The test environment - 
CBL app -
 -  android 2.0.0-291

SG - 
 - OS: Centos7
 - Build - 2.0.0-749
 - mode - cc

CBS -
 - OS: Centos7
 - Build - 5.0.0-3519
 - mode

Test method is available here - https://github.com/couchbaselabs/mobile-testkit/blob/feature/cbl20-ver1/testsuites/CBLTester/CBL_Functional_tests/test_no_conflicts_cbl.py#L96

[sgcollect_info.zip](https://github.com/couchbase/couchbase-lite-core/files/1674926/sgcollect_info.zip)
[CBL_app.log](https://github.com/couchbase/couchbase-lite-core/files/1674928/CBL_app.log)
[CBL_testrunner.log](https://github.com/couchbase/couchbase-lite-core/files/1674929/CBL_testrunner.log)

snej:
There's an unexpected warning just above the I/O error log: `W/LiteCore: SQLite error (code 28): file unlinked while open: /data/user/0/com.example.hemant.coucbaselitesampleapp/files/pull-test.cblite2/db.sqlite3`. Deleting an open SQLite database can certainly cause data corruption, but your test description doesn't say anything about deleting the DB.

snej:
@hrajput89 Is it possible you're deleting the database on one thread, and opening it on another before the deletion has finished? I found a race condition in LiteCore that could allow this to happen. (But even if I fix that, it's still a bad idea for you to do a simultaneous delete and open. Those should be serialized.)

djpongh:
@snej , is the race condition fix, checked in? I want to see if QE can re-run with the latest LiteCore.

snej:
It's on the dev branch. I don't know if this is the problem with this test, though. I was hoping for an answer to my question above.

[...]

PR #394, opened 2018-02-07
Small tweak to yesterday's zombie-db fix

snej:
We don't need to set the NO_CKPT_ON_CLOSE flag unless the database still
has open statements and will become a zombie. So we can move that code
into the `if (!_sqlDb->closeUnlessStatementsOpen())` block that handles
that case.

(This isn't going to affect what the code does, since the
`if (_sqlDb->getHandle() != NULL)` test effectively does the same thing.
But it's clearer this way. I've also added some lengthy comments
explaining what's going on.)

For #381

### DIFF
--- a/LiteCore/Storage/SQLiteDataFile.cc
+++ b/LiteCore/Storage/SQLiteDataFile.cc
@@ -263,21 +263,24 @@ namespace litecore {
         _setLastSeqStmt.reset();
         if (_sqlDb) {
             optimizeAndVacuum();
-            // Log if there are still statements open. This does happen in real use sometimes in
-            // CBL-Java due to Query finalizers being delayed.
+            // Close the SQLite database:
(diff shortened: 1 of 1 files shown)


snej:
Jim — this is a minor change, but since the problem was so nasty and you put a lot of time into testing, I'd like to run it past you to make sure I'm not making a mistake. Thx!

borrrden:
I double checked the behavior of `sqlite3_close` on `nullptr` and the documentation indicates it is a harmless no-op, so I am assuming that `nullptr` will return `true` for `closeUnlessStatementsOpen`.  If that is not the case then the check is still needed, or some logic in `closeUnlessStatementsOpen` to return true in the face of `nullptr`

snej:
The only way `_sqlDb->getHandle()` can be null is if `closeUnlessStatementsOpen` already ran and returned true. (The pre-existing code in that class never nulls it out.)

PR #395, opened 2018-02-07
Convert to asciidoc

tleyden:
@jamiltz converted this doc to "Asciidoc" format, which is more powerful than markdown.   

The main value-adds are:

- Automatic TOC generation
- Easier to read table format syntax
- As human readable as markdown

@jamiltz and I were planning on to do a presentation on asciidoc, but I think this PR could go in in the meantime (or we could at least have a discussion about it)

Also in this PR I added some ascii diagrams of the blip replication flow, which helped me get my head around the replication protocol and I use it as a reference.


jamiltz:
@tleyden looks good. I'm going to add another commit to make this work with the front-end layout.

jamiltz:
@tleyden I changed where files are located. Will have to check if this impacts ease of use. On the plus side, we now have a [staging URL](http://docs-build.sc.couchbase.com/antora/site/couchbase-lite-core/docs/replication-protocol.html) which gets automatically triggered when there's a new commit on this repo.

PR #421, opened 2018-02-24
Reduce number of file descriptors opened while pulling blobs

### LINKED ISSUE #420: IncomingBlob quickly saturates file descriptor count
One of the functional tests has a replication test that ends up with 2000 blobs.  The limit for number of open file descriptors for a process on Windows is 2048, and it seems that they all get used up during this process (Each DLL, etc, opened by the dotnet process itself also counts against this limit for .NET).  This causes mkTempFile to start returning invalid null handles and the result is that the handles cannot be closed later and the underlying numeric file descriptors are stuck open.  Specifically `fdopen` is setting `errno` to `EMFILE`.  It looks like these need to be throttled somehow.

snej:
* Changed IncomingRev to pull only one blob at a time.
* Changed IncomingBlob to delay opening the write stream until it
  receives the first bytes from the response.

Fixes #420

### DIFF
--- a/Replicator/IncomingBlob.cc
+++ b/Replicator/IncomingBlob.cc
@@ -19,11 +19,17 @@
 #include "IncomingBlob.hh"
 #include "StringUtil.hh"
 #include "MessageBuilder.hh"
+#include <atomic>
 
 using namespace fleece;
 using namespace litecore::blip;
 
 namespace litecore { namespace repl {
+
+#if DEBUG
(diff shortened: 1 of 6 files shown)


borrrden:
This seems to take care of the issue, but introduces a new one.  After the replication goes idle, and then the test client calls stop the replicator switches to busy, closes the web socket connection, but then never seems to enter the stopped state.

borrrden:
Nevermind about the above, it appears to be unrelated and due to a C# change

PR #482, opened 2018-03-20
 Stop c4doc_update() from clobbering remote-revID state

snej:


snej:
the cookie commit shouldn't be here; going to close and re-create this.

PR #483, opened 2018-03-20
Stop c4doc_update() from clobbering remote-revID state

### LINKED ISSUE #478: c4doc_update can clobber current-remote-rev info
From @tleyden's hard work on couchbase/couchbase-lite-android#1664 "doc update is not getting replicated when android is in offline":

> ### Environment 
> * Version: Android 2.0.0-435
> * Client app: Todo app android
> * Server:    sg - 823 [config](https://gist.github.com/tleyden/533d4a180af8962669c554332b7e54b3)
> 
> ### steps to reproduce : 
> 1. login to todo app android 
> 2. create list on android
> 3. Create doc in android 
> 4. update doc in android
> 
> Result : The initial doc will get pushed to sync gateway, but updates will be rejected w/ the error mentioned in https://github.com/couchbase/couchbase-lite-android/issues/1664#issuecomment-374343980

After step 3, the doc is pushed and its kSynced flag is set, marking that revision 1 is the current server version.

After step 4, the doc is pushed again, but by this time **the current-server-version marker is gone**, so the pusher tells the server there's no ancestor rev, triggering the 409 from the server.

I'm pretty sure what's clobbering the marker is that the second save is done using `c4doc_update`, which takes the C4Document resulting from the first save. This C4Document has revision 1 in it but doesn't have the kSynced flag or the remote-revision marker, which hadn't been set yet. CBL adds revision 2. Then c4doc_update saves that to the db; the overwrite succeeds because the sequences match. But the kSynced flag is lost (because setting that doesn't bump the sequence.)

snej:
I had to replace the implementation of c4doc_update with one that just
uses c4doc_put, so it won’t be as fast. (It’ll do a read-modify-write
instead of just modify-write.) But it’s the only thing I can think of
that’s safe to do for 2.0.0.

Still needs a unit test.

Fixes #478

### DIFF
--- a/C/c4.def
+++ b/C/c4.def
@@ -203,6 +203,7 @@ c4doc_getForPut
 c4_getObjectCount
 c4_dumpInstances
 c4_shutdown
+c4db_markSynced
 gC4InstanceCount
 gC4ExpectExceptions
 
--- a/C/c4.exp
+++ b/C/c4.exp
@@ -209,6 +209,7 @@ _c4error_return
 _c4doc_getForPut
 _c4_getObjectCount
 _c4_shutdown
+_c4db_markSynced
 _c4_dumpInstances
 _gC4InstanceCount
 _gC4ExpectExceptions
--- a/C/c4Document.cc
+++ b/C/c4Document.cc
@@ -517,6 +517,28 @@ C4Document* c4doc_update(C4Document *doc,
                          C4RevisionFlags revFlags,
                          C4Error *outError) noexcept
 {
+#if 1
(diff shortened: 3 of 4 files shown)


tleyden:
LGTM.  Please review unit test and then merge.

tleyden:
I'm going to retest the original steps in https://github.com/couchbase/couchbase-lite-android/issues/1664

tleyden:
> I'm going to retest the original steps in couchbase/couchbase-lite-android#1664

Done.  Can't reproduce using same steps.

PR #494, opened 2018-03-27
fix compilation using gcc

KraemerDEM:
depends on https://github.com/couchbaselabs/BLIP-Cpp/pull/10 and
https://github.com/couchbaselabs/fleece/pull/25

Most of the changes are missing headers and avoiding ambiguities regarding overloaded [] operators.

I'm not sure about the exclusion of the static C4SocketFactory constructor.
gcc complains that "declaration of ‘static C4SocketFactory litecore::websocket::CivetProvider::C4SocketFactory()’ [-fpermissive]  static C4SocketFactory C4SocketFactory();
changes meaning of ‘C4SocketFactory’ from ‘typedef struct C4SocketFactory C4SocketFactory’ [-fpermissive]
 } C4SocketFactory; "
Is there another way to solve that?


snej:
We appreciate that you did the work, but we don't have a need to support GCC, since Clang and MSVC work on all our target OS's. We actually did support GCC earlier, but gave up because it was troublesome to support the idiosyncrasies of three compilers.

Is there a reason you can't use Clang?

KraemerDEM:
Sorry to hear that.
Is it possible to include the patches, but put a warning in the CMakeLists.txt that gcc is not officially supported?
I think the patches are (currently) not that intrusive (except for the C4SocketFactory issue I mentioned above).

We are currently using gcc for building our codebase on Linux and on Android, and we have to support older Android devices where AFAIK clang is not an option(?) .

snej:
Let me take a closer look at the diffs. Some of them seem like they could be resolved more simply (like the enum scope ones.)

borrrden:
FYI There is now CI running checks for GCC buildability, so this PR may not be needed.

KraemerDEM:
Does that mean there is some branch where you are working on gcc buildability?
I've not found anything.
The master branch still needs my fixes to build with gcc

borrrden:
oh My mistake I forgot that it is only fleece that is running through GCC. I’ll reopen this.

snej:
Oops, I forgot that too :) GCC compatibility would probably be a boon for Raspberry Pi builds, given the pain of having to install up-to-date Clang on Raspbian :-p

Dushistov:
Any progress? I also need this, because of I still have  to use gcc for Android build,
plus there is old Linux distro, that have gcc as default compiler and I have to have compatible ABI,
so I have to build with specific version of gcc.

snej:
No progress on this to my knowledge. Currently we (Couchbase) don't need a GCC build, but that might change if/when we decide to support a C API to Couchbase Lite.

>I still have to use gcc for Android build

We already build LiteCore for Android with Clang...

Dushistov:
@snej 

> We already build LiteCore for Android with Clang.

Yes, I know. But I have several ndk parts in my android project.
And all of them are built with gcc. I can explain why I have to spend time on
couchbase-lite-android integration into my project, but I have no idea how to explain why I have to heavily re-test (especially third part libraries) all other ndk parts (because they are already fully tested with gcc, mix clang and gcc built code into one `.apk` walks on the thread over the abyss).

I undertand that supporting build with one more compiler is expensive,
but what about a patch that warns about gcc not being officially supported,
and accepts PRs with obvious cleanups, like in this PR:

```C
C4ErrorCode::kC4ErrorNotFound to kC4ErrorNotFound
    if (NOT LIBCXX_INCLUDE)
    if (NOT LIBICU_INCLUDE)
```

Replicator/CivetWebSocket.hh
```diff
@@ -44,7 +44,8 @@ namespace litecore { namespace websocket {
         /** Registers a WebSocket handler for an HTTP server. */
         void setServerWebSocketHandler(mg_context*, std::string uri, ServerWebSocketHandler);
 
-        static C4SocketFactory C4SocketFactory();
+// this doesn't compile when using GCC, is it needed anywhere ?
+//        static C4SocketFactory C4SocketFactory();
```
borrrden: I think it is used in the automated tests (CppTests / C4Tests)

snej:
I'd merge a conflict-free patch that fixed GCC issues, while still passing unit tests in Clang of course.

(I've backed off from my initial pooh-pooing of GCC, having now spent some time working with embedded OS's that only support GCC toolchains...)

KraemerDEM:
I've rebased the pull request to the current master and added some stuff needed for gcc on Android.
This also depends on https://github.com/couchbaselabs/fleece/pull/38

CMakeLists.txt
```diff
@@ -94,33 +94,35 @@ if("${CMAKE_SYSTEM_NAME}" STREQUAL "Linux")
     # Enable relative RPATHs for installed bits
     set (CMAKE_INSTALL_RPATH "\$ORIGIN")
 
-    if(NOT "${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
-        message(FATAL_ERROR "${CMAKE_CXX_COMPILER_ID} is not supported for building!")
+    if(NOT "${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
```
snej: This logic is kind of convoluted. Please change it to (pseudocode) `if (clang) { ... } else if (!gnu) {fatal error}`. Thanks.

CMakeLists.txt
```diff
@@ -226,6 +228,9 @@ aux_source_directory(LiteCore/Storage         STORAGE_SRC)
 aux_source_directory(LiteCore/VersionVectors  VERSIONVECTORS_SRC)
 aux_source_directory(LiteCore/Support         SUPPORT_SRC)
 list(REMOVE_ITEM SUPPORT_SRC LiteCore/Support/Logging_Stub.cc)
+if (ANDROID)
```
snej: should be `if (!APPLE)`

LiteCore/Support/StringUtil.hh
```diff
@@ -23,12 +23,26 @@
 #include <string>
 #include <vector>
 #include <sstream>
+#if defined __ANDROID__ && !defined __clang__
```
snej: Add a blank line after the last `#include`.

Replicator/CivetWebSocket.cc
```diff
@@ -166,7 +166,7 @@ namespace litecore { namespace websocket {
             Log("CivetWebSocket connecting to <%.*s>...", SPLAT(_address.url()));
 
             stringstream extraHeaders;
-            for (Dict::iterator header(_options[kC4ReplicatorOptionExtraHeaders].asDict());
+                for (Dict::iterator header(_options[kC4ReplicatorOptionExtraHeaders].asDict());
```
snej: Indent level is wrong -- make sure there are no tabs, please. (Same for the two diffs below.) Actually looks like there are no changes aside from the indents, so you could remove the diffs.

Replicator/DBWorker.cc
```diff
@@ -208,7 +208,7 @@ namespace litecore { namespace repl {
     string DBWorker::effectiveRemoteCheckpointDocID(const C4UUID *localUUID, C4Error *err) {
         // Derive docID from from db UUID, remote URL, channels, filter, and docIDs.
         Array channels = _options.channels();
-        Value filter = _options.properties[kC4ReplicatorOptionFilter];
+            Value filter = _options.properties[kC4ReplicatorOptionFilter];
```
snej: Indent level is wrong -- make sure there are no tabs, please. Actually looks like there is no changes aside from the indent, so you could remove this diff.

Replicator/tests/ReplicatorAPITest.hh
```diff
@@ -241,8 +241,8 @@ public:
 
 
     c4::ref<C4Database> db2;
-    C4Address _address {kDefaultAddress};
-    C4String _remoteDBName {kScratchDBName};
+    C4Address _address = kDefaultAddress;
```
snej: Any idea why GCC doesn't like the brace syntax? It and the '=' syntax should be equivalent...

Dushistov:
@KraemerDEM @snej 

It is possible do not do changes like this:

```diff
+ CHECK(error.code == C4ErrorCode::kC4ErrorNotFound);
- CHECK(error.code == kC4ErrorNotFound);
```

in include/c4Compat.h 

```diff
-    #if (__cplusplus && _MSC_VER) || (__cplusplus && __cplusplus >= 201103L && (__has_extension(cxx_strong_enums) || __has_feature(objc_fixed_enum))) || (!__cplusplus && __has_feature(objc_fixed_enum))
+    #if (__cplusplus && _MSC_VER) || (__cplusplus && __GNUC__) || (__cplusplus && __cplusplus >= 201103L && (__has_extension(cxx_strong_enums) || __has_feature(objc_fixed_enum))) || (!__cplusplus && __has_feature(objc_fixed_enum))
+#if defined(__GNUC__)
+#define C4_ENUM(_type, _name) enum _name : _type
+#else
         #define C4_ENUM(_type, _name)     enum _name : _type _name; enum _name : _type
+#endif
```

But actually I think that this is bug in clang and MS compilers,
I create bug in clang's issue tracker https://bugs.llvm.org/show_bug.cgi?id=39426 and
I hope that in future clang stops accept this invalid code.

snej:
The explicit `C4ErrorCode::` qualifier in those lines was just a mistake -- it only happens a few times in the code, and I'm happy to take it out.

The definition of C4_ENUM was copied from Apple's definition of CF_ENUM. I don't claim to understand it in detail :)

snej:
Merged/rebased as 902209cd12043d06c668d21a95801cd23ab8ad72. Thanks very much!

Issue #525, opened 2018-05-30
Litecore does not retry subChanges request that fails with temporary 503 error


tleyden:
As part of https://github.com/couchbase/sync_gateway/issues/3595, I'm trying to run the Litecore replication tests against Sync Gateway that's configured to return 503 errors on every nth request.  

To give a broader picture, the motivation behind https://github.com/couchbase/sync_gateway/issues/3595 is to enable changes such as https://github.com/couchbase/sync_gateway/issues/3509, which will propagate more temporary errors from Couchbase Server back to CBLite, rather than retrying them internally in Sync Gateway.  

This change would put more of the onus for retrying onto Couchbase Lite.  Before making this the default behavior in Sync Gateway, I wanted to verify the CBLite retry behavior.

## Actual behavior

In an initial test, I'm seeing a case where CBLite sends a `subChanges` request to SG in frame 193:

```
BLIP Couchbase Mobile
    BLIP Message Number: 2
    BLIP Frame Flags: 0
    BLIP Properties Length: 45
    BLIP Properties: Profile:subChanges:batch:500:activeOnly:true
    BLIP Message Body: 
``` 

and SG responds in frame 197 with:

```
BLIP Couchbase Mobile
    BLIP Message Number: 2
    BLIP Frame Flags: 2
    BLIP Properties Length: 33
    BLIP Properties: Error-Domain:HTTP:Error-Code:503
    BLIP Message Body: Database server is over capacity (gocb.ErrTmpFail)
```

which causes CBLite to close the connection in frame 201:

```
WebSocket
    1... .... = Fin: True
    ...
```

## Expected behavior

When encountering 503 errors, CBLite would do a backoff / retry of the `subChanges` request until it succeeded or failed with a non-temporary error.  

Does this expectation match up w/ the [replication spec](https://github.com/couchbase/couchbase-lite-core/blob/master/modules/docs/pages/replication-protocol.adoc#subchanges)?  I'm not seeing temp errors addressed here, so maybe the first step should be to nail down exactly what the expected behavior is.

## Steps to repro

- Litecore commit 3657e009a12c1dbfad5f9571b7042df175c3fe56 (WARNING: old version, from March 21, 2018)
- Build Sync Gateway from source on the `feature/issue_3595_random_temp_blip_errors` SG branch
- Start SG with [this SG config](https://gist.github.com/0ef8eb9d4c3db5170e54d500ebe9536e)
[...]

PR #526, opened 2018-05-30
Fix the kC4Replicator2Scheme to match the values used in other parts …

tleyden:
…of the codebase

See https://github.com/couchbase/couchbase-lite-core/issues/525#issuecomment-393344030

A better fix would be to update https://github.com/couchbase/couchbase-lite-core/blob/058d1218aaa52c9f2ab438c986d468d8b8da4f2e/Replicator/c4Replicator.hh#L78-L80 to use these constants in the comparison.  

Also I didn't do a comprehensive search of the codebase.

### DIFF
--- a/C/include/c4Replicator.h
+++ b/C/include/c4Replicator.h
@@ -28,8 +28,8 @@ extern "C" {
     /** \defgroup Replicator Replicator
         @{ */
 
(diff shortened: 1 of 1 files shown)


snej:
There were a couple other changes necessary, so I made them all as 0c3d1c48d3eb3fefda0a92ef887bf20de4a72a44.

PR #534, opened 2018-06-30
Purge old log files

### LINKED ISSUE #521: Discard old binary logs
Otherwise they will just keep on building up, so there needs to be some kind of log rotation or something.

borrrden:
Deletes all but 10 log files in the given directory by examining each file for the logging magic number, and then sorting the entries by modification date and deleting the (x - 10) oldest.

Note: An actual log rotation implementation is on the `feature/log-rotation` branch.

### DIFF
--- a/LiteCore/Support/FilePath.cc
+++ b/LiteCore/Support/FilePath.cc
@@ -426,6 +426,17 @@ namespace litecore {
         return s.st_size;
     }
 
+    time_t FilePath::lastModified() const {
+        struct stat s;
+        if (stat_u8(path().c_str(), &s) != 0) {
+            if (errno == ENOENT)
+                return -1;
+            error::_throwErrno();
+        }
+        return s.st_mtime;
+    }
+
+
     bool FilePath::exists() const {
         struct stat s;
         return stat_u8(path().c_str(), &s) == 0;
--- a/LiteCore/Support/Logging.cc
+++ b/LiteCore/Support/Logging.cc
@@ -20,14 +20,12 @@
 #include "StringUtil.hh"
 #include "LogEncoder.hh"
 #include "LogDecoder.hh"
-#include <stdio.h>
-#include <stdlib.h>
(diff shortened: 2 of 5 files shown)


LiteCore/Support/FilePath.cc
```diff
@@ -426,6 +426,17 @@ namespace litecore {
         return s.st_size;
     }
 
+    time_t FilePath::lastModified() const {
+        struct stat s;
+        if (stat_u8(path().c_str(), &s) != 0) {
+            if (errno == ENOENT)
+                return -1;
```
snej: Seems weird to me that ENOENT isn't considered an error here. Is it useful to have this call succeed if the file is missing?

LiteCore/Support/Logging.cc
```diff
@@ -60,6 +58,31 @@ namespace litecore {
...
     static LogEncoder* sLogEncoder = nullptr;
     static mutex sLogMutex;
+    static const int kNumLogFiles = 10;
+
+    static void purgeOldLogs(const string& filePath)
+    {
+        FilePath filePathObj(filePath);
+        FilePath logDir = filePathObj.dir();
+        multimap<time_t, FilePath> logFiles;
+        const int magicNumber = *(int*)LogEncoder::kMagicNumber;
```
snej: You're assuming 'int' is 32-bit, which isn't necessarily true.
I think it would be cleaner to compare slices:
`slice(LogEncoder::kMagicNumber, 4) == slice(magicBuffer, 4)`

LiteCore/Support/FilePath.cc
borrrden: My basis for this was on the `dataSize` method

LiteCore/Support/Logging.cc
borrrden: OMG!  I've been in .NET for too long

PR #535, opened 2018-07-03
import replicator last seq when copying database

### LINKED ISSUE #504: Support importing replicator last sequence when copy the database
This doesn't seem to be the case for CBSE-5099 as the customer uses the pre-built database directly without copying. However from the Hipchat, @snej suggested that this [logic](https://github.com/couchbase/couchbase-lite-ios/pull/562) needed to be implemented in LC.

borrrden:
(Support importing replicator last sequence when copy the database)

### DIFF
--- a/LiteCore/Database/PrebuiltCopier.cc
+++ b/LiteCore/Database/PrebuiltCopier.cc
@@ -27,32 +27,6 @@
 
 namespace litecore {
     using namespace std;
-
-    static void createLocalCheckpointDocument(const unique_ptr<C4Database> &db)
-    {
-        C4UUID privateUUID;
-        C4Error err;
-        if(!c4db_getUUIDs(db.get(), nullptr, &privateUUID, &err)) {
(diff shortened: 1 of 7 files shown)


LiteCore/Database/PrebuiltCopier.cc
```diff
@@ -27,6 +27,32 @@
...
+
+    static void createLocalCheckpointDocument(const unique_ptr<C4Database> &db)
+    {
+        C4UUID privateUUID;
+        C4Error err;
+        if(!c4db_getUUIDs(db.get(), nullptr, &privateUUID, &err)) {
+            error::_throw(error::Domain::LiteCore, err.code);
+        }
+
+        fleeceapi::Encoder enc;
```
snej: Isn't it overkill to store this in Fleece format? The public and private UUIDs are just stored as raw bytes. If you stored this one that way too, you could maybe reuse the Database methods getUUIDIfExists and generateUUID.

PR #536, opened 2018-07-04
Fixes Android emoji issue

borrrden:
See https://github.com/couchbase/couchbase-lite-android/issues/1742 (Couchbase mobile 2.0.2 Emoji Support).  This is a bug in the Android NDK that we have to work around...

### DIFF
--- a/Java/jni/native_fleece.cc
+++ b/Java/jni/native_fleece.cc
@@ -268,7 +268,8 @@ JNIEXPORT jlong JNICALL
 Java_com_couchbase_litecore_fleece_FLValue_fromTrustedData(JNIEnv *env, jclass clazz,
                                                            jbyteArray jdata) {
     jbyteArraySlice data(env, jdata, true);
-    return (jlong) FLValue_FromTrustedData({((slice) data).buf, ((slice) data).size});
+    slice s = (slice)data;
+    return (jlong) FLValue_FromTrustedData({s.buf, s.size});
 }
 
 /*
@@ -339,7 +340,20 @@ Java_com_couchbase_litecore_fleece_FLValue_asDouble(JNIEnv *env, jclass clazz, j
 JNIEXPORT jstring JNICALL
 Java_com_couchbase_litecore_fleece_FLValue_asString(JNIEnv *env, jclass clazz, jlong jvalue) {
     FLString str = FLValue_AsString((FLValue) jvalue);
-    return toJString(env, {str.buf, str.size});
+    const char* newBytes;
+    size_t len = UTF8ToModifiedUTF8((const char*)str.buf, &newBytes, str.size);
+    if(len == 0) {
+        jclass c = env->FindClass("java/lang/OutOfMemoryError");
+        env->ThrowNew(c, "Out of memory in FLValue_AsString");
+        return nullptr;
+    }
+
+    jstring retVal = toJString(env, {newBytes, len});
+    if(len != str.size) {
+        free((void *)newBytes);
+    }
+
+    return retVal;
 }
 
 /*
--- a/Java/jni/native_fleece.cc
+++ b/Java/jni/native_fleece.cc
@@ -341,17 +341,21 @@ JNIEXPORT jstring JNICALL
 Java_com_couchbase_litecore_fleece_FLValue_asString(JNIEnv *env, jclass clazz, jlong jvalue) {
     FLString str = FLValue_AsString((FLValue) jvalue);
     const char* newBytes;
-    size_t len = UTF8ToModifiedUTF8((const char*)str.buf, &newBytes, str.size);
-    if(len == 0) {
+    ssize_t len = UTF8ToModifiedUTF8((const char*)str.buf, &newBytes, str.size);
+    if(len == -1) {
         jclass c = env->FindClass("java/lang/OutOfMemoryError");
         env->ThrowNew(c, "Out of memory in FLValue_AsString");
         return nullptr;
     }
 
-    jstring retVal = toJString(env, {newBytes, len});
-    if(len != str.size) {
-        free((void *)newBytes);
+    C4Slice endStr {str.buf, str.size};
+    if(newBytes != nullptr) {
+        endStr.buf = newBytes;
+        endStr.size = len;
     }
+
+    jstring retVal = toJString(env, endStr);
+    free((void *)newBytes);
 
     return retVal;
 }
--- a/Java/jni/native_glue.cc
+++ b/Java/jni/native_glue.cc
@@ -25,13 +25,13 @@ using namespace litecore::jni;
 using namespace std;
 
 namespace litecore { namespace jni {
-        void UTF8ToModifiedUTF8(const char *input, char *output);
+        void UTF8CharToModifiedUTF8(const char *input, char *output);
         void ModifiedUTF8ToUTF8(char* input);
         void ModifiedUTF8CharToUTF8(char* input);
     }
 }
 
-void litecore::jni::UTF8ToModifiedUTF8(const char *input, char *output) {
+void litecore::jni::UTF8CharToModifiedUTF8(const char *input, char *output) {
     char c = input[0];
     char c1 = input[1] & 0x3F;
     char c2 = input[2] & 0x3F;
@@ -50,7 +50,7 @@ void litecore::jni::UTF8ToModifiedUTF8(const char *input, char *output) {
     output[5] = surrogates[1] & 0x3F | 0x80;
 }
 
-size_t litecore::jni::UTF8ToModifiedUTF8(const char* input, const char** output, size_t len){
+ssize_t litecore::jni::UTF8ToModifiedUTF8(const char* input, const char** output, size_t len){
     // https://github.com/android-ndk/ndk/issues/283
     size_t extraBytes = 0;
     const auto unsignedInput = (const uint8_t *)input;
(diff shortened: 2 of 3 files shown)


snej:
Please use descriptive titles for PRs! If you're feeling stuck you can just copy/paste the title of the bug with "Fixed:" in front of it...

Java/jni/native_glue.cc
```diff
@@ -68,12 +70,15 @@ namespace litecore {
...
+                _env = env;
+
                 const char *cstr = env->GetStringUTFChars(js, &isCopy);
                 if (!cstr)
                     return; // Would it be better to throw an exception?
                 _slice = slice(cstr);
-                _jstr = js;
-                _env = env;
+
+                fixBrokenModifiedUTF8(cstr);
```
snej: Should we #ifdef this to make it Android-only?

Java/jni/native_glue.cc
```diff
@@ -84,6 +89,68 @@ namespace litecore {
...
+            // https://github.com/android-ndk/ndk/issues/283
+            size_t len = strlen(input);
+            size_t newLen = len;
+            queue<uint8_t*> modified;
+            const auto unsignedInput = (const uint8_t *)input;
+
+            // Need to figure out the actual length since each
+            // 4-bytes of real UTF-8 is 6 bytes of modified UTF-8
+            // but convert the bytes while we are at it
+            for(size_t i = 0; i < len; i++) {
```
snej: This function scans through the string twice. You can eliminate the `strlen` call by just changing the middle clause of the `for` to `unsignedInput[i] != 0`.

(I'm being picky about performance because it looks like this gets called on every single Java string passed into LiteCore.)

Java/jni/native_glue.cc
```diff
@@ -84,6 +89,68 @@ namespace litecore {
...
+                    // 0xF0 and above marks 4-byte sequences
+                    newLen += 2;
+                    int unicodePoint = (unsignedInput[i]-240<<18) +
+                            (unsignedInput[i+1]-128<<12) +
+                            (unsignedInput[i+2]-128<<6) +
+                            (unsignedInput[i+3]-128) - 0x10000;
+                    uint16_t surrogates[2] { (uint16_t)0xD800 | (uint16_t)(unicodePoint >> 10),
+                                             (uint16_t)0xDC00 | (uint16_t)(unicodePoint & 0x3FF) };
+
+                    modified.emplace({
```
snej: I think there's a use-after-free bug here. This call implicitly creates a stack-based C array of uint8_t, then adds its address to `modified`. But the C array is invalidated at the end of this `for` loop iteration, so `modified` ends up being full of pointers to garbage.

Java/jni/native_glue.cc
```diff
@@ -84,6 +89,68 @@ namespace litecore {
...
+                }
+            }
+
+            if(len == newLen) {
+                // No modifications necessary
+                return;
+            }
+
+            // Will be freed by destructor
+            char* newBytes = (char *)malloc(newLen);
```
snej: Check for failure ... on some platforms `malloc` can return `nullptr`.

Java/jni/native_glue.cc
borrrden: Emplacement should do an std::move I thought?

snej:
I think there's a simpler way to do this without having to buffer substrings to be inserted later:

1. Scan the string counting the number of bytes >= 0xF0. (This also gives you the string length.)
2. If the count is zero, return.
3. Allocate a new buffer whose size is the old length plus 2 x the count.
4. Copy bytes from the string to the new buffer using `src` and `dst` pointers; when `*src >= 0xF0` do the translation, appending the translated sequence to `dst`.

Java/jni/native_glue.cc
snej: Sure, but it's moving a _pointer_. If you wanted to copy the bytes, you'd need to use a `queue<string>` or something like that.

Java/jni/native_glue.cc
borrrden: If possible, I guess.  I am not sure if this works on Java desktop or not so if we support that someday it would need to be rechecked (from the linked bug it seems that the NDK is just lagging behind and that desktop JRE fixed this issue, but my lack of Java expertise makes me unsure).

Java/jni/native_glue.cc
borrrden: Oh that's correct yeah, newb mistake on my part.

borrrden:
Nice, that also takes care of the use after free because there will be nothing to use anymore!

Java/jni/native_glue.cc
```diff
@@ -84,6 +92,74 @@ namespace litecore {
...
+                if(unsignedInput[i] >= 0xF0) {
+                    // 0xF0 and above marks 4-byte sequences
+                    newStrLen += 2;
+
+                    // 3 + 1 from next loop iteration = 4 bytes advance
+                    i += 3;
+                }
+
+                newStrLen++;
+                inputByteCount++;
```
snej: Isn't inputByteCount the same as i?

snej: Oh wait, it's not, because inputByteCount doesn't include 3 of the bytes from the 4-byte sequences.
I think that means the second `for` loop will go wrong...

Java/jni/native_glue.cc
```diff
@@ -84,6 +92,74 @@ namespace litecore {
...
+
+            if(newStrLen == inputByteCount) {
+                // No modifications necessary
+                return;
+            }
+
+            // Will be freed by destructor
+            char* newBytes = (char *)malloc(newStrLen);
+            if(newBytes == nullptr) {
+                throw bad_alloc();
```
snej: Double-check that jstringslice is always constructed inside a `try...catch` block, since the prior code never threw exceptions

borrrden:
Ok no more exception throwing at the C++ level.  If the malloc fails, I use the JNI to throw a Java out of memory error instead.  I added in a modified UTF-8 -> UTF-8 conversion process but it's slightly awkward.  Suggestions welcome on how to make it better but it scans for modified UTF-8 surrogate encodings and rewrites them in place.  Since this makes things shorter, it then scans again for UTF-8 4 byte entries and shifts the following bytes backwards in place.  Would a copy actually be better here?  I wonder...

I added tests in the Android repo for both decoding UTF-8 fleece (as opposed to modified UTF-8 fleece that Java would produce) and round tripping emoji.

Java/jni/native_fleece.cc
```diff
@@ -339,7 +340,20 @@ Java_com_couchbase_litecore_fleece_FLValue_asDouble(JNIEnv *env, jclass clazz, j
 JNIEXPORT jstring JNICALL
 Java_com_couchbase_litecore_fleece_FLValue_asString(JNIEnv *env, jclass clazz, jlong jvalue) {
     FLString str = FLValue_AsString((FLValue) jvalue);
-    return toJString(env, {str.buf, str.size});
+    const char* newBytes;
+    size_t len = UTF8ToModifiedUTF8((const char*)str.buf, &newBytes, str.size);
+    if(len == 0) {
```
snej: What if it's an empty string? Then returning 0 is success, not failure.
(Make sure the unit tests include an empty string!)

Java/jni/native_fleece.cc
```diff
@@ -339,7 +340,20 @@ Java_com_couchbase_litecore_fleece_FLValue_asDouble(JNIEnv *env, jclass clazz, j
...
+        jclass c = env->FindClass("java/lang/OutOfMemoryError");
+        env->ThrowNew(c, "Out of memory in FLValue_AsString");
+        return nullptr;
+    }
+
+    jstring retVal = toJString(env, {newBytes, len});
+    if(len != str.size) {
```
snej: This makes the assumption that if`newBytes` is allocated, `len` will be different from `str.size`, which isn't obvious. I think it'd be safer to initialize `newBytes` to nullptr, then you could just test for that; in fact you don't even need to because `free` allows a null parameter.

Java/jni/native_glue.cc
```diff
@@ -17,9 +17,134 @@
...
+using namespace std;
+
+namespace litecore { namespace jni {
+        void UTF8ToModifiedUTF8(const char *input, char *output);
+        void ModifiedUTF8ToUTF8(char* input);
+        void ModifiedUTF8CharToUTF8(char* input);
+    }
+}
+
+void litecore::jni::UTF8ToModifiedUTF8(const char *input, char *output) {
```
snej: This confused me because it has the same name as the public function, but operates on only one character (which isn't obvious from the signature.) How about renaming it `UTF8CharToModifiedUTF8`?

Java/jni/native_glue.cc
```diff
@@ -17,9 +17,134 @@
...
+        // No modifications necessary
+        *output = input;
+        return len;
+    }
+
+    size_t newStrLen = len + extraBytes;
+    char* newBytes = (char *)malloc(newStrLen);
+    if(newBytes == nullptr) {
+        *output = nullptr;
+        return 0;
```
snej: 0 is a legal return value (for an empty input string.) Maybe return -1 on error (and change type to `ssize_t`?)

Java/jni/native_glue.cc
```diff
@@ -68,12 +193,15 @@ namespace litecore {
             assert(env != nullptr);
             if (js != nullptr) {
                 jboolean isCopy;
-                const char *cstr = env->GetStringUTFChars(js, &isCopy);
+                _jstr = js;
+                _env = env;
+
+                // Can we assume that a copy is always made here for Android?  It seems like the case.
+                char *cstr = (char *)env->GetStringUTFChars(js, &isCopy);
```
snej: This is pretty dangerous if the string _isn't_ copied! Add an assertion of some kind that `isCopy` is true.

Issue #565, opened 2018-09-12
Blob causing CBL 2.1 to crash during pull replication


HodGreeley:
CBL 2.1.0 crashes during replication of a doc with a blob attached.  The same document without the blob appears to work.  The crash does not happen with CBL 2.0.0.

Steps to repro:

- Create bucket `debug`
- Load provided data in CBS
- Run SG, Android app with simple replication
- Seg fault in libc

Environment:
Mac, Couchbase Server 6.0.0 beta, Pixel AVD, Android API 25, Sync Gateway 2.0.0

I believe I've seen the same crash on other devices, but have not reverified it.

App to repro crash: https://github.com/HodGreeley/Zebra-Inventory (under android).

CBS data (from cbbackup): [cbbackup.tar.gz](https://github.com/couchbase/couchbase-lite-core/files/2384677/cbbackup.tar.gz)

Restore with, e.g., `cbrestore /Users/hod/work/debug/ http://localhost:8091 -u Administrator -p Password -v -x rehash=1`

Sync Gateway config: [sg-debug.json.zip](https://github.com/couchbase/couchbase-lite-core/files/2377166/sg-debug.json.zip)

Logcat output:
```
09-12 16:17:52.112 18521-18521/? I/art: Not late-enabling -Xcheck:jni (already on)
09-12 16:17:52.113 18521-18521/? W/art: Unexpected CPU variant for X86 using defaults: x86
09-12 16:17:52.137 18521-18527/? E/art: Failed sending reply to debugger: Broken pipe
09-12 16:17:52.137 18521-18527/? I/art: Debugger is no longer active
    Starting a blocking GC Instrumentation
09-12 16:17:52.325 18521-18521/? W/ActivityThread: Application com.couchbase.mobile.zebra is waiting for the debugger on port 8100...
09-12 16:17:52.326 18521-18521/? I/System.out: Sending WAIT chunk
09-12 16:17:54.161 18521-18527/com.couchbase.mobile.zebra I/art: Debugger is active
09-12 16:17:54.340 18521-18521/com.couchbase.mobile.zebra I/System.out: Debugger has connected
09-12 16:17:54.341 18521-18521/com.couchbase.mobile.zebra I/System.out: waiting for debugger to settle...
09-12 16:17:54.543 18521-18521/com.couchbase.mobile.zebra I/System.out: waiting for debugger to settle...
[...]

PR #570, opened 2018-09-26
Handle blob from 2.0.0 when stripping legacy _attachments

pasin:
* In `_attachments`, 2.0 blob key will be in `blob_<number>` format, while 2.1 blob key will be in `blob_/<property-pointer>` format.

* In LegacyAttachments.encodeStrippingOldMetaProperties(root, key), detect 2.0 blob key by checking whether the suffix after `blob_` is a digit or not. If 2.0 blob key is detected, remove the attachment and continue.

* Made c4doc_encodeStrippingOldMetaProperties(doc, sk, outError) returns out error.

* When handle incoming rev, check the error when calling c4doc_encodeStrippingOldMetaProperties() and return 500 error code if there is an error occurs. Similar thing was done in RESTListener+Handlers.cc

* Updated c4DocumentTest

#565

### DIFF
--- a/C/tests/c4DocumentTest.cc
+++ b/C/tests/c4DocumentTest.cc
@@ -696,7 +696,7 @@ N_WAY_TEST_CASE_METHOD(C4Test, "Document Legacy Properties 2", "[Database][C]") 
     auto sk = c4db_getFLSharedKeys(db);
     auto dict = json2dict(db, "{_id:'foo', _rev:'1-2345', x:17}");
     CHECK(c4doc_hasOldMetaProperties(dict, sk));
-    alloc_slice stripped = c4doc_encodeStrippingOldMetaProperties(dict, sk);
+    alloc_slice stripped = c4doc_encodeStrippingOldMetaProperties(dict, sk, NULL);
     CHECK(fleece2json(stripped, sk) == "{x:17}");
 }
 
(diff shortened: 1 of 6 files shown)


C/tests/c4DocumentTest.cc
```diff
@@ -710,22 +710,20 @@ N_WAY_TEST_CASE_METHOD(C4Test, "Document Legacy Properties 3", "[Database][C]")
...
+    alloc_slice stripped = c4doc_encodeStrippingOldMetaProperties(dict, sk, NULL);
     CHECK(fleece2json(stripped, sk) == "{_attachments:{oldie:{digest:\"sha1-xVVVVVVVVVVVVVVVVVVVVVVVVVU=\"}},foo:[0,{\"@type\":\"blob\",digest:\"sha1-VVVVVVVVVVVVVVVVVVVVVVVVVVU=\"}]}");
 }
 
 
 N_WAY_TEST_CASE_METHOD(C4Test, "Document Legacy Properties 4", "[Database][C]") {
-    // Check that a translated attachment whose digest is different than its blob (i.e. the
-    // attachment was probably modified by a non-blob-aware system) has its digest transferred to
-    // the blob before being deleted. See #507. (Also, the _attachments property should be deleted.)
+    // Check that the 2.0.0 blob_<number> gets removed:
```
snej: This should be a new test, not a replacement of the existing one.

REST/RESTListener+Handlers.cc
```diff
@@ -287,8 +287,11 @@ namespace litecore { namespace REST {
 
         // Encode body as Fleece (and strip _id and _rev):
         alloc_slice encodedBody = c4doc_encodeStrippingOldMetaProperties(body,
-                                                                         c4db_getFLSharedKeys(db));
-
+                                                                         c4db_getFLSharedKeys(db),
+                                                                         outError);
+        if (!encodedBody && outError && outError->domain != 0 && outError->code != 0)
```
snej: No need to check `outError`; if `encodedBody` is null, that means the call failed. (And even if the caller passed null for `outError`, this method should still return false.)

pasin:
Thanks @snej  for the review. I have made changes accordingly.

pasin:
I fixed an indentation error and rebased. It's ready for review.

PR #573, opened 2018-10-03
Fixed #562: Add try/catch to c4socket_received and c4socket_gotHTTPResponse

### LINKED ISSUE #562: Add try/catch to c4socket_received to prevent crash / abort
From @snej's comment in https://github.com/couchbase/couchbase-lite-ios/issues/2196:

> c4socket_received's implementation doesn't have a try/catch block, so throwing an exception causes an abort/crash ... we should fix that in LiteCore.

From https://github.com/couchbase/couchbase-lite-ios/issues/2197, there could be an exception thrown from `c4socket_gotHTTPResponse` as well. We should consider add try/catch there as well.

pasin:
* Added try/catch to c4socket_received and c4socket_gotHTTPResponse and close socket when catching an exception.
* Exposed factory object on C4SocketImpl.

#562

### DIFF
--- a/Replicator/c4Socket+Internal.hh
+++ b/Replicator/c4Socket+Internal.hh
@@ -42,10 +42,10 @@ namespace litecore { namespace repl {
                      void *nativeHandle_ =nullptr);
 
         ~C4SocketImpl();
-        
-        C4SocketFactory factory() { return _factory; }
 
         void connect() override;
+
+        void closeWithException(const std::exception&);
 
     protected:
         virtual void requestClose(int status, fleece::slice message) override;
--- a/Replicator/c4Socket+Internal.hh
+++ b/Replicator/c4Socket+Internal.hh
@@ -42,6 +42,8 @@ namespace litecore { namespace repl {
                      void *nativeHandle_ =nullptr);
 
         ~C4SocketImpl();
+        
+        C4SocketFactory factory() { return _factory; }
 
         void connect() override;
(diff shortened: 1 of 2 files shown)


pasin:
I did some tests yesterday and closing the socket after getting an exception on c4socket_received and c4socket_gotHTTPResponse resulted to different error codes. Especially the error from c4socket_gotHTTPResponse will cause the replicator to stopped without retrying. I need to investigate more.

pasin:
@snej What do you think about making kCodeUnexpectedCondition (1010) a transient error?

snej:
That's 1011, not 1010. Seems equivalent to an HTTP 500 Internal Server Error ... do we treat that as transient?

Do you know where in the code that status is set?

pasin:
Yes, we made 500 error as a transient error (see [here](https://github.com/couchbase/couchbase-lite-core/blob/cb9014da72c489cbc6705be55cde9f4f90aab439/C/c4Base.cc#L233)).

We set the kCodeUnexpectedCondition [here](https://github.com/couchbase/couchbase-lite-core/pull/573/commits/59ceb1a0a727c3b13b5d4c9601016bf88bf47336#diff-9853624ce47071d7413c3a3b83c5eb3cR129).

pasin:
Anyway I'm merging this PR now. I will have another commit to make kCodeUnexpectedCondition a transient error.

PR #577, opened 2018-10-05
Fixed log-rotation deleting non-log files

### LINKED ISSUE #576: Log pruning can delete non-log files/dirs in same parent dir
LiteCore's log-rotation code can accidentally delete non-log items that happen to be in the same directory as the current log file, as long as those items are either (a) a directory, (b) a non-readable file, or (c) an empty file.

I just had C++Tests throw an exception trying to delete some directory from my temp folder — rmdir failed because the dir wasn't empty. This dir doesn't even belong to LiteCore. What's happening is:

* The C++Tests put LiteCore logs in the system temp directory ($TMPDIR).
* `purgeOldLogs()` looks at every item in that directory to decide if it's a log file
* It reads the first 4 bytes of the file to see if they match the log-file magic number; if so, that file is added to the candidate set for deletion.
* Unfortunately there is no error checking, so if the file can't be opened, _or is a directory_, nothing gets read into the buffer and its contents are garbage. 
* And since this is in a loop, what appears to be happening is that the same location in the stack is used every time, and once a log file is found the that location now contains the magic number ... so from then on until another file is successfully read, any unreadable items are mistakenly thought to be log files.

snej:
purgeOldLogs() would get fooled if it couldn’t open a file in the log directory, and might assume it was a log file, then try to delete it. Bad news if logging is going to the system temp directory with lots of unrelated files belonging to other processes in it…

Fixes #576

### DIFF
--- a/LiteCore/Support/Logging.cc
+++ b/LiteCore/Support/Logging.cc
@@ -60,20 +60,21 @@ namespace litecore {
     static mutex sLogMutex;
     static const int kNumLogFiles = 10;
 
+
     static void purgeOldLogs(const string& filePath)
     {
         FilePath filePathObj(filePath);
         FilePath logDir = filePathObj.dir();
+        string ext = filePathObj.extension();
         multimap<time_t, FilePath> logFiles;
-        const uint32_t magicNumber = *(uint32_t*)LogEncoder::kMagicNumber;
-        logDir.forEachFile([&logFiles, &magicNumber](const FilePath& f)
-        {
-            char magicBuffer[4];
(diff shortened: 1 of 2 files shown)


snej:
I was leery of having this code try to open every item in the directory (the macOS temp directory has hundreds of items in it) so I made it filter by the file extension of the log file being created.

borrrden:
Will this work with a log that has no extension?

snej:
Yes, if the log file has no extension, the loop examines only other files with no extension. (Which won't only be other log files, but it's better than nothing...)

snej:
Nice addition to the unit test!

PR #579, opened 2018-10-09
Make alternate tempDirectory based on move target

pasin:
Made alternate of the tempDirectory based on moving (or renaming) target instead of source.

### DIFF
--- a/LiteCore/Database/PrebuiltCopier.cc
+++ b/LiteCore/Database/PrebuiltCopier.cc
@@ -42,8 +42,8 @@ namespace litecore {
         
         FilePath backupPath;
         Log("Copying prebuilt database from %s to %s", from.path().c_str(), to.path().c_str());
-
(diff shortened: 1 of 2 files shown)


borrrden:
Oops that was a silly mistake of me.  Has this been tested?  It looks ok to me but I don't have the means to test I think.  The Nexus 5 doesn't seem to have an SD card slot that I can find.

pasin:
I tested with using Nexus 5x that comes with an external storage inside. It took sometime to know that I need to explicitly ask for the read/write permission using code (settings the permissions in the AndroidManifest file is not enough.

pasin:
Weird that from this [page](https://developer.android.com/training/data-storage/files#ExternalStoragePermissions ), I just need to ask for the WRITE permission in the AndroidManifest file. Not sure why I need to ask using Java Code as well.

PR #581, opened 2018-10-10
Fixed #575  Build a more robust tempDirectory call

### LINKED ISSUE #575: Android: copying prepackaged db and upgrading db may not work with external storage
The logic to copy prepackaged db or upgrade db involves moving files (via rename) from temp directory (system storage) and a target directory which could be an external storage (e.g. SD Card). The rename will fail with the EXDEV error code if that condition happens.

pasin:
* It will require a "hint" directory, and check that the default temp directory is both on the same device as the hint, and also writeable.

* Made `/data/local/tmp` default temp directory for Android.


pasin:
@borrrden I rebased with master and push the fix. Can you review and check if the fix and parentDir() also work on Windows? Feel free to fix and add some windows tests for parentDir(). Thanks!

Issue #592, opened 2018-10-26
Since " Predictive query implementation" build broken on Linux


Dushistov:
16e58345466d1abe8f03e473dc0b01110aa64f21 cause build failure on Linux machine:

```
$ git submodule update --init --recursive && make -k -j33 -C build_cmake/unix/
...
vendor/fleece/API/fleece/Fleece+CoreFoundation.h:20:10: fatal error: 'CoreFoundation/CFBase.h' file not found
#include <CoreFoundation/CFBase.h>
         ^~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```

b421761eb24259c3350f177fa5b3c63e68733bc3 builds without any problems.

snej:
What's the full error message? I.e. what file `#include`d `Fleece+CoreFoundation.h`? 

I've searched the source, and the only includes of that header are either from Objective-C code or guarded with `#ifdef __OBJC__`, so they should be ignored in a Linux build.

Dushistov:
I did `git pull` and got more errors.
To fix build of a36949f1d648287abb6fb5f2180d1b1128e6cab3 I add `#include <time.h>`

Because of `Fleece/Support/ParseDate.cc` uses `gmtime_r` without proper include.

```diff
vendor/fleece $ git diff
diff --git a/Fleece/Support/ParseDate.cc b/Fleece/Support/ParseDate.cc
index 53870fd..cb89acf 100644
--- a/Fleece/Support/ParseDate.cc
+++ b/Fleece/Support/ParseDate.cc
@@ -69,6 +69,7 @@
 #include <stdarg.h>
 #include <ctype.h>
 #include <math.h>
+#include <time.h>
 
 typedef uint8_t u8;
 typedef int64_t sqlite3_int64;

```

After that I ran ./build_cmake/scripts/build_unix.sh

```
In file included from /home/evgeniy/projects/study/couch/couchbase-lite-core/C/tests/c4PredictiveQueryTest+CoreML.mm:19:
In file included from /home/evgeniy/projects/study/couch/couchbase-lite-core/C/include/c4PredictiveQuery.h:9:
In file included from /home/evgeniy/projects/study/couch/couchbase-lite-core/vendor/fleece/API/fleece/Fleece.h:1058:
/home/evgeniy/projects/study/couch/couchbase-lite-core/vendor/fleece/API/fleece/Fleece+CoreFoundation.h:20:10: fatal error: 
[...]

PR #594, opened 2018-10-29
use travis for testing linux build on Ubuntu 14.04.5 LTS

Dushistov:
- fix build_cmake/scripts/build_unix.sh to return not 0 on error

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,22 @@
+dist: trusty
+sudo: false
+language: cpp
+
+addons:
+  apt:
+    sources:
+      - llvm-toolchain-trusty-7
+      - ubuntu-toolchain-r-test
+    packages:
+      - clang-7
+      - libc++-7-dev
+      - libc++abi-7-dev
+before_script:
+  - uname -a
+  - cmake --version
+  - export PATH=/usr/lib/llvm-7/bin:$PATH
+  - mkdir -p build_cmake/unix/tlm/deps
+  - ln -s /usr/lib/llvm-7 build_cmake/unix/tlm/deps/libcxx.exploded
+  - clang --version
+script:
+  - ./build_cmake/scripts/build_unix.sh
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -128,7 +128,7 @@ if("${CMAKE_SYSTEM_NAME}" STREQUAL "Linux")
     find_path(LIBICU_INCLUDE unicode/ucol.h
         HINTS "${CMAKE_BINARY_DIR}/tlm/deps/icu4c.exploded"
         PATH_SUFFIXES include)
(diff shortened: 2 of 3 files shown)


Dushistov:
After fix of #592 build stills failed, at now because of a36949f1d648287abb6fb5f2180d1b1128e6cab3 :

```c++
diff --git a/Replicator/tests/ReplicatorLoopbackTest.cc b/Replicator/tests/ReplicatorLoopbackTest.cc
index 548ef3a6..fa8dfdc2 100644
--- a/Replicator/tests/ReplicatorLoopbackTest.cc
+++ b/Replicator/tests/ReplicatorLoopbackTest.cc
@@ -12,6 +12,7 @@
 #include "Database.hh"
 #include "PrebuiltCopier.hh"
 #include <chrono>
+#include "betterassert.hh"
 
 using namespace litecore::actor;
 
@@ -778,7 +779,8 @@ TEST_CASE_METHOD(ReplicatorLoopbackTest, "Push Validation Failure", "[Push]") {
     auto pullOptions = Replicator::Options::passive();
     atomic<int> validationCount {0};
     pullOptions.callbackContext = &validationCount;
-    pullOptions.pullValidator = [](FLString docID, FLDict body, void *context)->bool {
+    pullOptions.pullValidator = [](FLString docID, C4RevisionFlags flags, FLDict body, void *context)->bool {
+        assert_always(flags == 0);      // can't use CHECK on a bg thread
         ++(*(atomic<int>*)context);
         return (Dict(body)["birthday"].asstring() < "1993");
     };
``` 

looks like library and test code build with `NDEBUG` and without `NDEBUG` correspondingly. 

So I suggest use travis CI to check linux build, it is free for public repositories,
but you need to login to https://travis-ci.org .

In this PR I add build against ubuntu 14 lts, with clang-7.

It is failed because of a36949f1d648287abb6fb5f2180d1b1128e6cab3,
you can see full log here:

https://api.travis-ci.org/v3/job/448020342/log.txt

snej:
I fixed the assert_always issue in Fleece commit ef619cff yesterday; sorry about that.

snej:
FYI, we have our own internal CI using Jenkins, including a [build for Android](http://mobile.jenkins.couchbase.com/view/Core_Based_Builds/job/couchbase-lite-core-android/), but no Ubuntu builds. It took us a while to realize the build was broken, because a typo in the Jenkins config was causing the emails to fall on the floor.

Having Travis too seems like a good idea, esp. since it seems to turn up some build problems that don't happen on Android, even though Android is Linux (right?)

snej:
@borrrden, the changes look good to me, but you're more in tune with Linux and CMake, so could you double-check, and merge if it's ok? Then we can add our own Travis build.

PR #614, opened 2018-11-16
c4log_getBuildInfo: Return correct Git ID in EE build

### LINKED ISSUE #613: Jenkins EE builds don't identify their Git commit correctly
In EE builds from Jenkins, the Git commit ID in the output of `c4log_getBuildInfo()` is not meaningful. It's not an actual commit ID, rather [IIRC] a digest generated from the commit IDs of both the couchbase-lite-core and couchbase-lite-core-ee repos.

This means that one can't look at unit test logs and identify what source code they came from. For example, in [this failed build's logs](http://mobile.jenkins.couchbase.com/view/Core_Based_Builds/job/couchbase-lite-core-linux/880/consoleText):
```
19:55:01.969347| This is LiteCore EE build number 1545 from commit 00337b2c ... short version  (1545)
```
There is no commit `00337b2c` anywhere.

I know the computed digest is needed to work around some issue with how we trigger Jenkins builds, but IMO it shouldn't be returned from `c4log_getBuildInfo`. Or if it is, the actual Git commit ID(s) should be returned too.

snej:
In EE builds, return the actual Git commit ID as well as the build ID ($VERSION).

Tested by modifying the script to set VERSION and BLD_NUM, then verifying the unit tests log the right info.

Fixes #613

### DIFF
--- a/C/c4Base.cc
+++ b/C/c4Base.cc
@@ -51,21 +51,28 @@ extern "C" {
 
 // LCOV_EXCL_START
 static string getBuildInfo() {
+    static string commit;
 #ifdef COUCHBASE_ENTERPRISE
+    if(commit.empty()) {
+        commit = format("%.8s+%.8s", GitCommitEE, GitCommit);
+    }
     static const char *ee = "EE ";
 #else
+    if(commit.empty()) {
+        commit = format("%.8s", GitCommit);
+    }
     static const char *ee = "";
 #endif
 #if LiteCoreOfficial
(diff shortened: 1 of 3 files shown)


borrrden:
The "actual commit ID" is a bit more complicated since there are two commits that you need to know about (which EE was built and which CE was built to make a given EE build since there is no submodule relationship here).  In .NET I do a SHA+SHA style entry (e.g. 123456+abcdef).  Perhaps that can also be done here?

borrrden:
@snej I had a crack at it.  Let me know what you think.

PR #617, opened 2018-11-20
Support for encoding java.lang.Short numeric values

dtchepak:
I was going to ask this as an issue but thought it might be easier as a PR with a diff. 

I got the following exception saving data containing `Short` values using CBL 2.1.0 on Android:

    java.lang.ClassCastException: java.lang.Short cannot be cast to java.lang.Float

I think it is just a missing case in these two encoders. I've haven't been able to test it properly and may have completely misdiagnosed the issue, in which case please just close this. 😄 

(Maybe also consider adding `Byte`, `BigInteger`, `BigDecimal`?)


pasin:
Thanks for the PR.

dtchepak:
Thanks @pasin. Did you want me to send one through for `Byte`, `BigInteger`, and/or `BigDecimal`?

PR #618, opened 2018-11-27
Update JNI Binding base on the latest change in LiteCore

Sandychuang8:
- c4db_free is now not returning any value 
- doc constructor takes alloc_slice as fleece data instead of FLSLice 
- documentEndedCallback accepts two more parameters (C4HeapString revID, C4RevisionFlags flags) and parameter docID is C4HeapString instead of C4String

### DIFF
--- a/Java/jni/native_c4database.cc
+++ b/Java/jni/native_c4database.cc
@@ -82,11 +82,11 @@ Java_com_couchbase_litecore_C4Database_copy(JNIEnv *env, jclass clazz,
 /*
  * Class:     com_couchbase_litecore_C4Database
  * Method:    free
- * Signature: (J)Z
+ * Signature: (J)
  */
 JNIEXPORT jboolean JNICALL
 Java_com_couchbase_litecore_C4Database_free(JNIEnv *env, jclass clazz, jlong jdb) {
-    return (jboolean) c4db_free((C4Database *) jdb);
+    c4db_free((C4Database *) jdb);
 }
 
 /*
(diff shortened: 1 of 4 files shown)


pasin:
@Sandychuang8 Can you change the PR's title to match the actual PR?

Java/jni/native_c4database.cc
```diff
@@ -82,11 +82,11 @@ Java_com_couchbase_litecore_C4Database_copy(JNIEnv *env, jclass clazz,
 /*
  * Class:     com_couchbase_litecore_C4Database
  * Method:    free
- * Signature: (J)Z
+ * Signature: (J)
```
pasin: 1. Please fix the signature [here](https://github.com/couchbase/couchbase-lite-core/blob/master/Java/src/com/couchbase/litecore/C4Database.java#L339) to return void as well.
2. To return void, the Signature should be `(J)V`

Java/jni/native_c4database.cc
```diff
@@ -82,7 +82,7 @@ Java_com_couchbase_litecore_C4Database_copy(JNIEnv *env, jclass clazz,
 /*
  * Class:     com_couchbase_litecore_C4Database
  * Method:    free
- * Signature: (J)
+ * Signature: (J)V
```
pasin: Please fix the signature below to return void as well.

https://github.com/couchbase/couchbase-lite-core/blob/master/Java/src/com/couchbase/litecore/C4Database.java#L339

pasin:
@Sandychuang8 Can you fix the signature to return void on this method as well so that we can merge the PR?

https://github.com/couchbase/couchbase-lite-core/blob/master/Java/src/com/couchbase/litecore/C4Database.java#L339

PR #620, opened 2018-12-01
Fix KeyStore.now() function for Android

pasin:
The time() function doesn’t return the correct time since Epoch for android. Used std::chrono::system_clock::now().time_since_epoch() instead for Android.


borrrden:
Did you confirm that this return milliseconds?  If this works I'd prefer this for all platforms.  We should stop using these old C methods if there is a C++11 equivalent that is better.

pasin:
Yes.

PR #622, opened 2018-12-05
New Couchbase Lite logging system

borrrden:
In order to support the new console / file / custom logging system in Couchbase Lite, there are some changes in LiteCore:

- Since there are many options for binary files now, a class was created to store them
- c4slog doesn't trigger callback to avoid needless marshalling of strings (these log messages originate in Couchbase Lite, and so will have already been sent to the other loggers)
- Binary / text encoding togglable
- Rotation of logs enabled
- JNI implemented for wiring into Java


LiteCore/Support/Logging.cc
```diff
@@ -55,36 +55,131 @@ namespace litecore {
...
-        FilePath logDir = filePathObj.dir();
-        string ext = filePathObj.extension();
+        FilePath logDir(sLogDirectory);
         multimap<time_t, FilePath> logFiles;
+        const char* levelStr = kLevelNames[(int)level];
+
         logDir.forEachFile([&](const FilePath& f) {
-            if (f.extension() == ext) {
-                char magicBuffer[4] = {};
+            if (f.fileName().find(levelStr) != string::npos) {
```
snej: Why not just use a custom extension like `.cbllog`? Then you wouldn't have to be paranoid and check the magic number in the file.

LiteCore/Support/Logging.hh
```diff
@@ -145,7 +176,7 @@ extern LogDomain DBLog, QueryLog, SyncLog, &ActorLog;
 #ifdef _MSC_VER
 #define LogToAt(DOMAIN, LEVEL, FMT, ...) \
     {if (_usuallyFalse((DOMAIN).willLog(LogLevel::LEVEL))) \
-        (DOMAIN).log(LogLevel::LEVEL, FMT, ##__VA_ARGS__);}
+        (DOMAIN).log(LogLevel::LEVEL, true, FMT, ##__VA_ARGS__);}
```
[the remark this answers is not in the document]
borrrden: Probably.  Somehow this area got git reverted several times.  It should be true everywhere.  I'll fix that.

LiteCore/Support/LogEncoder.hh
```diff
@@ -36,20 +37,22 @@ namespace litecore {
     public:
         static const uint8_t kMagicNumber[4];
 
-        LogEncoder(std::ostream &out);
+        LogEncoder(std::ostream &out, LogLevel level);
```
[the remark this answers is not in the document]
borrrden: I thought that was exactly the plan (Having a tool that interleaves all of the logs).  Isn't that what was indicated in the PRD review?  If it's unclear we should definitely confirm ASAP.

LiteCore/Support/Logging.cc
borrrden: I would still be paranoid that someone for some reason had inserted a file with the same extension for the lolz.  It would then be deleted and it would be our fault.  Is that overboard?

LiteCore/Support/LogEncoder.cc
```diff
@@ -47,9 +47,10 @@ namespace litecore {
     static const uint64_t kSaveInterval = 1 * kTicksPerSec;
 
 
-    LogEncoder::LogEncoder(ostream &out)
-    :_out(out)
-    ,_flushTimer(bind(&LogEncoder::performScheduledFlush, this))
+    LogEncoder::LogEncoder(ostream &out, LogLevel level)
+    :_out(out),
+    _flushTimer(bind(&LogEncoder::performScheduledFlush, this)),
```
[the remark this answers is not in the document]
borrrden: I like that reasoning.  I will use that method for C++ starting now.

LiteCore/Support/Logging.cc
snej: Yeah, that's overboard ;-)

LiteCore/Support/LogEncoder.hh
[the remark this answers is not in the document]
snej: When separate log files were first proposed (back in 1.x I think), I was all "WTF", but the rejoinder was that each file would include the higher levels, and this wouldn't be much overhead since there are pretty few Warning or Error logs. (But I guess with a Verbose log file, including all the Info logs _would_ be expensive.) I hadn't realized that this got taken out of the current proposal.

I worry that if the app crashes, the log levels might not be in sync. If the Info file got flushed more recently than the Warning file (because it's written more often), then afterwards you'd be missing some warnings that came before the last info messages...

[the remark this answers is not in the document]
borrrden: I don't think I have ever heard about the proposal you mentioned to include all of that in each successive level.  If the app crashes, then wouldn't the `atexit` handler flush all the logs?  Or does that not run for crashes?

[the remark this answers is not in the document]
snej: Yeah, you're right, I forgot about the atexit handler.

LiteCore/Support/Logging.hh
```diff
@@ -87,8 +118,8 @@ public:
 
     bool willLog(LogLevel lv) const                 {return _effectiveLevel <= lv;}
 
-    void log(LogLevel level, const char *fmt, ...) __printflike(3, 4);
-    void vlog(LogLevel level, const char *fmt, va_list);
+    void log(LogLevel level, bool callback, const char *fmt, ...) __printflike(4, 5);
```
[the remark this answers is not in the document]
borrrden: That worked out nicely that way!

LiteCore/Support/Logging.cc
```diff
@@ -294,24 +406,66 @@ namespace litecore {
...
+        if(sLogEncoder[(int)level]) {
+            sLogEncoder[(int)level]->vlog(domain, (LogEncoder::ObjectRef)objRef, fmt, args);
+        } else if(sFileOut[(int)level]) {
+            static char formatBuffer[2048];
+            size_t n = 0;
+            LogDecoder::writeTimestamp(LogDecoder::now(), *sFileOut[(int)level]);
+            LogDecoder::writeHeader(kLevels[(int)level], domain, *sFileOut[(int)level]);
+            if (objRef)
+                n = snprintf(formatBuffer, sizeof(formatBuffer), "{%s#%u} ", objName, objRef);
+            vsnprintf(&formatBuffer[n], sizeof(formatBuffer) - n, fmt, args);
```
pasin: Not sure if I'm pointing to the right spot or not. Based on my testing, log format from Objective-C that has %@ doesn't get formatted correctly. Can we use NSString to format the plain text log messages for Apple?

borrrden: Can you give an example of the logging line that does such a thing?  I thought that kind of stuff was built in already.  It seems like it would be following different logic since c4slog already takes a C4Slice which is supposed to be an already formatted UTF-8 string.

pasin: iOS only calls c4vlog with the format that could include `%@` gets passed to the function directly. For callback log, the format will be passed back to the callback directly. Given https://github.com/couchbase/couchbase-lite-core/issues/631, is the right approach is to pass the already-formatted UTF-* string to c4vlog?

pasin: Actually iOS can create a formatted UTF-8 string and use c4vlog(). We don't do a lot of logging from CBL side, probably that is OK.

borrrden: It might even be preferable.  Otherwise you are doing the formatting process three separate times, right?

pasin: No, the formatting will be done twice (before calling c4log and in callback with "%s") instead of once in callback. I will ping you to see if there is a way to do it only once.

borrrden:
@snej I think all outstanding comments have been resolved.  If so, then once you change your review I will rebase and merge.

LiteCore/Support/Logging.cc
pasin: As discussed with @borrrden, iOS will formatted the message before calling c4log so there will be no "%@" to worry about. This will inline with what the other platform is doing and the same thing as what @snej suggested in https://github.com/couchbase/couchbase-lite-ios/issues/2281.

PR #628, opened 2018-12-12
Make sure time zones are set up for N1QL

borrrden:
Note that without https://github.com/couchbaselabs/fleece/pull/39 this PR is somewhat meaningless, so I'll say that this is dependent on that.


snej:
I'm assuming `tzset` is idempotent, so it's not a problem if it's already been called elsewhere?

Wouldn't this be better to do in ParseDate itself (wrapped in `call_once`), since that's the code that needs the time-zone to be set?

PR #635, opened 2018-12-15
Feature/Replication filters

Sandychuang8:
Replication filters: Add validation callback function in JNI and LiteCore related classes
Replication filter: add push filter and pull filter as part of replicator parameters, add replicator filter interface, add a new table to store replicator with replicator context key, remove filter method from replicator listener interface.
Replication filter fix: make sure return the value from validationFunction.
Replication filters fix: JNI methods requires returning j values

### DIFF
--- a/Java/src/com/couchbase/litecore/C4Replicator.java
+++ b/Java/src/com/couchbase/litecore/C4Replicator.java
@@ -85,7 +85,7 @@ public class C4Replicator {
                 pullFilter,
                 options);
         reverseLookupTable.put(handle, this);
-        reverseLookupTable2.put((long)socketFactoryContext, this);
+        reverseLookupTable2.put((long)replicatorContext.hashCode(), this);
     }
 
     C4Replicator(C4Database db, C4Socket openSocket, int push, int pull, byte[] options, Object replicatorContext) throws LiteCoreException {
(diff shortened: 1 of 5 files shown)


Java/src/com/couchbase/litecore/C4Replicator.java
```diff
@@ -61,19 +66,26 @@
...
                 otherLocalDB,
                 push, pull,
                 replicatorContext.hashCode(),
                 framing,
                 socketFactoryContext,
+                pushFilter,
+                pullFilter,
                 options);
         reverseLookupTable.put(handle, this);
+        reverseLookupTable2.put((long)socketFactoryContext, this);
```
pasin: IMO, the original code before this PR is very confusing. It seems like `replicatorContext.hashCode() == socketFactoryContext`. It's hard to follow from the code review where both replicatorContext and socketFactoryContext get used. We really need to clean this up.

As Java_com_couchbase_litecore_C4Replicator_createV2 sets the callback context to the replicatorContext.hashCode(), the correct value to set here is `replicatorContext.hashCode()`.

Java/src/com/couchbase/litecore/C4Replicator.java
```diff
@@ -172,6 +184,21 @@ private static void documentEndedCallback(long handle, boolean pushing,
         }
     }
 
+    private static boolean validationFunction(String docID, int flags, long dict, boolean isPush, long ctx) throws Exception {//
+        Log.e(TAG, "validationFunction() " +
```
pasin: Please don't use Log.e as Log.e means error.

Java/src/com/couchbase/litecore/C4Replicator.java
```diff
@@ -172,6 +184,21 @@ private static void documentEndedCallback(long handle, boolean pushing,
...
+                ", docID -> " + docID +
+                ", flags -> " + flags +
+                ", isPush -> " + isPush);
+         C4Replicator repl = reverseLookupTable2.get(ctx);
+         if(repl != null && repl != null) {
```
pasin: Look like checking the same thing here?

Java/src/com/couchbase/litecore/C4Replicator.java
```diff
@@ -172,6 +184,21 @@ private static void documentEndedCallback(long handle, boolean pushing,
         }
     }
 
+    private static boolean validationFunction(String docID, int flags, long dict, boolean isPush, long ctx) throws Exception {//
```
pasin: Please remove //.

pasin: I don't think the Exception is ever thrown from this method. If it's true, please remove throws Exception.

PR #639, opened 2018-12-18
Switch document ended to be a batched callback

### LINKED ISSUE #636: C4ReplicatorDocumentEndedCallback should be batched
A client app may want to purge docs that have been pushed, and it should do that in one batch operation for performance. So we've decided that replicator document notifications should also be batched, i.e. one call for a set of docs.

This means changing `C4ReplicatorDocumentEndedCallback` into a form that takes a C array.

borrrden:
Fixes #636


borrrden:
I wanted to get a head start on this since after this platforms will require work.

snej:
Oops, I already implemented this yesterday afternoon [mostly] and pushed it this morning :(

borrrden:
No problem, this was a good insurance anyway :)

PR #651, opened 2019-01-03
Update PredictiveIndex to support only-cache and compound index

pasin:
* Allowed to create a PredictiveIndex without properties specified so that the index can be used as a result cache only.

* Allowed to create a compound PredictiveIndex with multiple properties specified.


pasin:
Note: please don't review the PR yet. I will need to add unit tests to the PR.

PR #659, opened 2019-01-15
Add an explicit flush method to DBWorker

### LINKED ISSUE #658: Push deletions rejected with 409
Reproduction steps using DB->DB replication:

1. Create document in DB1
2. Push document to DB2
3. Delete document in DB1
4. Push document to DB2

This results in a null `remoteAncestorID` causing a false conflict somehow.  The end result is that the document is rejected from DB2.

[Logs for test run](https://gist.github.com/borrrden/4b6fcffeb79fab081f49adc04ab596a2)

borrrden:
This allows the replicator to ensure that all revisions have been marked as synced (among other possible future things) before the connection is closed and no more actions are allowed.  Fixes #658

### DIFF
--- a/Replicator/Replicator.cc
+++ b/Replicator/Replicator.cc
@@ -85,6 +85,7 @@ namespace litecore { namespace repl {
 
     void Replicator::_stop() {
         logInfo("Told to stop!");
+        _dbActor->flush();
         _disconnect(websocket::kCodeNormal, {});
     }
 
--- a/Replicator/DBWorker.cc
+++ b/Replicator/DBWorker.cc
@@ -1132,4 +1132,9 @@ namespace litecore { namespace repl {
         return level;
     }
 
+    void DBWorker::flush()
+    {
+        _markRevsSyncedNow();
+    }
+
 } }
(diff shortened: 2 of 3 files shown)


borrrden:
I imagine there are cases where we don't want to call this, so let me know what criteria I can use.  As is, the issue I filed is fixed but I don't know what implications it has for (e.g.) passive replication.

Replicator/Replicator.cc
```diff
@@ -85,6 +85,7 @@ namespace litecore { namespace repl {
 
     void Replicator::_stop() {
         logInfo("Told to stop!");
+        _dbActor->flush();
```
snej: You can't call an actor method like this directly -- you're on the wrong thread/queue. `_markRevsSyncedNow` uses internal state, so it has to run on the DBWorker's actor queue. But that means the call will be asynchronous, which won't achieve the goal of running it before the replicator finishes. A different solution is necessary.

snej:
I think I know what the problem is, and it requires a very different solution, so I'm going to close this PR.

PR #662, opened 2019-01-16
Fix remoteAncestorIDs not getting updated by replicator

### LINKED ISSUE #658: Push deletions rejected with 409

snej:
The Replicator could sometimes stop before its DBWorker had a chance
to call _markRevsSyncedNow and update all docs' remoteAncestorID.

I fixed this in the Batcher class in BLIP-Cpp. Batcher now lets its
parent Actor know that there is work to be done, by enqueuing a
message immediately when the 1st item is added.

I also fixed an issue I saw in Replicator, where it could decide it's
stopped even if it still has events to be processed.

Fixes #658 (I hope)

### DIFF
--- a/Replicator/DBWorker.cc
+++ b/Replicator/DBWorker.cc
@@ -974,11 +974,8 @@ namespace litecore { namespace repl {
     // Insert all the revisions queued for insertion, and sync the ones queued for syncing.
     void DBWorker::_insertRevisionsNow() {
         auto revs = _revsToInsert.pop();
-        if (!revs) {
-            // No insertions scheduled, only syncs, so just do those:
-            _markRevsSyncedNow();
+        if (!revs)
             return;
-        }
 
         logVerbose("Inserting %zu revs:", revs->size());
         Stopwatch st;
--- a/Replicator/Replicator.cc
+++ b/Replicator/Replicator.cc
@@ -173,8 +173,11 @@ namespace litecore { namespace repl {
                 break;
             case Connection::kDisconnected:
             case Connection::kClosed:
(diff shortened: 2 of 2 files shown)


Replicator/DBWorker.cc
```diff
@@ -974,11 +974,8 @@ namespace litecore { namespace repl {
     // Insert all the revisions queued for insertion, and sync the ones queued for syncing.
     void DBWorker::_insertRevisionsNow() {
         auto revs = _revsToInsert.pop();
-        if (!revs) {
-            // No insertions scheduled, only syncs, so just do those:
-            _markRevsSyncedNow();
```
snej: FYI, this isn't related to the bug fix, but I realized this call is obsolete. It used to be there because there was a single trigger for inserting and syncing revs.

borrrden:
Also I can't make comments on the BLIP commit but it seems like if the batcher is full then it will call "enqueue" after "enqueueAfter" resulting in a double enqueue.

snej:
> resulting in a double enqueue

That's why I didn't initially implement it this way. :/ But I looked at what happens in the unit tests, and it doesn't seem like a problem. There are a couple of extra calls, but it's not going to cause any noticeable slowdown.

Issue #663, opened 2019-01-16
Create Cache Only Predictive Index twice will cause the cache table dropped


pasin:
Create Cache Only Predictive Index twice will cause the cache table dropped. This is because this [line](https://github.com/couchbase/couchbase-lite-core/blob/master/LiteCore/Query/SQLiteDataFile%2BIndexes.cc#L97) results to false as there was no actual index created.

PR #664, opened 2019-01-16
Fix predictive cache table getting dropped

pasin:
* The cache-only predictive index doesn’t create an actual index so the index schema doesn’t exist. The `same` flag needs to be reverted when checking whether the index schema with index SQL exists or not for the cache-only predictive index.

* Updated the predictive index unit tests to also test creating index twice.

#663


pasin:
The fix is not correct so I'm taking this PR out. Will submit a new PR with slightly different fix.

PR #666, opened 2019-01-16
Fix predictive cache table getting dropped when creating the same index again

pasin:
* The cache-only predictive index doesn’t create an actual index so the index schema doesn’t exist.

* Updated the schemaExistsWithSQL() function to check if the given sql is empty or not. If the given sql is empty, return true if the scheme doesn’t exist, otherwise return false.

* Updated the predictive index unit tests to also test creating index twice.

#663

### DIFF
--- a/LiteCore/Storage/SQLiteDataFile.cc
+++ b/LiteCore/Storage/SQLiteDataFile.cc
@@ -458,11 +458,13 @@ namespace litecore {
     }
 
 
-    // Returns true if an index/table exists in the database with the given type and SQL schema
+    // Returns true if an index/table exists in the database with the given type and SQL schema OR
+    // Returns true if the given sql is empty and the schema doesn't exist.
     bool SQLiteDataFile::schemaExistsWithSQL(const string &name, const string &type,
                                              const string &tableName, const string &sql) {
         string existingSQL;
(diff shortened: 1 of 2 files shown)


LiteCore/Storage/SQLiteDataFile.cc
```diff
@@ -458,11 +458,13 @@ namespace litecore {
...
 
-    // Returns true if an index/table exists in the database with the given type and SQL schema
+    // Returns true if an index/table exists in the database with the given type and SQL schema OR
+    // Returns true if the given sql is empty and the schema doesn't exist.
     bool SQLiteDataFile::schemaExistsWithSQL(const string &name, const string &type,
                                              const string &tableName, const string &sql) {
         string existingSQL;
-        return getSchema(name, type, tableName, existingSQL) && existingSQL == sql;
+        bool existed = getSchema(name, type, tableName, existingSQL);
+        return sql != "" ? existed && existingSQL == sql : !existed;
```
snej: This expression was hard for me to read -- could you add parens after the `?`, or even better IMO, turn this into an `if / else`?

pasin: Yes, I have change it to if / else.

PR #672, opened 2019-01-25
docs: update the signature of the writeToBinaryFile method

jayahariv:
* update the signature


borrrden:
Please leave me out of stuff like this.  This only needs to be reviewed by one person since it is so tiny.

PR #673, opened 2019-01-25
Fix for incorrect logging refs

### LINKED ISSUE #671: Logs show wrong object refs/classes
I'm seeing log messages (in Xcode running LiteCore unit tests) that have the wrong object class in them, like:
```
14:08:25.211403| [Sync]: {IncomingRev#7} Received revision 'newdoc3' #1-11 (seq '3')
14:08:25.211817| [Sync]: {Repl#8} Received revision 'newdoc4' #1-11 (seq '4')
14:08:25.212090| [Sync]: {Repl#9} Received revision 'newdoc5' #1-11 (seq '5')
14:08:25.212356| [Sync]: {Push#10} Received revision 'newdoc6' #1-11 (seq '6')
14:08:26.725118| [Sync]: {Push#11} Received revision 'newdoc7' #1-11 (seq '7')
14:08:26.725998| [Sync]: {IncomingRev#12} Received revision 'newdoc8' #1-11 (seq '8')
```
The message "`Received revision`..." is only logged by IncomingRev, so all but two of those lines have the wrong class name.

borrrden:
This PR will move the log object refs to a central location so that everything has a consistent view.
This will allow a consistent view of objects between all objects that need them
Fixes #671

### DIFF
--- a/LiteCore/Support/LogEncoder.cc
+++ b/LiteCore/Support/LogEncoder.cc
@@ -61,7 +61,6 @@ namespace litecore {
         _st.reset();
     }
 
-
     LogEncoder::~LogEncoder() {
         flush();
     }
@@ -95,14 +94,13 @@ namespace litecore {
         _writer.write(&_level, sizeof(_level));
         _writeStringToken(domain ? domain : "");
 
-        _writeUVarInt((unsigned)object);
-        if (object != ObjectRef::None) {
-            auto i = _objects.find(unsigned(object));
-            if (i != _objects.end() && i->second != "\x1") {
-                _writer.write(slice(i->second));
-                _writer.write("\0", 1);
-                i->second = string("\x1"); // Remove this from memory, but keep so we can track as "seen"
-            }
+        const auto objRef = (unsigned)object;
+        _writeUVarInt(objRef);
+        if (object != ObjectRef::None && _seenObjects.find(objRef) == _seenObjects.end()) {
+            _seenObjects.insert(objRef);
+            auto i = LogDomain::getObject(objRef);
+            _writer.write(slice(i));
+            _writer.write("\0", 1);
         }
 
         _writeStringToken(format);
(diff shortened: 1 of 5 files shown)


LiteCore/Support/LogEncoder.cc
```diff
@@ -95,14 +94,13 @@ namespace litecore {
...
-            if (i != _objects.end() && i->second != "\x1") {
-                _writer.write(slice(i->second));
-                _writer.write("\0", 1);
-                i->second = string("\x1"); // Remove this from memory, but keep so we can track as "seen"
-            }
+        const auto objRef = (unsigned)object;
+        _writeUVarInt(objRef);
+        if (object != ObjectRef::None && _seenObjects.find(objRef) == _seenObjects.end()) {
+            _seenObjects.insert(objRef);
+            auto i = LogDomain::getObject(objRef);
```
snej: `name` would be a clearer variable name. (I just called it `i` before because it was an iterator)

LiteCore/Support/Logging.cc
```diff
@@ -57,6 +57,8 @@ namespace litecore {
     static LogDomain::Callback_t sCallback = LogDomain::defaultCallback;
     static bool sCallbackPreformatted = false;
     LogLevel LogDomain::sFileMinLevel = LogLevel::None;
+    unsigned LogDomain::slastObjRef {0};
+    map<unsigned, string> LogDomain::sobjNames;
```
snej: Nitpick: Please capitalize as `sObjNames`.

LiteCore/Support/Logging.cc
```diff
@@ -513,33 +512,35 @@ namespace litecore {
         #endif
     }
 
+    string LogDomain::getObject(unsigned ref)
+    {
```
snej: Needs to lock the mutex.
Or is it guaranteed that this is only called when the mutex is already locked?

LiteCore/Support/Logging.cc
borrrden: That was a straight up typo!  I would normally do that and I don't think it's a nit pick.

PR #699, opened 2019-02-14
Use RandomNumber wrapper fn instead of arc4random

snej:


CMakeLists.txt
```diff
@@ -397,6 +397,7 @@ add_library(Support STATIC
     LiteCore/Support/Error_android.cc
     LiteCore/Support/FilePath.cc
     LiteCore/Support/Logging_Stub.cc
+    LiteCore/Support/SecureRandomize.cc
```
borrrden: Without defining _CRYPTO_MBEDTLS in the Support library on non-Apple this will always result in a non-functioning `SecureRandomize`

LiteCore/Support/SecureRandomize.hh
```diff
@@ -43,73 +54,31 @@
...
     #define SECURE_RANDOMIZE_AVAILABLE 1
 
 
 #else
 
-    static inline void SecureRandomize(fleece::slice s) {
-        error::_throw(error::Unimplemented);
+    namespace litecore {
+        static inline void SecureRandomize(fleece::slice s) {
+            error::_throw(error::Unimplemented);
```
borrrden: This won't compile on Windows without adding `#include "Error.hh"`

CMakeLists.txt
snej: It gets added to `CMAKE_CXX_FLAGS` up on line 194; wouldn't that apply when building Support?

borrrden: That's only for Unix actually.  I don't know how it ended up like this but down on line 330 it defines it specifically for LiteCoreStatic on Windows (and also Android).

PR #705, opened 2019-02-20
Fixed stop-word processing in MATCH target strings

### LINKED ISSUE #626: Stop-words with "*" in Match expression cause error
From the [forum](https://forums.couchbase.com/t/fts-query-logical-and-wildcards-and-stop-words-causes-malformed-search-string-exception/19397):

> When I FTS search with a string like so: "anything* AND stopword*", the query execution fails with a SQLite error saying the query was given a "malformed search string".
> 
> Some examples:
> "usfs* AND he*"
> "banana* AND she*"
> "orange* AND i*"
> 
> I'm guessing this is caused by stop words (common English words) being extracted from the search string *before* taking logical operands and wildcards into account, resulting in:
> "usfs* AND *"
> "banana\* AND \*"
> "orange\* AND \*"
> 
> All of which are indeed "malformed search strings". Because I don't need any stemming or NLP features, I just set the index language to null, and the issue went away, so my problem is solved, but I figured I'd make a post as this is some very odd behavior!

snej:
Stop-words should not be removed in the target of a MATCH if immediately followed by a “`*`”, since in that case they’re just prefixes. For example, “`he*`” should not be removed.

Fixes #626

### DIFF
--- a/LiteCore/Query/SQLiteQuery.cc
+++ b/LiteCore/Query/SQLiteQuery.cc
@@ -32,6 +32,10 @@
 #include <sstream>
 #include <iostream>
 
+extern "C" {
+#include "sqlite3_unicodesn_tokenizer.h"        // for unicodesn_tokenizerRunningQuery()
+}
+
 using namespace std;
 using namespace fleece;
 using namespace fleece::impl;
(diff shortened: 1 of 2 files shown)


borrrden:
CppTests fine on MSVC with this branch

PR #716, opened 2019-03-08
Refactor JNI code out of Core

bmeike:
Large, multi-repo project to refactor the Java specific JNI code out of Core

See https://github.com/couchbase/couchbase-lite-android/pull/1875
See https://github.com/couchbaselabs/couchbase-lite-android-ee/pull/33


bmeike:
It is likely that I have broken this file:
couchbase-lite-core/Xcode/LiteCore.xcodeproj/project.pbxproj

I'm not clear on how it is used, so I'm not clear about how to fix it.

snej:
Did you make any changes in Xcode? It looks like a few script related settings in the project were changed, which doesn't seem like something that would happen on its own. But the changes seem harmless if the Java code is indeed being removed, since they won't have any effect.

bmeike:
done.

PR #729, opened 2019-03-29
fix possible crash on empty document

KraemerDEM:
I've encountered a crash when pulling a document that contains a blob with length 0.
This small patch fixes the crash.
Maybe it would be better to not allow blobs with length 0 ?

### DIFF
--- a/Replicator/Puller+DB.cc
+++ b/Replicator/Puller+DB.cc
@@ -324,7 +324,8 @@ namespace litecore { namespace repl {
                         put.revFlags |= kRevKeepBody;
                     } else {
                         // Encode doc body using database's real sharedKeys:
-                        bodyForDB = _db->reEncodeForDatabase(rev->doc);
+                        if (rev->doc)
+                            bodyForDB = _db->reEncodeForDatabase(rev->doc);
                         rev->doc = nullptr;
                         // Preserve rev body as the source of a future delta I may push back:
                         if (bodyForDB.size >= tuning::kMinBodySizeForDelta && !_disableDeltaSupport)


snej:
IIRC, `rev->doc` should never be null. So something else seems to be going wrong here.
Let me check if we have any unit tests with zero-length blobs...

snej:
The tests "Push Attachments" and "Pull Attachments" include a zero-length blob. But it's not the only blob in the doc. I modified "Push Attachments" to use only the zero-length blob, and it still worked.

@KraemerDEM, what's the exact scenario that triggered the crash? Is it reproducible?

KraemerDEM:
The zero-length blobs were saved during an import, I'm not exactly sure how.
The documents look like this
```
{
  "FILE_EXTENSION": ".pdf",
  "FILE_NAME": "test.pdf",
  "FILE_SIZE": "61554",
  "FIRST_PAGE_FILE": {
    "@type": "blob",
    "digest": "sha1-mriFETuy+31Wm5469oBhmeiDLUo="
  },
  "PDF_FILE": "/original/test.pdf",
  "_attachments": {
    "blob_/FIRST_PAGE_FILE": {
      "digest": "sha1-2jmj7l5rSw0yVb/vlWAYkK/YBwk=",
      "length": 0,
      "revpos": 1,
      "stub": true
    }
  },
  "type": "DOKUMENT"
}
```
It seems odd that the FIRST_PAGE_FILE entry doesn't have the same digest as the attachment, is that correct?
When pulling the document, the warning "Failed to strip legacy attachments: error 4/4" is printed.

borrrden:
What do the documents *out of* the database look like? (i.e. inside of Sync Gateway)

KraemerDEM:
Inside sync gateway they look the same (plus entries for "_rev" and "_id")

snej:
>It seems odd that the FIRST_PAGE_FILE entry doesn't have the same digest as the attachment, is that correct?

That indicates the document was modified via the SG REST API (or else CBL 1.x) and the attachment was changed to zero-length. SG and CBL 1.x don't know about the newer `"@type":"blob"` schema, only the `_attachments` property, so CBL 2 creates that property when it pushes to SG. When the doc is pulled back to CBL 2 it will detect the difference in digests and copy the metadata back to the newer blob.

snej:
I'm trying to find a code path that could lead to this crash. @KraemerDEM, in this situation is there a warning logged of the form `"Failed to strip legacy attachments: error %d/%d"`? If so, I understand the crash and have a simple fix.

KraemerDEM:
Yes, I get "Failed to strip legacy attachments: error 4/4", as mentioned above.

snej:
Ok, the commit I pushed yesterday should fix the crash. But it seems wrong that there was an error pulling the document; I’ll look at that.

snej:
Closing the PR as the issue seems to have been fixed by my commit.

PR #733, opened 2019-03-30
Begin transaction before applying delta

### LINKED ISSUE #732: Delta sync throws C++ exception with P2P
Discovered via https://github.com/couchbase/couchbase-lite-net/issues/1135

The original issue talks about not being able to filter properly, but the reason it is not able to filter properly is because it doesn't survive long enough to make it to the filter.  The reason it happens only with attachments is because without attachments it is too small to create a delta.

The offending function is here

https://github.com/couchbase/couchbase-lite-core/blob/db48f678adc748f877def59636f069977f574d16/Replicator/DBWorker%2BPull.cc#L284

The error code that gets thrown is SharedKeysStateError with the message "incorrect use of persistent shared keys: not in transaction" which bubbles up from the JSON converter.

borrrden:
Fixes #732


snej:
Starting a transaction isn't the best solution. The underlying problem is that the encoder shouldn't be using the database's SharedKeys at this point.

PR #735, opened 2019-03-30
fix clang warnings about wrong fmt arguments

Dushistov:
There is problem with 64 bit types and printf specification. 
Normally `int64_t` is `long long` on 32 bit platform, and `long` on 64 bit platform.
So it impossible to choose right variant from `%ld` or `%lld`, that's why there is 
https://en.cppreference.com/w/cpp/types/integer and `PRIdx` and similar specifiers.

This PR should makes code more portable, plus fix bunch of warnings.
Plus after merge PR against vendor sub repositories it would be possible to turn on `-Werror=format` for clang based platforms.

### DIFF
--- a/LiteCore/Database/SequenceTracker.cc
+++ b/LiteCore/Database/SequenceTracker.cc
@@ -60,7 +60,7 @@ namespace litecore {
 
 
     void SequenceTracker::beginTransaction() {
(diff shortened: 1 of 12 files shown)


borrrden:
I've been meaning to do something like this for a while and didn't realize that everything was simply defined already!

Issue #744, opened 2019-04-03
Delta Push+Push failing on Windows / Linux


borrrden:
There seem to be some erroneous connections left over that are leaked to the second database in the test.  This happens on the second iteration of the test, and is usually not the same number of connections each time.  However, the result is this:

```
10:24:06.566534| [DB]: Waiting for 7 other connection(s) to close before deleting C:\Users\JimBorden\AppData\Local\Temp\Litecore_C_Tests\cbl_core_test_2.cblite2\db.sqlite3
10:24:06.718683| ERROR: LiteCore throwing LiteCore error 16: Can't delete db file while other connections are open
at litecore::error::_throw in c:\development\couchbase-lite-net-ee\couchbase-lite-net\vendor\couchbase-lite-core\litecore\support\error.cc: line: 398: address: 0x7FF70E8439E0
        at litecore::DataFile::deleteDataFile in c:\development\couchbase-lite-net-ee\couchbase-lite-net\vendor\couchbase-lite-core\litecore\storage\datafile.cc: line: 207: address: 0x7FF70E84CD20
        at litecore::DataFile::deleteDataFile in c:\development\couchbase-lite-net-ee\couchbase-lite-net\vendor\couchbase-lite-core\litecore\storage\datafile.cc: line: 179: address: 0x7FF70E84BC90
        at c4Internal::Database::deleteDatabase in c:\development\couchbase-lite-net-ee\couchbase-lite-net\vendor\couchbase-lite-core\litecore\database\database.cc: line: 202: address: 0x7FF70E8D7C70
        at std::_Invoker_pmf_pointer::_Call<void (__cdecl c4Internal::Database::*)(void) __ptr64,c4Database * __ptr64 & __ptr64> in c:\program files (x86)\microsoft visual studio 14.0\vc\include\type_traits: line: 1342: address: 0x7FF70E8F7600
        at std::invoke<void (__cdecl c4Internal::Database::*& __ptr64)(void) __ptr64,c4Database * __ptr64 & __ptr64> in c:\program files (x86)\microsoft visual studio 14.0\vc\include\type_traits: line: 1445: address: 0x7FF70E8F92F0
        at std::_Invoke_ret<void (__cdecl c4Internal::Database::*& __ptr64)(void) __ptr64,c4Database * __ptr64 & __ptr64> in c:\program files (x86)\microsoft visual studio 14.0\vc\include\type_traits: line: 1476: address: 0x7FF70E8F8310
[...]

PR #746, opened 2019-04-03
travis-ci: build with clang-8 plus run tests

Dushistov:
I updated travis config to use clang-8 instead of clang-7,
plus added call of `test_unix.sh`.

It fails, but this is because of tests are really failed, it also crashes on my machine.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,3 +20,4 @@ before_script:
   - clang --version
 script:
   - ./build_cmake/scripts/build_unix.sh
+  - cd build_cmake/unix && ../scripts/test_unix.sh > tests.log 2>&1


borrrden:
We don't use travis for any heavy commit validation so I'm not sure we'd want to run the tests here every time.  What is it about clang 8 that messes everything up though....

Dushistov:
@borrrden 

> We don't use travis for any heavy commit validation so I'm not sure we'd want to run the tests here every time. What is it about clang 8 that messes everything up though....

Sorry, I don't make it clear. My main concern that tests are failed, and you CI do not catch this.
That's why I add `scripts/test_unix.sh` to travis CI.

Tests failed locally on my machine, test failed inside clean ubuntu image in travis.
Test fails with clang-7 and clang-8.

I change to  `clang-8`, because of afraid that you do bother to support clang-7 after clang-8 release.

See for example  build with `clang-7`:

https://api.travis-ci.org/v3/job/515923149/log.txt

```
[0;37m/home/travis/build/Dushistov/couchbase-lite-core/C/tests/c4Test.cc:268: [0m[1;31mFAILED:
[0m[0;36m  CHECK( leaks == 0 )
[0mwith expansion:
[0;33m  5 == 0
[0m
terminating with uncaught exception of type Catch::TestFailureException


********** CRASH: SIGABRT - Abort (abnormal termination) signal **********
	 6  0x544085
	 7  0x7fbd339e5cb0
	 8  libc.so.6                 gsignal + 55
	 9  libc.so.6                 abort + 328
	10  0x7fbd35075980
	11  0x7fbd3505f21c
	12  0x7fbd35074bb3
	13  libc++abi.so.1            std::terminate() + 51
	14  0x41fd1b
	15  0x516e25
	16  0x50b7a4
	17  0x543d87
	18  0x543635
	19  0x5380a5
	20  0x52c62c
	21  0x54bf75
	22  0x53738b
	23  libc.so.6                 __libc_start_main + 245
	24  0x41eca5
```

borrrden:
This bug *was* caught by our CI.  Our CI is just post commit validation instead of pre.  See #744

Dushistov:
@borrrden 

> This bug was caught by our CI. 
Oh, I see. 

> Our CI is just post commit validation instead of pre

I did not know that such things still exist.

> We don't use travis for any heavy commit validation

It runs 9 mins 1 sec with tests, and 6 min 46 secs without tests.
Do you think this is "heavy"?

borrrden:
We'll have to see what becomes of Travis.  The rest of the team doesn't like it much because it's understandably slow for things like Android emulators, etc.  Also it only tests *one* platform out of the 6 that we support so it is hardly definitive for us.

borrrden:
@Dushistov Just to ensure you that this was not being ignored, you might notice a new check in this PR.  Finally we have pre-commit validation set up for PRs that will build and test on Windows, Mac, and Linux (Ubuntu) and the results are visible on a public Jenkins.  Perhaps we will continue to use travis to check various compilers to make sure things don't break, but the main point of this (to add testing to travis) has been worked around.  The result will soon be (there needs to be a `set -e` in the script) that if the tests fail for a PR on any platform that a red X will show.

PR #747, opened 2019-04-03
add option to disable LTO via cmake configuration

Dushistov:
- it breaks build with clang 8.0.0 / libc++ 8.0.0 / Arch Linux
  /usr/bin/ld: ../../libLiteCoreStatic.a: error adding symbols: file format not recognized
  clang-8: error: linker command failed with exit code 1 (use -v to see invocation)
- it breaks build with gcc

and I suppose it is not big deal to have option to disable this optimization (by default option is off, so `couchbase-lite-core` builds with LTO).


borrrden:
I saw this error and the solution is to not try to do LTO on the static libraries and only do it on the final shared library.  I'd like to know more about what the error actually means though because it makes zero sense.

borrrden:
But wait, how does the PR to bump to clang 8 work if clang 8 cannot compile this correctly?

PR #749, opened 2019-04-04
do not change timestamp of source file every time of cmake run

Dushistov:
I use some tool `XYZ` that care about modification time of source files.
And every rerun of cmake cause `empty.cpp` update,
and cause `XYZ` rerun for nothing.

I don't know why generate this source file via cmake and don't just add it into repository,
but do not touch if file already exists seems good thing for me and for SSD disk lifetime.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -221,7 +221,9 @@ if(WIN32)
         vendor/SQLiteCpp/sqlite3
     )
 else()
-    file(WRITE empty.cpp)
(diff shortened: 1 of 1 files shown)


borrrden:
It's that way because I felt like committing an empty file to the repo was silly.  The reason an empty file is needed is because Windows won't compile a shared library without any source files inside of it.  Ironically, Windows doesn't use this part anymore but I didn't realize it would create a file every time.  Ideally I want to switch to object libraries in the long term but this will do for now.

PR #751, opened 2019-04-10
Account for datasource aliases in column names

### LINKED ISSUE #750: c4query_columnTitle doesn't account for data source aliases
In other words, if I am doing a query like the following:

```js
{"WHAT":[[".main."],[".secondary."]],
"FROM":[{"AS":"main"},{"AS":"secondary","ON":["=",[".main.number1"],[".secondary.theone"]]}]}
```

Then `c4query_columnTitle` simply reports back the column names as `*` and `* #2`.  I would expect `main` and `secondary`

borrrden:
Fixes #750

### DIFF
--- a/C/tests/c4QueryTest.cc
+++ b/C/tests/c4QueryTest.cc
@@ -774,6 +774,20 @@ N_WAY_TEST_CASE_METHOD(QueryTest, "Delete index", "[Query][C][!throws]") {
     }
 }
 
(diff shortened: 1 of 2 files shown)


snej:
Add some tests, please?

borrrden:
Added a test so if things look ok I will merge as soon as I get the review approval from @snej

PR #754, opened 2019-04-12
Cap 32-bit linux cookie date to the max value

### LINKED ISSUE #510: Cookie test failures on 32-bit Linux due to year 2038 bug
The cookie implementation uses mktime(), the 32-bit version of which can't cope with the far-future date of 2099 that the test code uses. This causes a few assertions to fail when running on 32-bit platforms, for instance

```
Cookie Parser
  Expires
-----------------------------------------------------------------------------------------------------------------------
/var/data/rlc/poky/build-nanoscope/tmp/work/cortexa9hf-neon-oe-linux-gnueabi/couchbase-lite-core/2.0-r0/git/Replicator/tests/CookieStoreTest.cc:21

/var/data/rlc/poky/build-nanoscope/tmp/work/cortexa9hf-neon-oe-linux-gnueabi/couchbase-lite-core/2.0-r0/git/Replicator/tests/CookieStoreTest.cc:90: FAILED:
  CHECK( c.expires == 4084683494 )
with expansion:
  -1 == 4084683494 (0xf37752e6)

```

borrrden:
Alter the test to check such.  Fixes #510.

### DIFF
--- a/Replicator/CookieStore.cc
+++ b/Replicator/CookieStore.cc
@@ -52,29 +52,48 @@ namespace litecore { namespace repl {
     static void adjust_for_overflow(struct tm* inputTime)
     {
         if(sizeof(time_t) > 4 || inputTime->tm_year < 138) {
+            // Don't do anything if time_t is big enough or if the year is 2037 or earlier
             return;
         }
 
+        // February and later is out of bounds
+        // January needs more analysis
         if(inputTime->tm_mon > 0) {
             cap_to_max(inputTime);
             return;
         }
 
-        if(inputTime->tm_mday > 19) {
-            cap_to_max(inputTime);
+        // January 20 and later is out of bounds, January 18 and earlier is normal
+        // January 19 needs more analysis
+        if(inputTime->tm_mday != 19) {
+            if(inputTime->tm_mday > 19) {
+                cap_to_max(inputTime);
+            }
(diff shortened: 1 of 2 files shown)


Replicator/CookieStore.cc
```diff
@@ -36,8 +36,54 @@ using namespace fleece;
...
+        inputTime->tm_isdst = 0;
+    }
+
+    static void adjust_for_overflow(struct tm* inputTime)
+    {
+        if(sizeof(time_t) > 4 || inputTime->tm_year < 138) {
+            return;
+        }
+
+        if(inputTime->tm_mon > 0) {
```
snej: Please combine all five of these tests into one with `||`. Then since there's only one call to `cap_to_max`, you can move that into the body of the `if`.
(Sorry to be picky, but this code can be so much smaller without losing clarity.)

borrrden: I wanted to do that but this is not just one condition.  For example, I only care if the seconds are greater than 7 if the minutes are <= 14.  Otherwise any time with seconds greater then 7 will get capped.  Am I having a brain fart?

borrrden: This just made me realize that what I wrote doesn't actually achieve what I want...hold on that

PR #764, opened 2019-04-20
 Account for result aliases when creating SQL

### LINKED ISSUE #761: AS in WHAT clause doesn't define an alias for properties
If a JSON query uses an AS operator in the WHAT clause, to name a result column, this doesn't define that name as an alias for use in property lookup.

For example, if the query returns `foo.bar.baz` and names it AS `baz`, and then in the WHERE clause the property `baz` is used, that `baz` will refer to a document property named `baz`, though it really should be `foo.bar.baz`.

This isn't just to make the query shorter — it has an effect on performance. SQLite's optimizer doesn't have common-subexpression elimination, so if you can't use the result alias you have to repeat the result expression again in the WHERE clause, which means it gets evaluated twice.

borrrden:
Fixes #761 by remembering aliases and keeping track of things that use them

### DIFF
--- a/LiteCore/Query/QueryParser.cc
+++ b/LiteCore/Query/QueryParser.cc
@@ -1240,11 +1240,11 @@ namespace litecore {
         string alias;
         auto iType = _aliases.end();
         if(!property.empty()) {
+            // Check for result alias before 'alias' gets reassigned below
             alias = string(property[0].keyStr());
             iType = _aliases.find(alias);
         }
 
-        // Check for result alias before 'alias' gets reassigned below
         if (_propertiesUseSourcePrefix) {
             // Interpret the first component of the property as a db alias:
             require(property[0].isKey(), "Property path can't start with array index");
--- a/LiteCore/Query/QueryParser.cc
+++ b/LiteCore/Query/QueryParser.cc
@@ -95,6 +95,10 @@ namespace litecore {
         return isAlphanumericOrUnderscore(str) && !isdigit(str[0]);
     }
 
+    static bool isValidAlias(const string& alias) {
+        return alias.find('"') == string::npos && alias.find('\\') == string::npos;
+    }
+
     static const Value* getCaseInsensitive(const Dict *dict, slice key) {
         for (Dict::iterator i(dict); i; ++i)
             if (i.key()->asString().caseEquivalent(key))
@@ -139,14 +143,7 @@ namespace litecore {
             return string("\"") + name + "\"";
     }
 
-    static string unescapeAlias(const string& alias) {
-        string unescaped = alias;
-        unescaped.erase(remove(unescaped.begin(), unescaped.end(), '\\'), unescaped.end());
-        unescaped.erase(remove(unescaped.begin(), unescaped.end(), '"'), unescaped.end());
-        return unescaped;
-    }
 
-    
 #pragma mark - QUERY PARSER TOP LEVEL:
 
 
@@ -399,6 +396,15 @@ namespace litecore {
 #pragma mark - "FROM" / "JOIN" clauses:
 
 
+    void QueryParser::addAlias(const string &alias, aliasType type) {
+        require(isValidAlias(alias),
+                "Invalid AS identifier '%s'", alias.c_str());
+        require(_aliases.find(alias) == _aliases.end(),
+                "duplicate AS identifier '%s'", alias.c_str());
+        _aliases.insert({alias, type});
+    }
+
+
     void QueryParser::parseFromClause(const Value *from) {
         _aliases.clear();
         bool first = true;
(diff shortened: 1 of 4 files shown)


LiteCore/Query/QueryParser.cc
```diff
@@ -717,6 +718,8 @@ namespace litecore {
                 }
                 if (title.empty())
                     title = "*";        // for the property ".", i.e. the entire doc
+            } else {
+                _sql << " AS \"" << title << "\"";
```
snej: Use `quoteTableName(title)` instead of putting quotes around it. This will handle edge cases like if there are quotes inside `title`.

LiteCore/Query/QueryParser.cc
```diff
@@ -1268,6 +1271,30 @@ namespace litecore {
             return;
         }
 
+        // If the property in question is identified as an alias, emit that instead of
```
snej: In the existing code, if (db or join) aliases are in use, `_propertiesUseAliases` is true, and all properties must start with an alias. The code above from lines 1243-1251 identifies the alias and pulls it out of the path, and then line 1264 looks up the alias type, making your loop unnecessary.

The problem is that it would break compatibility to set `_propertiesUseAliases` when a result alias is defined, so we can't do that. So now we have a new state where the path might or might not begin with a result alias. Hm.

How about moving lines 1243-1251 up out of the `if` block, then looking up `alias` to see if it's a result alias and handling it if so. Otherwise proceed to the `if` block and so forth.

borrrden: The problem is that `_propertiesUseAliases` triggers logic in aliases that all apply to the way that *databases* are written.  What I want here is to change the way *results* are written.

LiteCore/Query/QueryParser.cc
borrrden: Sure, I wasn't sure if this was limited to only table names.

borrrden: Wait, doesn't `quoteTableName` do the same thing?  I don't see any edge cases handled...

LiteCore/Query/QueryParser.cc
borrrden: The path could either begin with a result alias or it could actually *be* a result alias (two cases to handle, the latter is not handled in that logic you mentioned since it doesn't contain a period)

LiteCore/Query/QueryParser.cc
snej: Oops, I thought it did, sorry!

LiteCore/Query/QueryParser.cc
```diff
@@ -1236,17 +1239,20 @@ namespace litecore {
...
-            string rest;
-            if (dot == string::npos) {
-                dot = property.size();
-            } else {
-                rest = property.substr(dot+1);
-            }
-            alias = property.substr(0, dot);
+        // Interpret the first component of the property as an alias:
+        auto dot = property.find('.');
+        string rest;
```
snej: This is going to misinterpret strings like `foo[0].bar` — alias will be `foo[0]` when it should be `foo`. The code at lines 1257-59 checks for this but always rejects brackets which isn't right in general. I think you need to take the minimum position of `.` and `[`.

borrrden:
@snej I updated this 20 days ago, I don't know if you got a notice about another review request.

snej:
Sorry! It fell between the cracks. I'll leave this tab open and review it in the morning.

LiteCore/Query/QueryParser.cc
```diff
@@ -706,8 +713,9 @@ namespace litecore {
             if (expr && expr[0]->asString().caseEquivalent("AS"_sl)) {
                 // Handle 'AS':
                 require(expr.count() == 3, "'AS' must have two operands");
-                title = string(requiredString(expr[2], "'AS' alias"));
+                title = unescapeAlias(string(requiredString(expr[2], "'AS' alias")));
```
snej: I think it's better to fail if the alias contains invalid characters (quotes or backslashes), rather than just removing them.
Also, for consistency the same rules should also apply to other "AS" aliases, i.e. in joins.

LiteCore/Query/QueryParser.cc
```diff
@@ -719,14 +727,17 @@ namespace litecore {
...
-                    title = columnTitleFromProperty(propertyFromNode(result), _propertiesUseAliases);
+                    title = columnTitleFromProperty(propertyFromNode(result), _propertiesUseSourcePrefix);
                 } else {
                     title = format("$%u", ++anonCount); // default for non-properties
                 }
                 if (title.empty())
                     title = "*";        // for the property ".", i.e. the entire doc
+            } else {
+                _sql << " AS ";
+                writeSQLString(title);
```
snej: This is going to write a string, with single-quotes; does SQLite even allow that? To get double-quotes, add a double-quote character as a 2nd parameter.

LiteCore/Query/QueryParser.cc
```diff
@@ -1247,6 +1269,26 @@ namespace litecore {
             return;
         }
 
+        if(iType->second == kResultAlias && property[0].keyStr().asString() == iType->first) {
+            // If the property in question is identified as an alias, emit that instead of
+            // a standard getter since otherwise it will probably be wrong (i.e. doc["alias"]
+            // vs alias -> doc["path"]["to"]["value"])
+            if(property.size() == 1) {
+                // Simple case, the alias is being used as-is
+                // Note: no need to check for string equality since two cases are now true
```
snej: I think this note is obsolete (from when you were comparing strings not Paths)?

snej:
Also, the Xcode compile fails:
```
couchbase-lite-core/LiteCore/Query/QueryParser.cc:458:25: error: enumeration value 'kResultAlias' not handled in switch [-Werror,-Wswitch]
                switch (_aliases[alias]) {
                        ^
```

PR #770, opened 2019-05-02
Workaround for AES128 encryption in existing 2.1-created dbs

snej:
When opening an encrypted database, if the AES256 key doesn’t work,
try again with AES128 (using the 1st 16 bytes of the key). If that
succeeds, rekey the database to the full AES256 key.

Also added unit tests for
- Opening such an AES128-encrypted db
- Opening a db created with 2.5 (for future compatibility testing)
- Opening a db with the wrong encryption key


LiteCore/Storage/SQLiteDataFile.cc
```diff
@@ -293,6 +296,33 @@ namespace litecore {
...
+            // If using AES256, retry with AES128 for backward compatibility with earlier versions:
+            logInfo("Retrying decryption with AES128...");
+            reopenSQLiteHandle();
+            success = _decrypt(kAES128, slice(key.buf, kEncryptionKeySize[kAES128]));
+            if (success) {
+                logInfo("Success! Database is decrypted.");
+                if (options().writeable && options().upgradeable) {
+                    // Now rekey with the full AES256 key:
+                    logInfo("Rekeying db to full AES256 encryption; this may take time...");
+                    int rc = sqlite3_rekey_v2(_sqlDb->getHandle(), nullptr, key.buf, (int)key.size);
```
pasin: I'm trying to find the information about how safe this operation is but I couldn't find any useful information of SQLite / SEE. I found one question from the SQLCipher forum asking what will happen if the app killed during rekey [here](https://groups.google.com/forum/#!topic/sqlcipher/rfWoWoFJPdg). 

From the answer, SQLCipher starts a transaction before rekeying if using pragma to rekey so that the database will not be changed until the transaction is committed. Should we do the same here?

LiteCore/Storage/SQLiteDataFile.cc
```diff
@@ -293,6 +296,33 @@ namespace litecore {
...
                 error::_throw(error::InvalidParameter);
         }
+        bool success = _decrypt(alg, key);
+#if __APPLE__
+        if (!success && alg == kAES256) {
+            // If using AES256, retry with AES128 for backward compatibility with earlier versions:
+            logInfo("Retrying decryption with AES128...");
+            reopenSQLiteHandle();
+            success = _decrypt(kAES128, slice(key.buf, kEncryptionKeySize[kAES128]));
+            if (success) {
```
pasin: How do we propagate the error up if the success == false? Should we throw an exception here?

snej: There is supposed to be a `throw` at the end if `success` is false ... I did a bunch of editing of that function and it looks like it got lost somehow. I'll restore it.

LiteCore/Storage/SQLiteDataFile.cc
snej: I don't know if a transaction will help. The docs say "every page of the database file [is] read, decrypted, reencrypted with the new key, then written out again". If that's literally all it does, then it's working at the page level not the database level, so a transaction wouldn't affect it.

pasin: I just sent an email to sqlite mailing list to ask about sqlite3_rekey_v2(). We could improve this later if it becomes an issue.

Issue #275, opened 2017-10-17
Remove C4DocEnumerator; it’s unnecessary & unused


snej:
The C4DocEnumerator API isn't necessary anymore, since queries can do the same thing more flexibly. Couchbase Lite 2 doesn't use it. So we could remove the API and implementation from LiteCore.

However, C4DocEnumerator is currently being used by the implementation of document expiration. So that makes #146 a prerequisite for removing it.

snej:
I ripped unused functionality out of C4DocEnumerator as part of improving code coverage; see commits 295f8a775c506eb63ac067e1f5a0ca16da59d588 and 412df3d32c0354f6710cfe89cbd64e43fa0cd0c4.

But note that the replicator also uses C4DocEnumerator.

snej:
Remaining users of C4DocEnumerator are:
- [ ] `DBWorker::_getChanges` (in the replicator)
- [ ] `RESTListener::handleGetAllDocs` (REST listener)
- [ ] `CBLiteTool::listDocs` (cblite tool)
- [ ] `DbEndpoint::exportTo` (cblite tool)

These can all be replaced by C4Query.

borrrden:
Migrated to [CBL-51](https://issues.couchbase.com/browse/CBL-51)

PR #779, opened 2019-05-14
feat: unresolved documents when restarting replicator

jayahariv:
* adds the unresolved docs to the docs-ended queue so that it tries to resolve when status changes.
* add method in DBAccess to return the docEnumerator, instead of vector<docInfo>, which can avoid one iteration of all docs.
* RevToSend was push, which will not work in docsEnded method in CBL side.
* include c4Error with conflict, so that it is handled like wise in CBL.

### DIFF
--- a/Replicator/DBAccess.cc
+++ b/Replicator/DBAccess.cc
@@ -119,7 +119,6 @@ namespace litecore { namespace repl {
             C4EnumeratorOptions options = kC4DefaultEnumeratorOptions;
             options.flags &= ~kC4IncludeNonConflicted;
             options.flags &= ~kC4IncludeBodies;
-            options.flags &= kC4IncludeDeleted;
             e = c4db_enumerateAllDocs(db, &options, &error);
         });
         return e;
--- a/Replicator/DBAccess.cc
+++ b/Replicator/DBAccess.cc
@@ -115,12 +115,11 @@ namespace litecore { namespace repl {
     C4DocEnumerator* DBAccess::unresolvedDocsEnumerator(C4Error *outError) {
         C4DocEnumerator* e;
         use([&](C4Database *db) {
-            C4Error error = {};
             C4EnumeratorOptions options = kC4DefaultEnumeratorOptions;
             options.flags &= ~kC4IncludeBodies;
             options.flags &= ~kC4IncludeNonConflicted;
             options.flags |= kC4IncludeDeleted;
-            e = c4db_enumerateAllDocs(db, &options, &error);
+            e = c4db_enumerateAllDocs(db, &options, outError);
         });
         return e;
     }
--- a/Replicator/DBAccess.cc
+++ b/Replicator/DBAccess.cc
@@ -117,8 +117,9 @@ namespace litecore { namespace repl {
         use([&](C4Database *db) {
             C4Error error = {};
             C4EnumeratorOptions options = kC4DefaultEnumeratorOptions;
-            options.flags &= ~kC4IncludeNonConflicted;
             options.flags &= ~kC4IncludeBodies;
+            options.flags &= ~kC4IncludeNonConflicted;
+            options.flags |= kC4IncludeDeleted;
             e = c4db_enumerateAllDocs(db, &options, &error);
         });
         return e;
--- a/Replicator/DBAccess.cc
+++ b/Replicator/DBAccess.cc
@@ -110,6 +110,20 @@ namespace litecore { namespace repl {
         else
             return {};
     }
+    
+    
+    C4DocEnumerator* DBAccess::unresolvedDocsEnumerator(C4Error *outError) {
+        C4DocEnumerator* e;
+        use([&](C4Database *db) {
+            C4Error error = {};
+            C4EnumeratorOptions options = kC4DefaultEnumeratorOptions;
+            options.flags &= ~kC4IncludeNonConflicted;
+            options.flags &= ~kC4IncludeBodies;
+            options.flags &= kC4IncludeDeleted;
+            e = c4db_enumerateAllDocs(db, &options, &error);
+        });
+        return e;
+    }
 
 
     static bool containsAttachmentsProperty(slice json) {
--- a/Replicator/DBAccess.hh
+++ b/Replicator/DBAccess.hh
@@ -74,6 +74,7 @@ namespace litecore { namespace repl {
         /** Returns the remote ancestor revision ID of a document. */
         alloc_slice getDocRemoteAncestor(C4Document *doc);
         
+        /** Returns the document enumerator for all unresolved docs present in the DB */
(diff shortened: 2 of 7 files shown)


Replicator/DBAccess.cc
```diff
@@ -110,6 +110,19 @@ namespace litecore { namespace repl {
...
             return {};
     }
+    
+    
+    C4DocEnumerator* DBAccess::unresolvedDocsEnumerator(C4Error *outError) {
+        C4DocEnumerator* e;
+        use([&](C4Database *db) {
+            C4Error error = {};
+            C4EnumeratorOptions options = kC4DefaultEnumeratorOptions;
+            options.flags &= ~kC4IncludeNonConflicted;
```
jayahariv: hoping this will include "conflicted + non-deleted", and "conflicted + deleted". some how when included the `kC4IncludeDeleted` it was not returning the non-deleted.

Replicator/DBAccess.hh
```diff
@@ -73,6 +73,9 @@ namespace litecore { namespace repl {
 
         /** Returns the remote ancestor revision ID of a document. */
         alloc_slice getDocRemoteAncestor(C4Document *doc);
+        
+        /** Returns the document enumerator for all unresolved docs present in the DB */
+        C4DocEnumerator* unresolvedDocsEnumerator(C4Error *outError);
```
jayahariv: returning the docEnumerator here can avoid help avoid unwanted conversion to docs and then to RevToInsert.

Replicator/Replicator.cc
```diff
@@ -82,11 +82,32 @@ namespace litecore { namespace repl {
...
         // Now wait for _onConnect or _onClose...
+        
+        // handle the unresolved docs
+        C4Error err = { };
+        C4DocEnumerator* e = _db->unresolvedDocsEnumerator(&err);
+        if (e) {
+            while(c4enum_next(e, &err)) {
+                C4DocumentInfo info;
+                c4enum_getDocumentInfo(e, &info);
+                auto rev = retained(new RevToInsert(nil,            /* incoming rev */
```
jayahariv: hope passing of incoming rev to nil, won't cause any issue!

Replicator/Replicator.cc
```diff
@@ -82,11 +82,32 @@ namespace litecore { namespace repl {
...
+                                                    info.docID,
+                                                    info.revID,
+                                                    nullslice,      /* history buf */
```
jayahariv: same with passing `nullslice`

Replicator/DBAccess.cc
borrrden: `IncludeDeleted` will also return non-deleted.  It's just a flag to toggle inclusion, not to exclude the alternate.

jayahariv: yes. correct.

Replicator/tests/ReplicatorLoopbackTest.cc
```diff
@@ -1306,6 +1306,56 @@ TEST_CASE_METHOD(ReplicatorLoopbackTest, "Pull Doc Notifications", "[Push]") {
...
+        if (info.docID == C4STR("conflict")) {
+            CHECK((info.flags & kDocDeleted) != kDocDeleted);
+        } else {
+            CHECK((info.flags & kDocDeleted) == kDocDeleted);
+        }
+        count++;
+    }
+    c4enum_free(e);
+    // TODO: This fails, since the `db2-deleted` is not showing up in the `unresolved`
+    CHECK(count == 3);
```
jayahariv: test fails, need to discuss about this

jayahariv:
taking the ownership back, when #781 is fixed, will continue with this.

jayahariv:
verified using this [unit test](https://gist.github.com/jayahariv/e0ab72322a3a27b6951255672cb2a978).

jayahariv:
will update that!

Replicator/DBAccess.cc
```diff
@@ -110,6 +110,20 @@ namespace litecore { namespace repl {
...
+    
+    C4DocEnumerator* DBAccess::unresolvedDocsEnumerator(C4Error *outError) {
+        C4DocEnumerator* e;
+        use([&](C4Database *db) {
+            C4Error error = {};
+            C4EnumeratorOptions options = kC4DefaultEnumeratorOptions;
+            options.flags &= ~kC4IncludeBodies;
+            options.flags &= ~kC4IncludeNonConflicted;
+            options.flags |= kC4IncludeDeleted;
+            e = c4db_enumerateAllDocs(db, &options, &error);
```
snej: You should pass outError here instead so the caller can see the error. (The local var `error` isn't needed at all.)

Replicator/Replicator.cc
```diff
@@ -82,11 +82,32 @@ namespace litecore { namespace repl {
         _connectionState = Connection::kConnecting;
         connection()->start();
         // Now wait for _onConnect or _onClose...
+        
+        // handle the unresolved docs
+        C4Error err = { };
```
snej: FYI, it's not necessary to initialize C4Error instances. They're initialized by the called function when an error occurs.

Replicator/Replicator.cc
```diff
@@ -82,11 +82,32 @@ namespace litecore { namespace repl {
...
+        if (e) {
+            while(c4enum_next(e, &err)) {
+                C4DocumentInfo info;
+                c4enum_getDocumentInfo(e, &info);
+                auto rev = retained(new RevToInsert(nullptr,        /* incoming rev */
+                                                    info.docID,
+                                                    info.revID,
+                                                    nullslice,      /* history buf */
+                                                    info.flags & kDocDeleted,
+                                                    !(info.flags & kDocConflicted)));
```
snej: This should be `false` (it doesn't mean the doc has no conflicts.)

Replicator/Replicator.cc
jayahariv: since this is an internal logic, and we need to make sure the enumeration didn't cause any error. can we use this?

jayahariv: understood now. I was thinking not straight. corrected

jayahariv:
> Would it be possible to refactor this to not use C4DocEnumerator and use the C++ equivalent instead? I noticed we have another open ticket to remove the C4DocEnumerator class all together so that will be harder if we add new stuff that uses it.

as per the discussion, we will do this in future. 

Ref: https://github.com/couchbase/couchbase-lite-core/issues/275

Replicator/Replicator.cc
```diff
@@ -82,11 +82,32 @@ namespace litecore { namespace repl {
         _connectionState = Connection::kConnecting;
         connection()->start();
         // Now wait for _onConnect or _onClose...
+        
+        // handle the unresolved docs
+        C4Error err;
```
snej: Please make this a separate method, something like `findExistingConflicts`.

PR #783, opened 2019-05-15
Add support for 4 date based N1QL functions

borrrden:
- DATE_DIFF_STR
- DATE_DIFF_MILLIS
- DATE_ADD_STR
- DATE_ADD_MILLIS

### DIFF
--- a/LiteCore/Query/SQLiteN1QLFunctions.cc
+++ b/LiteCore/Query/SQLiteN1QLFunctions.cc
@@ -863,8 +863,8 @@ namespace litecore {
         return year_day.count();
     }
 
-    inline system_clock::time_point to_time_point(DateTime& dt) {
-        const auto millis = ToMillis(dt);
+    inline system_clock::time_point to_time_point(DateTime& dt, bool no_tz = false) {
+        const auto millis = ToMillis(dt, no_tz);
         return system_clock::time_point(milliseconds(millis));
     }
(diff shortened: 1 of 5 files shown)


LiteCore/Query/SQLiteN1QLFunctions.cc
```diff
@@ -813,28 +837,223 @@ namespace litecore {
 
 #pragma mark - DATES:
 
+    inline int64_t getQuarter(const DateTime& t) {
```
snej: This is a lot of code being added — could you modularize this by creating new source/header files for the date-time logic, and then having just the SQLite glue code in here?

LiteCore/Query/SQLiteN1QLFunctions.cc
```diff
@@ -813,28 +837,223 @@ namespace litecore {
...
+        return (t.M + 2) / 3;
+    }
+
+    inline int64_t leapYearsWithin(int64_t year) {
```
snej: What's the source of these algorithms/formulas? Did they come from existing working code?

borrrden: I think the source was the N1QL go code from the server team.

borrrden:
This PR appears to be unimportant.  I'll leave the branch around.

Issue #776, opened 2019-05-10
c4db_endTransaction sometimes results in invalid memory access


borrrden:
Noticing this during debug testing of master branch for .NET (post-Iridium).  Sometimes a call to `c4db_endTransaction` results in an invalid memory access, probably due to a garbage deleted object.  Here is the stack trace

> LiteCore.dll!c4Internal::Database::externalTransactionCommitted(const litecore::SequenceTracker & sourceTracker) Line 525
> LiteCore.dll!c4Internal::Database::_cleanupTransaction::__l8::<lambda>(litecore::DataFile * other) Line 514
> LiteCore.dll!fleece::function_ref<void __cdecl(litecore::DataFile * __ptr64)>::callback_fn<void <lambda>(litecore::DataFile *) >(__int64 callable, litecore::DataFile * <params_0>) Line 61
> LiteCore.dll!fleece::function_ref<void __cdecl(litecore::DataFile * __ptr64)>::operator()(litecore::DataFile * <params_0>) Line 73
> LiteCore.dll!litecore::DataFile::Shared::forOpenDataFiles(litecore::DataFile * except, fleece::function_ref<void __cdecl(litecore::DataFile *)> fn) Line 88
> LiteCore.dll!litecore::DataFile::forOtherDataFiles(fleece::function_ref<void __cdecl(litecore::DataFile *)> fn) Line 161
> LiteCore.dll!c4Internal::Database::_cleanupTransaction(bool committed) Line 516
> LiteCore.dll!c4Internal::Database::endTransaction(bool commit) Line 502

The final LiteCore line of code that is executed is:

https://github.com/couchbase/couchbase-lite-core/blob/3c98e820e93fadd4aab985ae14f2ce836b112f96/LiteCore/Database/Database.cc#L525

Examining `_sequenceTracker` shows that it is full of garbage values.  Another thread is destroying a `DBAccess` object at the same time (notice the Database destructor has already been passed):

> LiteCore.dll!litecore::DataFile::Shared::removeDataFile(litecore::DataFile * dataFile) Line 72
> LiteCore.dll!litecore::DataFile::close() Line 136
> LiteCore.dll!litecore::SQLiteDataFile::~SQLiteDataFile() Line 175
> LiteCore.dll!c4Internal::Database::~Database() Line 187
> LiteCore.dll!c4Database::~c4Database() Line 51	
> LiteCore.dll!fleece::RefCounted::_careful_release() Line 87
[...]

PR #785, opened 2019-05-18
Alternative solution for close vs transaction race

borrrden:
Instead of trying to lock, use another method that indicates that a data file is in the process of closing.
See #776


borrrden:
Are there other locations that might be affected by this?

Issue #788, opened 2019-05-22
save method is not failing when same doc is updated 


jayahariv:
1. conflicted doc A 
2. while resolving conflict, save a new revision. I have updated the same document inside conflict resolver method. 
3. when calling the `void resolveConflict(C4String winningRevID, C4String losingRevID,
                             C4Slice mergedBody, C4RevisionFlags mergedFlags)`, this is returning success. 

expecting the `if (!winningRev->isLeaf() || !losingRev->isLeaf())` condition gets called and return failure with CBLConflict error.

adding the below unit test in Swift will show the error, 
```
- (void) testConflictResolverCalledTwice {
    NSString* docId = @"doc";
    NSDictionary* localData = @{@"key1": @"value1"};
    NSDictionary* remoteData = @{@"key2": @"value2"};
    [self makeConflictFor: docId withLocal: localData withRemote: remoteData];
    
    TestConflictResolver* resolver;
    CBLReplicatorConfiguration* pullConfig = [self pullConfig];
    
    resolver = [[TestConflictResolver alloc] initWithResolver: ^CBLDocument* (CBLConflict* con) {
        
        // update the doc will cause a second conflict
        CBLMutableDocument* savedDoc = [[self.db documentWithID: docId] toMutable];
        if (![savedDoc booleanForKey: @"secondUpdate"]) {
            NSError* error;
            [savedDoc setBoolean: YES forKey: @"secondUpdate"];
            [self.db saveDocument: savedDoc error: &error];
            AssertNil(error);
        }
        
        CBLMutableDocument* mDoc = con.localDocument.toMutable;
        [mDoc setString: @"local" forKey: @"edit"];
        return mDoc;
    }];
    pullConfig.conflictResolver = resolver;
    
    [self run: pullConfig errorCode: 0 errorDomain: nil];
    
    AssertEqual(self.db.count, 1u);
    NSMutableDictionary* exp = [NSMutableDictionary dictionaryWithDictionary: localData];
    [exp setValue: @"local" forKey: @"edit"];
    [exp setValue: @YES forKey: @"secondUpdate"];
    AssertEqualObjects([self.db documentWithID: docId].toDictionary, exp);
}
```

jayahariv:
savedDoc revID(3-3b580c270ad4412a50fc38bcdafdf2b1475f1270) and sequence(4)
localDocument revID(2-4d55da5a6671a685a75717f7307503ba87cfb87c) and sequence(2) 
[...]

PR #789, opened 2019-05-22
fix: conflict error not returning

jayahariv:
not sure, whether this is the best way to fix this issue. 

Ref: #788


snej:
Perfect.

PR #795, opened 2019-06-07
Fix a compiler error on macOS

farktronix:
When building couchbase-lite-ios using Xcode 11 I got this error:

```couchbase-lite-core/LiteCore/Storage/UnicodeCollator_Apple.cc:94:22: error: implicit conversion loses integer precision: 'CFComparisonResult' to 'int' [-Werror,-Wshorten-64-to-32]
        int result = CFStringCompareWithOptionsAndLocale(cfstr1, cfstr2,
            ~~~~~~   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```

This could be fixed with an explicit cast, but instead I made this match the Windows version of `compareStringsUnicode`.

### DIFF
--- a/LiteCore/Storage/UnicodeCollator_Apple.cc
+++ b/LiteCore/Storage/UnicodeCollator_Apple.cc
@@ -91,12 +91,21 @@ namespace litecore {
         }
 
 
-        int result = CFStringCompareWithOptionsAndLocale(cfstr1, cfstr2,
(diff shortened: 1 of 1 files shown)


borrrden:
@snej Simple enough, but now conflicting

snej:
Already fixed in commit 7b19da4c3b3be1cc554843049b7cd35585774391.

PR #800, opened 2019-06-21
[CBL-108] Fix a regression regarding handleSetCheckpoint

borrrden:
The introduction of _db->use made it so that any error response was incorrectly followed by another response for the same request.


snej:
Good catch! -- I hadn't noticed that the `return` statements just returned from the lambda, not from the method. (I only noticed the missing `return` on line 248 and thought that was the bug.)

It would be simpler to use a `C4Error` instead of the flag `needsResponse`. Then the error cases can set the error, and the test after the lambda can call `respondWithError` (once) and return.

And then it's even more concise if you use the special value-returning variant of `use`, e.g.

```c++
    C4Error error = _db->use<C4Error>([&](C4Database *db) { ...
```

snej:
Jim pointed out it wasn't that simple :) Let's merge this as it is.

PR #808, opened 2019-07-11
Add symbols used by JNI to c4.exp

pasin:
Added symbols used by JNI to c4.exp


pasin:
Will add those in .def in a separate PR.

snej:
I'm confused ... I thought the .exp file was only used on Apple platforms, where we don't use JNI.

And what does JNI use `c4blob_freeStore` (or `c4blob_openStore`) for? I can't think of a scenario where you have to open/close your own BlobStore.

PR #811, opened 2019-07-24
CBL-171: fix: `_activeIncomingRevs` was not decremented for rejected docs

jayahariv:
* `_activeIncomingRevs` are not decremented in case of rejected docs. 
* decrement logic added to revWasHandled method. 

* feel free to close, if this is not the right solution. But seems to be working in case of 1000 docs.

### DIFF
--- a/Replicator/Puller.cc
+++ b/Replicator/Puller.cc
@@ -263,7 +263,6 @@ namespace litecore { namespace repl {
 
     // Callback from an IncomingRev when it's been written to the db but before the commit
     void Puller::_revWasProvisionallyHandled() {
-        decrement(_activeIncomingRevs);
         if (_activeIncomingRevs < tuning::kMaxActiveIncomingRevs
                     && _unfinishedIncomingRevs < tuning::kMaxUnfinishedIncomingRevs
                     && !_waitingRevMessages.empty()) {
@@ -278,6 +277,7 @@ namespace litecore { namespace repl {
 
     // Callback from an IncomingRev when it's finished (either added to db, or failed)
     void Puller::revWasHandled(IncomingRev *inc) {
+        decrement(_activeIncomingRevs);
         _incomingDocIDs.remove(inc->rev()->docID);       // this is thread-safe
         _returningRevs.push(inc);
     }


snej:
This patch has a couple of problems (which aren't at all obvious, so don't feel bad!) 

1. `_activeIncomingRevs` is not atomic, but is being modified in `revWasHandled()` which happens to be called directly from another thread (the IncomingRev's.) Normally Actors don't call each other like this, but this call is an exception for performance reasons. (I should add a warning in a comment.)
2. When decrementing `_activeIncomingRevs` lowers it below `kMaxActiveIncomingRevs`, a new `IncomingRev` should be activated. That's the logic of `_revWasProvisionallyHandled()`. Decrementing the counter elsewhere defeats this.

@jayahariv, could you add a link to the issue? I'll try and come up with a better fix.

jayahariv:
https://issues.couchbase.com/browse/CBL-171

snej:
@jayahariv, are you certain the bug is that `_activeIncomingRevs` doesn't end up at zero? And are you certain that a rejected doc causes it not to be decremented? I'm not seeing a problem like that from inspecting the code.

jayahariv:
this happens, if there are more than 100 docs which are rejected by the validator function. then `_activeIncomingRevs` becomes 100 and doesn't go down. 

If you need to reproduce the issue, use the unit test in CBL side. [unit-test.txt](https://github.com/couchbase/couchbase-lite-core/files/3429448/unit-test.txt)

jayahariv:
more insights. 

in method `void Puller::handleRev(Retained<MessageIn> msg)`. 
- when `_activeIncomingRevs` == 100, it will keep adding the `msg` to `_waitingRevMessages`. which was not getting being handled.  
- one issue i can think of is, `_activeIncomingRevs` is not decrementing in case of rejected revision. which i tried to resolve. 
- also, if we could call the `void Puller::handleMoreChanges()`, i am assuming this will tries to pop the messages from the `_waitingChangesMessages`. But i was not sure, where to call this.

snej:
OK, I found the problem: `IncomingRev::_provisionallyInserted` is not cleared when the IncomingRev instance is reused. So if on its second (or later) use the `IncomingRev` doesn't get inserted, the flag is still true, and the `Puller` won't call `_revWasProvisionallyHandled()` on it.

I haven't verified this with testing, but I'm pretty certain of it.

Also, for safety's sake, `_provisionallyInserted` should be a `std::atomic<bool>` since it gets accessed from outside the actor's event loop, but that's not likely to have caused a problem in practice.

snej:
I pasted `testPullFilter1000` into ReplicatorTests.swift and verified that it passes now. I'll create a new PR for you to try.

snej:
See #812 for my fix.

PR #812, opened 2019-07-25
CBL-171: Fix Puller not stopping after filter rejects docs

snej:
The replicator never went to Stopped after >100 docs were rejected by
the client's validation filter.
The bug was that IncomingRev didn't clear its _provisionallyInserted
flag when it was reused a second time; this then caused the Puller to
not decrement _activeIncomingRevs when that IncomingRev finished,
which meant the Puller remained in Busy state.

Fixes CBL-171

### DIFF
--- a/Replicator/Puller.cc
+++ b/Replicator/Puller.cc
@@ -276,15 +276,21 @@ namespace litecore { namespace repl {
         }
     }
 
-    // Callback from an IncomingRev when it's finished (either added to db, or failed)
+    // Called from an IncomingRev when it's finished (either added to db, or failed.)
+    // The IncomingRev will be processed later by _revsFinished().
     void Puller::revWasHandled(IncomingRev *inc) {
-        _incomingDocIDs.remove(inc->rev()->docID);       // this is thread-safe
-        _returningRevs.push(inc);
+        // CAUTION: For performance reasons this method is called directly, without going through the
+        // Actor event queue, so it runs on the IncomingRev's thread, NOT the Puller's! Thus, it needs
+        // to pay attention to thread-safety.
+        _incomingDocIDs.remove(inc->rev()->docID);      // this is thread-safe
+        _returningRevs.push(inc);                       // this is thread-safe
     }
 
     void Puller::_revsFinished(int gen) {
         auto revs = _returningRevs.pop(gen);
         for (IncomingRev *inc : *revs) {
+            // If it was provisionally inserted, it'll have called _revWasProvisionallyHandled
+            // already. If not, call that method now:
             if (!inc->wasProvisionallyInserted())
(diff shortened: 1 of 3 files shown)


Replicator/Puller.cc
```diff
@@ -276,15 +276,21 @@ namespace litecore { namespace repl {
...
+        _incomingDocIDs.remove(inc->rev()->docID);      // this is thread-safe
+        _returningRevs.push(inc);                       // this is thread-safe
     }
 
     void Puller::_revsFinished(int gen) {
         auto revs = _returningRevs.pop(gen);
         for (IncomingRev *inc : *revs) {
+            // If it was provisionally inserted, it'll have called _revWasProvisionallyHandled
+            // already. If not, call that method now:
             if (!inc->wasProvisionallyInserted())
```
jayahariv: does this mean, all the revisions, including the rejected ones(which are not inserted to db), will get a flag called `wasProvisionallyInserted` on it?

jayahariv: got confused with the inserted name, which i assumed, will only get callbacks for those are inserted to the db. Since the rejected revisions are not inserted, i assume, we should probably get back a callback which is like `handled` and not inserted.

snej: "Provisionally inserted" means the revision's been written to the database but the enclosing transaction hasn't been committed yet. (Revs get added in bulk.) This is a boolean flag on the IncomingRev class used by the replicator. When a rev is provisionally inserted, `Puller::_revWasProvisionallyHandled` is called.

If a rev is not inserted, it still goes through `_revsFinished`, and the bookkeeping needs to happen then.

jayahariv: yea. understood now.

Replicator/Puller.cc
```diff
@@ -276,15 +276,21 @@ namespace litecore { namespace repl {
...
+        // Actor event queue, so it runs on the IncomingRev's thread, NOT the Puller's! Thus, it needs
+        // to pay attention to thread-safety.
+        _incomingDocIDs.remove(inc->rev()->docID);      // this is thread-safe
+        _returningRevs.push(inc);                       // this is thread-safe
     }
 
     void Puller::_revsFinished(int gen) {
         auto revs = _returningRevs.pop(gen);
         for (IncomingRev *inc : *revs) {
+            // If it was provisionally inserted, it'll have called _revWasProvisionallyHandled
```
borrrden: What happened here to the comments?

jayahariv:
Curious, why Android doesn't fail due to this.

Replicator/Puller.cc
snej: I don't understand; the comment above looks fine to me.

PR #815, opened 2019-08-03
CBL-160: fix: java issues on windows 

jayahariv:
* return type for the `CopyFileW();` success return value is 1 where as `CopyFile2` which is zero. 
* add references to remove gradle build failure.

### DIFF
--- a/C/c4.def
+++ b/C/c4.def
@@ -110,7 +110,6 @@ c4blob_keyFromString
 c4blob_keyToString
 c4blob_openStore
 c4blob_deleteStore
-c4blob_freeStore
 c4blob_getSize
 c4blob_getContents
 c4blob_getFilePath
@@ -157,7 +156,6 @@ c4pred_unregisterModel
 
 c4_getObjectCount
 c4_shutdown
-c4_setTempDir
 
 FLSlice_Equal
 FLSlice_Compare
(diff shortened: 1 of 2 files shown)


C/c4.def
```diff
@@ -110,6 +110,7 @@ c4blob_keyFromString
 c4blob_keyToString
 c4blob_openStore
 c4blob_deleteStore
+c4blob_freeStore
```
borrrden: Please only add these below the `; C4Tests` line.  You can rename this comment if you like but I use all lines above this for the C# bindings script to generate bindings only for what is needed.

LiteCore/Support/FilePath.cc
```diff
@@ -124,7 +124,7 @@ static int copyfile(const char* from, const char* to)
     err = CopyFile2(wideFrom, wideTo, nullptr);
 #endif
 
-    if(err != S_OK) {
+    if(err < S_OK) {
```
borrrden: I kind of understand but this seems like only half of the needed change.  CopyFileW should not be used anymore because we don't support Windows 7 anyway so that is some small debt that needs to be paid back by removing the dynamic search for CopyFile2.  I think most error codes here are greater than 0 so this needs either a removal of CopyFileW (recommended) or both values need to be coalesced to the same value.

PR #817, opened 2019-08-09
CBL-221 Keep Puller in busy status longer

borrrden:
It needs to wait until all the unfinished incoming revisions are done.  These are interjected in a thread-suspicious way by the Inserter.

### DIFF
--- a/Replicator/Puller.cc
+++ b/Replicator/Puller.cc
@@ -349,12 +349,14 @@ namespace litecore { namespace repl {
     
     Worker::ActivityLevel Puller::computeActivityLevel() const {
         ActivityLevel level;
-        if (_fatalError || !connection()) {
-            level = _unfinishedIncomingRevs > 0 ? kC4Busy : kC4Stopped;
+        if (_unfinishedIncomingRevs > 0) {
+            // CBL-221: Crash when scheduling document ended events
+            level = kC4Busy;
+        } else if (_fatalError || !connection()) {
+            level = kC4Stopped;
         } else if (Worker::computeActivityLevel() == kC4Busy
                 || (!_caughtUp && nonPassive())
                 || _pendingRevMessages > 0
-                || _unfinishedIncomingRevs > 0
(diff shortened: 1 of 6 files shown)


borrrden:
I'll see if I can add a test for this.  I think it should be possible with the loopback connection tests.

Replicator/Pusher.cc
```diff
@@ -637,6 +637,9 @@ namespace litecore { namespace repl {
     Worker::ActivityLevel Pusher::computeActivityLevel() const {
         ActivityLevel level;
         if (!connection()) {
+            // Does this need a similar guard to what Puller has?  It doesn't
+            // seem so since the Puller has stuff that happens even after the
+            // connection is closed, while the Pusher does not seem to.
```
snej: I agree, I don't think Pusher needs a change.

Replicator/Puller.cc
```diff
@@ -350,7 +350,7 @@ namespace litecore { namespace repl {
     Worker::ActivityLevel Puller::computeActivityLevel() const {
         ActivityLevel level;
         if (_fatalError || !connection()) {
-            level = kC4Stopped;
+            level = _unfinishedIncomingRevs > 0 ? kC4Busy : kC4Stopped;
```
snej: I like the logic, but I think it'd be clearer if done by moving the existing `_unfinishedIncomingRevs` test to the start of the `if` chain:

```c++
if (_unfinishedIncomingRevs > 0) {
    level = kBusy;
} else if (_fatalError || !connection()) {
    level = kStopped;
} else ...

snej:
Oh hey, while you're fixing this, it'd be good to change the bad cast that triggered the bug into a `dynamic_cast<Replicator*>`.

borrrden:
The above seems to be taken care of in another PR, should I put it in this one too?

snej:
No need to; yeah, I started thinking about unsafe casts and null pointers and cleaned up the replicator code.

PR #818, opened 2019-08-09
Close existing Query objects when closing DataFile

snej:
We can't leave any active sqlite3_stmt when closing a DataFile,
because if the database is deleted its files will still be open,
which causes errors on Windows. This situation occurs when a Query
instance still exists, since SQLiteQuery owns a sqlite3_stmt.

Freeing the C4Query objects isn't always possible because if CBL is
implemented in a garbage collected language the objects' finalizers
may not yet have run.

The fix is to have the DataFile track all open Query instances and
tell them to close when it closes. SQLiteQuery closes its statement
when it closes.

Added some logic to DataFile::deleteFile to enforce that no handles
are left open when closing a db that's about to be deleted. This
allowed me to write a unit test to verify the fix even on non-
Windows platforms.

Fixes CBL-214

### DIFF
--- a/LiteCore/Query/Query.cc
+++ b/LiteCore/Query/Query.cc
@@ -18,6 +18,7 @@
 
 #include "Query.hh"
 #include "KeyStore.hh"
+#include "DataFile.hh"
 #include "Logging.hh"
 #include "StringUtil.hh"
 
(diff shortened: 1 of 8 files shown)


snej:
No, you don't need to worry about the QueryEnumerator; it doesn't have any SQLite references.

PR #819, opened 2019-08-09
Merge feature/libwebsockets

snej:


snej:
Looks like I need to fix the CMake build script...

borrrden:
Also the PR validation will not be able to run with merge conflicts now that it is set up to run on the result of merging the PR

snej:
I'm reimplementing (again) without libwebsockets.

PR #825, opened 2019-08-15
CBL-197 c4doc_get returns null when the document is expired. Add test to verify.

Sandychuang8:


snej:
This would hurt performance of every document load, by doing another SQLite lookup of the expiration date. Instead the expiration test should be folded into the SQLite query that fetches the document data (down in SQLiteKeyStore.cc.)

Actually I thought we had decided not to implement this, because it creates an inconsistency with query results, which don't filter expired documents (for performance reasons, again.)

Note that there is only a very small time window where an expired document exists in the db before it gets purged...

borrrden:
If we decided not to do this, it was unclear to me because there is still an open ticket for it.  @rajagp What are we supposed to be doing for [CBL-197](https://issues.couchbase.com/browse/CBL-197)

borrrden:
Closing this PR since it cannot be merged unless another area of the library is altered.  Let's keep the work so far though, so don't delete the branch.

PR #829, opened 2019-08-22
Fixed race condition starting replicators

snej:
It's possible for a replicator to start and call a progress callback
before the c4repl_new function returns; in that case the client
hasn't yet set the C4Replicator handle, so if the callback references
it, it will get null (or worse, garbage.)

This is a pretty rare situation given the timing required, but it
does happen to the ReplicatorAPITest occasionally on Jenkins and
causes a test failure.

To address this, I added a flag to the options that if set prevents
c4repl_new from starting the replicator. Then the client, once it has
the C4Replicator handle, can call a new function c4repl_start.

### DIFF
--- a/C/c4.def
+++ b/C/c4.def
@@ -139,6 +139,7 @@ c4docobs_free
 c4repl_new
 c4repl_newWithSocket
 c4repl_free
+c4repl_start
 c4repl_stop
 c4repl_getStatus
 
--- a/C/c4.exp
+++ b/C/c4.exp
@@ -161,6 +161,7 @@ _c4queryobs_getEnumerator
 _c4repl_new
 _c4repl_newWithSocket
 _c4repl_free
+_c4repl_start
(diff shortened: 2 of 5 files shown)


C/c4.def
```diff
@@ -139,6 +139,7 @@ c4docobs_free
 c4repl_new
 c4repl_newWithSocket
 c4repl_free
+c4repl_start
```
borrrden: This file has a special extra requirement by me that I should probably write as a comment on top of the file, but if this is not going to be auto slurped by the bindings script, then please include it below the line `; C4Tests`

snej: Hm. It's probably a good idea to create a binding, since the platform code could benefit from using this. It's a very unlikely race condition, but it could conceivably happen in real life.

borrrden: Ok I will leave it there and file a ticket to make use of it then.

PR #831, opened 2019-08-23
CBL-126 Add 503 to the list of errors to stop replication

borrrden:
Need to come up with a testing method, but this will stop the replicator when a BLIP message containing 503 is received.  CBL should pick this up and react to it by retrying later.

### DIFF
--- a/Replicator/Replicator.cc
+++ b/Replicator/Replicator.cc
@@ -36,9 +36,16 @@ using namespace fleece;
 
 namespace litecore { namespace repl {
 
-    static C4Error StoppingErrors[] = {
-        { LiteCoreDomain, kC4ErrorUnexpectedError,0 },
-        { WebSocketDomain, 503, 0 }
+    struct C4StoppingErrorEntry
+    {
+        C4Error err;
+        bool isFatal;
+        slice msg;
+    };
+
+    static C4StoppingErrorEntry StoppingErrors[] = {
+        {{ LiteCoreDomain, kC4ErrorUnexpectedError,0 }, true, "An exception was thrown"_sl},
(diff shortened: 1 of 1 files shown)


borrrden:
@snej This is the right place to be adding this correct?

borrrden:
Confirmed via proxy testing that this works.  The replication disconnects and goes offline.

Replicator/Replicator.cc
```diff
@@ -241,11 +246,14 @@ namespace litecore { namespace repl {
...
-        if (error.domain == LiteCoreDomain && error.code == kC4ErrorUnexpectedError) {
-            // Treat an exception as a fatal error for replication:
-            alloc_slice message( c4error_getDescription(error) );
-            logError("Stopping due to fatal error: %.*s", SPLAT(message));
-            _disconnect(websocket::kCodeUnexpectedCondition, "An exception was thrown"_sl);
+        for(const C4Error& stoppingErr : StoppingErrors) {
+            if(stoppingErr.domain == error.domain && stoppingErr.code == error.code) {
+                // Treat an exception as a fatal error for replication:
+                alloc_slice message( c4error_getDescription(error) );
+                logError("Stopping due to fatal error: %.*s", SPLAT(message));
```
snej: I'd rather not report these the same way, since they're quite different conditions. How about just adding an `else` condition after the existing `if` to check for 503, and handle it by logging a warning and setting the disconnect message to "The server is over capacity".

borrrden: You mean replace the message received from the other side with a hardcoded one?  I think it might be a good idea to keep the original around.  Although I could make a distinction between "fatal" and not "fatal"

PR #833, opened 2019-08-25
RevTree.hh: UINT_MAX used, but there is no climits include

Dushistov:
fix build with gcc 9


borrrden:
Sorry about the failing PR validation.  New machine comes with new and unique ways for things to crash, but that stuff is being addressed in *other* PRs so for now failures are not immediately bad.

PR #841, opened 2019-09-11
CBL-218 Make FTS indexes work on array properties

borrrden:
Make a special function for changing arrays into a proper searchable string


borrrden:
@snej I'm confident in the results, but not in the restrictions I've put into place (only string / array of string allowed).  Perhaps I should relax them?

LiteCore/Query/SQLiteFleeceFunctions.cc
```diff
@@ -103,6 +107,77 @@ namespace litecore {
         }
     }
 
+    static void handle_fts_value(const Value* data, alloc_slice& result) {
```
snej: `alloc_slice::append` isn't efficient: it calls realloc every time. (Because alloc_slice doesn't remember capacity separately from size.) It'd be better to use a `stringstream`.

LiteCore/Query/SQLiteFleeceFunctions.cc
```diff
@@ -103,6 +107,77 @@ namespace litecore {
...
+        switch(data->type()) {
+            case kBoolean:
+                result.append(data->asBool() ? "T"_sl : "F"_sl);
```
snej: Why "T" and "F", not "true" and "false"? Compatibility with Server?

LiteCore/Query/SQLiteFleeceFunctions.cc
```diff
@@ -103,6 +107,77 @@ namespace litecore {
...
+                break;
+            case kNumber:
+            {
```
snej: For the entire `kNumber` case, you can call `toString()` to convert it to string form.

LiteCore/Query/SQLiteFleeceFunctions.cc
```diff
@@ -103,6 +107,77 @@ namespace litecore {
...
+            case kArray:
+                handle_fts_array(data, result);
+                break;
+            case kDict:
+                handle_fts_dict(data, result);
+                break;
+        }
+    }
+
+    static void handle_fts_array(const Value* data, alloc_slice& result) {
```
snej: Consider using a DeepIterator to walk through the Fleece. It'll make the code simpler, and avoid using unbounded stack space too. (Use the one in DeepIterator.hh, _not_ Fleece.hh.)

LiteCore/Query/SQLiteFleeceFunctions.cc
```diff
@@ -103,6 +107,77 @@ namespace litecore {
...
+            result.append(" "_sl);
+        }
+    }
+
+    // fl_fts_value(body, propertyPath) -> blob data
+    static void fl_fts_value(sqlite3_context* ctx, int argc, sqlite3_value **argv) noexcept {
+        try {
+            QueryFleeceScope scope(ctx, argv);
+            if(scope.root == nullptr) {
+                setResultFromValue(ctx, scope.root);
```
snej: Why set any result when there's no value (i.e. it's MISSING?)

LiteCore/Query/SQLiteFleeceFunctions.cc
```diff
@@ -103,6 +107,77 @@ namespace litecore {
         }
     }
 
+    static void handle_fts_value(const Value* data, alloc_slice& result) {
+        switch(data->type()) {
```
snej: This will crash if `data` is NULL, i.e. MISSING.

LiteCore/Query/SQLiteFleeceFunctions.cc
borrrden: Yes, that's where I got the values from

LiteCore/Query/SQLiteFleeceFunctions.cc
borrrden: What happens if you don't set a result?  Simply no row gets emitted....I guess that would make sense.

Issue #845, opened 2019-09-16
hiding sqlite3* on Linux


Dushistov:
I use sqlite in my project and libLiteCore.so, 
it would be nice if libLiteCore doesn't export sqlite symbols, that prevents mixing symbols with system sqlite3 shared library :

```
$ nm -D libLiteCore.so | grep sqlite3_open
000000000016ddb0 T sqlite3_open
000000000016dca0 T sqlite3_open16
000000000016dda0 T sqlite3_open_v2
```

Ideally if libLiteCore.so hide all symbols except exported via C/include,
but hiding sqlite3_* functions will be nice.

Also this is theoretically make LTO more effective, see https://clang.llvm.org/docs/LTOVisibility.html if you hide sqlite3_ with `-fvisibility=hidden`

borrrden:
Seems like the downstream is busy and hasn't considered the other suggestion on the PR (which is fine), so I filed [CBL-430](https://issues.couchbase.com/browse/CBL-430) to get this into a sprint in the near future.

borrrden:
This enhancement is in 2.7.0. Will close if I don't hear back in 7 days.

PR #846, opened 2019-09-16
mark sqlite3 symbols as hidden for linux build

Dushistov:
Before this patch stripped build_cmake/unix/libLiteCore.so 3407504 bytes,
after 3373424. Related to #845

### DIFF
--- a/LiteCore/Storage/SQLiteChooser.c
+++ b/LiteCore/Storage/SQLiteChooser.c
@@ -19,6 +19,10 @@
 // Compiles the appropriate version of SQLite, depending on whether we are building the
 // Consumer Edition or Enterprise Edition of LiteCore.
 
+#ifdef __linux__
+#define SQLITE_API __attribute__ ((visibility ("hidden")))
+#endif
+
 #ifdef COUCHBASE_ENTERPRISE
     // These source files are NOT in this repository, rather in a Couchbase-private
     // repository that needs to be checked out next to this one.


borrrden:
Are you familiar with ld files?  I believe this same thing could be solved in a better way (i.e. hiding all the things that we *actually* hide on non-Linux) by setting that up with CMake (i.e. by making `c4.gnu` with contents similar to [this](https://github.com/Yubico/libfido2/blob/master/src/export.gnu)).  Is that something you can put in the PR, or should I file a ticket to do it later?

Dushistov:
> Are you familiar with ld files? 
> I believe this same thing could be solved in a better way

Not sure that this is really better way.
Yes, GNU ld script will hide all unnecessary symbols,
but will this cooperate with LTO? May be

```
set(CMAKE_C_VISIBILITY_PRESET hidden)
set(CMAKE_CXX_VISIBILITY_PRESET hidden)
set(CMAKE_VISIBILITY_INLINES_HIDDEN 1)
```

and explicit marking public symbols with

```
__attribute__ ((visibility ("default")))
```

gives much better result.

I will try both approaches and compare result when I have time.

borrrden:
Perhaps I should say "more consistent" rather than "better" because we have these lists put together already for macOS and Windows (c4.def and c4.exp in the C folder)

snej:
3MB is pretty high! The macOS x86-64 LiteCore dylib is ~1.6MB. It restricts exported symbols by using a `.exp` file (`C/4c.exp`), which is I think a feature specific to Apple's linker.

borrrden:
> which is I think a feature specific to Apple's linker.

We do this for MSVC too, and GNU ld also has this feature (i.e. it will only export what you fill it out with)

borrrden:
This should be taken care of as part of PR #853

PR #847, opened 2019-09-17
CBL-324: fix: support source build

jayahariv:
* support source for android was not building with ICU
* create section for including the directory and pass the argument
* complaining about arc4random and strlcat, which resolved by including linux support configs.


cmake/platform_android.cmake
```diff
@@ -59,7 +59,7 @@ function(setup_litecore_build)
     )
 
     target_link_libraries(
-        LiteCore PRIVATE 
+        LiteCore PRIVATE
```
jayahariv:
```suggestion
        LiteCore PRIVATE 
```

cmake/platform_android.cmake
```diff
@@ -47,7 +47,7 @@ endfunction()
 function(setup_litecore_build)
     setup_litecore_build_linux()
 
-    
+
```
jayahariv:
```suggestion
    
```

cmake/platform_android.cmake
```diff
@@ -70,6 +70,18 @@ function(setup_litecore_build)
     )
 endfunction()
 
+function(setup_support_build)
+    setup_support_build_linux()
+    target_compile_definitions(
+        Support PRIVATE
+        -DLITECORE_USES_ICU=1
```
borrrden: If this is needed in Android then it's needed in Linux as well, so better to move this compile definition into `platform_linux` (which both Linux and Android derive from)

cmake/platform_linux.cmake
```diff
@@ -53,9 +53,13 @@ function(setup_litecore_build_linux)
     )
 endfunction()
 
-function(setup_support_build)
+function(setup_support_build_linux)
```
borrrden: If you rename this then you need to add `setup_support_build` to `platform_linux_desktop` which just calls this method.

pasin:
@bmeike Maybe this also fixes the swedish collator issue in one of the unit tests?

cmake/platform_linux.cmake
```diff
@@ -53,9 +53,13 @@ function(setup_litecore_build_linux)
...
     target_include_directories(
         Support PRIVATE
         LiteCore/Unix
     )
+endfunction()
+
+function(setup_support_build)
```
borrrden: This will conflict with the Android version in `platform_android`

cmake/platform_linux_desktop.cmake
```diff
@@ -103,13 +103,17 @@ function(setup_litecore_build)
     endif()
 
     target_link_libraries(
-        LiteCore PRIVATE 
-        z 
-        ${ICU4C_COMMON} 
+        LiteCore PRIVATE
```
jayahariv:
```suggestion
        LiteCore PRIVATE 
```

cmake/platform_linux_desktop.cmake
```diff
@@ -103,13 +103,17 @@ function(setup_litecore_build)
...
+        z
```
jayahariv:
```suggestion
        z 
```

cmake/platform_linux_desktop.cmake
```diff
@@ -103,13 +103,17 @@ function(setup_litecore_build)
...
+        ${ICU4C_COMMON}
```
jayahariv:
```suggestion
        ${ICU4C_COMMON} 
```

bmeike:
@pasin That would be cool!  I'll check, when this becomes available.

PR #848, opened 2019-09-18
CBL-52 Make the LIKE and CONTAINS functions collation aware

borrrden:
A few things go into making this possible:

- Add a new entry into the parser table function spec indicating if a function can take a collation, if offered.  It will always be the last thing passed to it.
- Replace LIKE with FL_LIKE() (breaking change for CBL!)
- FL_LIKE() and CONTAINS() now have both 2 and 3 argument variants
- Split UTF8Length into two functions, the latter returning the number of bytes in the next character (useful for comparing slices char by char in a multibyte friendly way)

### DIFF
--- a/LiteCore/Query/QueryParserTables.hh
+++ b/LiteCore/Query/QueryParserTables.hh
@@ -165,10 +165,10 @@ namespace litecore {
         {"regexp_like"_sl,      2, 2},
         {"regexp_position"_sl,  2, 2},
         {"regexp_replace"_sl,   3, 9},
-        {"fl_like"_sl,          2, 3, nullslice, false, true},
+        {"fl_like"_sl,          2, 2, nullslice, false, true},
 
         // Strings:
-        {"contains"_sl,         2, 3, nullslice, false, true},
+        {"contains"_sl,         2, 2, nullslice, false, true},
         {"length"_sl,           1, 1, "N1QL_length"_sl},
         {"lower"_sl,            1, 1, "N1QL_lower"_sl},
         {"ltrim"_sl,            1, 2, "N1QL_ltrim"_sl},
--- a/LiteCore/Query/QueryParserTables.hh
+++ b/LiteCore/Query/QueryParserTables.hh
@@ -63,6 +63,7 @@ namespace litecore {
         {"IS"_sl,      2, 2,  3,  &QueryParser::infixOp},
         {"IS NOT"_sl,  2, 2,  3,  &QueryParser::infixOp},
         {"IN"_sl,      2, 9,  3,  &QueryParser::inOp},
+        {"LIKE"_sl,    2, 3,  3,  &QueryParser::likeOp},
         {"NOT IN"_sl,  2, 9,  3,  &QueryParser::inOp},
         {"MATCH"_sl,   2, 2,  3,  &QueryParser::matchOp},
         {"BETWEEN"_sl, 3, 3,  3,  &QueryParser::betweenOp},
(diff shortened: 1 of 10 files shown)


LiteCore/Storage/UnicodeCollator.cc
```diff
@@ -50,6 +50,48 @@ namespace litecore {
         });
     }
 
+    bool LikeUTF8(slice comparand, slice pattern, const Collation& col) {
```
snej: It doesn't look like this handles backtracking in the pattern, i.e. false matches. For example, can it handle `'(oo)oo)' LIKE '(%)' ?

LiteCore/Query/QueryParserTables.hh
```diff
@@ -165,9 +165,10 @@ namespace litecore {
         {"regexp_like"_sl,      2, 2},
         {"regexp_position"_sl,  2, 2},
         {"regexp_replace"_sl,   3, 9},
+        {"fl_like"_sl,          2, 3, nullslice, false, true},
 
         // Strings:
-        {"contains"_sl,         2, 2},
+        {"contains"_sl,         2, 3, nullslice, false, true},
```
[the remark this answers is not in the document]
borrrden: It's not invisible though, it's passed in as the collation "sqlite name" in a third argument (see `fl_like` for example:  

```
if(argc > 2) {
    col.readSQLiteName((const char *)sqlite3_value_text(argv[2]));
}
```

)

LiteCore/Query/SQLiteN1QLFunctions.cc
```diff
@@ -505,10 +505,34 @@ namespace litecore {
...
+        if(argc > 2) {
+            col.readSQLiteName((const char*)sqlite3_value_text(argv[2]));
+        }
+
         auto str = stringSliceArgument(argv[0]);
         auto substr = stringSliceArgument(argv[1]);
-        if (str)
-            sqlite3_result_int(ctx, str.find(substr).buf != nullptr);
+        auto current = substr;
+        while(str.size > 0) {
```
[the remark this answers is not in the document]
borrrden: Good idea, I'll have to look into the Windows variant.  `ICU` could use `usearch_first`

LiteCore/Storage/UnicodeCollator.cc
borrrden: You mean % is supposed to be greedy?

LiteCore/Query/QueryParserTables.hh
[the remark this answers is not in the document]
borrrden: Nevermind, I got it now!

LiteCore/Storage/UnicodeCollator.cc
```diff
@@ -22,12 +22,20 @@
 #include "StringUtil.hh"
 #include <sqlite3.h>
 #include <algorithm>
+#define CBL_MATCH             0
+#define CBL_NOMATCH           1
+#define CBL_NOWILDCARDMATCH   2
```
snej: These should be declared in the header, since they're used as the return value.
Also, it'd be better to give them CamelCase names and put them in an enum.

LiteCore/Storage/UnicodeCollator.cc
```diff
@@ -50,6 +58,59 @@ namespace litecore {
         });
     }
 
+    int LikeUTF8(slice comparand, slice pattern, const Collation& col) {
```
snej: It's a good idea to add a comment saying where this code came from (inside SQLite, right?)

PR #853, opened 2019-10-01
CBL-430 Hide non-exported symbols on Linux

borrrden:
Via gnu linker file, and add a script to generate the list of symbols to export for each platform

### DIFF
--- a/C/scripts/generate_export_list.py
+++ b/C/scripts/generate_export_list.py
@@ -1,0 +1,85 @@
+import os
+
+def write_def_file(cbl_export_list, c4_export_list):
+    with open(output_dir + os.sep +  'c4.def', 'w') as fout:
+        fout.write("EXPORTS\n\n")
+        fout.write("; Couchbase Lite .NET\n")
+        for cbl_export in cbl_export_list:
+            fout.write("{0}\n".format(cbl_export))
+
+        fout.write("; C4Tests\n")
+        for c4_export in c4_export_list:
+            fout.write("{0}\n".format(c4_export))
+
+        if(os.path.isfile('c4_def.txt')):
+            fout.write("\n; Windows Specific\n")
+            for line in open('c4_def.txt', 'r'):
+                fout.write(line)
+
+        fout.write("\n")
+
+def write_exp_file(cbl_export_list, c4_export_list):
+    with open(output_dir + os.sep + 'c4.exp', 'w') as fout:
+        fout.write("# Couchbase Lite .NET\n")
+        for cbl_export in cbl_export_list:
+            if len(cbl_export) > 0:
+                fout.write("_{0}\n".format(cbl_export))
+            else:
+                fout.write("\n")
+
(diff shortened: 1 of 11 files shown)


cmake/platform_apple.cmake
```diff
@@ -64,17 +64,10 @@ function(setup_support_build)
 endfunction()
 
 function(setup_rest_build)
+    setup_rest_build_unix()
+    
     set_target_properties(
         LiteCoreREST PROPERTIES LINK_FLAGS
         "-exported_symbols_list ${CMAKE_CURRENT_SOURCE_DIR}/c4REST.exp"
     )
-
```
jayahariv: assuming this was resolved only by importing to apple platform, but now seems to be imported to all platforms, and how did that resolved?

C/scripts/generate_export_list.py
```diff
@@ -0,0 +1,85 @@
+import os
```
jayahariv: is this not used now?

cmake/platform_apple.cmake
borrrden: This one is less relevant because it is not a product that we ship, it is only used interally.  The main target of this PR is LiteCore (not LiteCoreREST)

C/scripts/generate_export_list.py
borrrden: It's used several times in the script (any time you see a line that uses `os`)

jayahariv: sorry, i mean whole script. `generate_export_list.py`

borrrden: It is used to keep the three export files in sync.  The new process should be edit c4.txt, then run the python script to generate the three export lists.  This eliminates forgetting to update one or more of them.

jayahariv: whenever we add a new file to lite-core, we will adding them to c4.txt, and run the script `generate_export_list.py`.

borrrden: Not a new file, necessarily, but when you want to add an exported function.  Only functions in this list can be used by consumers of the shared LiteCore library, the rest are hidden.

cmake/platform_linux_desktop.cmake
```diff
@@ -110,5 +110,5 @@ function(setup_support_build)
 endfunction()
 
 function(setup_rest_build)
-    # No-op
+    setup_rest_build_unix()
```
jayahariv: this makes the linux_desktop.cmake fails with to build LiteCoreREST project

C/tests/cmake/platform_linux.cmake
```diff
@@ -24,6 +24,11 @@ function(setup_build)
     target_link_libraries(
         C4Tests PRIVATE
         pthread
+        FleeceBase
```
jayahariv: i believe this caused the C4Tests to fail in linux_desktop

PR #855, opened 2019-10-08
CBL-455 [API] Pending document ID APIs

borrrden:
c4repl_isDocumentPending and c4repl_pendingDocIDs

### DIFF
--- a/C/include/c4Replicator.h
+++ b/C/include/c4Replicator.h
@@ -208,6 +208,26 @@ extern "C" {
     /** Returns the HTTP response headers as a Fleece-encoded dictionary. */
     C4Slice c4repl_getResponseHeaders(C4Replicator *repl C4NONNULL) C4API;
 
+    /** Gets a fleece encoded list of IDs of documents who have revisions pending push.  This
+     *  API is a snapshot and results may change between the time the call was made and the time
+     *  the call returns.
+     *
+     *  @param outErr Records error information, if any
+     *  @return A fleece encoded array of document IDs, each of which has one or more pending
+     *  revisions
+     */
+    C4SliceResult c4repl_getPendingDocIDs(C4Replicator* repl C4NONNULL, C4Error* outErr) C4API;
+
+    /** Checks if the document with the given ID has revisions pending push.  This
+     *  API is a snapshot and results may change between the time the call was made and the time
+     *  the call returns.
+     *
+     * @param docID The ID of the document to check
+     * @param outErr Records error information, if any
+     * @return true if the document has one or more revisions pending, false otherwise
+     */
+    bool c4repl_isDocumentPending(C4Replicator* repl C4NONNULL, C4String docID, C4Error* outErr) C4API;
+
 
 #pragma mark - COOKIES:
 
--- a/Replicator/Pusher.hh
+++ b/Replicator/Pusher.hh
@@ -42,6 +42,16 @@ namespace litecore { namespace repl {
             _checkpointValid = false;
         }
 
+        // Checks if a given document should be filtered by this pusher.  The revision body
+        // must be loaded prior to this call.
+        bool documentShouldBeFiltered(C4Document* doc) const {
+            const auto body = doc->selectedRev.body;
+            Assert(body.buf);
+            const auto bodyContent = FLValue_FromData(body, kFLTrusted);
+            return _options.pushFilter
+                && _options.pushFilter(doc->docID, doc->selectedRev.flags, FLValue_AsDict(bodyContent), _options.callbackContext);
+        }
+
     private:
         void _start(C4SequenceNumber sinceSequence);
         bool passive() const                         {return _options.push <= kC4Passive;}
--- a/Replicator/Replicator.cc
+++ b/Replicator/Replicator.cc
@@ -169,7 +169,7 @@ namespace litecore { namespace repl {
             }
 
             Encoder retEncoder;
-            retEncoder.beginArray(dbLastSequence - replLastSequence);
+            retEncoder.beginArray(size_t(dbLastSequence - replLastSequence));
             C4DocumentInfo info;
             outErr->code = 0;
             while(c4enum_next(e, outErr)) {
--- a/Replicator/Replicator.cc
+++ b/Replicator/Replicator.cc
@@ -147,6 +147,95 @@ namespace litecore { namespace repl {
         _disconnect(websocket::kCodeNormal, {});
     }
 
+    alloc_slice Replicator::pendingDocumentIDs(C4Error* outErr) const {
+        if(!_pusher) {
+            // Couchbase Lite should not allow this case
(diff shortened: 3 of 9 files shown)


Replicator/Replicator.hh
```diff
@@ -83,6 +83,11 @@ namespace litecore { namespace repl {
         void start(bool synchronous =false); 
         void stop()                             {enqueue(&Replicator::_stop);}
 
+        /** Returns a fleece encoded list of the IDs of documents which have revisions pending push */
+        alloc_slice pendingDocumentIDs(C4Error* outErr) const;
+
+        /** Checks if the document with the given ID has any pending revisions */
```
jayahariv:
```suggestion
        /** Checks if the document with the given ID has any pending revisions to push*/
```

Replicator/Replicator.cc
```diff
@@ -147,6 +147,95 @@ namespace litecore { namespace repl {
...
+            // Couchbase Lite should not allow this case
+            return nullslice;
+        }
+
+        return _db->use<alloc_slice>([this, outErr](C4Database* db)
+        {
+            const auto dbLastSequence = c4db_getLastSequence(db);
+            const auto replLastSequence = _checkpoint.sequences().local;
+            if(replLastSequence >= dbLastSequence) {
+                return alloc_slice(nullslice);
```
jayahariv: is this how an empty encoded fleece array created?

borrrden: No, this is a null value.  It's easier than encoding an empty array (which needs an FLEncoder)

borrrden:
> how do you think about adding more corner cases in the unit tests? like
> 
>     1. puller only returns empty
> 
>     2. when finished pushing, return empty array

1. Seems like not worth it.  I'll call it "undefined behavior".  CBL should not allow it to be called this way.

2. This is already checked in the test (by checking if it is a null slice).

Replicator/Replicator.cc
jayahariv: but then whats the difference between `return nullslice;` and `return alloc_slice(nullslice);`

borrrden: Just to satisfy the return type of this method since `nullslice` is not of type `alloc_slice`

Replicator/Replicator.cc
```diff
@@ -147,6 +147,95 @@ namespace litecore { namespace repl {
...
+                    if(!c4doc_loadRevisionBody(nextDoc, outErr)) {
+                        Warn("Error loading revision body in pending document IDs (%d / %d)", outErr->domain, outErr->code);
+                        continue;
+                    }
+
+                    if(!_pusher->documentShouldBeFiltered(nextDoc)) {
+                        retEncoder.writeString(nextDoc->docID);
+                    }
+                } else {
+                    const bool success = c4enum_getDocumentInfo(e, &info);
```
snej: I don't think there's any need to call this. You don't use any of the info, and the function won't possibly return false unless `e` is NULL.

Replicator/Replicator.cc
```diff
@@ -147,6 +147,95 @@ namespace litecore { namespace repl {
...
+                    }
+
+                    retEncoder.writeString(info.docID);
+                }
+            }
+
+            retEncoder.endArray();
+            FLError flErr;
+            auto retVal = retEncoder.finish(&flErr);
+            if(retVal == nullslice) {
```
snej: I don't usually bother with error checking here, TBH. It only happens if you've messed up the structure, like not nesting begin/end calls correctly, not due to any runtime issue.

snej:
I think this also needs to take `kC4ReplicatorOptionDocIDs` into account, as another type of filtering. I.e. ignore all docs that aren't in that set, if it exists.

Replicator/Replicator.cc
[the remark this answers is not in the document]
borrrden: Compiler nuance.  At least on Windows, it won't let you return a different type than the one specified in the template arguments.  If I change the closure to something else then it will complain about the return type of the closure itself.  This seemed like the neatest way.

Replicator/Pusher.hh
```diff
@@ -42,6 +42,16 @@ namespace litecore { namespace repl {
             _checkpointValid = false;
         }
 
+        // Checks if a given document should be filtered by this pusher.  The revision body
+        // must be loaded prior to this call.
+        bool documentShouldBeFiltered(C4Document* doc) const {
+            const auto body = doc->selectedRev.body;
+            Assert(body.buf);
+            const auto bodyContent = FLValue_FromData(body, kFLTrusted);
+            return _options.pushFilter
```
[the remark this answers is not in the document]
borrrden: Taken care of in the new PR.

C/include/c4Replicator.h
```diff
@@ -208,6 +208,26 @@ extern "C" {
...
+     */
+    C4SliceResult c4repl_getPendingDocIDs(C4Replicator* repl C4NONNULL, C4Error* outErr) C4API;
+
+    /** Checks if the document with the given ID has revisions pending push.  This
+     *  API is a snapshot and results may change between the time the call was made and the time
+     *  the call returns.
+     *
+     * @param docID The ID of the document to check
+     * @param outErr Records error information, if any
+     * @return true if the document has one or more revisions pending, false otherwise
```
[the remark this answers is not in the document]
borrrden: Taken care of in the new PR as well.

Replicator/Replicator.cc
```diff
@@ -147,6 +147,95 @@ namespace litecore { namespace repl {
         _disconnect(websocket::kCodeNormal, {});
     }
 
+    alloc_slice Replicator::pendingDocumentIDs(C4Error* outErr) const {
```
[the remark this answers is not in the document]
borrrden: Moved this in the new PR.

Replicator/Replicator.cc
```diff
@@ -147,6 +147,95 @@ namespace litecore { namespace repl {
...
+
+        return _db->use<alloc_slice>([this, outErr](C4Database* db)
+        {
+            const auto dbLastSequence = c4db_getLastSequence(db);
+            const auto replLastSequence = _checkpoint.sequences().local;
+            if(replLastSequence >= dbLastSequence) {
+                return alloc_slice(nullslice);
+            }
+
+            C4EnumeratorOptions opts { kC4IncludeNonConflicted | kC4IncludeDeleted };
```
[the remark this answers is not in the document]
borrrden: Added this as well as made it work with docIDs in the new PR.

Replicator/Replicator.cc
borrrden: I use it to get the docID here, as well as the sequence in the new PR.

Replicator/Replicator.cc
borrrden: Noted, and adjusted in the new PR.

PR #856, opened 2019-10-09
2.6.2 release candidate

bmeike:
CBL-468: Null dereference during pull-only replication

_pusher will be null if push is disabled

Change-Id: I2f1686e8785fb74439a53e15e002ca037245cb77

### DIFF
--- a/Xcode/xcconfigs/Project.xcconfig
+++ b/Xcode/xcconfigs/Project.xcconfig
@@ -15,7 +15,7 @@ GCC_WARN_ABOUT_MISSING_FIELD_INITIALIZERS = NO
 GCC_WARN_SIGN_COMPARE = NO
 GCC_TREAT_WARNINGS_AS_ERRORS                       = YES
 
-LITECORE_VERSION_STRING                            = 2.6.1
+LITECORE_VERSION_STRING                            = 2.6.2
 LITECORE_BUILD_NUMBER                              = 0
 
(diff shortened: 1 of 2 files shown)


bmeike:
I believe Jim wrote a test to validate this change.  I don't know where it is.  It would be neat to get it into this release...

Xcode/xcconfigs/Project.xcconfig
```diff
@@ -15,7 +15,7 @@ GCC_WARN_ABOUT_MISSING_FIELD_INITIALIZERS = NO
 GCC_WARN_SIGN_COMPARE = NO
 GCC_TREAT_WARNINGS_AS_ERRORS                       = YES
 
-LITECORE_VERSION_STRING                            = 2.6.1
+LITECORE_VERSION_STRING                            = 2.6.2
```
jayahariv: i was under the impression, this is only used by iOS.

bmeike: Yeah... I don't know how that number is used.

Core has changed: any Cobalt build (including iOS) will now be different than it was for 2.6.1.  I should think that would require updating that number.

Am I mistaken?

snej: This string gets written into the Info.plist if LiteCore is built as a (iOS/Mac) framework. That's all. Not a bad idea to keep it updated, but it won't affect our build process.

PR #859, opened 2019-10-18
MB-36238: Numerous build cleanups, partly for Centos 6

ceejatec:
- Use CMake tools to find zlib, rather than depending on compiler
  (because on Centos 6 we need to use version from cbdeps)
- Find additional ICU libs
- Fix issue looking up git commit SHA
- Add -D__STDC_FORMAT_MACROS
- Only use cbdeps-provided headers for libcxx if they don't exist on system
  (because Centos 6 build slave provides them and gets confused if we use
  a different set)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -64,13 +64,14 @@ option(LITECORE_DISABLE_ICU "Disables ICU linking" OFF)
 option(DISABLE_LTO_BUILD "Disable build with Link-time optimization" OFF)
 
 add_definitions(
-    -DCMAKE                             # Let the source know this is a CMAKE build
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -65,6 +65,7 @@ option(DISABLE_LTO_BUILD "Disable build with Link-time optimization" OFF)
 
 add_definitions(
     -DCMAKE                             # Let the source know this is a CMAKE build
+    -D__STDC_FORMAT_MACROS
```
borrrden: I'd prefer to have a comment here to remind others of what this actually does.  "Enables printf format macros for variable sized types (e.g. size_t)"

PR #860, opened 2019-10-18
CBL-123: Keep track of the retry count of pushable revisions

borrrden:
At least in some cases.  The goal here is to have certain error types being transient a limited number of times, but permanent after that.

### DIFF
--- a/Replicator/Pusher.cc
+++ b/Replicator/Pusher.cc
@@ -641,6 +641,21 @@ namespace litecore { namespace repl {
         }
     }
 
+    void Pusher::afterEvent() {
+        Worker::afterEvent();
+
+        if(!_revsToRetry.empty()) {
+            _caughtUp = false;
+            auto revsToRetry = move(_revsToRetry);
+            _revsToRetry.clear();
+            for(const auto& revToRetry : revsToRetry) {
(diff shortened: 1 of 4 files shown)


Replicator/Pusher.cc
```diff
@@ -650,6 +658,18 @@ namespace litecore { namespace repl {
...
+            level = kC4Busy;
+            Pusher* pusher = const_cast<Pusher *>(this);
+            pusher->_caughtUp = false;
+            auto revsToRetry = move(_revsToRetry);
+            pusher->_revsToRetry.clear();
+            for(const auto& revToRetry : revsToRetry) {
+                pusher->_pushingDocs[revToRetry->docID] = revToRetry;
+            }
+            
+            pusher->gotChanges(make_shared<RevToSendList>(revsToRetry), _maxPushedSequence, {});
```
jayahariv: not much context, but function name suggest`computeActivityLevel` is meant to compute the activity level and no retry handling. is it possible to separate this method out?

borrrden: I can make it into another method but this function is called often from many different places and so it's not like I can bring it "up one level" or something like that.

PR #866, opened 2019-10-30
CBL-242 Switch to CRC32C for revision ID

borrrden:
Switching the revision ID algorithm from SHA-1 to CRC32-C.  This revision ID is not required to be secure, so this algorithm will be faster.  It is especially fast in the presence of SSE4.2.  It also has hardware instructions on armv8 but it is still unclear how to compile them in a safe way (to check for older devices that may not support the CRC extensions).  I will leave in the implementation because by default the compiler defines are off and thus it is compiled out.  

I have a feeling that we won't be able to take advantage of the armv8 instructions until we raise our minimum versions of Android / iOS in the future.


adamcfraser:
@borrrden Can you confirm whether the body used when generating the rev id on CBL has any special properties?  (_deleted, _attachments, etc)?

borrrden:
If any of those properties are present in the body then they will be used.  `_deleted` is not used anymore but if it is present from before it will be used.  The body is fed through as a collection of bytes representing the fleece data without any analysis.

PR #867, opened 2019-10-31
CBL-429 Add a way to track logically grouped mailbox calls

borrrden:
Log out the progression of calls when there is an error.  This PR depends on https://github.com/couchbaselabs/BLIP-Cpp/pull/18

### DIFF
--- a/LiteCore/Database/LiveQuerier.cc
+++ b/LiteCore/Database/LiveQuerier.cc
@@ -52,7 +52,7 @@ namespace litecore {
 
     void LiveQuerier::run(Query::Options options) {
         _lastTime = clock::now();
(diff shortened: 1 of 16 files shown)


borrrden:
This will be useful for debugging a freeze or an exception (but not a straight crash) because if you are frozen, you can attach a debugger and see if there is a stuck actor and if so follow the chain of messages back through the mailbox.  On an exception these will be written to the logs (up to 100 frames)

borrrden:
The closer the cutoff date gets the less I want to merge this into Mercury.  This change is somewhat impactful so if I don't hear from @snej by the end of the week I will close this and reopen it later.

borrrden:
Deferring until Hydrogen

PR #869, opened 2019-11-01
Generate doc revID based on canonical JSON, not Fleece

snej:
It doesn't actually create the JSON data in memory; it just accumulates the bytes into a CRC32 digest as they're emitted by the JSON encoder. (This required [a change in Fleece](https://github.com/couchbaselabs/fleece/commit/810029b2c1da442dbe217b9fa486c23063d1ba0c).)


borrrden:
This is getting pushed, unfortunately, so closing it until later.

PR #871, opened 2019-11-02
Skip upgrading docs with underscore prefix

benjaminglatzeder:
Any document ID which starts with an underscore will be skipped when
upgrading the database from CBL 1.x to 2.x.

Discussion and suggestion for this change can be found here:
https://forums.couchbase.com/t/doc-with-invalid-id-from-cbl-1-x-crashes-app-when-upgrading-to-cbl-2-x/23534/11


snej:
Issue filed as [CBL-525](https://issues.couchbase.com/browse/CBL-525).

snej:
Thanks!

PR #873, opened 2019-11-05
Pending Document ID API fixes

borrrden:
Based on testing with iOS platform

### DIFF
--- a/Replicator/Pusher.cc
+++ b/Replicator/Pusher.cc
@@ -73,28 +73,15 @@ namespace litecore { namespace repl {
 
 
     bool Pusher::isDocumentAllowed(C4Document* doc) const {
-        if(!isDocumentIDAllowed(doc->docID)) {
-            return false;
-        }
-
-        if(!_options.pushFilter) {
-            return true;
-        }
-
-        const auto body = doc->selectedRev.body;
-        Assert(body.buf);
-        const auto bodyContent = FLValue_FromData(body, kFLTrusted);
-        return _options.pushFilter(doc->docID, doc->selectedRev.flags, FLValue_AsDict(bodyContent), _options.callbackContext);
+        return isDocumentIDAllowed(doc->docID)
+            && (!_options.pushFilter || _options.pushFilter(doc->docID, doc->selectedRev.flags,
+                                                            DBAccess::getDocRoot(doc),
+                                                            _options.callbackContext));
     }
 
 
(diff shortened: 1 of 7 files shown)


Replicator/Pusher.hh
```diff
@@ -45,13 +46,36 @@ namespace litecore { namespace repl {
         // Checks if a given document should be filtered by this pusher.  The revision body
         // must be loaded prior to this call.
         bool documentShouldBeFiltered(C4Document* doc) const {
```
snej: I find it confusing that this method's return value has the opposite meaning from the filter functions. Could you reverse it to be consistent? I.e. return true to use the doc, false to skip it.

Replicator/Pusher.hh
```diff
@@ -45,13 +46,36 @@ namespace litecore { namespace repl {
...
+            if(!isDocumentIDAllowed(doc->docID)) {
```
snej: Pusher.hh didn't previously have any nontrivial inline methods, so please move these bodies into the .cc file for consistency.

Replicator/Replicator.cc
```diff
@@ -514,7 +426,7 @@ namespace litecore { namespace repl {
 
 
     // Start off by getting the local checkpoint, if this is an active replicator:
-    void Replicator::getLocalCheckpoint() {
+    void Replicator::getLocalCheckpoint(bool noRemote) {
```
snej: Since there's only one existing caller of `getLocalCheckpoint`, I think it'd be clearer to remove the `getRemoteCheckpoint` call from the end of this method, and add it to there. Then you don't need to add a parameter. (It's already kind of weird that I had `getLocalCheckpoint` get the remote checkpoint too. No idea why I did that...)

Replicator/Pusher.cc
```diff
@@ -72,6 +72,32 @@ namespace litecore { namespace repl {
...
+    bool Pusher::isDocumentAllowed(C4Document* doc) const {
+        if(!isDocumentIDAllowed(doc->docID)) {
+            return false;
+        }
+
+        if(!_options.pushFilter) {
+            return true;
+        }
+
+        const auto body = doc->selectedRev.body;
```
snej: For consistency, call `DBAccess::getDocRoot(doc)` instead of doing the work yourself.

snej: Did it myself :)

PR #875, opened 2019-11-05
CBL-521 Add revID to replicator filter callback

borrrden:


borrrden:
I am not sure passing the c4doc here is a good idea because it does not belong to the document instance, and the document will try to free it.  That will lead to bad things.

Replicator/Pusher+DB.cc
```diff
@@ -209,7 +209,7 @@ namespace litecore { namespace repl {
             }
 
             if (_options.pushFilter) {
-                if (!_options.pushFilter(doc->docID, doc->selectedRev.flags,
+                if (!_options.pushFilter(doc->docID, doc->revID, doc->selectedRev.flags,
```
snej: should be `doc->selectedRev.revID`

Replicator/Pusher.hh
```diff
@@ -49,7 +49,7 @@ namespace litecore { namespace repl {
             Assert(body.buf);
             const auto bodyContent = FLValue_FromData(body, kFLTrusted);
             return _options.pushFilter
-                && _options.pushFilter(doc->docID, doc->selectedRev.flags, FLValue_AsDict(bodyContent), _options.callbackContext);
+                && _options.pushFilter(doc->docID, doc->revID, doc->selectedRev.flags, FLValue_AsDict(bodyContent), _options.callbackContext);
```
snej: should be `doc->selectedRev.revID`

Replicator/Pusher+DB.cc
borrrden: What is the difference?  I thought this was shorthand for that.

PR #876, opened 2019-11-06
Don't leak objects if C4Replicator is released w/o being started

snej:
Make sure everything is torn down cleanly if a C4Replicator is
released without having ever been started. There are reference
cycles in the underlying objects that need to be untangled.
Fixes CBL-524

Part of the fix is in BLIP-Cpp -- [here's the commit](https://github.com/couchbaselabs/BLIP-Cpp/commit/8bbba50dcb3d023eb07df0f0649389e7c8c003a9)

### DIFF
--- a/Replicator/Replicator.cc
+++ b/Replicator/Replicator.cc
@@ -147,6 +147,20 @@ namespace litecore { namespace repl {
         _disconnect(websocket::kCodeNormal, {});
     }
 
+
+    void Replicator::terminate() {
+        auto conn = connection();
+        if (conn) {
+            Assert(_connectionState == Connection::kClosed);
+            conn->terminate();
+            _delegate = nullptr;
+            _pusher = nullptr;
+            _puller = nullptr;
+            _db.reset();
+        }
+    }
+
+
     alloc_slice Replicator::pendingDocumentIDs(C4Error* outErr) const {
         if(!_pusher) {
             // Couchbase Lite should not allow this case
--- a/Replicator/tests/ReplicatorAPITest.cc
+++ b/Replicator/tests/ReplicatorAPITest.cc
@@ -115,6 +115,25 @@ TEST_CASE("URL Generation") {
 }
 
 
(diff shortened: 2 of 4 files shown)


Replicator/Replicator.cc
```diff
@@ -147,6 +147,20 @@ namespace litecore { namespace repl {
         _disconnect(websocket::kCodeNormal, {});
     }
 
+
+    void Replicator::terminate() {
+        auto conn = connection();
+        if (conn) {
```
borrrden: Just to confirm:  This is a safe assumption right?  Either we've never started and the connection is still closed, or we've already stopped and there is no connection anymore.  There are no other scenarios?

Replicator/tests/ReplicatorAPITest.cc
```diff
@@ -115,6 +115,25 @@ TEST_CASE("URL Generation") {
...
+    params.push = kC4OneShot;
+    params.pull = kC4Disabled;
+    params.callbackContext = this;
+    params.socketFactory = _socketFactory;
+    params.dontStart = true;
+
+    _repl = c4repl_new(db, _address, kC4SliceNull, (C4Database*)db2,
+                       params, &err);
+    C4Log("---- Releasing C4Replicator ----");
+    _repl = nullptr;
```
borrrden: Not sure this will add to the test since this will get released later.  Is there something visible that happens when it is done early here?

Replicator/Replicator.cc
snej: Well, it's a precondition (see the header comment.) We know this condition is true in the C4Replicator destructor, because C4Replicator retains itself while the replicator is running, so once it's been started it can't be destructed until after the Replicator has stopped.

Replicator/tests/ReplicatorAPITest.cc
snej: Mostly I just wanted to have an explicit, visible place where it's being released, rather than down inside auto-generated code in the test destructor. It made it easier to set breakpoints.

PR #879, opened 2019-11-08
CBL-444 Add a purge count to the database

borrrden:
Use it in conjunction with last sequence in order to properly refresh a C4Query after a purge operation on the database

### DIFF
--- a/LiteCore/Storage/SQLiteDataFile.cc
+++ b/LiteCore/Storage/SQLiteDataFile.cc
@@ -193,21 +193,22 @@ namespace litecore {
                      "BEGIN; "
                      "CREATE TABLE IF NOT EXISTS "      // Table of metadata about KeyStores
                      "  kvmeta (name TEXT PRIMARY KEY, lastSeq INTEGER DEFAULT 0, purgeCnt INTEGER DEFAULT 0) WITHOUT ROWID; "
+                     "PRAGMA user_version=202; "
+                     "END;"
                      );
                 // Create the default KeyStore's table:
                 (void)defaultKeyStore();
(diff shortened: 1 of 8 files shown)


borrrden:
@bmeike This is the LiteCore fix for the query refresh problem.

LiteCore/Storage/SQLiteDataFile.cc
```diff
@@ -192,12 +192,18 @@ namespace litecore {
...
-                     "  kvmeta (name TEXT PRIMARY KEY, lastSeq INTEGER DEFAULT 0) WITHOUT ROWID; "
+                     "  kvmeta (name TEXT PRIMARY KEY, lastSeq INTEGER DEFAULT 0, purgeCnt INTEGER DEFAULT 0) WITHOUT ROWID; "
+                     "PRAGMA user_version=202; "
+                     "END;"
                      );
                 // Create the default KeyStore's table:
                 (void)defaultKeyStore();
-                _exec("PRAGMA user_version=201; "
-                      "END;");
+            } else if(userVersion == 201) {
```
snej: The latest userVersion is actually 301; see SQLiteDataFile+Indexes.cc:49. That code bumps the userVersion to prevent an older LiteCore from opening a database with the `indexes` table. So it's possible for an existing db to have userVersion 201 or 301.

I _think_ the best thing to do here is `if (userVersion <= 301)...` and then set `user_version` to 302.

borrrden: Alright I did that, but I had to make sure it wasn't older than the min version first as the only caveat

PR #880, opened 2019-11-08
update version for mercury

jayahariv:


borrrden:
Is there not a way to automatically do this on the build server instead of manually updating it all the time?

jayahariv:
we could override the [userAgent information](https://github.com/couchbase/couchbase-lite-ios/blob/34d73df44357539a3b380167a79708bd40831eb9/Objective-C/Internal/CBLVersion.m#L50) for the lite core easily from lite side. This means, there won't be any use for `c4_getVersion` API.

@pasin what do you think?

PR #881, opened 2019-11-14
Fixed problems with Pusher::isSequencePending()

snej:
* It wasn't thread-safe, because SequenceSet wasn't
* The logic was wrong in that it would return false for a sequence newer (greater) than any seen by Pusher::gotChanges.


borrrden:
Hmmm let me try the validation again

borrrden:
Ok all green

PR #882, opened 2019-11-14
Optimized pre-pull scan for conflicted docs

snej:
C4DocEnumerator:
* Added option flag for unordered results.
* Conflicts-only option is passed through to the RecordEnumerator.
* Stripped out unused/unneeded functionality, like filtering.

RecordEnumerator:
* Added options for conflicts-only & unordered results.
* With Debug-level query logging, it explains its query plan.

Database:
* Added SQLite partial indexes for conflicted docs and docs with blobs.
* During compaction, enumeration of docs with blobs is now unordered.

Replicator:
* The conflicted-docs scan is unordered, to speed it up.
* Logging before and after the scan, including the time taken.

Fixes [CBL-536](https://issues.couchbase.com/browse/CBL-536) (but hasn't been benchmarked on large dbs!)

### DIFF
--- a/C/c4DocEnumerator.cc
+++ b/C/c4DocEnumerator.cc
@@ -26,7 +26,6 @@
 #include "RecordEnumerator.hh"
 #include "Logging.hh"
 #include "InstanceCounted.hh"
-#include <set>
 
 
 #pragma mark - DOC ENUMERATION:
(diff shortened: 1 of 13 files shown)


C/c4DocEnumerator.cc
```diff
@@ -36,89 +35,51 @@ CBL_CORE_API const C4EnumeratorOptions kC4DefaultEnumeratorOptions = {
...
-        do {
-            if (!_e.next())
-                return false;
-        } while (!useDoc());
-        return true;
-    }
-
     Retained<Document> getDoc() {
-        return _e ? _database->documentFactory().newDocumentInstance(_e.record()) : nullptr;
+        return *this ? _database->documentFactory().newDocumentInstance(record()) : nullptr;
```
borrrden: boolean test for `*this`?  Is there any way to be clearer about what this is actually checking?

LiteCore/Storage/RecordEnumerator.cc
```diff
@@ -82,7 +74,7 @@ namespace litecore {
                 close();
                 return false;
             }
-            LogToAt(EnumLog, Debug, "enum:     --> [%s]", _record.key().hexCString());
+            //LogToAt(QueryLog, Debug, "RecordEnumerator %p  --> '%.*s'", this, SPLAT(_record.key()));
```
borrrden: Is there a reason for leaving this in as a comment?

C/c4DocEnumerator.cc
snej: I know, it's kind of ugly. The class has an `operator bool` to test for EOF, but not a named method. I'll add a named method so it's clearer.

LiteCore/Storage/RecordEnumerator.cc
snej: It's very useful in specific circumstances, when debugging RecordEnumerator as I was yesterday, but it spews an AWFUL LOT of stuff during unit tests. I guess Debug logging isn't going to be on often, though, so I could leave it there.

PR #886, opened 2019-11-20
Implemented N1QL concat() function

snej:
Takes 2+ parameters, converts each to a string and concatenates them.
Returns MISSING if any parameter is MISSING.

The existing '||' operator now translates into a 'concat()' call
because a regular SQLite '||' expression doesn't know what to make of
nulls or collections.

### DIFF
--- a/LiteCore/Query/QueryParser.cc
+++ b/LiteCore/Query/QueryParser.cc
@@ -831,6 +831,11 @@ namespace litecore {
     }
 
 
+    // Handles "x || y", turning it into a call to the concat() function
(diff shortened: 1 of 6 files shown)


snej:
I did this on the dev branch but it can probably be rebased to master without conflicts...

borrrden:
It looks alright but the validation cannot complete because the CMake project has gotten stale on the branch.  This might be an interesting opportunity to clean up the branch so you can have some assurance that all the platforms are still behaving.

borrrden:
PR validation looks good, it just timed out from taking too long since more platforms were introduced recently.

PR #889, opened 2019-11-23
Restore ability to open pre-2.7 dbs read-only

snej:
The schema upgrade required for purgeCount in 2.7 was causing an
exception when opening a database read-only.

* Made some logic changes to skip the upgrade if the db is read-only.
  The purgeCount() method will return 0; it isn't needed in read-only
  mode anyway because the db won't change.
* Also catch + ignore the SQLITE_READONLY exception thrown if the
  db _file_ is read-only.
* Throw a better exception (CantUpgradeDatabase) if the database is
  writeable but not upgradeable.
* Improved readability & reliability with a snazzy custom enum class
  for the schema versions. ^_^
* After all that, updated the test to skip the kC4DB_NoUpgrade mode
  since it still fails. (But at least kC4DB_ReadOnly works now!)

Fixes CBL-562

### DIFF
--- a/LiteCore/Query/SQLiteDataFile+Indexes.cc
+++ b/LiteCore/Query/SQLiteDataFile+Indexes.cc
@@ -42,7 +42,7 @@ namespace litecore {
         if (indexTableExists())
             return;
 
-        if (!options().upgradeable)
+        if (!options().upgradeable && _schemaVersion < SchemaVersion::WithIndexTable)
             error::_throw(error::CantUpgradeDatabase,
                           "Accessing indexes requires upgrading the database schema");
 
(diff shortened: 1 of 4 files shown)


LiteCore/Query/SQLiteDataFile+Indexes.cc
```diff
@@ -42,11 +42,15 @@ namespace litecore {
         if (indexTableExists())
             return;
 
+        if (!options().upgradeable)
+            error::_throw(error::CantUpgradeDatabase,
+                          "Accessing indexes requires upgrading the database schema");
```
borrrden: This is not usually valid.  It's only valid to throw this error if we are opening an older DB (2xx level).  In the usual case, we've created a 301+ database but the table doesn't exist yet.

borrrden:
Something is wrong with the PR validation job mac machine, but this looks good now.

PR #891, opened 2019-11-28
Fixes needed for both dev branch regressions

borrrden:
and building CBL-C for Windows (probably Linux too)


borrrden:
Apologies that this PR is so massive.  It fixes months worth of breakage creep.

LiteCore/Support/Error.cc
```diff
@@ -56,6 +56,87 @@ namespace litecore {
...
+            return win32_message(err);
+        }
+
+        // Hope the POSIX definitions don't change...
+        if(err < 100 || err > 140) {
+            return "Unknown Error";
+        }
+
+        static long wsaEquivalent[] = {
+            WSAEADDRINUSE,
```
snej: Could we just map the WSA codes to the corresponding POSIX codes when the error is detected?

Networking/Poller.cc
```diff
@@ -16,26 +16,37 @@
 // limitations under the License.
 //
 
+#define NOMINMAX
```
snej: What's this for? Could you add a comment? And does it have to be at the very top of the file?0

Networking/Poller.cc
```diff
@@ -16,26 +16,37 @@
...
-#include <poll.h>
 
 #ifndef _WIN32
 #include <unistd.h>
+#include <poll.h>
+#define tcp_read ::read
+#define tcp_write ::write
+#define tcp_poll ::poll(fdArray, nfds_t(fdSize), timeout)
+#else
+#define tcp_read(s, buf, len) ::recv(s, (char *)(buf), len, 0)
```
snej: Since`tcp_read` and `tcp_write` are only called once apiece, I think it'd be cleaner to just use an `#ifdef` at the point of the call, rather than defining these macros. That way it matches the ifdef where we create the interrupting FDs.

Networking/Poller.cc
```diff
@@ -16,26 +16,37 @@
...
+#define tcp_write(s, buf, len) ::send(s, (const char *)(buf), len, 0)
+#define tcp_poll(fdArray, fdSize, timeout) WSAPoll(fdArray, ULONG(fdSize), timeout)
```
snej: `tcp_poll` doesn't seem to be used anywhere. Remove it?

Networking/TCPSocket.hh
```diff
@@ -10,6 +10,7 @@
 #include "HTTPTypes.hh"
 #include "function_ref.hh"
 #include "fleece/Fleece.hh"
+#include "sockpp/socket.h"
```
snej: This doesn't seem necessary since nothing else was added to this header. (I'd like to avoid including sockpp into too many source files.)

Networking/Poller.cc
borrrden: It stops the annoying min and max macros from being defined and clobbering std::min, etc.  I'll make that more clear.

LiteCore/Support/Error.cc
borrrden: That's exactly what happens, but now we need a reverse map to get the actual message from the OS.  Otherwise you just end up with a bunch of "Unknown Error" strings.  Alternatively I guess we could look for places that we map to POSIX, grab the message first, and then save the message with the C4Error.

PR #893, opened 2019-12-18
CBL-594 Correct the URL of LoopbackSocket

borrrden:
For example, db1 -> db2 should use db2 URL as the remote an db1 -> db3 should use db3 URL but they are both using db1 URL


snej:
Ouch. This patch will not need to be merged into the dev branch, because the rewritten C4Replicator class doesn't have this bug (when I rewrote it, I did this the correct way, without noticing it was backwards in the old code.)

PR #894, opened 2019-12-19
CBL-590: Remove HTTP 500 from transient error list

borrrden:
There are quite a few cases where it is permanent (e.g. sync function exception)

### DIFF
--- a/C/c4Base.cc
+++ b/C/c4Base.cc
@@ -234,7 +234,6 @@ bool c4error_mayBeTransient(C4Error err) C4API {
     static CodeList kTransientWebSocket = {
         408, /* Request Timeout */
         429, /* Too Many Requests (RFC 6585) */
-        500, /* Internal Server Error */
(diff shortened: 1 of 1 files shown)


C/c4Base.cc
```diff
@@ -234,7 +234,6 @@ bool c4error_mayBeTransient(C4Error err) C4API {
     static CodeList kTransientWebSocket = {
         408, /* Request Timeout */
         429, /* Too Many Requests (RFC 6585) */
-        500, /* Internal Server Error */
```
Sandychuang8: I wonder if sync gateway should return bad data error (400) instead of 500 for Invalid UTF-8 character?

borrrden: Maybe but that's far from the only exception the function can get.

borrrden:
Travis issues with toolchains (unable to download, a problem on their side I guess) is responsible for the red checkmark here.  Safe to ignore.

PR #899, opened 2020-01-15
API: Added c4queryobs_setEnabled, to prevent a race condition

snej:
After a thread creates an observer, it was possible for the callback
to be called (on another thread) before c4queryobs_create returned.
In this case the app's reference to the observer wouldn't be set yet,
which could cause problems (as in one of our unit tests.)

To prevent this case, a query observer now has to be enabled before
its callback will be invoked.

This API also makes it possible to disable an observer without
freeing it, which could be useful in situations like a backgrounded
app.

### DIFF
--- a/C/c4.def
+++ b/C/c4.def
@@ -345,6 +345,7 @@ c4queryenum_getRowCount
 c4query_fullTextMatched
 
 c4queryobs_create
+c4queryobs_setEnabled
 c4queryobs_getEnumerator
 c4queryobs_free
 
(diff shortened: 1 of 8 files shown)


snej:
Oh, I see — the problem is that the background query still starts when the observer is created. It should start when it's enabled. I'll fix that.

borrrden:
I thought about that, but what if you have two observers on the same BG query?  The second one will face the same issue even after that fix.

snej:
The second observer added isn't guaranteed to get any notifications until the next change. That's how live queries have worked since 1.0.

borrrden:
Really?  That makes things easier then but there should definitely be some documentation around that because I never knew about that!

snej:
I don't think it's ever come up. The LiveQuery object would start the bg query task when the first observer was added, so the first observer would always get the initial notification.

PR #900, opened 2020-01-18
fix build: missed include functional

Dushistov:
Fix compilation under linux/gcc, but I suppose it is not bad to include proper header if code uses `std::function`

### DIFF
--- a/Networking/TCPSocket.hh
+++ b/Networking/TCPSocket.hh
@@ -13,6 +13,7 @@
 #include <memory>
 #include <mutex>
 #include <vector>
+#include <functional>
 
 namespace litecore { namespace crypto {
     class Cert;


borrrden:
Why is it that Travis passed the build without this but yours fails to build?  That seems to defeat the point of having Travis...

Dushistov:
> Why is it that Travis passed the build without this but yours fails to build?

My wild guess that on Ubuntu libstdc++ is separate package, and travis only test with gcc 9.x, not with gcc 9.x + libstdc++ relevant to gcc 9.x .

But this is my development machine with arch linux I can live with build failures for arch linux, 
in productions obviously I use other linux distributions and some kind of CI that validate build against old debian/centos/ubuntu would be not bad idea for users. Though travis is almost out of data and new trend is github CI.

But for me looks like it all becomes irrelevant because of  #901.

borrrden:
I didn't forget about this, but I'm in the middle of switching our CI to use the GNU toolchain going forward and I'm having lots of issues getting it to work correctly (Linux and I are not exactly friends...), but this will certainly be part of the changes

borrrden:
This commit was included as part of #910

PR #906, opened 2020-01-28
Fix null-Checkpoint logging crash

snej:
Only occured in P2P replication (or unit tests), and only when "SyncBusy" logging is enabled.
This is a regression from the dev-branch merge, so won't affect any released versions.

Fixes CBL-643

### DIFF
--- a/Replicator/Checkpointer.cc
+++ b/Replicator/Checkpointer.cc
@@ -109,7 +109,7 @@ namespace litecore { namespace repl {
 
     size_t Checkpointer::pendingSequenceCount() const {
         LOCK();
-        return _checkpoint->pendingSequenceCount();
+        return _checkpoint ? _checkpoint->pendingSequenceCount() : 0;
     }
 
 
@@ -242,13 +242,6 @@ namespace litecore { namespace repl {
 
 
 #pragma mark - READING THE CHECKPOINT:
-
-
-    void Checkpointer::forgetCheckpoint() {
-        LOCK();
-        Assert(!_changed);
-        _checkpoint.reset();
-    }
 
 
     static inline bool isNotFoundError(C4Error err) {
--- a/Replicator/Checkpointer.hh
+++ b/Replicator/Checkpointer.hh
@@ -93,10 +93,6 @@ namespace litecore { namespace repl {
             the remote peer. */
         bool write(C4Database *db NONNULL, slice checkpointData, C4Error *outError);
 
-        /** Clears the in-memory checkpoint; the next call to \ref read will read it from the
-            database again. (Not used by the replicator, only by \ref C4PendingPush.) */
-        void forgetCheckpoint();
-
         // Autosave:
 
         using duration = std::chrono::nanoseconds;


snej:
(BTW, removing the unused `forgetCheckpoint` method is just something I noticed while examining the code ... not related to the fix, but has no effect on the code.)

jayahariv:
does it mean, its fine to have `_checkpoint` instance null? 

Also this is causing a mutex lock failure 
```
* thread #7, queue = 'Push->file:////var/folders/bz/vsvhjks11qncrj3_mzsstvj80000gn/T/CouchbaseLite_EE/otherdb.cblite2/', stop reason = breakpoint 1.1
  * frame #0: 0x00007fff6c94d3ec libc++abi.dylib`__cxa_throw
    frame #1: 0x00007fff6c93f2db libc++.1.dylib`std::__1::__throw_system_error(int, char const*) + 77
    frame #2: 0x00007fff6c8f944d libc++.1.dylib`std::__1::mutex::lock() + 29
    frame #3: 0x0000000109e558a9 CouchbaseLite`std::__1::lock_guard<std::__1::mutex>::lock_guard(this=0x000000010f4a6260, __m=0x000061900007d8b0) at __mutex_base:104:27
    frame #4: 0x0000000109e26c77 CouchbaseLite`std::__1::lock_guard<std::__1::mutex>::lock_guard(this=0x000000010f4a6260, __m=0x000061900007d8b0) at __mutex_base:104:21
    frame #5: 0x0000000109e29998 CouchbaseLite`litecore::repl::Checkpointer::pendingSequenceCount(this=0x000061900007d860) const at Checkpointer.cc:111:9
    frame #6: 0x000000010a15fb46 CouchbaseLite`litecore::repl::Pusher::computeActivityLevel(this=0x0000615000050a80) const at Pusher.cc:662:13
    frame #7: 0x000000010a7cd9d4 CouchbaseLite`litecore::repl::Worker::afterEvent(this=0x0000615000050a80) at Worker.cc:265:25
    frame #8: 0x000000010a15dbfd CouchbaseLite`litecore::repl::Pusher::afterEvent(this=0x0000615000050a80) at Pusher.cc:622:17
    frame #9: 0x000000010b11ea50 CouchbaseLite`litecore::actor::GCDMailbox::afterEvent(this=0x0000615000050a90) at GCDMailbox.cc:126:17
    frame #10: 0x000000010b11e91f CouchbaseLite`::___ZN8litecore5actor10GCDMailbox7enqueueEU13block_pointerFvvE_block_invoke(.block_descriptor=0x00006040000bbf50) at GCDMailbox.cc:102:13
    frame #11: 0x000000010016e6e6 libclang_rt.asan_osx_dynamic.dylib`__wrap_dispatch_async_block_invoke + 214
    frame #12: 0x00007fff6f788583 libdispatch.dylib`_dispatch_call_block_and_release + 12
    frame #13: 0x00007fff6f78950e libdispatch.dylib`_dispatch_client_callout + 8
    frame #14: 0x00007fff6f78eace libdispatch.dylib`_dispatch_lane_serial_drain + 597
    frame #15: 0x00007fff6f78f452 libdispatch.dylib`_dispatch_lane_invoke + 363
    frame #16: 0x00007fff6f798a9e libdispatch.dylib`_dispatch_workloop_worker_thread + 598
    frame #17: 0x00007fff6f9e36fc libsystem_pthread.dylib`_pthread_wqthread + 290
    frame #18: 0x00007fff6f9e2827 libsystem_pthread.dylib`start_wqthread + 15
```

Replicator/Checkpointer.cc
```diff
@@ -109,7 +109,7 @@ namespace litecore { namespace repl {
 
     size_t Checkpointer::pendingSequenceCount() const {
         LOCK();
-        return _checkpoint->pendingSequenceCount();
+        return _checkpoint ? _checkpoint->pendingSequenceCount() : 0;
```
borrrden: What reason would a checkpoint be null?  I would think we should read the checkpoint whenever we need it or we might not get an accurate count.

snej:
> does it mean, its fine to have _checkpoint instance null?

Yes, if it's a passive replicator, as you pointed out.

> Also this is causing a mutex lock failure

What is the exception being thrown?

Replicator/Checkpointer.cc
snej: It's a passive replicator, so it doesn't care about the checkpoint and never loads it.

PR #908, opened 2020-01-29
Fixed two crashes (Checkpointer & C4PredictiveModel)

snej:
1. Don't call `_checkpointer` after the Replicator (which owns it) is gone.
2. Incorrect conversion of DataFileDelegate* to c4Database* (due to the intricacies of C++ multiple inheritance), manifesting as a bogus C4BlobStorage* passed to c4blob_getContents. (CBL-648)

Fixes CBL-648

### DIFF
--- a/C/c4PredictiveQuery.cc
+++ b/C/c4PredictiveQuery.cc
@@ -20,7 +20,7 @@
 //
 
 #include "c4PredictiveQuery.h"
-#include "Database.hh"
+#include "c4Database.hh"
 #include "PredictiveModel.hh"
 
 using namespace litecore;
@@ -40,7 +40,7 @@ public:
         try {
             return _c4Model.prediction(_c4Model.context,
                                        (FLDict)input,
-                                       (c4Database*)dfDelegate,
+                                       dynamic_cast<c4Database*>(dfDelegate),
                                        outError);
         } catch (const std::exception &x) {
             if (outError)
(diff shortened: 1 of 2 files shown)


C/c4PredictiveQuery.cc
```diff
@@ -40,7 +40,7 @@ class C4PredictiveModelInternal : public PredictiveModel {
         try {
             return _c4Model.prediction(_c4Model.context,
                                        (FLDict)input,
-                                       (c4Database*)dfDelegate,
+                                       dynamic_cast<c4Database*>(dfDelegate),
```
bmeike: What is the distinction, here?  Is dynamic cast checked at runtime?

borrrden: dynamic cast will set the element to null if it does not exceed, so I guess the answer to your question is "yes" ?

bmeike: Oh!  That's interesting!  If it fails the value is null!
Is that what we want here?  What are the circumstances under which that thing isn't a c4Database?

pasin: I guess we don't expect that to be null here.

snej: The reason I added it is that `dynamic_cast<T>` will fail to compile if it hasn't seen the full declaration of `T`. The bug was that `c4Database` had been forward-referenced but not declared in this source file, because it was including `Database.hh` not `c4Database.hh`.

(The reason that's a bug is that casting from `DataFile::Delegate*` to `c4Database*` requires subtracting an offset from the pointer value, because the former is not the first base class of the latter. But the compiler doesn't know to do that unless it's seen the declaration.)

borrrden: That definitely seems comment worthy

PR #913, opened 2020-02-07
API: Added c4log_enableFatalExceptionBacktrace()

snej:
C API wrapper around Backtrace::installTerminateHandler(), so the bindings can call it. It writes the backtrace to the log. (I had to add a callback to the Backtrace API, because Backtrace is in Fleece and doesn't know about LiteCore's log API.)


borrrden:
Windows build is failing 👎

snej:
Yup, changed PURE --> FLPURE in Fleece.
Merging this PR by hand since Fleece has also changed...

PR #915, opened 2020-02-12
Fixed BuiltinWebSocket leak

snej:
The ad-hoc self-retain/release calls in BuiltinWebSocket didn't work in all circumstances, causing leaks.
Use a Retained<> instance instead, which is safer, and clear it in closeWithError() which is always called on close.

### DIFF
--- a/Networking/BuiltInWebSocket.cc
+++ b/Networking/BuiltInWebSocket.cc
@@ -95,7 +95,7 @@ namespace litecore { namespace websocket {
     void BuiltInWebSocket::connect() {
         // Spawn a thread to connect and run the read loop:
         WebSocketImpl::connect();
-        retain(this);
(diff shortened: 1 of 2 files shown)


snej:
The Windows unit tests failed, but the only sign of the failure is
```
d:\agent\_work\2\s\src\vctools\crt\crtw32\stdcpp\thr\mutex.c(51): mutex destroyed while busy
```
which presumably aborted the process, but there's no backtrace. I've looked through the code but can't figure out where an object gets freed while its mutex is held. Any way to get Windows to log a backtrace, @borrrden ?

borrrden:
In general Windows doesn't "log backtraces" but rather creates dump files.  Unfortunately, Catch intercepts the process before this happens, or so it seems.  This is the equivalent to a SIGSEGV on Apple though (a destroyed object was likely accessed)

PR #916, opened 2020-02-13
Put deleted docs in a separate KeyStore

snej:
* Create a new KeyStore named `deleted` (SQLite table `kv_deleted`.)
* All docs with the 'deleted' flag set are saved to this table and deleted from `kv_default`. And vice-versa if the flag is removed.
* A new KeyStore subclass `BothKeyStore` is used for the defaultKeyStore. It acts as the union of the two underlying KeyStores, operating on one or both as necessary.
* Queries no longer need to add "`... AND (flags & 1)=0`" to the `WHERE` clause.
* But queries for deleted docs do need to `JOIN` the `kv_deleted` table.
* This is a schema change, of course. On upgrade the new table is created and all deleted docs moved to it.

CBL-219


snej:
Yay, the Jenkins tests seem more stable now.

Still, this is a pretty nontrivial change, so I'm thinking of waiting till the build's been stable longer before letting this loose on it.

borrrden:
Is this done / going to be done on the dev branch?

PR #918, opened 2020-02-14
Fixed C4QueryObserver (LiteCore's LiveQuery)

snej:
I reported the issue as "C4QueryObserver does not trigger when a doc is purged", but that turned out not to be the case; the problem is a race condition that makes it miss some changes. To fix that I changed the way that it gets triggered: now it's only on transaction commits, not on individual changes.

**NOTE:** It's probably best to read through the individual commits rather than the merged view.

Fixes CBL-688

### DIFF
--- a/C/c4PredictiveQuery.cc
+++ b/C/c4PredictiveQuery.cc
@@ -1,5 +1,3 @@
-#ifdef COUCHBASE_ENTERPRISE
-
 //
 // c4PredictiveQuery.cc
 //
@@ -20,6 +18,9 @@
 //
 
 #include "c4PredictiveQuery.h"
+
+#ifdef COUCHBASE_ENTERPRISE
+
 #include "c4Database.hh"
 #include "PredictiveModel.hh"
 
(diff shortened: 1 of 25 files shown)


borrrden:
I'll remove myself for now but when the tests are settled then add me back.

C/c4PredictiveQuery.cc
```diff
@@ -59,15 +60,23 @@ class C4PredictiveModelInternal : public PredictiveModel {
...
 
+#endif // COUCHBASE_ENTERPRISE
+
 
 void c4pred_registerModel(const char *name, C4PredictiveModel model) C4API {
+#ifdef COUCHBASE_ENTERPRISE
     auto context = retained(new C4PredictiveModelInternal(model));
     context->registerAs(name);
+#else
+    abort();
```
jayahariv: when fails, should we need to include an error message for easy debugging?

C/c4PredictiveQuery.cc
```diff
@@ -59,15 +60,23 @@ class C4PredictiveModelInternal : public PredictiveModel {
...
+#endif
 }
 
 
 bool c4pred_unregisterModel(const char *name) C4API {
+#ifdef COUCHBASE_ENTERPRISE
     return PredictiveModel::unregister(name);
+#else
+    abort();
```
jayahariv: same, error log?

C/tests/c4Test.cc
```diff
@@ -159,6 +159,18 @@ void CheckError(C4Error error,
...
+        if (predicate())
+            return;
+        this_thread::sleep_for(chrono::milliseconds(100));
+    }
+    FAIL("Wait timed out after " << timeoutMillis << "ms");
+}
+
+
+
+
```
jayahariv: extra lines?

LiteCore/Database/LiveQuerier.cc
```diff
@@ -41,24 +51,83 @@ namespace litecore {
     ,_continuous(continuous)
     ,_delegate(delegate)
     {
-        logInfo("Created on Query %p", query);
+        logInfo("Created on Query %s", query->loggingName().c_str());
```
jayahariv: is this `loggingName` or `loggingIdentifier`?

C/c4PredictiveQuery.cc
snej: ok.

LiteCore/Database/LiveQuerier.cc
snej: `loggingName` is correct. It returns the `{Query#123}` string that shows up in logs.

borrrden:
There is something wrong with this.  I think the CE version is missing an include (maybe it is guarded by `COUCHBASE_ENTERPRISE` or something)

PR #922, opened 2020-02-22
Fixed db compaction (auto_vacuum)

snej:
Databases were not being created with incremental vacuum enabled,
because of an undocumented ordering dependency of pragmas when setting
up a new db. Fixed this by making "PRAGMA auto_vacuum" come first.

Unfortunately there is no way to correct this for existing databases,
because auto_vacuum can only be enabled on an empty database :(

This means that `SQLiteDataFile::compact()` has been doing nothing. I added a
workaround: if auto_vacuum isn't enabled, run a full VACUUM command
instead.

Finally, I extended the unit test of compaction to verify that the
database file actually gets smaller. Both the auto compaction on
close, and the explicit compaction.

Fixes CBL-707

### DIFF
--- a/LiteCore/Storage/SQLiteDataFile.cc
+++ b/LiteCore/Storage/SQLiteDataFile.cc
@@ -77,10 +77,10 @@ namespace litecore {
     static const int kMMapSize = 50 * MB;
 #endif
 
-    // If this fraction of the database is composed of free pages, vacuum it
+    // If this fraction of the database is composed of free pages, vacuum it on close
     static const float kVacuumFractionThreshold = 0.25;
-    // If the database has many bytes of free space, vacuum it
-    static const int64_t kVacuumSizeThreshold = 50 * MB;
+    // If the database has many bytes of free space, vacuum it on close
+    static const int64_t kVacuumSizeThreshold = 10 * MB;
 
     // Database busy timeout; generally not needed since we have other arbitration that keeps
     // multiple threads from trying to start transactions at once, but another process might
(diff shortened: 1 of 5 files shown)


LiteCore/Storage/SQLiteDataFile.cc
```diff
@@ -600,37 +606,62 @@ namespace litecore {
...
+            if (SQL.willLog(LogLevel::Verbose)) {
+                // Log the details of what the optimize will do, before actually doing it:
+                SQLite::Statement stmt(*_sqlDb, "PRAGMA optimize(3)");
+                while (stmt.executeStep()) {
+                    LogVerbose(SQL, "PRAGMA optimize ... %s", stmt.getColumn(0).getString().c_str());
+                    logged = true;
+                }
             }
+            if (!logged)
+                LogVerbose(SQL, "PRAGMA optimize");
```
borrrden: This is never going to be hit I think.  If the level is verbose, the previous verbose line will be logged.  If the level is not verbose, this is a no-op.  Should this be info level?

borrrden:
There actually is a way, according to the SQLite docs:

> However, changing from "none" to "full" or "incremental" can only occur when the database is new (no tables have yet been created) **or by running the VACUUM command.**  To change auto-vacuum modes, first use the auto_vacuum pragma to set the new desired mode, then invoke the VACUUM command to reorganize the entire database file. (emphasis mine)

LiteCore/Storage/SQLiteDataFile.cc
borrrden: @snej Is there a path I am missing here?

snej: > the previous verbose line will be logged

Not if `executeStep` never returns true, which in my experience happens most of the time.

snej: (anyway, that logic isn't new; it's been around a long time. It just looks new in the diff.)

PR #923, opened 2020-02-22
CBL-704: handle the websocket + 4001 = transient error

jayahariv:
* websocketdomain + 4001 = transient error with websocket disconnect
* this logic was handled in platform before, but missed when added to lite core

details > https://issues.couchbase.com/browse/CBL-704

### DIFF
--- a/Replicator/c4RemoteReplicator.hh
+++ b/Replicator/c4RemoteReplicator.hh
@@ -170,7 +170,7 @@ namespace c4Internal {
 
             // If this is a transient error, or if I'm continuous and the error might go away with
             // a change in network (i.e. network down, hostname unknown), then go offline.
-            bool transient = c4error_mayBeTransient(c4err);
+            bool transient = c4error_mayBeTransient(c4err) || (c4err.domain == WebSocketDomain && c4err.code == 4001);
             if (transient || (continuous() && c4error_mayBeNetworkDependent(c4err))) {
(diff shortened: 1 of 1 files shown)


Replicator/c4RemoteReplicator.hh
```diff
@@ -170,7 +170,7 @@ namespace c4Internal {
 
             // If this is a transient error, or if I'm continuous and the error might go away with
             // a change in network (i.e. network down, hostname unknown), then go offline.
-            bool transient = c4error_mayBeTransient(c4err);
+            bool transient = c4error_mayBeTransient(c4err) || (c4err.domain == WebSocketDomain && c4err.code == 4001);
```
borrrden: I have trouble with this because now we are basically forcing the consumer to rely on an implementation detail.  I think it would be better to document that 4001 is a special code and then handle it inside of `c4error_mayBeTransient`

snej:
What's a 4001 status? I've never seen that before. It's not a standard HTTP or WebSocket status code.

snej:
Code 4001 is something defined in the iOS codebase. It's defined in CBLReplicator+Internal.h, and set in CBLMessageSocket.mm.

LiteCore shouldn't be using nonstandard codes defined at a higher layer. Seems like this should be fixed in the iOS code. Or if we need a specific code for this, we should define it in LiteCore.

PR #924, opened 2020-02-25
CBL-704: Fix: Include user generated websocket closure codes

jayahariv:
* transient and permanent status codes by user was previously handled in platform side, now moving it to lite core. 
* include the transient code maybe_transient method. 

linking closed PR: https://github.com/couchbase/couchbase-lite-core/pull/923

### DIFF
--- a/C/c4Base.cc
+++ b/C/c4Base.cc
@@ -246,7 +246,7 @@ bool c4error_mayBeTransient(C4Error err) C4API {
         websocket::kCodeGoingAway,
         websocket::kCodeAbnormal,
         websocket::kCodeUnexpectedCondition,
-        kWebSocketCloseUserTransient,
+        websocket::kCloseAppTransient,
         0};
     static ErrorSet kTransient = { // indexed by C4ErrorDomain
         nullptr,
--- a/C/c4Base.cc
+++ b/C/c4Base.cc
@@ -18,6 +18,7 @@
 
 #include "c4Internal.hh"
 #include "c4Private.h"
+#include "c4Socket.h"
 
 #include "Actor.hh"
 #include "Backtrace.hh"
@@ -245,6 +246,7 @@ bool c4error_mayBeTransient(C4Error err) C4API {
         websocket::kCodeGoingAway,
         websocket::kCodeAbnormal,
         websocket::kCodeUnexpectedCondition,
+        kWebSocketCloseUserTransient,
         0};
     static ErrorSet kTransient = { // indexed by C4ErrorDomain
         nullptr,
--- a/C/include/c4Socket.h
+++ b/C/include/c4Socket.h
@@ -42,7 +42,9 @@ extern "C" {
         kWebSocketCloseCantFulfill      = 1011, // Can't fulfill request due to "unexpected condition"
         kWebSocketCloseTLSFailure       = 1015, // Never sent, only received
 
-        kWebSocketCloseFirstAvailable   = 4000, // First unregistered code for freeform use
+        kWebSocketCloseUserTransient    = 4001, // transient websocket closure error initiated by user
+        kWebSocketCloseUserPermanent    = 4002, // permanent websocket closure error initiated by user
+        kWebSocketCloseFirstAvailable   = 4003, // First unregistered code for freeform use
     };
 
 
--- a/C/include/c4Socket.h
+++ b/C/include/c4Socket.h
@@ -42,8 +42,8 @@ extern "C" {
         kWebSocketCloseCantFulfill      = 1011, // Can't fulfill request due to "unexpected condition"
         kWebSocketCloseTLSFailure       = 1015, // Never sent, only received
 
-        kWebSocketCloseUserTransient    = 4001, // transient websocket closure error initiated by user
-        kWebSocketCloseUserPermanent    = 4002, // permanent websocket closure error initiated by user
+        kWebSocketCloseUserTransient    = 4001, // transient websocket closure error by user
+        kWebSocketCloseUserPermanent    = 4002, // permanent websocket closure error by user
         kWebSocketCloseFirstAvailable   = 4003, // First unregistered code for freeform use
     };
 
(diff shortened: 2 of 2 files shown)


C/include/c4Socket.h
```diff
@@ -42,7 +42,9 @@ extern "C" {
         kWebSocketCloseCantFulfill      = 1011, // Can't fulfill request due to "unexpected condition"
         kWebSocketCloseTLSFailure       = 1015, // Never sent, only received
 
-        kWebSocketCloseFirstAvailable   = 4000, // First unregistered code for freeform use
+        kWebSocketCloseUserTransient    = 4001, // transient websocket closure error initiated by user
```
jayahariv:
```suggestion
        kWebSocketCloseUserTransient    = 4001, // transient websocket closure error by user
```

C/include/c4Socket.h
```diff
@@ -42,7 +42,9 @@ extern "C" {
...
+        kWebSocketCloseUserPermanent    = 4002, // permanent websocket closure error initiated by user
```
jayahariv:
```suggestion
        kWebSocketCloseUserPermanent    = 4002, // permanent websocket closure error by user
```

C/include/c4Socket.h
```diff
@@ -42,7 +42,9 @@ extern "C" {
         kWebSocketCloseCantFulfill      = 1011, // Can't fulfill request due to "unexpected condition"
         kWebSocketCloseTLSFailure       = 1015, // Never sent, only received
 
-        kWebSocketCloseFirstAvailable   = 4000, // First unregistered code for freeform use
+        kWebSocketCloseUserTransient    = 4001, // transient websocket closure error by user
```
jayahariv:
```suggestion
        kWebSocketCloseUserTransient    = 4001, // transient websocket closure by user
```

C/include/c4Socket.h
```diff
@@ -42,7 +42,9 @@ extern "C" {
...
+        kWebSocketCloseUserPermanent    = 4002, // permanent websocket closure error by user
```
jayahariv:
```suggestion
        kWebSocketCloseUserPermanent    = 4002, // permanent websocket closure by user
```

snej:
I'm not clear on what these codes are for. "closure by user" is kind of vague. @pasin, could you explain?

borrrden:
These are used in the enterprise `MessageEndpoint` implementation to indicate that an exception occurred during the user implementation of their transport.  We don't know what kinds of exceptions they will face and the ones we define may not make sense, so we created two custom ones to use.  One (4002) is for an exception that should not be retried, and one (4001) is for an exception that should trigger a retry according to the retry logic (similar to a connection time out, etc).

pasin:
Yes, it is for what @borrrden has explained.

C/include/c4Socket.h
```diff
@@ -42,7 +42,9 @@ extern "C" {
         kWebSocketCloseCantFulfill      = 1011, // Can't fulfill request due to "unexpected condition"
         kWebSocketCloseTLSFailure       = 1015, // Never sent, only received
 
-        kWebSocketCloseFirstAvailable   = 4000, // First unregistered code for freeform use
+        kWebSocketCloseUserTransient    = 4001, // transient websocket closure by user
```
jayahariv:
```suggestion
        kWebSocketCloseUserTransient    = 4001, // any transient message endpoint web socket closure by user
```

jayahariv:
```suggestion
        kWebSocketCloseUserTransient    = 4001, // any transient message endpoint websocket closure by user
```

C/include/c4Socket.h
```diff
@@ -42,7 +42,9 @@ extern "C" {
...
+        kWebSocketCloseUserPermanent    = 4002, // permanent websocket closure by user
```
jayahariv:
```suggestion
        kWebSocketCloseUserPermanent    = 4002, // any permanent message endpoint websocket closure by user
```

snej:
Do you mean an error, like some problem with their transport, or an unexpected problem? Is there a way the app code could provide us an error code to use? Otherwise the error comes out the other end (back from the Replicator) as 4001/4002 code that may not mean anything to them.

borrrden:
That's the way things have been now, but we've been saving the actual message from the error to go along with it so that the user has an idea of what is going on.  We could take an error code from them, but what if it overlaps with one of our numerous other error codes?

C/c4Base.cc
```diff
@@ -245,6 +246,7 @@ bool c4error_mayBeTransient(C4Error err) C4API {
         websocket::kCodeGoingAway,
         websocket::kCodeAbnormal,
         websocket::kCodeUnexpectedCondition,
+        kWebSocketCloseUserTransient,
```
snej: The error codes in WebSocketInterface.hh should be kept in sync with the ones in c4Socket.h.

C/include/c4Socket.h
```diff
@@ -42,7 +42,9 @@ extern "C" {
         kWebSocketCloseCantFulfill      = 1011, // Can't fulfill request due to "unexpected condition"
         kWebSocketCloseTLSFailure       = 1015, // Never sent, only received
 
-        kWebSocketCloseFirstAvailable   = 4000, // First unregistered code for freeform use
+        kWebSocketCloseUserTransient    = 4001, // any transient message endpoint websocket closure by user
```
snej: How about `// App-defined transient error`, and `// App-defined permanent error`

C/include/c4Socket.h
```diff
@@ -42,7 +42,9 @@ extern "C" {
...
+        kWebSocketCloseUserPermanent    = 4002, // any permanent message endpoint websocket closure by user
+        kWebSocketCloseFirstAvailable   = 4003, // First unregistered code for freeform use
```
snej: Bump this up to 5000 so it won't be affected if we have to add more errors above it. (Also check that none of the platforms are using codes in the 4000 range.)

C/c4Base.cc
jayahariv: https://github.com/couchbaselabs/BLIP-Cpp/pull/20

PR #925, opened 2020-02-25
CBL-708: Don't leave obsolete revs behind after a merge

snej:
Unnecessary stuff gets left behind after a conflict is resolved: the revision metadata of the conflict branch (up to about 1KB), and worse, the body of the losing revision (could be big!) I have a copy of a customer database that has about 4MB of unnecessary overhead due to this.

This PR has two commits:
1. Prunes the entire losing branch during a merge.
2. Makes sure older revisions on the surviving branch don't have the kKeepBody flag set, which keeps their bodies around.

I updated the "Document Conflict" unit test to check this. (Most of the source changes there are just to use constants for the revIDs instead of repeating the strings over and over.)

Fixes CBL-708

### DIFF
--- a/C/tests/c4DocumentTest.cc
+++ b/C/tests/c4DocumentTest.cc
@@ -772,16 +772,18 @@ N_WAY_TEST_CASE_METHOD(C4Test, "Document Conflict", "[Database][C]") {
     if (!isRevTrees())
         return;
 
+    static constexpr slice kRev3ID("3-aaaaaa"), kRev3ConflictID("3-ababab"), kRev4ConflictID("4-dddd");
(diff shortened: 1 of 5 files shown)


borrrden:
PR validation results look very suspicious

snej:
Unix failed "TLS REST pinned cert" with an unexpected socket disconnect. Unrelated to this PR.
Mac failed "Continuous Push From Both Sides" with an unexpected conflict. Possibly related...

borrrden:
I ran the validation again, and Windows and Linux both passed, but macOS got the same error again.

snej:
Will rebase & retry this after the crash-fix PR #975 is merged. 🤞

PR #930, opened 2020-03-12
CBL-722 / CBL-729: Fix corner cases with suspending and unsuspending

borrrden:
Introduce a new replicator status "stopping" to indicate that we are in transit so that we can ignore or properly handle subsequent calls happening between the impetus for the stop or offline, and the actual stop / offline status change.  In particular two cases are handled better now:

1. Suspending while a stop is in progress
2. Unsuspending while suspending is in progress

### DIFF
--- a/C/include/c4Replicator.h
+++ b/C/include/c4Replicator.h
@@ -41,6 +41,7 @@ extern "C" {
     /** The possible states of a replicator. */
     typedef C4_ENUM(int32_t, C4ReplicatorActivityLevel) {
         kC4Stopped,     ///< Finished, or got a fatal error.
+        kC4Stopping,    ///< Stopping or going offline
         kC4Offline,     ///< Connection failed, but waiting to retry. */
         kC4Connecting,  ///< Connection is in progress.
         kC4Idle,        ///< Continuous replicator has caught up and is waiting for changes.
--- a/C/include/c4Replicator.h
+++ b/C/include/c4Replicator.h
@@ -40,12 +40,15 @@ extern "C" {
 
     /** The possible states of a replicator. */
     typedef C4_ENUM(int32_t, C4ReplicatorActivityLevel) {
+        /* EXTERNAL STATES */
         kC4Stopped,     ///< Finished, or got a fatal error.
-        kC4Stopping,    ///< Stopping or going offline (INTERNAL STATE)
         kC4Offline,     ///< Connection failed, but waiting to retry. */
         kC4Connecting,  ///< Connection is in progress.
         kC4Idle,        ///< Continuous replicator has caught up and is waiting for changes.
-        kC4Busy         ///< Connected and actively working.
+        kC4Busy,         ///< Connected and actively working.
+        
+        /* INTERNAL STATES */
+        kC4Stopping,    ///< Stopping or going offline
     };
 
     /** For convenience, an array of C strings naming the C4ReplicatorActivityLevel values. */
--- a/C/include/c4Replicator.h
+++ b/C/include/c4Replicator.h
@@ -41,7 +41,7 @@ extern "C" {
     /** The possible states of a replicator. */
     typedef C4_ENUM(int32_t, C4ReplicatorActivityLevel) {
         kC4Stopped,     ///< Finished, or got a fatal error.
-        kC4Stopping,    ///< Stopping or going offline
+        kC4Stopping,    ///< Stopping or going offline (INTERNAL STATE)
         kC4Offline,     ///< Connection failed, but waiting to retry. */
         kC4Connecting,  ///< Connection is in progress.
         kC4Idle,        ///< Continuous replicator has caught up and is waiting for changes.
--- a/C/include/c4Replicator.h
+++ b/C/include/c4Replicator.h
@@ -49,7 +49,7 @@ extern "C" {
     };
 
     /** For convenience, an array of C strings naming the C4ReplicatorActivityLevel values. */
-    CBL_CORE_API extern const char* const kC4ReplicatorActivityLevelNames[5];
+    CBL_CORE_API extern const char* const kC4ReplicatorActivityLevelNames[6];
 
 
     /** A simple parsed-URL type. */
--- a/Replicator/c4Replicator.cc
+++ b/Replicator/c4Replicator.cc
@@ -34,8 +34,8 @@ using namespace c4Internal;
 constexpr unsigned C4RemoteReplicator::kMaxRetryDelay;
 
 
-CBL_CORE_API const char* const kC4ReplicatorActivityLevelNames[5] = {
-    "stopped", "offline", "connecting", "idle", "busy"
+CBL_CORE_API const char* const kC4ReplicatorActivityLevelNames[6] = {
+    "stopped", "stopping", "offline", "connecting", "idle", "busy"
 };
 
 
--- a/Replicator/c4Replicator.cc
+++ b/Replicator/c4Replicator.cc
@@ -35,7 +35,7 @@ constexpr unsigned C4RemoteReplicator::kMaxRetryDelay;
 
 
 CBL_CORE_API const char* const kC4ReplicatorActivityLevelNames[6] = {
-    "stopped", "stopping", "offline", "connecting", "idle", "busy"
+    "stopped", "offline", "connecting", "idle", "busy", "stopping"
 };
 
 
--- a/Replicator/c4Replicator.hh
+++ b/Replicator/c4Replicator.hh
@@ -55,6 +55,12 @@ struct C4Replicator : public RefCounted,
 
     virtual void start() {
         LOCK(_mutex);
+		if(_status.level == kC4Stopping) {
+            logWarn("Rapid call to start() (stop() is not finished yet), scheduling a restart after stop() is done...");
+            _cancelStop = true;
+            return;
+        }
+
         if (!_replicator) {
             if(!_start()) {
                 UNLOCK();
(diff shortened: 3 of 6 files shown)


borrrden:
There is going to be a sort of "indirect API change" here in that an enum is being modified.  I notice some code relying on there being 5 enum states, which seems like a smell to begin with but let me know if adding one is too disruptive.  LiteCore, at least, needs to know about it so it can know how to react to calls in that situation.

Replicator/c4Replicator.hh
```diff
@@ -72,8 +78,39 @@ struct C4Replicator : public RefCounted,
...
+            // Suspending a stopped replicator?  Get outta here...
+            Warn("Ignoring a suspend call on a stopped replicator...");
+            return;
+        }
+        
+        if(_status.level == kC4Stopping && !statusFlag(kC4Suspended)) {
+            // CBL-722: Stop was already called or Replicator is stopped,
+            // making suspending meaningless (stop() should override any
+            // suspending or unsuspending)
+            Warn("Ignoring a suspend call on a stopping replicator...");
```
bmeike: Do these really want to be Warnings?  There is no way to know the exact state of the replicator.  It seems like this will happen all the time...

Replicator/c4Replicator.hh
```diff
@@ -72,8 +78,39 @@ struct C4Replicator : public RefCounted,
...
+            _cancelStop = !suspended;
+            if(_cancelStop) {
+                Warn("Request to unsuspend, but Replicator is already suspending.  Will restart after suspending process is completed.");
+            } else {
+                Warn("Replicator suspension process being spammed (request to suspend followed by at least one request to unsuspend and then suspend again), attempting to cancel restart.");
+            }
             return;
+        }
+        
         logInfo("%s", (suspended ? "Suspended" : "Un-suspended"));
```
bmeike: "resumed"?

Replicator/c4Replicator.hh
```diff
@@ -97,7 +134,15 @@ struct C4Replicator : public RefCounted,
...
+        if(_status.level == kC4Stopping) {
+            // Already stopping, this call is spammy so ignore it
+            logVerbose("Duplicate call to stop() ignored...");
+            return;
+        }
+        
         if (_replicator) {
+            _status.level = kC4Stopping;
             _replicator->stop();
         } else if (_status.level != kC4Stopped) {
```
bmeike: That seems unnecessarily complex?  Just set it to kC4Stopped.  Nobody cares what it was, before.

C/include/c4Replicator.h
```diff
@@ -41,14 +41,15 @@ extern "C" {
     /** The possible states of a replicator. */
     typedef C4_ENUM(int32_t, C4ReplicatorActivityLevel) {
         kC4Stopped,     ///< Finished, or got a fatal error.
+        kC4Stopping,    ///< Stopping or going offline
```
bmeike: Whoah!!!   New state???   Reported to the platforms?  This needs discussion.

bmeike:
Oddly enough, that enum thing seemed like a smell to me too.  I just happened to fix it, just now.

On the other hand, the values of those enums are just Java constants, so I *will* have to update those...

Replicator/c4Replicator.hh
borrrden: Can't, it's not all the way stopped yet.  This next call to stop is async.

Replicator/c4Replicator.hh
borrrden: It won't.  It only happens if there are rapid calls to stop / start / suspend

bmeike: I see.  Still... is spamming the restart button a warning?

Replicator/c4Replicator.hh
bmeike: No, not what I meant. `else if (_status.level != kC4Stopped) {`
sets the status to kC4Stopped unless the status is kC4Stopped.  Just set it to kC4Stopped.  I.e:

else `{ _status.level = kC4Stopped }`

Replicator/c4Replicator.hh
```diff
@@ -55,6 +55,12 @@ struct C4Replicator : public RefCounted,
 
     virtual void start() {
         LOCK(_mutex);
+        if(_status.level == kC4Stopping) {
+            Warn("Rapid call to start() (stop() is not finished yet), scheduling a restart after stop() is done...");
+            _cancelStop = true;
```
pasin: In iOS's replicator, I ignored the start as the replicator has not been stopped yet. On the public facing side, users don't get STOPPED notification yet so I think it sort of expects that the start will be ignored.

Replicator/c4Replicator.hh
borrrden: I'd like to keep the review within the scope of what I changed, but that would result in spurious notifications back to the delegate (inside that block calls `notifyStateChanged`)

Replicator/c4Replicator.hh
borrrden: That is in no way clear from our contract, but are you asserting that the correct behavior is to ignore start unless the replicator is stopped?

Replicator/c4Replicator.hh
borrrden: Maybe not, it just bothers me....and this is actually not the restart button at this location but rapid backgrounding and foregrounding and backgrounding again.  I doubt the user will come out and say "Oh yeah I was licking the home button to see what happens" so if possible I'd like to log when we are in this unstable situation.

Replicator/c4Replicator.hh
bmeike: Gotcha on both accounts.  This UI cut of the rest of the block.

Replicator/c4Replicator.hh
bmeike: ok.  sounds good to me.

Replicator/c4Replicator.hh
bmeike: I do not depend on Core to do that for me.  On the other hand, that is exactly how the Java bindings behave.  You will never see two calls to stop, without a call to start in between, and vice-versa.

pasin: I think it is worth to define the behavior. In other condition beside stopping, if the replicator is still running, calling start again will be no-ops. IMO, it should behave the same for stopping state. 

How will this affect the suspend and resume? In iOS, when the app goes to background, the replicator will be suspended. When the apps goes foreground, the replicator will be resume. There is also a case of switching b/w background and foreground is fast, the expectation is that the replicator should be resume when the app is foregrounded. When I implemented this in iOS, I did add an ability to cancel the suspend (auto resume).

borrrden: I took this implementation partially from the iOS implementation and spent quite a long time thinking about these situations.  If a rapid foreground is expected to resume the replicator from suspending then why is a rapid start not expected to behave similarly?  Here are the assumptions I am working on:

- Calling start when:
    - stopped -> start the replicator immediately
    - stopping -> start the replicator after it is stopped (should be ignore?)
    - all others -> ignore
- Calling stop when:
    - stopped -> ignore
    - stopping -> cancel any resume efforts, if applicable
    - offline -> switch to stopped immediately
    - all others -> transition to stopping and start the stop process
- Backgrounding when:
    - stopped or stopping (or connecting ?) -> ignore
    - All others -> suspend
 - Foregrounding when:
    - stopping -> if suspended flag is set, set resume flag otherwise ignore
    - offline -> resume
    - all others -> ignore

It's pretty frustrating that this has only been discussed now after a good deal of implementation has already been done (previous to even this PR).  Having the start be a no op for the stopping state would make me happy because it is easier to implement, but it also presents an inconsistency with the way rapid backgrounding and foregrounding is handled (I guess this could be explained away since those calls are automatic versus the other calls which are explicit)

Replicator/c4Replicator.hh
```diff
@@ -72,8 +78,37 @@ struct C4Replicator : public RefCounted,
...
+        
+        if(_status.level == kC4Stopping) {
+            // CBL-729: At this point, the suspended state has changed from a previous
+            // call that caused a suspension to start.  Register to restart later
+            // (or cancel the later restart) and move on
+            _cancelStop = !suspended;
+            if(_cancelStop) {
+                logInfo("Request to unsuspend, but Replicator is already suspending.  Will restart after suspending process is completed.");
+            } else {
+                logInfo("Replicator suspension process being spammed (request to suspend followed by at least one request to unsuspend and then suspend again), attempting to cancel restart.");
```
bmeike: Is that true???  Won't a stop followed quickly by a suspend, get you here?

Replicator/c4Replicator.hh
```diff
@@ -72,8 +78,37 @@ struct C4Replicator : public RefCounted,
...
+            }
+            return;
+        } else if (!setStatusFlag(kC4Suspended, suspended)) {
```
bmeike: why the 'else'?

Replicator/c4Replicator.hh
borrrden: Stop preempts suspend.  A few lines up you will see where that case is handled:

```
if(_status.level == kC4Stopping && !statusFlag(kC4Suspended)) {
    logInfo("Ignoring a suspend call on a stopping replicator...");
    return;
}
```

Replicator/c4Replicator.hh
borrrden: Uncovered during the suspend, unsuspend, suspend exercise.  Suspend will set the suspended flag, but unless this is an `else` then unsuspend will clear it, resulting in the next call to suspend returning early because it thinks `stop` was called because of the lack of suspended flag.

borrrden: Oh, you simply meant that I put in an `else` after a `return`?  Good catch, I don't like doing that either!

Replicator/c4Replicator.hh
bmeike: YEah... ok.   I see it now.

Replicator/tests/ReplicatorAPITest.cc
```diff
@@ -419,3 +427,79 @@ TEST_CASE_METHOD(ReplicatorAPITest, "Is Document Pending", "[Push]") {
...
+    }
+    
+    SECTION("Stop / Suspend") {
+        c4repl_stop(_repl);
+        c4repl_setSuspended(_repl, true);
+    }
+    
+    SECTION("Suspend / Stop") {
+        c4repl_setSuspended(_repl, true);
+        c4repl_stop(_repl);
```
pasin: Just curious if it needs to set expected to kC4Stopped?

borrrden: That's the default on line 441 `C4ReplicatorActivityLevel expected = kC4Stopped;`

Replicator/c4Replicator.cc
```diff
@@ -34,8 +34,8 @@ using namespace c4Internal;
 constexpr unsigned C4RemoteReplicator::kMaxRetryDelay;
 
 
-CBL_CORE_API const char* const kC4ReplicatorActivityLevelNames[5] = {
-    "stopped", "offline", "connecting", "idle", "busy"
+CBL_CORE_API const char* const kC4ReplicatorActivityLevelNames[6] = {
+    "stopped", "stopping", "offline", "connecting", "idle", "busy"
```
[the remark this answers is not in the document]
borrrden: I made a half-conscious choice to not do this to force us to get rid of the smell of relying on the values of the enums.  But it's a decent point.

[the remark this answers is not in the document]
borrrden: But I changed my mind because there are some other annoyances about ignoring an enum in the middle of as set

C/include/c4Replicator.h
```diff
@@ -40,12 +40,15 @@ extern "C" {
...
         kC4Stopped,     ///< Finished, or got a fatal error.
-        kC4Stopping,    ///< Stopping or going offline (INTERNAL STATE)
         kC4Offline,     ///< Connection failed, but waiting to retry. */
         kC4Connecting,  ///< Connection is in progress.
         kC4Idle,        ///< Continuous replicator has caught up and is waiting for changes.
-        kC4Busy         ///< Connected and actively working.
+        kC4Busy,         ///< Connected and actively working.
+        
+        /* INTERNAL STATES */
+        kC4Stopping,    ///< Stopping or going offline
```
pasin: In term of the Lite-Core API, I think putting kC4Stopping is better. Is the provide a guideline to CBL to change instead of moving kC4Stopping to the end of the enum? 

When moving the kC4Stopping like that, it doesn't allow to do something like `status > kC4Stopping` to check whether the replicator is in stopping or stopped state.

bmeike: I think that, perhaps, this is what Jim means by a "code smell".  Assuming that we know the order and exact number of states seems fragile and easy to break.  I think that a good way forward is to assure that, once a state is assigned a number, that number never changes.  Changing the number is changing the API and it is my opinion that adding or removing internal states is not a good reason for doing that.

borrrden: I looked over and didn't find any situations where we need to do that.  Since we aren't mapping C4 to CBL for this state I thought it would be better in the long run to split into external and internal sections.

bmeike: 👍🏻

PR #931, opened 2020-03-19
CBL-683: Don't reopen the DB so early in the replication process (CE part)

borrrden:
If the DB is reopened immediately in the call to the various c4repl_new functions, then it remains open for the life of the C4Replicator.  It only needs to be reopened for the actively replicating part of the replication, so reopen it in the C++ replicator class instead so that it goes away after stop.


borrrden:
Note: Depends on https://github.com/couchbase/couchbase-lite-core-EE/pull/11

PR #935, opened 2020-04-01
C4Listener: added auth callbacks & connection-count

snej:
* Optional cert-validation callback
* Optional HTTP auth callback
* c4listener_connectionCount() returns number of connected clients

Improved error handling for rejected TLS certs. When the peer rejects the TLS handshake, what comes back is a "fatal alert" code, which for some reason mbedTLS reports as a single error code even though there are many possible causes. I had to detect this error in sockpp and return a range of fake error codes to pass the actual alert codes up to TCPSocket.

Fixes CBL-679

### DIFF
--- a/C/include/c4Listener.h
+++ b/C/include/c4Listener.h
@@ -42,6 +42,25 @@ extern "C" {
     };
 
 
+    /** Called when a client connects, during the TLS handshake, if a client certificate is received.
+        @param listener  The C4Listener handling the connection.
+        @param clientCertData  The client's X.509 certificate in DER encoding.
+        @param context  The `tlsCallbackContext` from the `C4TLSConfig`.
+        @return  True to allow the connection, false to refuse it. */
+    typedef bool (*C4ListenerCertAuthCallback)(C4Listener *listener,
+                                               C4Slice clientCertData,
+                                               void *context);
+
+    /** Called when a client connects, after the TLS handshake (if any), when the initial HTTP request is
+        received.
+        @param listener  The C4Listener handling the connection.
+        @param authHeader  The "Authorization" header value from the client's HTTP request, or null.
+        @param context  The `callbackContext` from the listener config.
+        @return  True to allow the connection, false to refuse it. */
+    typedef bool (*C4ListenerHTTPAuthCallback)(C4Listener *listener,
+                                               C4Slice authHeader,
+                                               void *context);
+
     /** TLS configuration for C4Listener. */
     typedef struct C4TLSConfig {
         C4PrivateKeyRepresentation privateKeyRepresentation; ///< Interpretation of `privateKey`
@@ -49,6 +68,8 @@ extern "C" {
         C4Cert* certificate;            ///< X.509 certificate data
         bool requireClientCerts;        ///< True to require clients to authenticate with a cert
         C4Cert* rootClientCerts;        ///< Root CA certs to trust when verifying client cert
+        C4ListenerCertAuthCallback certAuthCallback; ///< Callback cor X.509 cert auth
+        void* tlsCallbackContext;
     } C4TLSConfig;
 
 
@@ -58,6 +79,9 @@ extern "C" {
         C4String networkInterface;      ///< name or address of interface to listen on; else all
         C4ListenerAPIs apis;            ///< Which API(s) to enable
         C4TLSConfig* tlsConfig;         ///< TLS configuration, or NULL for no TLS
+
+        C4ListenerHTTPAuthCallback httpAuthCallback; ///< Callback for HTTP auth
+        void* callbackContext;
 
         // For REST listeners only:
         C4String directory;             ///< Directory where newly-PUT databases will be created
@@ -110,6 +134,9 @@ extern "C" {
         This is useful if you didn't specify a port in the config (`port`=0), so you can find out which
         port the kernel picked. */
     uint16_t c4listener_getPort(C4Listener *listener C4NONNULL) C4API;
+
+    /** Returns the number of client connections. */
+    int c4listener_connectionCount(C4Listener *listener C4NONNULL) C4API;
 
     /** A convenience that, given a filesystem path to a database, returns the database name
         for use in an HTTP URI path.
--- a/C/include/c4Listener.h
+++ b/C/include/c4Listener.h
@@ -135,8 +135,11 @@ extern "C" {
         port the kernel picked. */
     uint16_t c4listener_getPort(C4Listener *listener C4NONNULL) C4API;
 
-    /** Returns the number of client connections. */
-    int c4listener_connectionCount(C4Listener *listener C4NONNULL) C4API;
+    /** Returns the number of client connections, and how many of those are currently active.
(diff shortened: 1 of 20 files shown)


borrrden:
Linux failed with an unexpected TLS error code
Windows failed because of a commit I made directly to master added in the SG replicator tests to CppTests (apparently they don't compile on Windows, but that's unrelated to this PR)
macOS failed because it got a mysterious SIGILL again :-S

C/include/c4Listener.h
```diff
@@ -58,6 +79,9 @@ extern "C" {
         C4ListenerAPIs apis;            ///< Which API(s) to enable
         C4TLSConfig* tlsConfig;         ///< TLS configuration, or NULL for no TLS
 
+        C4ListenerHTTPAuthCallback httpAuthCallback; ///< Callback for HTTP auth
+        void* callbackContext;
```
borrrden, line 83: Ambiguous name here, could it be renamed?

REST/RESTListener.cc
```diff
@@ -145,10 +153,21 @@ namespace litecore { namespace REST {
...
+            tlsContext->setCertAuthCallback([=](slice certData) {
+                return callback((C4Listener*)this, certData, context);
+            });
+        }
         return tlsContext;
     }
 
 
+    unsigned RESTListener::connectionCount() {
+        return _server->connectionCount();
```
borrrden, line 167: Why is one of these a signed int and the other an unsigned?  I think we should settle on one or the other.

REST/Request.cc
```diff
@@ -387,6 +387,11 @@ namespace litecore { namespace REST {
     }
 
 
+    void RequestResponse::onClose(std::function<void()> callback) {
+        _socket->onClose(move(callback));
```
borrrden, line 391: Does this move have any effect if the function `onClose` doesn't use the `&&` syntax?  I think if it did a copy would be saved.

REST/Server.cc
```diff
@@ -138,9 +138,9 @@ namespace litecore { namespace REST {
         }
         RequestResponse rq(this, move(responder));
         if (rq.isValid()) {
-            dispatchRequest(&rq);
-            rq.finish();
-        }
+        dispatchRequest(&rq);
+        rq.finish();
+    }
```
borrrden, line 143: Awkward spacing

C/include/c4Listener.h
```diff
@@ -41,6 +41,25 @@ extern "C" {
...
+                                               void *context);
+
+    /** Called when a client connects, after the TLS handshake (if any), when the initial HTTP request is
+        received.
+        @param listener  The C4Listener handling the connection.
+        @param authHeader  The "Authorization" header value from the client's HTTP request, or null.
+        @param context  The `callbackContext` from the listener config.
+        @return  True to allow the connection, false to refuse it. */
+    typedef bool (*C4ListenerHTTPAuthCallback)(C4Listener *listener,
+                                               C4Slice authHeader,
```
jayahariv, line 60: Instead of passing the `authHeader` itself, can we decode and pass back the username and password? Since the fleece has the decoder and all platform has to do it.

borrrden: The advantage of this way is that later we can extend it later if something besides username and password is used in the auth header.  The auth header is not fleece, but base64 anyway.

snej:
Linux test failure:
```
/var/lib/jenkins/mobile/workspace/eline_couchbase-lite-core_PR-935/couchbase-lite-core/REST/tests/RESTListenerTest.cc:449: FAILED:
  CHECK( r->error() == (C4Error{NetworkDomain, kC4NetErrTLSCertRejectedByPeer}) )
with expansion:
  C4Error(POSIX error 104 "Connection reset by peer")
  ==
  C4Error(Network error 10 "server rejected the TLS client certificate")
```
Damn, this is exactly an issue I fixed yesterday, where the server's `mbed_tls_socket` closes before it can write the remaining error info, so the client just sees that the socket was closed. No idea why my fix didn't work on Linux.

C/include/c4Listener.h
snej: It's a context for all the callbacks in that struct. There's currently just one, but we could add more. (See C4ReplicatorConfig, which has a field with the same name.)

The one in C4TLSConfig just has a different name to avoid confusion.

borrrden:
Well now that Windows is building correctly it has a similar error to Linux.  Probably some subtle threaded vs gcd mailbox differences....or something more nefarious maybe.  Being on Windows makes it easier to repro though.

C/include/c4Listener.h
snej: FYI, CBLHTTPLogic.m has code that shows how to generate & parse Authorization headers.

snej:
Rebased to current master (no conflicts.)

borrrden:
Was any specific action taken to solve the above connection reset error? If not then I feel a bit worried.

pasin:
@snej per @borrrden's question, has the connection reset error been fixed?

snej:
No; the error only seems to occur on non-Mac platforms, which I'm not set up for (or skilled at) debugging...

borrrden:
In that case let me look at it before it gets merged.

snej:
@borrrden noticed that some changes had gone missing from this PR since it was first created; in fact, when I rebased it onto master on the 16th, I ran into a serious bug in my git client app that lost all the changes in files that were in conflict. 😱

I've gone back to the pre-rebase commit, re-rebased from the CLI, resolved the conflicts, and force-pushed the result.

The "TLS REST client cert w/auth callback" unit test is failing somehow; the call succeeds even though it has the wrong credentials. I have to leave work now so either I can fix this in the morning, or maybe @borrrden wants to work on it tonight...

borrrden:
There seems to be a fundamental flaw in the way this is set up.  The added call to `shutdown` only hid the problem on macOS because of the incorrect way that it was added.  `shutdown` essentially incorrectly replaces `close` to keep the socket open indefinitely.  The correct way to do this is to:

1. Call `shutdown` on the write channel
2. Read until `recv` returns 0 bytes
3. Close the socket

snej:
### Changes:

* Use `SO_REUSEADDR` on listener sockets.
* Fixed a use-after-free bug in TCPSocket.
* Expanded `c4listener_connectionCount` into `c4listener_getConnectionStatus` (CBL-905).

borrrden:
The tests as they are cannot pass on Linux this way...

snej:
Assigning this to you, @borrrden , since it seems this requires Linux-specific changes.

borrrden:
Some tests are still failing on Linux and Windows is hanging....ugh

borrrden:
@snej I can't get the new unit test to work correctly, even on macOS.  Can you verify that it passes?  Even after I fixed the immediate issue the test never registers an active connection now and so fails when it expect max active count to be 1.

snej:
Oops, I'd forgotten to push my LiteCore-EE changes -- [here they are now](https://github.com/couchbase/couchbase-lite-core-EE/tree/feature/listener_auth_CBL-679).

PR #939, opened 2020-04-03
C4CertNameAttributeID is now a string (slice) not an enum [API CHANGE]

snej:
This allows clients to set arbitrary names if they want.

Fixes CBL-814

### DIFF
--- a/C/include/c4Certificate.h
+++ b/C/include/c4Certificate.h
@@ -46,31 +46,27 @@ extern "C" {
     /** Certificate subject name attributes, from RFC 4519 and RFC 5280 (sec. 4.2.1.6)
         Only the CommonName is required; it's used as the visible name of the certificate.
         If the cert is to be used for a TLS server, the CommonName must match its DNS name. */
-    typedef C4_ENUM(uint8_t, C4CertNameAttributeID) {
-        // Some common Distinguished Name attributes:
-        kC4Cert_CommonName,        // e.g. "Jane Doe", (or "jane.example.com")
-        kC4Cert_Pseudonym,         // e.g. "plainjane837"
-        kC4Cert_GivenName,         // e.g. "Jane"
-        kC4Cert_Surname,           // e.g. "Doe"
-        kC4Cert_Organization,      // e.g. "Example Corp."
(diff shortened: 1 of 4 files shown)


borrrden:
What is with this SIGILL on macOS?  It is happening more and more....

C/include/c4Certificate.h
```diff
@@ -46,31 +46,27 @@ extern "C" {
...
+    #define kC4Cert_PostalCode        C4STR("postalCode")   // e.g. "02134"
+    #define kC4Cert_StateOrProvince   C4STR("ST")           // e.g. "Massachusetts" (or "Quebec", ...)
+    #define kC4Cert_Country           C4STR("C")            // e.g. "us" (2-letter ISO country code)
+
+    // These are the Subject Alternative Name attributes:
+    #define kC4Cert_EmailAddress      C4STR("rfc822Name")   // 'rfc822Name', e.g. "jane@example.com"
+    #define kC4Cert_Hostname          C4STR("dNSName")      // 'dnsName', e.g. "www.example.com"
+    #define kC4Cert_URL               C4STR("uniformResourceIdentifier") // "https://example.com/jane"
+    #define kC4Cert_IPAddress         C4STR("iPAddress")    // *Binary* IP address, e.g. "\x0A\x00\x01\x01"
+    #define kC4Cert_RegisteredID      C4STR("registeredID") // Some sort of domain-specific ID?
```
borrrden, line 69: Is it possible to use `constexpr` here instead, or does that not work well with `C4Slice`?  Having so many public preprocessor defines makes me antsy...

snej: `constexpr` is C++. In C these would have to be `extern C4String`, and then each of them would be a global symbol that had to be exported from the DLL. :-p

snej:
Unit tests failed weirdly in Jenkins. Any idea, @borrrden ?
```
/Users/mobile/jenkins/mobile/workspace/eline_couchbase-lite-core_PR-939/couchbase-lite-core/C/tests/c4DatabaseTest.cc:409: FAILED:
  SIGILL - Illegal instruction signal
```

borrrden:
No I don't know what's going on but it is happening more and more as I noted in the above comment...

PR #940, opened 2020-04-08
Enable all warnings on CMake

borrrden:
`-Wall -Werror` for unix and `/W3 /WX` for Windows


borrrden:
This has gotten so bloated and there is almost nothing of value so.....close

PR #943, opened 2020-04-16
CBL-859: Fix for rekeyed database failing to save documents

borrrden:


borrrden:
I'm definitely fixing the failing test to check for two errors instead of 1 because this test failure wastes a lot of time, but since that is the only failure then this run is also good.

PR #949, opened 2020-04-22
Use C4KeyPair and C4Cert in C4TLSConfig

pasin:
Instead of using C4Slice for setting the private key and the certificate, use C4KeyPair and C4Cert respectively.

#CBL-897


pasin:
Taking the PR back as I think rootClientCerts should be C4Cert instead of C4Slice too.

PR #952, opened 2020-04-23
Fix/binary logging

snej:
C++Tests aren't generating binary logs anymore, because the
LogEncoderTest leaves LogDomain::fileLogLevel != None, which is what
LiteCoreTest checks to make sure no one else has turned logging on
yet.

* Made LogEncoderTest reset the log level to None to fix this.
* LiteCoreTest's logging config was wrong: it specifies a file path
  when it needs to specify an (existing) directory. Changed it to
  create a temp directory.
* Also changed the config's max log size from 1KB(!) to 1MB, plus
  added 5 logs of rollover.

Also fixed a crash in LogDecoder:
The string returned from readStringToken can move the next time a
token is added, since _tokens is a vector. That invalidates the
`format` pointer. Instead, copy the string.

### DIFF
--- a/LiteCore/Support/LogDecoder.cc
+++ b/LiteCore/Support/LogDecoder.cc
@@ -222,8 +222,8 @@ namespace litecore {
             }
 
             // Read the format string, then the parameters:
-            const char *format = readStringToken().c_str();
-            for (const char *c = format; *c != '\0'; ++c) {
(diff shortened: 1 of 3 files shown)


borrrden:
Not sure what's going on with that one test but it's not related to this PR.

PR #953, opened 2020-04-24
Enable iOS support in PublicKey.hh

pasin:
* Made iOS support PERSISTENT_PRIVATE_KEY_AVAILABLE
* Fixed PublicKey+Apple to support both iOS (10.3+) and MacOS.
* Tested manually from CBL side on iOS 13.4.1 (Simulator and Device).
* Plan to test on all supported iOS and MacOS versions on the AWS device farm when the device farm permission is granted.

CBL-860, CBL-861

### DIFF
--- a/Crypto/PublicKey+Apple.mm
+++ b/Crypto/PublicKey+Apple.mm
@@ -262,58 +262,75 @@ namespace litecore { namespace crypto {
 
 
     Retained<PersistentPrivateKey> PersistentPrivateKey::withCertificate(Cert *cert) {
-        SecCertificateRef certRef = SecCertificateCreateWithData(kCFAllocatorDefault,
-                                                        (CFDataRef)cert->data().copiedNSData());
-        if (!certRef)
-            throwMbedTLSError(MBEDTLS_ERR_X509_INVALID_FORMAT); // impossible?
-        CFAutorelease(certRef);
-        SecKeyRef publicKeyRef;
-        checkOSStatus(SecCertificateCopyPublicKey(certRef, &publicKeyRef),
-                      "SecCertificateCopyPublicKey", "get private key from keychain");
-
-        SecIdentityRef identityRef;
-        checkOSStatus(SecIdentityCreateWithCertificate(nullptr, certRef, &identityRef),
-                      "SecIdentityCreateWithCertificate", "get private key from keychain");
-        CFAutorelease(identityRef);
-        SecKeyRef privateKeyRef;
-        checkOSStatus(SecIdentityCopyPrivateKey(identityRef, &privateKeyRef),
-                          "SecIdentityCopyPrivateKey", "get private key from keychain");
-        auto keySize = unsigned(8 * SecKeyGetBlockSize(privateKeyRef));
-        return new KeychainKeyPair(keySize, publicKeyRef, privateKeyRef);
+        @autoreleasepool {
+            SecCertificateRef certRef = SecCertificateCreateWithData(kCFAllocatorDefault,
+                                                                     (CFDataRef)cert->data().copiedNSData());
+            if (!certRef)
+                throwMbedTLSError(MBEDTLS_ERR_X509_INVALID_FORMAT); // impossible?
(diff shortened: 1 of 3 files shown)


Crypto/PublicKey+Apple.mm
```diff
@@ -262,58 +262,75 @@ virtual int _sign(int/*mbedtls_md_type_t*/ mbedDigestAlgorithm,
...
-        return new KeychainKeyPair(keySize, publicKeyRef, privateKeyRef);
+        @autoreleasepool {
+            SecCertificateRef certRef = SecCertificateCreateWithData(kCFAllocatorDefault,
+                                                                     (CFDataRef)cert->data().copiedNSData());
+            if (!certRef)
+                throwMbedTLSError(MBEDTLS_ERR_X509_INVALID_FORMAT); // impossible?
+            CFAutorelease(certRef);
+            
+            // Get public key from the certificate:
+            SecKeyRef publicKeyRef;
```
jayahariv, line 273: CFAutoRelease?

Crypto/PublicKey+Apple.mm
```diff
@@ -262,58 +262,75 @@ virtual int _sign(int/*mbedtls_md_type_t*/ mbedDigestAlgorithm,
...
+            // Lookup private key by using the public key hash:
+            auto privateKeyRef = (SecKeyRef)findInKeychain(@{
+                (id)kSecClass:                  (id)kSecClassKey,
+                (id)kSecAttrKeyType:            (id)kSecAttrKeyTypeRSA,
+                (id)kSecAttrKeyClass:           (id)kSecAttrKeyClassPrivate,
+                (id)kSecAttrApplicationLabel:   publicKeyHash,
+                (id)kSecReturnRef:              @YES
+            });
+            if (!privateKeyRef) {
+                CFAutorelease(publicKeyRef);
```
jayahariv, line 307: why are we auto releasing here, instead of outside

Crypto/PublicKey+Apple.mm
```diff
@@ -262,58 +262,75 @@ virtual int _sign(int/*mbedtls_md_type_t*/ mbedDigestAlgorithm,
...
-
-            // Get the public key from the cert, not from the private key, because calling
-            // SecKeyCopyPublicKey results in a key ref that doesn't allow its data to be read,
-            // for some reason (bug?)
-
-            SecKeyRef publicKeyRef;
-            checkOSStatus(SecCertificateCopyPublicKey(certRef, &publicKeyRef),
-                          "SecCertificateCopyPublicKey", "get private key from keychain");
+            
+            auto publicKeyRef = SecKeyCopyPublicKey(privateKeyRef);
```
jayahariv, line 330: same, any reason why not auto release?

jayahariv:
This doesn't mean we completely drop support of iOS 9 right, but only the Certificate classes(P2P TLS).

Crypto/PublicKey+Apple.mm
pasin: Because we want to release it here only when there is an error.

Crypto/PublicKey+Apple.mm
pasin: Because we want to release it here only when there is an error.

pasin:
The target iOS version for CBL iOS was set to 10.0 in Hydrogen (Master) so I didn't see the build issue before. I'm taking the PR out and will check whether we need to support iOS 9 for the listener and TLS feature.

Crypto/PublicKey+Apple.mm
borrrden: Then why `CFAutoRelease` instead of `CFRelease` ?

pasin: It was supposed to be CFRelease. It's a bug.

Crypto/PublicKey+Apple.mm
pasin: It was supposed to be CFRelease even though CFAutoRelease when returning is fine too.

PR #954, opened 2020-04-24
CBL-848: Don't allow two listeners on the same port

borrrden:
This has the side effect of not being able to restart a listener for up to 60 seconds on some platforms as their sockets linger in TIME_WAIT or whatever it is called.

### DIFF
--- a/LiteCore/Support/Error.cc
+++ b/LiteCore/Support/Error.cc
@@ -140,6 +140,41 @@ namespace litecore {
 
     struct codeMapping { int err; error::Domain domain; int code; };
 
+#ifdef WIN32
+#define MAPWSA(code) { WSA##code,error::POSIX,code }
+#define MAPWSATO(from, to) { from,error::POSIX,to }
+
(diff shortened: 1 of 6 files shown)


borrrden:
@snej I modified the tests to not use the same port every time in the listener harness.  I don't think this should cause any problems, but do you think it will?

borrrden:
I only used SO_REUSEPORT becuase SO_REUSEADDR did not fix the problem of restarting the listener on Linux specifically.  I'll do another verification of what happens though, but it seems like Linux does not honor SO_REUSEADDR when it is considering another socket in the TIME_WAIT state as BSD seems to do.

borrrden:
Linux, at least the Linux I tried, is the one that has to wait before listening again for whatever reason....and Windows has no facility to allow this without allowing multiple listeners on the same address and port.

borrrden:
Interesting....the latest sockpp updates cause the P2P test to hang...

PR #956, opened 2020-04-28
Enable iOS support in PublicKey.hh

pasin:
* Made iOS support PERSISTENT_PRIVATE_KEY_AVAILABLE
* Fixed PublicKey+Apple to support both iOS (10.0+) and MacOS.
* Tested manually from CBL side on iOS 13.4.1 (Simulator and Device).
* Plan to test on all supported iOS and MacOS versions on the AWS device farm when the device farm permission is granted.

CBL-860, CBL-861

### DIFF
--- a/Crypto/CertificateTest.cc
+++ b/Crypto/CertificateTest.cc
@@ -170,14 +170,21 @@ TEST_CASE("Persistent key and cert", "[Certs]") {
     Cert::IssuerParameters issuerParams;
     issuerParams.validity_secs = 3600*24;
     Retained<Cert> cert = new Cert(DistinguishedName(kSubjectName), issuerParams, key);
-
+    
+    // Try to get private key with public key:
+    key = PersistentPrivateKey::withPublicKey(pubKey);
+    REQUIRE(key);
+    CHECK(pubKey->data(KeyFormat::Raw) == key->publicKeyData(KeyFormat::Raw));
+    CHECK(pubKey->data(KeyFormat::DER) == key->publicKeyData(KeyFormat::DER));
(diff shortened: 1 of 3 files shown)


pasin:
Tests in PR validation failed but they don't seem to be related to the PR.

```
/var/lib/jenkins/mobile/workspace/eline_couchbase-lite-core_PR-956/couchbase-lite-core/LiteCore/tests/../../Replicator/tests/ReplicatorAPITest.hh:304: FAILED:

  CHECK( status.error.code == 0 )
with expansion:
  5 == 0
```

pasin:
The MacOS Deployment target is set to 10.12 [here](https://github.com/couchbase/couchbase-lite-core/blob/master/Xcode/xcconfigs/Project.xcconfig#L28) so I didn't do the explicit check in the code. Is that the right place to check for the MacOS deployment target?

borrrden:
This is at odds with our [docs](https://docs.couchbase.com/couchbase-lite/current/swift.html#supported-versions) which say we need to go back to 10.11 (10.9 and 10.10 to be removed in 2.8)

pasin:
So which one is correct. CBL iOS Master branch also sets the target version to 10.12 (2.7.0 set to 10.11).

borrrden:
Don't ask me, I'm just the messenger :)

pasin:
Just confirm with @rajagp 10 times that we will do 10.12+ in 2.8.0. We will list 10.12 as a deprecating version.

pasin:
BTW, I did add macOS 10.12 check. Please review.

borrrden:
Looks good except for the error message only says iOS 10 right?

pasin:
I will update the message. Thanks for reviewing.

PR #958, opened 2020-05-05
Fix how to get public key from private key and update remove key logic

pasin:
* SecKeyCopyPublicKey from the private key doesn’t work when the public key is not stored in the KeyChain. Instead getting the public key from the certificate via trust.

* Updated remove key to also remove public key (if available).

* In findInKeychain(), ignore some internal C++ exception thrown inside the security framework which doesn’t seem to harm anything.

* In findInKeychain(), do not Assert the result so that it can be used to check if thing exists in the key chain or not.

### DIFF
--- a/Crypto/PublicKey+Apple.mm
+++ b/Crypto/PublicKey+Apple.mm
@@ -90,12 +90,13 @@ namespace litecore { namespace crypto {
 
     static CFTypeRef CF_RETURNS_RETAINED findInKeychain(NSDictionary *params NONNULL) {
         CFTypeRef result = NULL;
+        ++gC4ExpectExceptions;  // ignore internal C++ exceptions in Apple Security framework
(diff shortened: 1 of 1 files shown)


pasin:
An unrelated unit test failed:
```
/Users/mobile/jenkins/mobile/workspace/eline_couchbase-lite-core_PR-958/couchbase-lite-core/Replicator/tests/ReplicatorLoopbackTest.cc:218

/Users/mobile/jenkins/mobile/workspace/eline_couchbase-lite-core_PR-958/couchbase-lite-core/Replicator/tests/ReplicatorLoopbackTest.cc:218: FAILED:

  SIGTERM - Termination request signal
```

pasin:
Found a mistake of missing 'else' cause so I just re-push a fix.

PR #961, opened 2020-05-06
CBL-895: Provide reset option to c4repl_start

borrrden:
As a part of making the replicator restartable, instead of passing the option as part of initialization pass the reset flag in the start method

### DIFF
--- a/Replicator/Replicator.cc
+++ b/Replicator/Replicator.cc
@@ -470,7 +470,6 @@ namespace litecore { namespace repl {
                 stop();
                 return false;
             } else if (reset) {
-                _checkpointer.reset();
                 logInfo("Ignoring local checkpoint ('reset' option is set)");
             } else {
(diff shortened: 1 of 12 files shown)


Replicator/Replicator.cc
```diff
@@ -469,7 +469,8 @@ namespace litecore { namespace repl {
                 gotError(error);
                 stop();
                 return false;
-            } else if (_options.properties[kC4ReplicatorResetCheckpoint].asBool()) {
+            } else if (reset) {
+                _checkpointer.reset();
```
pasin, line 473: Doesn't the checkpoint has already been reset as of line 461 when calling _checkpointer.read() with reset = true?

borrrden: Oh right good catch.  This is leftover from a previous failed way of doing it.

PR #962, opened 2020-05-07
Database reindexing [CBSE-8175]

snej:
Added a new function `c4db_maintenance` and enum `C4MaintenanceType`, whose only value so far is `kC4Reindex`.

Reindexing simply executes the SQLite command `REINDEX`, which erases all indexes and rebuilds them from the table values.

### DIFF
--- a/C/c4.def
+++ b/C/c4.def
@@ -30,6 +30,7 @@ c4db_copy
 c4db_delete
 c4db_deleteAtPath
 c4db_compact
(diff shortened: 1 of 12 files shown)


snej:
Oops, Github is showing a commit that isn't relevant. Ignore "CBL-619: Try to plug...";

snej:
Wrong destination branch, sorry

PR #963, opened 2020-05-07
Database reindexing [CBL-8175]

snej:
New function `c4db_maintenance(db, kC4Reindex, &err)` will rebuild the SQLite indexes.

Fixes CBL-8175


borrrden:
For now, at least, this won't be merged into mercury but this code will live on in the `hotfix/2.7.1.1` branch

PR #964, opened 2020-05-08
(More) database maintenance

snej:
* Merged the `c4db_maintenance` commit from #963 into master.
* Added maintenance option for compaction, obsoleting `c4db_compact()` (now informally deprecated)
* Added maintenance option for SQLite integrity check.

### DIFF
--- a/C/c4.def
+++ b/C/c4.def
@@ -316,6 +316,7 @@ c4db_getRemoteDBID
 c4db_exists
 c4db_startHousekeeping
 c4db_findDocAncestors
+c4db_maintenance
 
 c4doc_removeRevisionBody
 c4doc_getForPut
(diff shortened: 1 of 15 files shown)


snej:
The Mac tests failed in sync listener code. Totally unrelated to this PR, so I say we merge it anyway. What do you think, @borrrden ?
```
	>>> P2P Sync connection count
16:35:47.203339| WARNING: Couldn't find C4Replicator in RESTSyncListener's _replicators
Assertion failed: (ec == 0), function unlock, file /BuildRoot/Library/Caches/com.apple.xbs/Sources/libcxx/libcxx-400.9.4/src/mutex.cpp, line 48.
```

borrrden:
Yes, but macOS has become dangerously unstable lately.  Almost every single PR validation it fails for one reason or another....

snej:
FYI, I just pushed PR #971 to speculatively fix the above crash.

PR #969, opened 2020-05-16
CBL-960 Add enableDeltaSync parameter to listener config

borrrden:


borrrden:
Depends on [EE branch](https://github.com/couchbase/couchbase-lite-core-EE/tree/PR-969)

borrrden:
This new mac crash is back again, but since it showed up in another PR that makes it not related to this one.....

PR #970, opened 2020-05-18
Fix withPublicKey() when the public key is not in the KeyChain

pasin:
The public key may not be in the KeyChain as a separate entity, for example, when using SecPKCS12Import to import the identity from P12 created from openSSL tool on macOS, the public key entity will not be presented in the KeyChain. When that happens, create the SecKeyRef of the public key from the public key data instead.

CBL-947

### DIFF
--- a/Crypto/PublicKey+Apple.mm
+++ b/Crypto/PublicKey+Apple.mm
@@ -392,12 +392,26 @@ namespace litecore { namespace crypto {
                 if (!privateKeyRef)
                     return nullptr;
                 
+                ++gC4ExpectExceptions;
                 auto publicKeyRef = SecKeyCopyPublicKey(privateKeyRef);
+                --gC4ExpectExceptions;
                 if (!publicKeyRef) {
-                    CFRelease(privateKeyRef);
-                    throwMbedTLSError(MBEDTLS_ERR_X509_INVALID_FORMAT); // Impossible?
+                    // If the public key is not in the KeyChain, SecKeyCopyPublicKey() will return null.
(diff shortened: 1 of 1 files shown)


pasin:
Unrelated test failed:

```
/Users/mobile/jenkins/mobile/workspace/eline_couchbase-lite-core_PR-970/couchbase-lite-core/C/tests/c4DatabaseTest.cc:402: FAILED:

  SIGILL - Illegal instruction signal
```

Crypto/PublicKey+Apple.mm
```diff
@@ -392,12 +392,26 @@ virtual int _sign(int/*mbedtls_md_type_t*/ mbedDigestAlgorithm,
...
+                --gC4ExpectExceptions;
                 if (!publicKeyRef) {
-                    CFRelease(privateKeyRef);
-                    throwMbedTLSError(MBEDTLS_ERR_X509_INVALID_FORMAT); // Impossible?
+                    // If the public key is not in the KeyChain, SecKeyCopyPublicKey() will return null.
+                    // Create a SecKeyRef directly from the publicKey data instead:
+                    NSDictionary* attrs = @{
+                        (id)kSecAttrKeyType:            (id)kSecAttrKeyTypeRSA,
+                        (id)kSecAttrKeyClass:           (id)kSecAttrKeyClassPublic,
+                        (id)kSecAttrKeySizeInBits:      @2048
```
snej, line 404: This shouldn't be here, should it? The key size isn't fixed, it depends on the actual key data.

PR #976, opened 2020-05-21
Compile LiteCoreREST and LiteCoreWebsocket into shared lib

borrrden:
.NET bindings need all symbols to be exported from a dynamic library.  This has the side effect of putting in some logic that is unused in CBL into the CE build, but to get around that there would have to be a lot of rewrites to the `c4listener` API to remove the unused HTTP REST mode.

### DIFF
--- a/C/c4.def
+++ b/C/c4.def
@@ -163,6 +163,15 @@ c4socket_completedWrite
 c4socket_received
 c4socket_gotHTTPResponse
 
+c4listener_availableAPIs
+c4listener_start
+c4listener_free
+c4listener_shareDB
+c4listener_unshareDB
+c4listener_getPort
+c4listener_getURLs
+c4listener_getConnectionStatus
+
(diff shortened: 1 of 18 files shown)


snej:
I get a Python error when building in Xcode:
```
  File "/Users/snej/Couchbase/CBL/couchbase-lite-ios/vendor/couchbase-lite-core/C/scripts/generate_export_list.py", line 11
    def write_line_to(line: str, files: list):
                          ^
SyntaxError: invalid syntax
Command PhaseScriptExecution failed with a nonzero exit code
```

borrrden:
Don't call it using `python` (which is the now defunct python 2.x).  Use `python3` or better yet just execute it directly.

snej:
Fixed the error (Xcode was launching the script with Python2), and got the conditional export files working.

I also added a commit that removes `c4repl_newLocal` from the CE exports.

Issue #978, opened 2020-05-25
c4db_findDocAncestors usage = memory leak?


Dushistov:
valgrind reports for my program:

```
==132188== 20 bytes in 4 blocks are definitely lost in loss record 28 of 375
==132188==    at 0x483977F: malloc (vg_replace_malloc.c:307)
==132188==    by 0x4AB9A22: fleece::sharedBuffer::operator new(unsigned long, unsigned long) (FLSlice.cc:79)
==132188==    by 0x4AB9C0E: FLSlice_Copy (FLSlice.cc:132)
==132188==    by 0x4AF9CF4: fleece::alloc_slice::alloc_slice(fleece::pure_slice) (slice.cc:371)
==132188==    by 0x4A32D14: litecore::SQLiteKeyStore::withDocBodies(std::vector<fleece::slice, std::allocator<fleece::slice> > const&, std::function<fleece::alloc_slice (fleece::slice, fleece::slice, unsigned long)>) (SQLiteKeyStore.cc:463)
==132188==    by 0x48C0B99: c4Internal::TreeDocumentFactory::findAncestors(std::vector<fleece::slice, std::allocator<fleece::slice> > const&, std::vector<fleece::slice, std::allocator<fleece::slice> > const&, unsigned int, bool, unsigned int) (TreeDocument.cc:606)
==132188==    by 0x4887411: c4db_findDocAncestors::{lambda()#1}::operator()() const (c4Database.cc:305)
==132188==    by 0x4888E67: void fleece::function_ref<void ()>::callback_fn<c4db_findDocAncestors::{lambda()#1}>(long) (function_ref.hh:59)
==132188==    by 0x494C393: fleece::function_ref<void ()>::operator()() const (function_ref.hh:72)
==132188==    by 0x4B0C9F1: c4Internal::tryCatch(C4Error*, fleece::function_ref<void ()>) (c4ExceptionUtils.cc:35)
==132188==    by 0x48875CD: c4db_findDocAncestors (c4Database.cc:300)
==132188==    by 0x4AA410F: litecore::repl::RevFinder::_findOrRequestRevs(fleece::Retained<litecore::blip::MessageIn>, litecore::repl::DocIDMultiset*, std::function<void (std::vector<bool, std::allocator<bool> >)>)::{lambda(c4Database*)#2}::operator()(c4Database*) const (RevFinder.cc:143)
```

I can reproduce it with couchbase-lite-core tests:
```sh
[...]

PR #979, opened 2020-05-26
fix memory leak in RevFinder::_findOrRequestRevs

Dushistov:
ref #978

### DIFF
--- a/Replicator/RevFinder.cc
+++ b/Replicator/RevFinder.cc
@@ -153,7 +153,7 @@ namespace litecore { namespace repl {
                 for (size_t i = 0; i < nChanges; ++i) {
                     alloc_slice docID(docIDs[i]);
                     alloc_slice revID(revIDs[i]);
-                    alloc_slice anc(ancestors[i]);
+                    alloc_slice anc(std::move(ancestors[i]));
                     if (anc == kC4AncestorExistsButNotCurrent) {
                         // This means the rev exists but is not marked as the latest from the
                         // remote server, so I better make it so:


Dushistov:
This fix memory leak for me,
because of `vector<C4StringResult> ancestors(nChanges)` = `vector<FLSliceResult>`,
is type without destructor, so it should be freed manually. 
And converting to alloc_slice is nice way to free memory,
but should be used  `alloc_slice(FLSliceResult &&)`
instead of `alloc_slice(const FLSliceResult &)`.

This is done in the supposition that there is no exception possible in the 
`for (size_t i = 0; i < nChanges; ++i) {`,
if exception possible may be better convert `vector<C4StringResult> ancestors` to `vector<alloc_slice>` right after `c4db_findDocAncestors` call.

borrrden:
An elegant way to tackle this!  And to your point about exceptions, as a rule our C API functions are all `noexcept` (especially the public ones, since they serve as a bridge to other languages that have no way of dealing with native exceptions).  If there is a possibility for an error condition then the function will have a `C4Error *` parameter that will capture the error information.  I'm going to discuss about how to handle this situation more explicitly and then decide on this PR.

Dushistov:
> as a rule our C API functions are all noexcept

but all functions inside loop is not C API functions and not marked as noexcept
```
 _db->setDocRemoteAncestor(docID, revID);
 replicator()->docRemoteAncestorChanged(docID, revID);
```
So, if for example `nChanges == 2`, and exception will occur during the first element handling,
then memory leak happens.

snej:
Thanks @Dushistov, that's a good fix ... I'll look at the surrounding code to see if there could be an exception in that loop.

snej:
Exceptions won't be an issue in that loop. The replicator code tends to use `C4Error*` rather than throw exceptions.

PR #981, opened 2020-05-27
Bump Fleece and add test for lack of precision in some math results

borrrden:
See CBL 978

### DIFF
--- a/LiteCore/Query/SQLitePredictionFunction.cc
+++ b/LiteCore/Query/SQLitePredictionFunction.cc
@@ -108,14 +108,6 @@ namespace litecore {
         return (i1.count() == i2.count());
     }
 
-    static void writeResultNumeric(sqlite3_context* ctx, double num){
-        if(num <= INT64_MAX && num >= INT64_MIN && num == floor(num)) {
-            sqlite3_result_int64(ctx, (sqlite3_int64)num);
-        } else {
-            sqlite3_result_double(ctx, num);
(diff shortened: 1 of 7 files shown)


borrrden:
Most of the work here is downstream in Fleece

LiteCore/Query/SQLitePredictionFunction.cc
```diff
@@ -108,6 +108,14 @@ namespace litecore {
         return (i1.count() == i2.count());
     }
 
+    static void writeResultNumeric(sqlite3_context* ctx, double num){
```
snej, line 111: I don't think this is necessary, since these results are unlikely to be integral.

borrrden: More than half of our assertions on this method use integral results.  Is it ok to change all of them to float instead (query distance functions)

snej: Sure.

borrrden:
Latest failure is due to another issue being investigated.  However I noticed there are N1QL parser tests that fail that don't seem to be included in the Cmake project.  Need clarification on if they are needed.

snej:
N1QLParserTest.cc is in the CppTests target in the Xcode build, so yeah, it should be in CMake too.

borrrden:
@snej Take a look at the latest commit for the N1QL fixes

PR #982, opened 2020-05-29
Fixed some delta-sync race conditions

snej:
Fixing CBL-986, manifested by transient failures of the "Continuous Push From Both Directions" unit test.

### Commit 1: Improved delta handling on push and pull side:

* RevToSend now preserves the order of its `ancestorRevIDs` so that
  the peer's preferred rev can be used as the delta source. It used
  to use a `set`, which sorts the revIDs, meaning that the _oldest_
  one (lowest gen) would be tried first, which is bad!
* If c4doc_put can't reassemble the doc body because the delta source
  revision isn't available, _and_ the operation would fail anyway
  because it would create a conflict, then it now returns
  kC4ErrorConflict instead of kC4ErrorDeltaBaseUnknown. That helps
  the replicator & app handle the situation more appropriately.
* Added some more logging (mostly debug-level) to make it easier to
  debug situations like this in the future.

### Commit 2: Retry revs that couldn't be inserted due to delta failures

* Inserter flags a delta error as 'transient'.
* Pusher detects a delta error received from the peer and retries sending 
  the revision, this time without a delta.

NOTE: SG will need a change corresponding to the Pusher change above. (See CBG-881)

### Commit 3: Use HTTP 422 as BLIP error for delta failure

Use a less LiteCore-specific error code when sending a BLIP error for a failed delta insertion, since SG has to be able to recognize it too.

### DIFF
--- a/LiteCore/Database/TreeDocument.cc
+++ b/LiteCore/Database/TreeDocument.cc
@@ -340,7 +340,7 @@ namespace c4Internal {
                                 "Unknown source revision ID for delta", outError);
                 } else if (!selectedRev.body.buf) {
                     recordError(LiteCoreDomain, kC4ErrorDeltaBaseUnknown,
-                                "Unknown source revision body for delta", outError);
+                                "Missing source revision body for delta", outError);
                 } else {
                     slice delta = (rq.allocedBody.buf)? slice(rq.allocedBody) : slice(rq.body);
                     body = rq.deltaCB(rq.deltaCBContext, &selectedRev, delta, outError);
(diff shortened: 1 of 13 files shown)


borrrden:
I think it's possible that GCC 7 didn't get the memo about underscore:

> /home/travis/build/couchbase/couchbase-lite-core/LiteCore/Database/TreeDocument.cc: In member function ‘virtual int32_t c4Internal::TreeDocument::putExistingRevision(const C4DocPutRequest&, C4Error*)’:
>
> /home/travis/build/couchbase/couchbase-lite-core/LiteCore/Database/TreeDocument.cc:375:35: error: unused variable ‘_’ [-Werror=unused-variable]
>
>      auto [_, i] = _versionedDoc.findCommonAncestor(revIDBuffers, rq.allowConflict);

snej:
@borrrden The Windows tests failed but I don't know what happened; the log ends with
```
Sending interrupt signal to process
	    [10.5803 sec]
	>>> Continuous Super-Fast Push
script returned exit code -1
```

borrrden:
If you see a gray circle it means the job was cancelled, which in this case means it hit its 40 minute timeout.  I only recently bumped from 30 to 40 minutes so I'm perturbed that now the job is hitting even this new limit....

borrrden:
I re-ran it and Windows failed with the same error.  Does this PR require any EE commits?

snej:
No ... can't imagine why this would happen. Have you tried running this PR's tests on Windows locally? If they hang, then maybe you could find out why.

borrrden:
Sorry badly worded on my part.  I meant "the same error that this PR is trying to fix".  Not the hang error (there was no hang error, the problem was a rogue process on the mac machine sucking up all the resources and causing the iOS build to be so slow that Windows didn't have a chance to finish).

borrrden:
Oh interesting this time it is slightly different.  It expecting _push_ errors (probably conflict) and not getting any.  I assume if things happen in just the correct order that this can happen and the test should account for that?

snej:
Hm. The expectation of push errors predates this PR. I guess it's _possible_ that the test could finish without a conflict, but it seems pretty unlikely. I wonder if something else went wrong in that test run that isn't showing up as an assertion?

borrrden:
I re-ran the validation and this time I saw logs about conflicts and the test passed.  So I guess there are certain rare conditions in which the test can run completely without getting any push errors....

borrrden:
(I also ran the tests on Linux about 5 times in a row without getting any pull errors)

PR #984, opened 2020-05-29
[API] CBL-948: Make listener return correct scheme for server mode

borrrden:
The caller must now specify which API they are intending to get the server addresses for


borrrden:
Depends on EE as well (PR-948 branch).  Failures due to issue being investigated outside the scope of this PR (See #982).

PR #990, opened 2020-05-30
add option to enable sqlite r-tree extension

Dushistov:
In my project I uses sqlite3 directly (for spatial read-only data) and 
via couchbase-lite-core for other things, all was fine until I need link all libraries statically.

So because of I have to use sqlite3.o from LiteCoreStatic I need the way to build
sqlite with r-tree support, so I suggest to add this option to couchbase-lite-core.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -66,6 +66,7 @@ option(LITECORE_DISABLE_ICU "Disables ICU linking" OFF)
 option(DISABLE_LTO_BUILD "Disable build with Link-time optimization" OFF)
 option(LITECORE_BUILD_TESTS "Builds C4Tests and CppTests" ON)
 option(SANITIZE_FOR_DEBUG_ENABLED "Enable -fsanitize for Debug Build" ON)
+option(SQLITE_ENABLE_RTREE "Enable r-tree extension in sqlite3 module" OFF)
 
 add_definitions(
     -DCMAKE                  # Let the source know this is a CMAKE build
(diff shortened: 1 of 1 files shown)


borrrden:
The way this PR accomplishes this opens up a slippery slope into lack of maintainability.  I'm not wiling to start adding in CMake options for every SQLite option that an external project could need.  However, given that many projects probably do use SQLite, I think the solution could be to compile sqlite into its own unit and add an option to not make it a dependency of LiteCoreStatic.  That way you could link in your SQLite instead (with the caveat that if you don't compile in any of the options that LiteCore needs then you will get bad behavior)

snej:
We don't support use of LiteCore directly, only as a component of Couchbase Lite.
And after all the horrible issues with static libraries that we had with CBL 1.x, we don't support using our code as a static library.
The best resolution of this is to use Couchbase Lite For C, linked as a dynamic library.

Dushistov:
@snej 

> We don't support use of LiteCore directly
> The best resolution of this is to use Couchbase Lite For C, linked as a dynamic library.

But you do exactly the opposite here:
https://github.com/couchbaselabs/couchbase-lite-C/blob/37617f799c82bac33f2432e14cd60863b6c6ce5e/bindings/rust/CouchbaseLite/build.rs#L92

Or it is one of "Quod licet Iovi, non licet bovi" ?

snej:
It's in early development and not everything is kosher yet. ¯\_(ツ)_/¯ And if we ship a Rust binary, we might statically link that ... although I wasn't aware that a stable Rust ABI is that far away, so I guess that's not realistic.

I can guarantee that Couchbase will not provide support for customers who build their own copies of CBL from source, because frankly, some of those customers aren't smart enough to do it correctly. (Our support team has wasted a lot of time on a few developers who it turned out were changing build settings and even modifying library code but not telling us, then expecting us to fix their regressions.)

Obviously concerns like that don't apply to open-source users! But we do have architectural decisions that are driven by factors like these, because we're a business.

PR #977, opened 2020-05-22
add ability to build debug build without -fsanitize on macos

Dushistov:
Note that this changes nothing (the option turn on by default),
and you need explicity set to to OFF to change behaviour before patch.
To reproduce problem you need old enough cmake (I used cmake 3.9.6 in CI builds).

This mainly to fix CI issue in my attempt to migrate to current master (Dushistov/couchbase-lite-rust#22), but I suppose it is not bad idea to have this options to make possible to have Debug build without huge memory consumptions caused by santizers.

Related part of my CI failure that this patch fix:
```
[couchbase-lite-core-sys 0.1.0]   "___asan_report_load1", referenced from:
[couchbase-lite-core-sys 0.1.0]       fleece::impl::Value::isUndefined() const in libFleeceStatic.a(Fleece.cc.o)
[couchbase-lite-core-sys 0.1.0]       fleece::impl::Value::isUnsigned() const in libFleeceStatic.a(Fleece.cc.o)
[couchbase-lite-core-sys 0.1.0]       fleece::impl::Value::isDouble() const in libFleeceStatic.a(Fleece.cc.o)
[couchbase-lite-core-sys 0.1.0]       fleece::impl::Value::tag() const in libFleeceStatic.a(Fleece.cc.o)
[couchbase-lite-core-sys 0.1.0]       fleece::impl::FLEncoderImpl::~FLEncoderImpl() in libFleeceStatic.a(Fleece.cc.o)
[couchbase-lite-core-sys 0.1.0]       fleece::pure_slice::_strlen(char const*) in libFleeceStatic.a(Fleece.cc.o)
[couchbase-lite-core-sys 0.1.0]       fleece::impl::internal::HeapCollection::isChanged() const in libFleeceStatic.a(Fleece.cc.o)
```

PR #991, opened 2020-05-31
Allow build with clang++/libstdc++

Dushistov:
The problem is almost the same as in #977 . 
Such decision  as usage of libc++ instead of libstdc++ (or enabling ASAN)
 should be made on the level above and universally for all project,
it is almost impossible use libstdc++ and libc++ at the same time, and use ASAN only for part of build tree. In one of my CI steps I uses clang for test build, but I have boost/Qt librararies pre-build with 
libstdc++ so I can not use libc++.

To not introducing new option I suggest changing old one, because of purpose really the same,
make decision on the level above.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -66,6 +66,11 @@ option(LITECORE_DISABLE_ICU "Disables ICU linking" OFF)
 option(DISABLE_LTO_BUILD "Disable build with Link-time optimization" OFF)
 option(LITECORE_BUILD_TESTS "Builds C4Tests and CppTests" ON)
 option(SANITIZE_FOR_DEBUG_ENABLED "Enable -fsanitize for Debug Build" ON)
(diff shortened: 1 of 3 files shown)


Dushistov:
This also fixes work of clang based instruments like: clangd, clang static analyzer,
because of they used by default headers files from libstdc++ on Linux, 
so can not understand `CONCAT`.

borrrden:
I think I'd prefer it if the other options (minus SQLite RTree, see the other PR) were still there, but that the default were ON or OFF based on the value of the new maintainer variable.  Would that also meet your needs?

Dushistov:
@borrrden 

>  think I'd prefer it if the other options

I update PR to preserve `SANITIZE_FOR_DEBUG_ENABLED`.

borrrden:
Has this been addressed by #1010 ?

Dushistov:
Yes

PR #993, opened 2020-06-01
Fix cert validation on iOS (CBL-995)

snej:
iOS for some reason doesn't make the list of trusted root certs available to 3rd party apps. This means we can't give mbedTLS the list of roots, so it can't validate peer certificates.

However, mbedTLS _does_ have an API to register a callback that will be asked to find the root cert that signed a given peer cert, and we _can_ do that on iOS using SecTrust.

Includes tiny change to mbedTLS's config header to enable the function we need.
Includes addition to sockpp's mbedtls_context to provide a callback.

Fixes [CBL-995](https://issues.couchbase.com/browse/CBL-995)

### DIFF
--- a/Crypto/Certificate.cc
+++ b/Crypto/Certificate.cc
@@ -17,6 +17,7 @@
 //
 
 #include "Certificate.hh"
+#include "TLSContext.hh"
 #include "Defer.hh"
 #include "Logging.hh"
 #include "Error.hh"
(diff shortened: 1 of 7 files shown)


snej:
Travis builds failed to start, with `gpg: keyserver receive failed: keyserver error`. But I don't have privileges to restart them, for some reason.

snej:
Linux hit the not-uncommon "Continuous Push From Both Sides" failure.
macOS got the ~BuiltInWebSocket() crash, which I also saw a minute ago in PR #996.

PR #994, opened 2020-06-01
API: Added c4cert_isSelfSigned()

snej:


snej:
The usual Continuous Push From Both Sides failure, on Windows.

Issue #992, opened 2020-05-31
sporadic ReplicatorLoopbackTest.hh:117: FAILED


Dushistov:
Time to time I see such failure in github actions CI/linux:
```
couchbase-lite-core/Replicator/tests/ReplicatorLoopbackTest.cc:635
.......................................................................................................................

couchbase-lite-core/Replicator/tests/ReplicatorLoopbackTest.hh:117: FAILED:
  CHECK( asVector(_docPullErrors) == asVector(_expectedDocPullErrors) )
with expansion:
  { "doc" } == {  }
```

I was able to reproduce it locally vai sequentially running `./CppTests -r list`.
I tested current master (775893eddbfaff3569f1d351e12d1b27eb19fd40).
I ran `./CppTests -r list` 8 times, I got failure during 0, 4, 8 run.
I attach run 0 - bad, and  next one 1 - good.
[logs.zip](https://github.com/couchbase/couchbase-lite-core/files/4707911/logs.zip)

The similar problem was triggered in #988 in jenkins/linux,
but I run my test against vanilla master without my changes.

borrrden:
We are aware of this issue and currently working on it, there is a PR being worked on at the moment to address it ( #982 )

borrrden:
This should be fixed at this point.  I saw once that a failure that looks very similar happened (push error vs pull error) but not the pull error again.  Let us know if you see it again.

PR #995, opened 2020-06-02
build sqlite3 as separate library

Dushistov:
Build sqlite3 as a separate module for the possibility of replacement.

close #990

### DIFF
--- a/C/tests/CMakeLists.txt
+++ b/C/tests/CMakeLists.txt
@@ -62,6 +62,7 @@ target_link_libraries(
     FleeceBase
     mbedcrypto
     mbedtls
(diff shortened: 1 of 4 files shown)


Dushistov:
The windows/linux failures are related to #992

borrrden:
Thanks!  For future plans I plan to make this an exported dependency of static LiteCore, but only in the "maintainer mode" (still mulling the name of that one) so that things that link with LiteCore static will automatically link with sqlite as well unless they opt out of it.

borrrden:
I'll merge this in for now since other things keep taking my focus away from the bigger issue of removing the support lib etc

PR #996, opened 2020-06-02
Fixed memory leaks (most were in tests)

snej:
One leak was in mbedTLS's cert-signing request API; it was a bug in
an earlier change I'd made. Oops.

The one leak in LiteCore itself was just when generating a back-
trace; I didn't fully understand `__cxa_demangle`s very weird memory
management.

The others were all in tests, mostly failures to release a returned
C4StringResult. I fixed most of them by just assigning to an
alloc_slice, whose constructor will adopt the ref and make sure it
gets released.

c4DatabaseInternalTest had some very old code ported from 1.x that
did some sketchy malloc-based copying of slices, and forgot to free
a couple of those. I replaced that with alloc_slice.

c4ThreadingTest was calling c4dbobs_releaseChanges at the wrong time,
after the loop instead of within it.

----

I also made a small optimization to the Document constructor, that I discovered while doing the leak checking. It now re-uses an `alloc_slice` instead of making an unnecessary copy.

----

Tracking issue: [CBL-987](https://issues.couchbase.com/browse/CBL-987)

### DIFF
--- a/C/tests/c4BlobStoreTest.cc
+++ b/C/tests/c4BlobStoreTest.cc
@@ -66,7 +66,7 @@ public:
 TEST_CASE("parse blob keys", "[blob][C]") {
     C4BlobKey key1;
     memset(key1.bytes, 0x55, sizeof(key1.bytes));
-    C4SliceResult str = c4blob_keyToString(key1);
+    alloc_slice str = c4blob_keyToString(key1);
     CHECK(string((char*)str.buf, str.size) == "sha1-VVVVVVVVVVVVVVVVVVVVVVVVVVU=");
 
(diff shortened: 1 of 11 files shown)


snej:
All platforms passed, except macOS got a C++ terminate in ~BuiltInWebSocket, which I've seen before in other PRs, so I don't think there are problems related to this PR.

borrrden:
This crash sometimes happens when the machine is going bonkers like it just was....it's probably indicative of something but the prerequisite of spamming the CPU makes it not very diagnosable.

snej:
I rebased & merged it manually.

PR #997, opened 2020-06-02
Update NDK

bmeike:
Update NDK version

It seems to be working for Android builds.

Remember that you may have to install it...


borrrden:
My scripts have copied your excellent logic for installing it on demand! :)

PR #998, opened 2020-06-03
CBL-1036 : Rollback when fail to update label to cert on MacOS

pasin:
* If the cert is failed to update its label on MacOS, rollback by deleting the added cert.

* In “Persistent key and cert" and "Persistent save duplicate cert or id" test, delete cert label "Jane Doe" that causes the test failure on some machines.


pasin:
PR Validation failed as `stderr: fatal: unable to access 'https://github.com/couchbase/couchbase-lite-core.git/': Could not resolve host: github.com`

snej:
FYI, Jenkins failures were bogus: "unable to access 'https://github.com/couchbase/couchbase-lite-core.git/': Could not resolve host: github.com"

PR #1005, opened 2020-06-05
CBL-988: Stop replicator immediately in connecting state

borrrden:
If a stop is requested, just end the connection without waiting for the onConnect callback that might never happen.

### DIFF
--- a/Networking/WebSockets/WebSocketImpl.cc
+++ b/Networking/WebSockets/WebSocketImpl.cc
@@ -372,8 +372,8 @@ namespace litecore { namespace websocket {
         if(!_didConnect) {
             // The web socket is being requested to close before it's even connected, so just
             // shortcut to the callback and make sure that onConnect does nothing now
+            closeSocket();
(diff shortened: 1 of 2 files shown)


Networking/WebSockets/WebSocketImpl.cc
```diff
@@ -362,6 +369,14 @@ namespace litecore { namespace websocket {
 
     // Initiates a request to close the connection cleanly.
     void WebSocketImpl::close(int status, fleece::slice message) {
+        if(!_didConnect) {
+            // The web socket is being requested to close before it's even connected, so just
```
snej, line 373: This should just call `closeSocket()` so that the underlying socket disconnects. Then it doesn't have to do the stuff below, since `onClose` will be called.

borrrden: Also discussed this offline, but in the end yes I agree with this!  Had to modify the test for this but it's a sane adjustment.

PR #1006, opened 2020-06-05
Fixed potential deadlock in pending-docs API

snej:
Blake discovered that c4repl_isDocumentPending calls the replicator's
validation callback while holding the replicator mutex, which can lead
to a deadlock if the callback calls something else that locks a mutex.

Fixed this by changing the implementation to lock the replicator,
grab the necessary state, unlock, then use the state to get the
results. It actually adds a little helper class encapsulating the
state, which seemed the cleanest way to do it.

### DIFF
--- a/Replicator/c4Replicator.hh
+++ b/Replicator/c4Replicator.hh
@@ -175,34 +175,12 @@ struct C4Replicator : public RefCounted,
         _onBlobProgress   = nullptr;
     }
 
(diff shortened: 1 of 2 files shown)


bmeike:
... and now I know what `emplace` means....

snej:
>... and now I know what emplace means....

I know, right? Someone on the C++ standards committee must have dug through a thesaurus while they were deciding what to name that method.

PR #1007, opened 2020-06-11
CBL-954: Return 403 and stop for readonly violation

borrrden:
Now if a listener is set up to not allow a certain replication mode (push or pull), an attempt to exercise that path will result in a 403 error and replication termination.  The following things needed to happen to achieve that goal:

1. Instead of returning `websocket::kCodeUnexpectedCondition` for every error that can stop the replication (regardless or whether it is fatal or not), return `websocket::kCloseAppPermanent` if it is fatal (to avoid retrying) and `websocket::kCloseAppTransient` if the library should retry (for example, in the case of 503, which stops replication to back off and try later)
2. Modify an assert in ReplicatorAPITest that checks that an error never happens during a busy state
3. Call doneWithRev on all currently processing revs so if an error BLIP message is received so the library doesn't hang waiting for them to finish
4. Register a 403 returning handler for the BLIP messages specific to push and pull

For review: Are the above choices appropriate and are the messages for the errors OK?

### DIFF
--- a/REST/tests/SyncListenerTest.cc
+++ b/REST/tests/SyncListenerTest.cc
@@ -128,4 +128,42 @@ TEST_CASE_METHOD(C4SyncListenerTest, "P2P Sync connection count", "[Listener][C]
 }
 
 
+TEST_CASE_METHOD(C4SyncListenerTest, "P2P ReadOnly Sync", "[Push][Listener][C]") {
+    C4ReplicatorMode pushMode = kC4Disabled;
+    C4ReplicatorMode pullMode = kC4Disabled;
+    SECTION("Push") {
+        config.allowPull = false;
+        SECTION("Continuous") {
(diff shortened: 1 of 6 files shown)


REST/tests/SyncListenerTest.cc
```diff
@@ -128,4 +128,42 @@ TEST_CASE_METHOD(C4SyncListenerTest, "P2P Sync connection count", "[Listener][C]
...
+    if (pinnedCert)
+        _address.scheme = kC4Replicator2TLSScheme;
+    
+    _callbackStatus.error = {WebSocketDomain, 10403};
+    replicate(pushMode, pullMode, false);
+    CHECK(c4db_getDocumentCount(db2) == 0);
+}
+
+
+
```
jayahariv, line 168:
```suggestion
```

PR #1008, opened 2020-06-12
CBL-1008: Fix IPv6 addresses being reported, but not connectable

borrrden:
During this process I also uncovered that specifying an IPv6 address to bind to was essentially a no-op because only IPv4 "any" addresses were being considered.  The P2P Server Addresses test was added to make sure this does not regress.  The main fix here is to use inet6_address in our Server class whenever possible since it also handles IPv4 socket operations.  This was added in the following locations

- When no network interface is specified, an inet6_address instance should be created instead of inet_address
- When looking at the acceptor address to determine if the address is not constrained, the family of the socket needs to be considered

Note: The inspection of the "any" IPv6 address is a bit hacky, but the alternative is about 5 different ifdef due to the lack of consistency in the definition of in6_addr between platforms.

### DIFF
--- a/C/include/c4Listener.h
+++ b/C/include/c4Listener.h
@@ -124,6 +124,12 @@ extern "C" {
     /** Returns the URL(s) of a database being shared, or of the root, separated by "\n" bytes.
         The URLs will differ only in their hostname -- there will be one for each IP address or known
         hostname of the computer, or of the network interface.
+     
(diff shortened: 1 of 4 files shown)


borrrden:
This one took a bit more finesse because things basically were only working reliably on macOS with the new test I introduced.

PR #1013, opened 2020-06-18
Export C4Certificate API from dylib, etc.

snej:


snej:
Hm, something is wrong at runtime on Windows in the new C4Cert test...

PR #1015, opened 2020-06-23
CBL-1054: Don't do offline handling after c4repl_stop

borrrden:
This manifests in a very hard to hit situation where a replicator cannot connect to the other side because of a transient error, but since it has already switched to offline it has no connection to close and so the call to stop has no effect other than to change the status to stopping.  This needs to be detected in the status changed callback and the offline handling should be skipped.

### DIFF
--- a/Replicator/tests/ReplicatorAPITest.cc
+++ b/Replicator/tests/ReplicatorAPITest.cc
@@ -567,4 +567,32 @@ TEST_CASE_METHOD(ReplicatorAPITest, "Stop while connect timeout", "[C][Push][Pul
     waitForStatus(kC4Stopped, 8s);
     _socketFactory = nullptr;
 }
+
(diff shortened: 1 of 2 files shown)


Replicator/tests/ReplicatorAPITest.cc
```diff
@@ -567,4 +567,32 @@ TEST_CASE_METHOD(ReplicatorAPITest, "Stop while connect timeout", "[C][Push][Pul
     waitForStatus(kC4Stopped, 8s);
     _socketFactory = nullptr;
 }
+
+TEST_CASE_METHOD(ReplicatorAPITest, "Stop after transient connect failure", "[C][Push][Pull]") {
+    c4log_setLevel(c4log_getDomain("Sync", false), kC4LogDebug);
```
jayahariv, line 572: intentional to enable logging?

borrrden: No!  Thanks, that should be removed.

PR #1017, opened 2020-06-26
CBL-1044 Add self-signed validation mode

borrrden:
This needs to live in harmony with the other modes, and furthermore it needs to disrupt normal processing by ignoring the normal CA validation.  Only self signed certificates will be accepted.  The motivation for this is due to our commitment to make an insecure mode as difficult to achieve as possible.  If this were simply a mode to allow self signed certificates in addition to valid ones, then this could easily be overlooked and put into production thereby making a product using the library easily attackable.  On the other side, many P2P implementations will have dynamically generated self signed certificates floating around and those should work by default if that mode is desired.  So the basic thinking is that in P2P the self signed mode will be used, and with non-P2P the normal mode will be used.  It is possible to use the self signed mode with non-P2P scenarios but this would be a poor idea in production.  It could be used during development, but there would be an actual error when switching to production if this mode was left on, thereby preventing the accidental insecurity.

### DIFF
--- a/C/include/c4Replicator.h
+++ b/C/include/c4Replicator.h
@@ -336,7 +336,7 @@ extern "C" {
     // TLS options:
     #define kC4ReplicatorOptionRootCerts        "rootCerts"  ///< Trusted root certs (data)
     #define kC4ReplicatorOptionPinnedServerCert "pinnedCert"  ///< Cert or public key (data)
-    #define kC4ReplicatorOptionOnlySelfSignedServer "onlySelfSignedServer" ///< Only accept self signed server certs (for P2P, bool)
+    #define kC4ReplicatorOptionOnlySelfSignedServerCert "onlySelfSignedServer" ///< Only accept self signed server certs (for P2P, bool)
 
     // HTTP options:
     #define kC4ReplicatorOptionExtraHeaders     "headers"  ///< Extra HTTP headers (string[])
--- a/C/include/c4Replicator.h
+++ b/C/include/c4Replicator.h
@@ -51,12 +51,6 @@ extern "C" {
         kC4Stopping,    ///< Stopping or going offline
     };
 
-    /** Possible values for the kC4ReplicatorOptionServerCertVerifyMode option*/
-    typedef C4_ENUM(int32_t, C4ReplicatorServerCertVerificationMode) {
-        kC4ServerCertCACert = 0, ///< Verify by using trusted anchor CA certs or by using the configured pinned server certs (Default Mode).
-        kC4ServerCertSelfSigned  ///< Verify by accepting any and only self-signed certs. Any non-self-signed certs will be rejected.
-    };
-
     /** For convenience, an array of C strings naming the C4ReplicatorActivityLevel values. */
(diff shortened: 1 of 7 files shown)


C/include/c4Replicator.h
```diff
@@ -342,7 +336,7 @@ extern "C" {
     // TLS options:
     #define kC4ReplicatorOptionRootCerts        "rootCerts"  ///< Trusted root certs (data)
     #define kC4ReplicatorOptionPinnedServerCert "pinnedCert"  ///< Cert or public key (data)
-    #define kC4ReplicatorOptionServerCertVerifyMode "serverVerifyMode" ///< How to verify the server TLS cert (C4ReplicatorServerCertVerificationMode)
+    #define kC4ReplicatorOptionOnlySelfSignedServer "onlySelfSignedServer" ///< Only accept self signed server certs (for P2P, bool)
```
pasin, line 339: Not onlySelfSignedServerCerts to inline with kC4ReplicatorOptionRootCerts or kC4ReplicatorOptionPinnedServerCert?

Networking/TCPSocket.cc
```diff
@@ -494,9 +494,15 @@ namespace litecore { namespace net {
             if (flags != 0 && flags != UINT32_MAX) {
                 string message = tlsSocket->peer_certificate_status_message();
                 int code;
-                if (flags & MBEDTLS_X509_BADCERT_NOT_TRUSTED)
-                    code = kNetErrTLSCertUnknownRoot;
-                else if (flags & MBEDTLS_X509_BADCERT_REVOKED)
+                if (flags & MBEDTLS_X509_BADCERT_NOT_TRUSTED) {
+                    if(_tlsContext && _tlsContext->onlySelfSignedAllowed()) {
+                        code = kNetErrTLSCertUntrusted;
+                        message = "Self-signed only mode is active for P2P, and a non self-signed certificate was received";
```
pasin, line 500: Is this message only for P2P?

C/include/c4Replicator.h
borrrden: I was trying to avoid making the name comically long....

Networking/TCPSocket.cc
borrrden: Hmmm technically no, but that's what it is designed for.  I'm open on the wording, but I think that this mode is going to be EE only right?

pasin: @rajagp would like to have this mode available for both CE and EE.

borrrden: Unsure of what purpose it would serve in CE

pasin: For making the development with a self-signed Sg easier. @rajagp can provide the rational here.

I think making the message generic by removing 'for P2P' doesn't change the meaning of the message here.

borrrden: You're probably right, but I really wanted to drive home the point to the user that "this is for P2P" and if they are doing it somewhere else they are most likely "doing it wrong".  I won't take a stand on that position though.

PR #1022, opened 2020-07-01
Create the long and arduous path to extracting a peer certificate

borrrden:
This is a simple action to desire, but it needs to break through about 7 or so layers of indirection and decoupling in order to get there, so the PR is pretty massive.  The certificate is discarded very quickly in a failed handshake scenario and so the following needs to happen:

1. mbedtls_context needs to remember the certificate it received
2. mbedtls_socket needs to consult mbedtls_context if it can't get the peer certificate via the normal route of asking mbedTLS
3. BuiltInWebSocket needs to cache the certificate after it sends an HTTP request so that the certificate is not lost because the socket is not kept around if it fails to open
4. BLIPIO needs to to cache the certificate when the connection closes because otherwise it is lost when the mbedtls_socket is nulled
5. BLIPConnection needs to cache the result from BLIPIO for after it nulls it during the connection close
6. The Replicator needs to cache the result from BLIPConnection for after it nulls it during connection close
7. The c4replicator needs to wrap all of this

The TLS tests have been modified to make sure that the correct certificate is retrieved as a condition of passing

### DIFF
--- a/Replicator/Replicator.cc
+++ b/Replicator/Replicator.cc
@@ -420,6 +420,11 @@ namespace litecore { namespace repl {
         _connectionState = state;
 
         _checkpointer.stopAutosave();
+        
+        if(connected()) {
+            // The ability to get this goes away after the connection is closed, so cache it!
+            _peerTLSCertificate = connection().peerTLSCertificate();
+        }
 
         // Clear connection() and notify the other agents to do the same:
         _connectionClosed();
--- a/C/c4_ee.def
+++ b/C/c4_ee.def
@@ -316,6 +316,7 @@ c4keypair_removePersistent
 c4keypair_persistentWithPublicKey
 
 c4repl_newLocal
+c4repl_getPeerTLSCertificate
 
 ; C4Tests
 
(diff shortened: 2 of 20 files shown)


snej:
Re. the sockpp change, your comment
>Is there a better way to do this then parsing everything again?

All you need to do is copy the raw cert data, i.e. to a string. Then you can hand it back directly in the accessor.

snej:
The only other way I can think to do it is via notifications — adding something like `onPeerCertReceived(slice)` to the various delegate APIs. That way the intermediate layers wouldn't need to cache the cert, they'd just pass the notification up the chain. But the total amount of code changed probably wouldn't be any less.

Replicator/Replicator.cc
```diff
@@ -420,6 +420,11 @@ namespace litecore { namespace repl {
         _connectionState = state;
 
         _checkpointer.stopAutosave();
+        
+        if(connected()) {
+            // The ability to get this goes away after the connection is closed, so cache it!
+            _peerTLSCertificate = connection().peerTLSCertificate();
```
pasin, line 426: Where does _peerTLSCertificate get set when it's not connected such as having a TLS related error?

borrrden: It is still connected at this level, even after failure.  This variable just means that the BLIP connection is still non-null.

borrrden:
@bmeike All of them are `Retained` (i.e. ref-counted) instances, so they are the same object and it is freed when the last object is destroyed.

PR #1023, opened 2020-07-02
Replicator refactoring

snej:
* Split up Pusher.cc, creating Pusher+Revs.cc and Pusher+Attachments.cc
* Extracted changes-tracking code from Pusher into a new class ChangesFeed
* ChangesFeed starts its c4DatabaseObserver earlier — before it first queries the database for historical changes — to avoid a race condition where some changes are missed in the time between the query running and the observer being created.


snej:
Linux test failure is in the "Rapid Restarts" test in ReplicatorAPITests.cc:
```
		--- Suspend / Unsuspend
pure virtual method called
```
I believe this means an overridden method was called while the object was in the base class constructor or destructor. Maybe a race condition?

snej:
Help me @borrrden, you're my only hope! Windows tests still crash mysteriously. Can you investigate when you have a chance?

snej:
This time Windows passed, but Linux got:
```
## SyncListenerTest.cc:
	>>> P2P Sync
	    [0.432368 sec]
	>>> TLS P2P Sync self-signed cert
		--- Pinned, self-signed mode
	--- TLS P2P Sync self-signed cert
		--- Pinned, normal mode
14:35:35.478228| ERROR: Assertion failed: underflow decrementing a counter (/var/lib/jenkins/mobile/workspace/line_couchbase-lite-core_PR-1023/couchbase-lite-core/LiteCore/Support/Increment.hh:34, in decrement) (no backtrace available)
```
No backtrace, unfortunately.

snej:
Finally, Jenkins passes!

PR #1026, opened 2020-07-09
CBL-1061: Always set delegate to null in terminate()

borrrden:
This issue is not reproducible, but static analysis leads me to believe that what is happening is that C4LocalReplicator is getting destroyed via c4repl_free() at a point when the replicator's connection is already gone.  This means that since connected() will be false, the delegate will remain non-null.  Furthermore, static analysis shows that this is always the object which is getting freed so there is no point in keeping it around after a call to terminate.

### DIFF
--- a/Replicator/Replicator.cc
+++ b/Replicator/Replicator.cc
@@ -180,15 +180,21 @@ namespace litecore { namespace repl {
 
 
     void Replicator::terminate() {
+        logDebug("terminate() called...");
         if (connected()) {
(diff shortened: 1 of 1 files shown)


borrrden:
Fair point, I'm ok with this being the new behavior though.  The method name is `terminate` so it's a pretty destructive operation.  Maybe it warrants a mention in the doc comments.

borrrden:
Tried to merge this on the command line.....it didn't end up closing this PR

PR #1027, opened 2020-07-11
Fixed problems when getting public key data in ExternalKeyPair

pasin:
* The estimated DER data size is about 38 bytes longer than key lenth instead of 14 bytes. This has been verified from the public key in DER format created by Java’s KeyGenerator and OpenSSL.

* Fixed publicKeyRawData() that returns DER instead of RAW data.

### DIFF
--- a/C/c4Certificate.cc
+++ b/C/c4Certificate.cc
@@ -540,7 +540,7 @@ namespace c4Internal {
         }
 
         virtual fleece::alloc_slice publicKeyDERData() override {
-            alloc_slice data(_keyLength + 20);  // DER data is ~14 bytes longer than keyLength
+            alloc_slice data(_keyLength + 40);  // DER data is ~38 bytes longer than keyLength
             size_t len = data.size;
             if (!_callbacks.publicKeyData(_externalKey, (void*)data.buf, data.size, &len)) {
                 WarnError("C4ExternalKey publicKeyData callback failed!");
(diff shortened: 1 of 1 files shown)


C/c4Certificate.cc
```diff
@@ -552,7 +552,8 @@ namespace c4Internal {
         }
 
         virtual fleece::alloc_slice publicKeyRawData() override {
-            return publicKeyDERData();
+            Retained<PublicKey> publicKey = new PublicKey(publicKeyDERData());
```
borrrden, line 555: Separate these into two lines please.  Windows often has issues with inline `alloc_slice` passed to a `slice` argument because it will get deallocated before it is used.

pasin: Done

pasin:
Honestly I don't know but Java and OpenSSL always do that.

bmeike:
I think we need to know.  ... or to guarantee that the code will fail before overwriting random bits of memory.  This kind of thing is just a disaster waiting to happen.

pasin:
The code will fail with an assertion if the data turns to be bigger than MAX. It's hard to know as the size seems to depend on the security provider but at least Java and openSSL seem to be consistent. We could increase the max to the keysize + 100 which I think it could be enough.

pasin:
I created https://issues.couchbase.com/browse/CBL-1134 for figuring that out.

bmeike:
Cool.  Perhaps we should consider that a blocker?

Is there anything that an adversary could do to control the length of the converted key?

pasin:
Each provider could use different exponent value which could be the key difference.

bmeike:
So long as the code fails before overwriting stuff, it seems safe for production.  I wish there were a way to precompute the key size, tho...

pasin:
I made the issue blocker and I'm merging this now. Anything else could be discuss in a separate ticket.

PR #1031, opened 2020-07-15
CBL-1141: Defer SQLite temporary file unlinking for Android

borrrden:
Otherwise, it doesn't seem able to handle file operations on already unlinked files (such as index sorts that exceed the memory limit)


bmeike:
This probably isn't a big security risk, unless the user meddles with the temp directory location.  Temp directories (/sdcard/Android/<package> are usually accessible only to the application with which they are associated.  They are deleted if the app is deleted.

PR #1039, opened 2020-07-21
CBL-1112: Port backpressure hotfix to master

borrrden:
If too many revision build up in the replicator, there is a path where they remain in the backpressure list without ever getting triggered to move to the active incoming list.  This means that the replicator will never think it is finished because it is missing 100 or so messages.

### DIFF
--- a/Replicator/Puller.cc
+++ b/Replicator/Puller.cc
@@ -139,16 +139,17 @@ namespace litecore { namespace repl {
             handleChangesNow(req);
         }
 
-#ifdef LITECORE_SIGNPOSTS
         bool backPressure = !_waitingRevMessages.empty();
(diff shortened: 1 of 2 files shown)


snej:
Where is the original fix you ported? I wish you'd given either the commit ID or issue number.

snej:
I think I've found it: edb1b1bedbabae1ad03d4fe35df828b7c28d9c1f and f284f01f16cf038d1842bd38ffbdeb01f183d63f, right? (I ask because this new commit is causing terrible merge conflicts with my replicator refactoring branch, and I need to look at the original change to figure out what was being done...)

borrrden:
Those are correct, I mentioned the branch name I took from on the CBL ticket but forgot to put it here.

PR #1043, opened 2020-07-30
 warning log in case of forbidden connection

jayahariv:


jayahariv:
Yea. right, initial intention was to get some warnings in console(either replicator or server, but you are right, we don't need to log on server side for every replicator, instead only on replicator side), if there is something wrong. 

I was not seeing error from replicator side too, since there was no documents pushed from it. When I pushed a document, it started showing error from replicator side. Closing this PR.

PR #1046, opened 2020-08-07
CBL-936: Make sure Puller's counters react to a failed delta

borrrden:
When a delta failure happens, the error response triggers the revision to be sent again as a full revision, but without going through the changes message first.  The changes message both adds progress and also adds pending rev message count that is balanced when the revision is received, and so needs to be re-added as part of the revision re-request.


borrrden:
Whoops borked the name of the commit and PR

PR #1047, opened 2020-08-07
CBL-936: Make sure Puller's counters react to a failed delta

borrrden:
When a delta failure happens, the error response triggers the revision to be sent again as a full revision, but without going through the changes message first.  The changes message both adds progress and also adds pending rev message count that is balanced when the revision is received, and so needs to be re-added as part of the revision re-request.

### DIFF
--- a/Replicator/Puller.cc
+++ b/Replicator/Puller.cc
@@ -340,6 +340,15 @@ namespace litecore { namespace repl {
     }
 
 
+    void Puller::_revReRequested(IncomingRev * inc) {
+        // Regression from CBL-936 / CBG-881:  Because after a delta failure the full revision is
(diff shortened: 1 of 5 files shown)


Replicator/Puller.cc
```diff
@@ -340,6 +340,15 @@ namespace litecore { namespace repl {
     }
 
 
+    void Puller::_revReRequested(IncomingRev * inc) {
```
snej, line 343: Parameter should be `Retained<IncomingRev>`, to ensure it isn't freed before the async call occurs.

snej:
If you fix that one parameter type, everything looks 👍 otherwise, so go ahead and merge.

PR #1051, opened 2020-08-13
CBL-1178: Suppress the recording of kWebSocketCloseAppPermanent

borrrden:
And kWebSocketCloseAppTransient.  These are basically used to close the connection in the face of an underlying error, and that underlying error should be the one that persists in the consumer visible status.

### DIFF
--- a/Replicator/Replicator.cc
+++ b/Replicator/Replicator.cc
@@ -310,6 +310,13 @@ namespace litecore { namespace repl {
 
 
     void Replicator::onError(C4Error error) {
+        if(status().error.code != 0 && (error.code == kWebSocketCloseAppPermanent || error.code == kWebSocketCloseAppTransient)) {
+            // CBL-1178: If we already have an error code, it is more relevant than the web socket close code, so keep it
+            // intact so that the consumer can know what went wrong
+            logVerbose("kWebSocketCloseAppPermanent or kWebSocketCloseAppTransient received, ignoring (only relevant for underlying connection...)");
+            return;
+        }
+        
         Worker::onError(error);
         for(const C4StoppingErrorEntry& stoppingErr : StoppingErrors) {
             if(stoppingErr.err.domain == error.domain && stoppingErr.err.code == error.code) {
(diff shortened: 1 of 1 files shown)


borrrden:
No, you will still receive the 4002 there, this situation is actually preventing a side effect of the exact thing you want here:

1. BLIP sends the 403 message 
2. LiteCore recognizes this as a fatal message which should stop the replication, but it cannot use this code to close the connection
3. So therefore, it uses the arbitrary (but legal) 4002 code to close the web socket
4. However, since this is not a "clean" closure, it gets redetected by LiteCore and set as the replication error code.  This is what this PR is going to stop (keep the previous 403 intact for the consumer to deal with)

Replicator/Replicator.cc
```diff
@@ -310,6 +310,13 @@ namespace litecore { namespace repl {
 
 
     void Replicator::onError(C4Error error) {
+        if(status().error.code != 0 && (error.code == kWebSocketCloseAppPermanent || error.code == kWebSocketCloseAppTransient)) {
```
snej, line 313: Need to check `error.domain == WebSocketDomain` too.

Replicator/Replicator.cc
```diff
@@ -310,6 +310,14 @@ namespace litecore { namespace repl {
...
 
     void Replicator::onError(C4Error error) {
+        if(status().error.code != 0 && error.domain == WebSocketDomain &&
+           (error.code == kWebSocketCloseAppPermanent || error.code == kWebSocketCloseAppTransient)) {
+            // CBL-1178: If we already have an error code, it is more relevant than the web socket close code, so keep it
+            // intact so that the consumer can know what went wrong
+            logVerbose("kWebSocketCloseAppPermanent or kWebSocketCloseAppTransient received, ignoring (only relevant for underlying connection...)");
+            return;
+        }
+
```
jayahariv, line 320: i thought Listener will throw a 403, which will still send same `kCloseAppPermanent` error right? Since this handling is done for only `kWebSocketCloseAppPermanent` and `kWebSocketCloseAppTransient`? 

https://github.com/couchbase/couchbase-lite-core/blob/b0d1ffe8808bb71246d8736f1b5c7d5ecf7e82d0/Replicator/Replicator.cc#L726

jayahariv: i got it, thanks

jayahariv:
is it possible to validate the error code in the unit test too?

PR #1058, opened 2020-09-01
CBL-1243: Freeing listener should stop running replicators

borrrden:
Note, this requires removing web socket 1001 from the list of transient error codes because that is the code that is given when the other side closes the connection


pasin:
@borrrden Previously you have mentioned that the passive replicator will need to be stopped as well. Have that been taken care of as well?

borrrden:
Ah, sorry, note that this is depending on the PR-1058 branch in EE.

PR #1059, opened 2020-09-03
Xcode: Build arm64 for macOS ("Apple Silicon")

snej:
[CBL-1253](https://issues.couchbase.com/projects/CBL/issues/CBL-1253): Our Mac build of CouchbaseLite.framework needs to be a universal binary containing ARM64 for Apple's upcoming "Apple Silicon" Macs. Developers will need this from us before they can make their own apps universal.

This PR makes LiteCore a universal Intel/ARM binary, when built with Xcode 12.


snej:
OK, passed Jenkins on Apple platforms. Travis Linux build timed out for some unrelated reason.

This doesn't need to go into Hydrogen, but I don't think it'll affect the build at all on Xcode 11.

Xcode/xcconfigs/Project.xcconfig
```diff
@@ -31,10 +31,10 @@ ONLY_ACTIVE_ARCH                                   = NO
 SKIP_INSTALL                                       = YES
 SUPPORTED_PLATFORMS                                = macosx iphoneos iphonesimulator appletvos appletvsimulator
 
-VALID_ARCHS                                        = x86_64 arm64 armv7 armv7s i386
-VALID_ARCHS[sdk=macosx*]                           = x86_64
-VALID_ARCHS[sdk=iphonesimulator*]                  = x86_64                    i386
-VALID_ARCHS[sdk=appletvsimulator*]                 = x86_64                    i386
+VALID_ARCHS                                        = x86_64 arm64 armv7 armv7s
+VALID_ARCHS[sdk=macosx*]                           = x86_64 arm64
+VALID_ARCHS[sdk=iphonesimulator*]                  = x86_64
```
jayahariv, line 36: officially we still support 32 bit iPhone simulator architecture right?

borrrden (commented):
I'm going to put this in after Hydrogen branches

Xcode/xcconfigs/Project.xcconfig
snej: I was assuming that a 32-bit simulator build was needed only for 32-bit Macs, but I think that's wrong; apparently it's needed to run a 32-bit iOS app in the simulator.

macOS Catalina can't run 32-bit code anymore, so you'd have to be on 10.14 or earlier, and you'd have to install iOS 10 into the Simulator because that's the last that can run 32-bit, but it's doable. So I'll put it back for now.

But I'm really hoping that for 2.0/Lithium we can dump 32-bit and go to iOS 11+.

snej:
Jenkins builds passed except Windows, which doesn't matter because this commit obviously doesn't affect Windows. So I'm merging.

PR #1060, opened 2020-09-10
Don't include host name in listener URL list

borrrden:
It doesn't seem to function at all on iOS, and macOS it's questionable whether or not it will work if a test is run with separate machines

### DIFF
--- a/Networking/NetworkInterfaces.cc
+++ b/Networking/NetworkInterfaces.cc
@@ -239,6 +239,7 @@ namespace litecore::net {
 #pragma mark - PLATFORM SPECIFIC CODE:
 
 
+#if 0 // Doesn't appear to function outside the same machine by default
     optional<string> GetMyHostName() {
 #ifdef __APPLE__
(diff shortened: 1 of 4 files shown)


pasin (approved):
I'm fine to take this out or at least move the hostname url to end of the list as it may not work on some network, and it's not cross-platform.

snej:
> It doesn't seem to function at all on iOS

Then we should try to fix that.

> macOS it's questionable whether or not it will work if a test is run with separate machines

Why wouldn't it? As long as the other machine supports mDNS. 

Even if we do have to disable it, it'd be simpler to just make the function always return null, than removing it entirely and having to ifdef out references to it in three other source files.

borrrden:
We can certainly try to fix it, but the timeframe is tight.  This is the last day before RC testing begins.  The reason I say it is questionable is because the whole concept of mDNS is basically a black box, especially the area of "when does it become active?"  Does it need to be turned on?  If so how?  These unknowns make me uncomfortable.

jayahariv (approved):

PR #1064, opened 2020-09-11
CBL-977: Skip "hidden" columns when calculating missing columns

borrrden:
FTS queries have some metadata columns in front, but it mistakenly writes in the missing columns as the raw column index, instead of the column index starting at the first user requested value

### DIFF
--- a/LiteCore/Query/SQLiteQuery.cc
+++ b/LiteCore/Query/SQLiteQuery.cc
@@ -474,12 +474,14 @@ namespace litecore {
 
             unicodesn_tokenizerRunningQuery(true);
             try {
-                while (_statement->executeStep()) {
-                    uint64_t missingCols = 0;
-                    enc.beginArray(nCols);
-                    for (int i = 0; i < nCols; ++i) {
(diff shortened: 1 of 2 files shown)


LiteCore/Query/SQLiteQuery.cc
```diff
@@ -474,12 +474,14 @@ namespace litecore {
...
-                    enc.beginArray(nCols);
-                    for (int i = 0; i < nCols; ++i) {
-                        if (!encodeColumn(enc, i) && i < 64)
-                            missingCols |= (1ULL << i);
+                 auto firstCustomCol = _query->_1stCustomResultColumn;
+                 while (_statement->executeStep()) {
+                     uint64_t missingCols = 0;
+                     enc.beginArray(nCols);
+                     for (int i = 0; i < nCols; ++i) {
+                         if (!encodeColumn(enc, i) && i >= firstCustomCol && i < 64) {
```
snej, line 482: `i < 64` should be `i - firstCustomCol < 64`

snej (changes requested):

pasin (approved):

PR #1067, opened 2020-09-20
CBL-1291: Expanded failed delta recovery handling

borrrden:
The pending rev message count needs to be handled in the case of attachments, which follows a different path than non-attachments

### DIFF
--- a/Replicator/IncomingRev.cc
+++ b/Replicator/IncomingRev.cc
@@ -260,19 +260,22 @@ namespace litecore { namespace repl {
 
 
     void IncomingRev::_revisionInserted() {
+        Retained<IncomingRev> retainSelf = this;
         decrement(_pendingCallbacks);
-        if(_rev->error.domain == LiteCoreDomain &&
-           (_rev->error.code == kC4ErrorDeltaBaseUnknown ||
-            _rev->error.code == kC4ErrorCorruptDelta)) {
-            // CBL-936: Make sure that the puller knows this revision is coming again
-            _puller->revReRequested(this);
-        }
-        
         finish();
     }
 
 
     void IncomingRev::finish() {
+        if(_rev->error.domain == LiteCoreDomain &&
+           (_rev->error.code == kC4ErrorDeltaBaseUnknown ||
+            _rev->error.code == kC4ErrorCorruptDelta)) {
+            // CBL-936: Make sure that the puller knows this revision is coming again
+            // NOTE: Important that this be done before _revMessage->respond to avoid
+            // racing with the newly requested rev
+            _puller->revReRequested(this);
+        }
+        
         if (_revMessage) {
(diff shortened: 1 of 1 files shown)


bmeike (approved):
It looks to me as if this change will prevent and IncomingRev from being de-allocated, while `decrement` and `finish` are called.  If that's the problem, this looks like a fix.

I have two concerns:
1) when was the Incoming Rev freed?  Isn't there still a race from that point, until this method is called?
2) this is not the simplest change that accomplishes the intended goal.  I believe that the only necessary change in the insertion of line 263

borrrden:
You are correct that there are two components to this and one is not strictly necessary to correct the failed assertion behavior, but you have it backwards.  The retaining thing was added on later as it seemed to help the secondary problem that came out of the fix.  It was a fix in master that Jens indicated could be put back into Hydrogen (CBL-1281 if I'm not mistaken) and it seemed to fix some flaky behavior I was seeing after the initial problem with the test was gone (it didn't get the assertion failure even once after the primary fix, but it would hang with the active vs completed being off by one 2 out of the 4 times I ran it.  After putting the retain in it didn't happen for 4 attempts.  That might be anecdotal and so I could accept an argument that it doesn't belong in this PR.)  

The primary part of this fix (aka the "simplest") is moving that block of logic.  That block needs to be executed whenever a delta fails.  I fixed this before in CBL-936 but it turns out that not all deltas pass through the `_revisionInserted` method.  They do, however, all pass through the `finish()` method, which is the method that actually sends a response back over the wire.  So I moved it there to ensure that no matter where the delta is resolved, the puller is reminded to expected another rev message containing the full body instead of the delta.

bmeike:
Got it.  'finish()` is called elsewhere.

Replicator/IncomingRev.cc
```diff
@@ -260,19 +260,22 @@ namespace litecore { namespace repl {
 
 
     void IncomingRev::_revisionInserted() {
+        Retained<IncomingRev> retainSelf = this;
```
snej, line 263: Would it make more sense to move this self-retain into `finish()` itself?

borrrden: Maybe.  I'm not sure what scope it is trying to keep alive.

borrrden:
Well I didn't expect it to be self explanatory, but my explanation was very terse:  "follows a different path than non-attachments".  I'll give more background next time.

Replicator/IncomingRev.cc
snej: I don’t remember the details anymore. It may be that the code changed some after I added the self-retain, and that revisionInserted used to call something else that was triggering deletion. With the code as it is, it must be the finish method that could delete, so the self-retain could go there too, but no problem leaving it where it is.

snej (approved):

PR #1068, opened 2020-09-21
One-shot (unbuffered) queries

snej:
This implements a new query mode where the result set is not first copied into memory; instead the enumerator directly steps the SQLite cursor (statement) and returns the current row. This generally uses less memory and has less latency. On the other hand, it doesn't support seeking backwards or returning the number of rows, and it doesn't work with live queries because those need the buffered results for comparison purposes.

(The memory savings will vary depending on the query. In some cases SQLite itself needs to buffer data, most commonly when there's an ORDER BY that doesn't match the ordering of the primary index being used. I don't know whether SQLite needs to buffer all the result data (or just rowids, for example), or whether its buffered form is more or less compact than ours. It's possible we could introspect the query structure, or scrape the EXPLAIN results, to check whether this buffering will occur.)

### API Changes

There is an API change: adding a `oneShot` property to `C4QueryOptions`. (I also marked the rank-results property as obsolete, because it's not hooked up to anything in the implementation.)

I don't think we can avoid making this a visible option. There are pros and cons to one-shot vs buffered queries, and we don't know ahead of time what the developer will prefer. 

### Theory of operation

Basically there are now two different QueryEnumerator subclasses: one plays back rows from a pre-recorded result set, as before. The other keeps the SQLite statement handle and steps one row at a time.

The problem with this is that, if any rows are changed in the table (by the same SQLite connection), the statement gets messed up and will start returning incorrect data. (It's a case of the generic CS issue wherein modifying a collection breaks any current iterators.)

So the workaround is that the one-shot enumerator listens for any changes about to happen in that table (using a new listener interface I've added.) When it's notified, it creates a new pre-recorded enumerator, which records the remainder of the rows into memory. It then drops its reference to the SQLite statement and delegates future API calls to the pre-recorded enumerator.

### TBD

The one edge case not yet handled is if you create two enumerators on a single query instance. The one-shot enumerator uses the same SQLite statement as the query object, so creating a new enumerator will reset the statement and mess up the first one.

I think the fix to this is to create a new statement for the enumerator, if the first one is still in use. SQLite doesn't seem to have a way to duplicate a `sqlite3_stmt`, so I guess this requires creating a new one from the original SQL text.

(The other option is to do the same thing as when the table is about to change, i.e. tell the first enumerator to record the rest of the rows and stop using the statement.)


borrrden:
Superceded by https://github.com/couchbase/couchbase-lite-core/pull/1083

PR #1069, opened 2020-09-28
Fix fl_each() on non document data

pasin:
* When data is not document’s body, database’s fleece-accessor will return nullslice. If that is the case, fall back to use the original data.
* Added unit test on both c and c++. The c++ test passes without the fix as it’s use the test’s fleece accessor that always returns the orignal data (C++ test doesn’t save doc with revision created).
* Reference CBL-1248

### DIFF
--- a/C/tests/c4QueryTest.cc
+++ b/C/tests/c4QueryTest.cc
@@ -252,6 +252,26 @@ N_WAY_TEST_CASE_METHOD(C4QueryTest, "C4Query ANY of dict", "[Query][C]") {
 }
 
 
+N_WAY_TEST_CASE_METHOD(C4QueryTest, "C4Query Nested ANY of dict", "[Query][C]") {
+    C4Error error;
+    TransactionHelper t(db);
+    
+    // New Doc:
+    auto doc1 = c4doc_create(db, C4STR("doc1"),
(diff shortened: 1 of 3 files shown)


pasin:
Two things:

1. I created this PR against `fix/misc-fixes` as master branch requires `fix/misc-fixes` to be able to build by using XCode12. However, it turns out that creating a PR against `fix/misc-fixes` doesn't work either as there is no `fix/misc-fixes` on EE to be clone for validation.

2. I don't know if the fix is the right fix. I couldn't find an explicit way to check if the data is doc's body or not to decide whether to use fleece-accessor. Please advise me know if there is a better fix so I will fix the PR.

pasin:
Closed to create a new PR on master branch instead.

snej:
Yeah, this fix is not safe, because it's expecting the db's `fleeceAccessor` function to reject anything that isn't a real document body; but that function wasn't designed for that. All it takes is one bit to be set in the first couple of bytes, and it'll happily return a sub-range of the blob, and then everything will go into the weeds, including possibly crashing.

The query runtime code is not being careful enough to distinguish between blobs that are raw doc bodies (`_doc.body` in SQL) with blobs that just contain Fleece data. I'm trying to figure out a good fix for this...

PR #1070, opened 2020-09-28
Fix fl_each() on non document data

pasin:
* When data passed to the fl-each() function is already the fleece data not the document's body which is required to be extracted using fleece-accessor, the fleece-accessor will return nullslice. If that is the case, fall back to use the data passed to the function.
* Added unit test on both c and c++. The c++ test passes without the fix as it’s use the test’s fleece accessor that always returns the orignal data (C++ test doesn’t save doc with revision created).
* Reference CBL-1248

### DIFF
--- a/LiteCore/Query/SQLiteFleeceEach.cc
+++ b/LiteCore/Query/SQLiteFleeceEach.cc
@@ -214,14 +214,19 @@ private:
             return SQLITE_OK;
 
         // Parse the Fleece data:
-        slice data = valueAsSlice(argv[0]);
-        if (!data) {
+        slice body = valueAsSlice(argv[0]);
+        if (!body) {
             // Weird not to get a document; have to return early to avoid a crash.
             // Treat this as an empty doc. (See issue #379)
             Warn("fleece_each filter called with null document! Query is likely to fail. (#379)");
             return SQLITE_OK;
         }
-        data = _vtab->context.delegate->fleeceAccessor(data);
+        slice data = _vtab->context.delegate->fleeceAccessor(body);
+        
+        // CBL-1248: if the original data is not document body, getting data
+        // from db's fleece accessor will return nullslice; fallback to use the original data.
+        if (!data.buf)
+            data = body;
 
         if (size_t(data.buf) & 1) {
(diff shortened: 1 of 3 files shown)


pasin:
Two things:

1. I don't know if the fix is the right fix. I couldn't find an explicit way to check if the data is doc's body or not to decide whether to use fleece-accessor. Please advise me know if there is a better fix so I will fix the PR.

2. The PR validation will fail as it requires fix/misc-fixes to be able to build by using XCode12.

LiteCore/Query/SQLiteFleeceEach.cc
```diff
@@ -214,14 +214,19 @@ class FleeceCursor : public sqlite3_vtab_cursor {
...
             // Treat this as an empty doc. (See issue #379)
             Warn("fleece_each filter called with null document! Query is likely to fail. (#379)");
             return SQLITE_OK;
         }
-        data = _vtab->context.delegate->fleeceAccessor(data);
+        slice data = _vtab->context.delegate->fleeceAccessor(body);
+        
+        // CBL-1248: if the original data is not document body, getting data
+        // from db's fleece accessor will return nullslice; fallback to use the original data.
+        if (!data.buf)
```
borrrden, line 228: There is a better test for this.  If `idxNum` is `kPathIndex` then you should use the `fleeceAccessor`, otherwise do not.  So I think it would be better to simply guard the old line 224 with this condition.  If it is a nested value, then the index will be `kFleeceDataIndex`

borrrden (changes requested):

LiteCore/Query/SQLiteFleeceEach.cc
pasin: I'm trying to understand the difference b/w kPathIndex and kFleeceDataIndex. It seems like it's not necessary that when the index is PathIndex, the data always requires database's fleece accessor for the case like `fl_each(fl_nested_value(_V.body, 'sub_dict'), 'array_in_sub_dict')`.

pasin:
I'm closing this PR as @snej suggested that the current fix by the PR is not safe, and better to be fixed as part of a bigger clean up of the query code.

PR #1071, opened 2020-09-29
Fix crash using nested ANY/EVERY

snej:
When used on a nested array/dict inside an array/dict, `fl_each` is called with Fleece data, not a `body` column value (which is a serialized RevTree.) This breaks the way it tries to access the Fleece.

Fixed this by taking advantage of the fact that the body-column case always also passes a 2nd parameter (the property path), so it can use the presence of that to tell how to interpret the data.

Fixes CBL-1248

### DIFF
--- a/LiteCore/Query/SQLiteFleeceEach.cc
+++ b/LiteCore/Query/SQLiteFleeceEach.cc
@@ -153,6 +153,9 @@ private:
            hidden columns are unconstrained." */
         /* From json1.c: "The query strategy is to look for an equality constraint on the
            [`root_data`] column.  Without such a constraint, the table cannot operate." */
+        /* In other words: If fl_each() is called with an argument, that manifests here as a
+           constraint on the `root_data` column. A second argument results in a constraint on
+           the `root_path` column. */
         int rootDataIdx = -1, rootPathIdx = -1;
         auto constraint = info->aConstraint;
         for (int i = 0; i < info->nConstraint; i++, constraint++){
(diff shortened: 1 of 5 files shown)


LiteCore/Query/SQLiteFleeceEach.cc
```diff
@@ -221,8 +224,11 @@ class FleeceCursor : public sqlite3_vtab_cursor {
             Warn("fleece_each filter called with null document! Query is likely to fail. (#379)");
             return SQLITE_OK;
         }
-        data = _vtab->context.delegate->fleeceAccessor(data);
-
+        if (idxNum == kPathIndex) {
+            // If fl_each is called with a 2nd (property path) argument, then the first arg is the
+            // doc body, which we need to extract Fleece from:
+            data = _vtab->context.delegate->fleeceAccessor(data);
+        }
```
snej, line 231: FYI: This is the fix. The other changes are test cases that Pasin wrote for his earlier PR, or adding/deleting comments.

pasin: This is the same as what @borrrden has suggested. If fl_each() calls is like `fl_each(fl_nested_value(_V.body, 'sub_dict'), 'array_in_sub_dict')`, the idxNum will be `kFleeceDataIndex`. However I couldn't come up with the SQL that makes translate to the fl_each() call like that.

pasin (approved):

LiteCore/Query/SQLiteFleeceEach.cc
snej: The QueryParser never generates an `fl_each` call like that.

(I know this is sort of fragile, but the better fixes I can think of all involve significant changes/refactoring of the query code. I think we should do that too, but in the meantime this is a decent band-aid.)

snej:
The only Jenkins test failure was in the replicator on Linux, unrelated to this change.

PR #1072, opened 2020-10-08
Fix crash using nested ANY/EVERY

snej:
When used on a nested array/dict inside an array/dict, `fl_each` is
called with Fleece data, not a `body` column value (which is a
serialized RevTree.) This breaks the way it tries to access the
Fleece.

Fixed this by taking advantage of the fact that the body-column case
always also passes a 2nd parameter (the property path), so it can use
the presence of that to tell how to interpret the data.

Fixes CBL-1248


snej:
Sorry, I am a bear of very little brain. This was already merged (#1071).

PR #1073, opened 2020-10-12
Deprecating some C4 APIs

snej:
See "[Changing The LiteCore API](https://docs.google.com/document/d/1hYDbCH_yDIK5i78TDFT6DQgV-YCVTFO7JC9dDR4ctfI/edit#heading=h.5x0d5h95i329)".

* Deprecated c4db_open, c4db_copy, c4db_deleteAtPath, c4db_getConfig.  Instead use c4db_openNamed, c4db_copyNamed, c4db_deleteNamed,  c4db_getConfig2.
* Added c4db_getConfig2, c4db_getName.
* Deprecated c4db_getIndexes and C4QueryOptions.rankFullText.
* Updated all tests to stop using the deprecated functions.

### DIFF
--- a/C/c4Database.cc
+++ b/C/c4Database.cc
@@ -58,7 +58,7 @@ c4Database::~c4Database() {
 static FilePath dbPath(C4String name, C4String parentDir) {
     Assert(name.buf != nullptr && parentDir.buf != nullptr);
     return FilePath(string(slice(parentDir)), string(slice(name)))
-                .withExtension(kC4DatabaseFilenameExtension);
+                .addingExtension(kC4DatabaseFilenameExtension);
 }
 
 
@@ -98,6 +98,7 @@ bool c4key_setPassword(C4EncryptionKey *key, C4String password, C4EncryptionAlgo
 }
 
 
+// TODO - Remove deprecated function
 C4Database* c4db_open(C4Slice path,
                       const C4DatabaseConfig *configP,
                       C4Error *outError) noexcept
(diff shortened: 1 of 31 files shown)


C/c4Database.cc
```diff
@@ -116,18 +117,20 @@ C4Database* c4db_openNamed(C4String name,
...
+    });
 }
 
 
 C4Database* c4db_openAgain(C4Database* db,
                            C4Error *outError) noexcept
 {
-    string path = db->path();
-    return c4db_open({path.data(), path.size()}, c4db_getConfig(db), outError);
+    return c4db_openNamed(c4db_getName(db), c4db_getConfig2(db), outError);
```
borrrden, line 129: Will this still work if the db here was opened with the deprecated method?

snej: Yeah, because c4db_getName and c4db_getConfig2 return the right thing, no matter how the db was opened.

C/include/c4Database.h
```diff
@@ -45,11 +45,6 @@ extern "C" {
         kC4DB_NonObservable = 0x40, ///< Disable c4DatabaseObserver
     };
 
-    /** Document versioning system (also determines database storage schema) */
-    typedef C4_ENUM(uint32_t, C4DocumentVersioning) {
-        kC4RevisionTrees,           ///< Revision trees
-    };
```
borrrden, line 51: Does this mean we are committed to either not doing version vectors, or switching to them without the option to stay on rev trees if and when we do?

borrrden (approved):

C/include/c4Database.h
snej: The latter, probably.

PR #1074, opened 2020-10-13
add a before push hook and an after pull hook

CedricCouton:
The need is to 
- can encrypt some data (with a home encryption system) before push it to the sync gateway 
- can decrypt some data (with a home encryption system) after pull it from the sync gateway before insert it in the local cbl

This PR is a suggestion to add an after pull hook and before push hook

### DIFF
--- a/C/include/c4Replicator.h
+++ b/C/include/c4Replicator.h
@@ -126,6 +126,20 @@ extern "C" {
                                                      C4Error error,
                                                      void *context);
 
+    /** Callback that can process a document body before it's pushed to the remote */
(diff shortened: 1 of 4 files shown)


snej:
This would require adding a Couchbase Lite feature (or it wouldn't be any use), so it would need to go through our product manager and then through an API review. If you're a Couchbase customer, you could request this feature through your support engineer. Or you could propose it on the Couchbase forums.

(We did have a sort of experimental hook like this in the Objective-C version of 1.x, but it never made it into the official feature set.)

PR #1075, opened 2020-10-15
Integrate showfast export into c4PerfTest

borrrden:
This will allow data export to JSON that can be integrated into the showfast graphical dashboard to detect performance regressions

### DIFF
--- a/C/tests/c4PerfTest.cc
+++ b/C/tests/c4PerfTest.cc
@@ -26,21 +26,33 @@
 #include "Benchmark.hh"
 #include "FilePath.hh"
 #include "SecureRandomize.hh"
+#include "repo_version.h"
 #include <fcntl.h>
 #include <sys/stat.h>
 #include <iostream>
 #include <chrono>
 #include <thread>
+#include <fstream>
 #ifndef _MSC_VER
 #include <unistd.h>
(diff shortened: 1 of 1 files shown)


borrrden:
I'm thinking about if this would be better to be completely opt-in via cmake options.

C/tests/c4PerfTest.cc
```diff
@@ -152,19 +164,113 @@ class PerfTest : public C4Test {
...
+        }
+    }
+    
+    
+    string generateShowfast(Benchmark &mark, double scale, string title, const char* items = nullptr) {
+        if(isEncrypted()) {
+            title += "_encrypted";
+        }
+        
+        FLEncoder enc = FLEncoder_NewWithOptions(kFLEncodeJSON, 0, false);
```
snej, line 179: FYI, the Fleece C++ API (Fleece.hh) is friendlier...

snej (approved):
Looks fine. The Fleece C++ API would make the code cleaner, but no biggie.

C/tests/c4PerfTest.cc
borrrden: I'll keep that in mind for an update!

PR #1076, opened 2020-10-23
CBL-862: Reject invalid top level keys

borrrden:
BEHAVIORAL CHANGE

LiteCore will now reject documents which explicitly contain any of the following keys at the top level: _id, _rev, _deleted

@snej I'm unsure how much of this logic is useful outside of debug mode, so if I left in more than is needed let me know.


borrrden:
PR validation failed due to an existing bug, it seems.

snej (approved):

PR #1081, opened 2020-11-13
Switch to gcc for future 2.7.x builds

borrrden:
Without these changes the build will fail on CentOS 6 because the `clock_gettime` function used by CivetWeb will not be found.

### DIFF
--- a/C/tests/cmake/platform_linux.cmake
+++ b/C/tests/cmake/platform_linux.cmake
@@ -33,5 +33,6 @@ function(setup_build)
         ${LIBCXXABI_LIB}
         ${ZLIB_LIB}
         dl
+        rt
     )
 endfunction()
--- a/LiteCore/tests/CMakeLists.txt
+++ b/LiteCore/tests/CMakeLists.txt
@@ -56,8 +56,6 @@ file(GLOB FLEECE_FILES "../../vendor/fleece/Tests/*.json" "../../vendor/fleece/T
 file(COPY ${FLEECE_FILES} DESTINATION ${CMAKE_CURRENT_BINARY_DIR}/vendor/fleece/Tests)
 add_executable(CppTests ${TEST_SRC})
 
-setup_build()
-
 target_compile_definitions(
     CppTests PRIVATE
     -DLITECORE_CPP_TESTS=1
@@ -102,3 +100,5 @@ target_link_libraries(
     Support
     ${LITECORE_CRYPTO_LIB}
 )
+
+setup_build()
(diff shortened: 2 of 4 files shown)


borrrden:
Found a bit more after attempting to build on multiple machines.  My guess of the difference is that the PR validation machine has two versions of GCC installed and so is possibly sharing the linker between them as the system installed version.  Either way, I was able to fix it on that machine and the official docker build image is happy too.

LiteCore/tests/CMakeLists.txt
```diff
@@ -102,3 +100,5 @@ target_link_libraries(
     Support
     ${LITECORE_CRYPTO_LIB}
 )
+
```
bmeike, line 103: Why is this necessary?  ... or, perhaps, just desirable?

C/tests/cmake/platform_linux.cmake
```diff
@@ -33,5 +33,6 @@ function(setup_build)
         ${LIBCXXABI_LIB}
         ${ZLIB_LIB}
         dl
+        rt
```
bmeike, line 36: What is "rt".

bmeike (commented):
There only think I understand looks right.  Can you say what the other two changes are about?

bmeike (approved):

C/tests/cmake/platform_linux.cmake
borrrden: Yes, it is the library for [POSIX.1b realtime extensions](https://www.unix.com/man-page/sunos/3lib/librt/)

LiteCore/tests/CMakeLists.txt
borrrden: It is so that things like ICU and such show up _after_ everything else to satisfy GCC's linker which only supplies symbols to satisfy previously undefined ones in the linker input.  Without this things like deflate, and ICU functions show up as "undefined reference" and the link fails.  This was addressed in a more robust way in 2.8+.

bmeike: Got it.  ... and the 2.8+ way is to big a change to put here?

borrrden: It could be done but it didn't seem worth it because the 2.8 change was basically a CMake project overhaul.  This change accomplishes what is needed while keeping changes to a minimum.

bmeike: Yep.  Got it.  Approved.

PR #1083, opened 2020-11-20
CBL-117: Implement one-shot queries

borrrden:
Trading low memory usage for constraints on usage, this type of query trades the ability to seek and refresh for not buffering anything into memory.

This implements a new query mode where the result set is not first copied into memory; instead the enumerator directly steps the SQLite cursor (statement) and returns the current row. This generally uses less memory and has less latency. On the other hand, it doesn't support seeking backwards or returning the number of rows, and it doesn't work with live queries because those need the buffered results for comparison purposes.

(The memory savings will vary depending on the query. In some cases SQLite itself needs to buffer data, most commonly when there's an ORDER BY that doesn't match the ordering of the primary index being used. I don't know whether SQLite needs to buffer all the result data (or just rowids, for example), or whether its buffered form is more or less compact than ours. It's possible we could introspect the query structure, or scrape the EXPLAIN results, to check whether this buffering will occur.)

### API Changes

There is an API change: adding a `oneShot` property to `C4QueryOptions`. (I also marked the rank-results property as obsolete, because it's not hooked up to anything in the implementation.)

I don't think we can avoid making this a visible option. There are pros and cons to one-shot vs buffered queries, and we don't know ahead of time what the developer will prefer. 

### Theory of operation

Basically there are now two different QueryEnumerator subclasses: one plays back rows from a pre-recorded result set, as before. The other keeps the SQLite statement handle and steps one row at a time.

The problem with this is that, if any rows are changed in the table (by the same SQLite connection), the statement gets messed up and will start returning incorrect data. (It's a case of the generic CS issue wherein modifying a collection breaks any current iterators.)

So the workaround is that the one-shot enumerator listens for any changes about to happen in that table (using a new listener interface I've added.) When it's notified, it creates a new pre-recorded enumerator, which records the remainder of the rows into memory. It then drops its reference to the SQLite statement and delegates future API calls to the pre-recorded enumerator.

### TBD

The one edge case not yet handled is if you create two enumerators on a single query instance. The one-shot enumerator uses the same SQLite statement as the query object, so creating a new enumerator will reset the statement and mess up the first one.

I think the fix to this is to create a new statement for the enumerator, if the first one is still in use. SQLite doesn't seem to have a way to duplicate a `sqlite3_stmt`, so I guess this requires creating a new one from the original SQL text.

(The other option is to do the same thing as when the table is about to change, i.e. tell the first enumerator to record the rest of the rows and stop using the statement.)


borrrden:
Adding more reviewers than usual because this is a LiteCore API change, and I want to make sure it is visible to everyone.

C/include/c4Query.h
```diff
@@ -101,14 +102,11 @@ extern "C" {
...
-        /** The columns of this result, in the same order as in the query's `WHAT` clause. */
+        /** The columns of this result, in the same order as in the query's `WHAT` clause.
+            A column whose value is MISSING will have a Fleece 'undefined' value; this can be detected
+            by `FLValue_GetType(col) == kFLUndefined`. */
         FLArrayIterator columns;
 
-        /** A bitmap where a 1 bit represents a column whose value is MISSING.
-            This is how you tell a missing property value from a value that's JSON 'null',
-            since the value in the `columns` array will be a Fleece `null` either way. */
-        uint64_t missingColumns;
```
snej, line 110: Just pointing out this additional API change: instead of the `missingColumns` kludge, columns whose value is `MISSING` now appear as Fleece values of type `kFLUndefined`. (See the updated doc-comment for `columns`.)

snej (approved):

snej:
Since we're on version 3.0 and breaking changes are allowed, I suggest that we make one-shot queries the default, with some kind of flag in the public API to make seekable queries. I suspect most uses of queries don't require seeks, so this will improve their performance and memory usage.

bmeike (changes requested):
I request that, before you make an API breaking change, you write a one-pager that describes the change and, perhaps, speculates about downstream effects.  I do not understand simply from reading this code, the extent of change you are proposing.  (If the doc already exists, just link it here)

borrrden:
I think we probably cannot merge this until we have a CBL API spec anyway for what this will look like.  From a LiteCore perspective this just means "you cannot use the `missingColumns` property that used to be on `C4QueryEnumerator` and that `C4QueryOptions` now has a new property for selecting one shot or not.  @snej Is there a document describing the purpose of one shot queries that already exists?  As for what to do instead of missing columns, if that's not in the document it should be added (and if there is no document it should be created).

borrrden:
Also @pasin note Jens' suggestion above about making one shot the default.

Sandychuang8 (approved):

pasin:
@borrrden I can confirm that the current CBL API's ResultSet implements forward enumeration only so the API is already compatible with the single shot mode. We need to change the implementation for the `missingColumns`. To support both single shot and non-single shot is separate discussion.

pasin (approved):

PR #1086, opened 2020-11-25
Modify c4_setTempDir and eliminate FilePath::tempDirectory

borrrden:
Motivation:  There is no place to have a default temporary directory that is guaranteed to be on the same filesystem as it needs to be (move operations require the same filesystem to function)

Main changes:  c4Base.h c4_setTempDir signature has changed, and it is now an error to call it twice.

The rest of the commit happened because the tests were still using the tempDirectory call that I removed, and moving that logic up to the test level plowed me into some technical debt regarding `using namespace std` in headers which causes frustrating compile errors (especially on Windows).  There are a few fixups, including a Fleece bump because of an incompatibility I noticed.  The vast majority is just moving the using statement out of the headers and into the compilation files instead, and qualifying things as needed.

### DIFF
--- a/C/c4Base.cc
+++ b/C/c4Base.cc
@@ -43,6 +43,7 @@
 
 
 using namespace litecore;
+using namespace std;
 
 extern "C" {
     CBL_CORE_API std::atomic_int gC4ExpectExceptions;
@@ -480,7 +481,7 @@ bool c4_setTempDir(C4String path, C4Error* err) C4API {
     }
 
     
-    sqlite3_temp_directory = (char *)sqlite3_malloc(path.size + 1);
+    sqlite3_temp_directory = (char *)sqlite3_malloc((int)path.size + 1);
     memcpy(sqlite3_temp_directory, path.buf, path.size);
     sqlite3_temp_directory[path.size] = 0;
     return true;
--- a/C/c4Base.cc
+++ b/C/c4Base.cc
@@ -473,12 +473,17 @@ void c4_dumpInstances(void) C4API {
 #pragma mark - MISCELLANEOUS:
 
 
-void c4_setTempDir(C4String path) C4API {
-    auto pathStr = slice(path).asString();
-    FilePath::setTempDirectory(pathStr);
+bool c4_setTempDir(C4String path, C4Error* err) C4API {
+    if(sqlite3_temp_directory != nullptr) {
+        c4error_make(LiteCoreDomain, kC4ErrorUnsupported, C4STR("c4_setTempDir cannot be called more than once!"));
+        return false;
+    }
 
-    // Tell SQLite to use this temp directory. Note that the dup'd string will be leaked.
-    sqlite3_temp_directory = cbl_strdup(pathStr.c_str());
+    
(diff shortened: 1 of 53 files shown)


bmeike (approved):
It would be very nice to separate the wheat from the chaff...

borrrden:
Your wish is my command.  The PR is now 3 commits long.  One "super wheat" commit, which I think you will find interesting.  One "incidental wheat" commit which LiteCore developers should know about, and one "chaff" commit.

C/c4Base.cc
```diff
@@ -480,7 +481,7 @@ bool c4_setTempDir(C4String path, C4Error* err) C4API {
     }
 
     
-    sqlite3_temp_directory = (char *)sqlite3_malloc(path.size + 1);
+    sqlite3_temp_directory = (char *)sqlite3_malloc((int)path.size + 1);
```
bmeike, line 484: Woo!  Was that a unsigned byte?  Did it overflow?

LiteCore/tests/DataFileTest.cc
```diff
@@ -465,7 +465,7 @@ TEST_CASE("CanonicalPath") {
     const char* startPath = "C:\\folder\\..\\subfolder\\";
     string endPath = "C:\\subfolder\\";
 #else
-    auto tmpPath = FilePath::tempDirectory().path();
+    auto tmpPath = TestFixture::sTempDir.path();
```
bmeike, line 468: Cool!

bmeike (commented):
The main idea, here, is that one shouldn't use the std namespace in a .h file?

Just std?  Maybe anything not defined locally (fleece: ok, std: not ok)?

LiteCore/tests/LiteCoreTest.cc
```diff
@@ -50,6 +50,20 @@ string TestFixture::sFixturesDir = "../LiteCore/tests/data/";
 string TestFixture::sFixturesDir = "LiteCore/tests/data/";
 #endif
 
+static FilePath GetTempDirectory() {
```
bmeike, line 54: This is such gobblty good that it seems a shame to say it twice...

bmeike (commented):
Nice

bmeike (commented):
so much easier to grok.

C/c4Base.cc
borrrden: It is an unsigned 64-bit integer.  I have an idea for this on another branch but I consider this sort of thing technical debt.  Here it is probably harmless unless you have got the world's longest path.

LiteCore/tests/LiteCoreTest.cc
borrrden: Not quite following what you mean

bmeike: Sorry.  It looks like C4Test.cc and LiteCoreTest.cc have almost identical code...

borrrden: There are two distinct testing programs here, and they each have a copy of the code.  Maybe the testing programs need some reworking but that will go down the list.

bmeike: :+1:

PR #1089, opened 2020-11-30
Get us to 0 warnings 0 errors on all platforms

borrrden:
Some of the warnings come from how old our Catch is, so upgraded that as well.  Introduced a "narrow_cast" cast to safely (in a debug build anyway) cast a numeric type to a smaller size (if it is not possible, a LiteCore error is thrown, or if it is in Fleece a FleeceException is thrown). 

This is large because we have a lot of places where this has built up, especially the casting / comparing between different signedness / sizes.  There are still a few more warnings I'd like to enable but for now they cause too many mundane issues (like `-Weffc++` on Linux will warn about not having your initialization in the same order as declared variables in a class, but Apple has this warning on somehow)


borrrden:
This PR is a monster, and shows just how much "warning debt" we've run up.  This brings us back to good standing and shortly after I think we should enable `-Werror` on all platforms.  There are a lot of the same patterns here, mostly `narrow_cast` and switching out printf format specifiers.  A quick glance should be enough, since functionally nothing is changing.

C/tests/c4CertificateTest.cc
```diff
@@ -35,7 +35,7 @@ TEST_CASE("C4Certificate smoke test", "[Certs][C]") {
     for (int i = 0; c4cert_subjectNameAtIndex(certs.temporaryClientIdentity.cert, i, &name); ++i) {
         C4Log("  %.*s = '%.*s'", SPLAT(name.id), SPLAT(name.value));
         if (i == 0) {
-            CHECK(name.id == kC4Cert_CommonName);
+            CHECK((name.id == kC4Cert_CommonName));
```
Sandychuang8, line 38: Just curious, why change to double brackets?

borrrden: I guess I forgot to write about this....the upgraded version of Catch does some funky things to try to break down the expression in these macros, but it has trouble when using `==` to compare `C4Slice` instances.  In order to make it stop freaking out, I wrapped the whole expression in parenthesis so that it would stop trying to break it down.

pasin (approved):
LGTM.

PR #1090, opened 2020-12-03
CBL-220: Fix Windows issue with pre-epoch dates

borrrden:
And by extension, remove all traces of `localtime_r` and `gmtime_r` in favor of [date.h](https://github.com/HowardHinnant/date/tree/v3.0.0/) (Incidentally, this library was largely voted into the C++20 standard so if/when we switch we won't need it anymore).


borrrden:
See https://github.com/couchbase/couchbase-lite-core/pull/1090/commits/f915a45689798a60d8c30ce43c413f3b5e7d3879 for the main changes

LiteCore/Support/LogDecoder.cc
```diff
@@ -53,33 +57,26 @@ namespace litecore {
...
-        localtime_r(&t.secs, &tm);
-        char timestamp[100];
-        strftime(timestamp, sizeof(timestamp), "%T", &tm);
-        out << timestamp;
-        sprintf(timestamp, ".%06u| ", t.microsecs);
-        out << timestamp;
+        local_time<microseconds> tp { seconds(t.secs) + microseconds(t.microsecs) };
+        struct tm tmpTime = FromTimestamp(duration_cast<seconds>(tp.time_since_epoch()));
+        tp -= GetLocalTZOffset(&tmpTime, true);
+        out << format("%T| ", tp);
```
pasin, line 63: Do I understand correctly that the original code writes time in local timezone but the new code writes time in UTC?

pasin (approved):

LiteCore/Support/LogDecoder.cc
```diff
@@ -53,33 +57,26 @@ namespace litecore {
...
 
     string LogIterator::formatDate(Timestamp t) {
-        struct tm tm;
-        localtime_r(&t.secs, &tm);
-        char timestamp[100];
-        strftime(timestamp, sizeof(timestamp), "%c", &tm);
-        return timestamp;
+        local_time<microseconds> tp(seconds(t.secs) + microseconds(t.microsecs));
+        struct tm tmpTime = FromTimestamp(duration_cast<seconds>(tp.time_since_epoch()));
+        tp -= GetLocalTZOffset(&tmpTime, true);
```
Sandychuang8, line 76: Just curious why we use local time, not utc?

borrrden: I don't know the answer to that.

LiteCore/Support/LogDecoder.cc
borrrden: There should be no changes to the format.  timestamps are always in UTC, so to get the local time the offset needs to be applied. Despite the name "local_time" it doesn't mean "time in the local timezone".  It actually means "a time with no associated time zone"  Contrast this with "sys_time" which is always in UTC.  This plays out more when parsing strings with offsets (sys_time will reverse the offset to bring it to UTC while local time will just ignore the offset).

Sandychuang8 (approved):

PR #1091, opened 2020-12-04
Remove and replace platform filler functions

borrrden:
- memmem (Windows) was not used, so remove it from compilation
- Removed the final usages of the BSD specific arc4random (replaced with stdlib random implementation)


Sandychuang8 (approved):

borrrden:
I guess Jenkins has real trouble deciding where to link but this passed validation -> http://mobile.jenkins.couchbase.com/blue/organizations/jenkins/_cbl_github_pipeline%2Fcouchbase-lite-core/detail/PR-1091/4/pipeline

PR #1092, opened 2020-12-10
CBL-1232: Allow progress level to be set after replicator start

borrrden:
[API CHANGE] Addition of c4repl_setProgressLevel in c4Replicator.h

This change has to work its way through the various actors involved:  C4Replicator -> Replicator -> Pusher / Puller

The old method will still continue to function for the time being, but will be removed before 3.0.  It emits a warning to the logs when it is used now.

### DIFF
--- a/C/c4Internal.hh
+++ b/C/c4Internal.hh
@@ -33,6 +33,28 @@ using namespace litecore;
 #define LOCK(MUTEX)     std::unique_lock<std::mutex> _lock(MUTEX)
 #define UNLOCK()        _lock.unlock();
 
+#if defined(__clang__)
+    #define C4_START_WARNINGS_SUPPRESSION _Pragma( "clang diagnostic push" )
+    #define C4_STOP_WARNINGS_SUPPRESSION _Pragma( "clang diagnostic pop" )
+    #define C4_IGNORE_TAUTOLOGICAL _Pragma( "clang diagnostic ignored \"-Wtautological-pointer-compare\"" )
+    #define C4_IGNORE_NONNULL _Pragma( "clang diagnostic ignored \"-Wnonnull\"" )
+#elif defined(__GNUC__)
+    #define C4_START_WARNINGS_SUPPRESSION _Pragma( "GCC diagnostic push" )
+    #define C4_STOP_WARNINGS_SUPPRESSION _Pragma( "GCC diagnostic pop" )
+    #define C4_IGNORE_TAUTOLOGICAL
+    #define C4_IGNORE_NONNULL _Pragma( "GCC diagnostic ignored \"-Wnonnull\"" )
+#elif defined(_MSC_VER)
+    #define C4_START_WARNINGS_SUPPRESSION __pragma( warning(push) )
+    #define C4_STOP_WARNINGS_SUPPRESSION __pragma( warning(pop) )
+    #define C4_IGNORE_TAUTOLOGICAL
+    #define C4_IGNORE_NONNULL
+#else
+    #define C4_START_WARNINGS_SUPPRESSION
+    #define C4_STOP_WARNINGS_SUPPRESSION
+    #define C4_IGNORE_TAUTOLOGICAL
+    #define C4_IGNORE_NONNULL
+#endif
+
 
 namespace c4Internal {
 
--- a/C/c4_ee.gnu
+++ b/C/c4_ee.gnu
@@ -152,6 +152,7 @@ CBL {
 		c4repl_isDocumentPending;
 		c4repl_setOptions;
 		c4repl_setHostReachable;
+		c4repl_setProgressLevel;
 
 		c4socket_registerFactory;
 		c4socket_fromNative;
--- a/C/include/c4Replicator.h
+++ b/C/include/c4Replicator.h
@@ -80,6 +80,17 @@ extern "C" {
         kC4Suspended     = 0x4          ///< If true, will not connect until unsuspended
     };
 
+    /** An enumeration of the levels of progress callbacks the replicator can provide.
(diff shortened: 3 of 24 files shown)


C/c4Internal.hh
```diff
@@ -33,6 +33,28 @@ using namespace litecore;
 #define LOCK(MUTEX)     std::unique_lock<std::mutex> _lock(MUTEX)
 #define UNLOCK()        _lock.unlock();
 
+#if defined(__clang__)
```
bmeike, line 36: What the heck is this?

C/c4_ee.gnu
```diff
@@ -152,6 +152,7 @@ CBL {
 		c4repl_isDocumentPending;
 		c4repl_setOptions;
 		c4repl_setHostReachable;
+		c4repl_setProgressLevel;
 
 		c4socket_registerFactory;
 		c4socket_fromNative;
```
bmeike, line 158: Boy... I hope these files are all generated by a tool with a single source...

C/include/c4Replicator.h
```diff
@@ -80,6 +80,13 @@ extern "C" {
         kC4Suspended     = 0x4          ///< If true, will not connect until unsuspended
     };
 
+    /** An enumeration of the levels of progress callbacks the replicator can provide */
+    typedef C4_ENUM(int32_t, C4ReplicatorProgressLevel) {
+        kC4ReplProgressOverall,         ///< Callback about completion and estimated total
```
bmeike, line 85: Is this exactly two callbacks, one near the start that is a estimate and one after completion?

REST/tests/SyncListenerTest.cc
```diff
@@ -320,6 +320,10 @@ TEST_CASE_METHOD(C4SyncListenerTest, "P2P Server Addresses", "[Listener]") {
     }
     
     for(const auto& interface : litecore::net::Interface::all()) {
+        if(!interface.isRoutable()) {
```
bmeike, line 323: This change is unrelated to the base topic?

Replicator/ReplicatorOptions.hh
```diff
@@ -75,7 +75,14 @@ namespace litecore { namespace repl {
         bool skipDeleted() const  {return properties[kC4ReplicatorOptionSkipDeleted].asBool();}
         bool noIncomingConflicts() const  {return properties[kC4ReplicatorOptionNoIncomingConflicts].asBool();}
         bool noOutgoingConflicts() const  {return properties[kC4ReplicatorOptionNoIncomingConflicts].asBool();}
-        int progressLevel() const  {return (int)properties[kC4ReplicatorOptionProgressLevel].asInt();}
+        int progressLevel() const  {
+            if(properties[kC4ReplicatorOptionProgressLevel]) {
+                C4Warn("Passing in progress level via configuration is deprecated, use the setProgressLevel API");
```
bmeike, line 80: I think that comma should be either a period or a colon

Replicator/Worker.cc
```diff
@@ -258,7 +258,14 @@ namespace litecore { namespace repl {
         }
     }
 
+    void Worker::setProgressNotificationLevel(int level) {
+        if(_progressNotificationLevel.exchange(level) == level) {
```
bmeike, line 262: This caches the new level and returns the old cached level?

bmeike (approved):
FIne... fine...

C/c4Internal.hh
borrrden: This is a way to disable compiler warnings for specific blocks of code.  In particular, Apple would complain mightily about the fact that I check for a "tautological falsehood" of the "non null" replicator being null, and then again it would (rightfully) complain about me passing in null during testing to check that it rejects it.

C/c4_ee.gnu
borrrden: They are (See [generate_export_list.py](https://github.com/couchbase/couchbase-lite-core/blob/master/C/scripts/generate_export_list.py) )

C/include/c4Replicator.h
borrrden: No, it's the "running complete and total" counts callback that you get which periodically updates the states of how much progress has been made.

REST/tests/SyncListenerTest.cc
borrrden: Yes, it was bugging me though because the test fails every time on my machine without it...

Replicator/ReplicatorOptions.hh
borrrden: I'll keep that in mind for next time, but this message and method will be gone soon enough :)

Replicator/Worker.cc
borrrden: Yes, that's how all atomic exchanges work as far as I know.

C/include/c4Replicator.h
bmeike: Ok.  We need some kind of indicator, then (... and I think this is a good place for it) of the relative costs of each of these things.

borrrden: Cost in terms of what?  Time?  Complexity?

bmeike: I have no idea.  That's what I'm asking for.  You need to say what makes them different and why someone would choose one instead of the other.

REST/tests/SyncListenerTest.cc
borrrden: I took it out though, will put it back later

C/include/c4Replicator.h
borrrden: Ok I've added that

Sandychuang8 (approved):

jayahariv (approved):

PR #1094, opened 2020-12-15
Remove undefined setSqliteTempDirectory

ionutcosminmihaipfa:


borrrden:
It's definitely [defined](https://github.com/couchbase/couchbase-lite-core/blob/master/MSVC/SQLiteTempDirectory.cc), the question is how are you getting into a situation in which it is attempting to be used without being defined.  Seems more like a bug with the C build than an issue with Core.

PR #1096, opened 2020-12-18
CBL-1496: Add more robust logging

borrrden:
There are many places where we swallow exceptions, silently return, and otherwise provide no information about things that are going wrong.  Since usually things go right we don't notice this until things go wrong.  This is an attempt to provide more proactive logging about unusual / nontypical situations so that we can better reason about the logs we receive.

### DIFF
--- a/C/c4Document.cc
+++ b/C/c4Document.cc
@@ -32,6 +32,7 @@
 #include "RevTree.hh"   // only for kDefaultRemoteID
 #include "SecureRandomize.hh"
 #include "FleeceImpl.hh"
+#include "StringUtil.hh"
 
 using namespace fleece::impl;
 using namespace std;
@@ -223,6 +224,9 @@ C4RemoteID c4db_getRemoteDBID(C4Database *db, C4String remoteAddress, bool canCr
             }
 
             // Look up the doc in the db, and the remote URL in the doc:
+            // Tech Debt: Constants are not matched with their most used form
+            // (i.e. kC4InfoStore is usually used as a string, and kRemoteDBURLsDoc
+            // is always used as a slice)
             Record doc = db->getRawDocument(toString(kC4InfoStore), slice(kRemoteDBURLsDoc));
             const Dict *remotes = nullptr;
             C4RemoteID remoteID = 0;
@@ -239,10 +243,16 @@ C4RemoteID c4db_getRemoteDBID(C4Database *db, C4String remoteAddress, bool canCr
 
             if (remoteID > 0) {
                 // Found the remote ID!
+                WriteVerbose("c4db_getRemoteDBID: Found remote ID %" PRIu32 " for %.*s", remoteID, SPLAT(remoteAddress));
                 return remoteID;
-            } else if (!canCreate) {
+            } 
+
+            if (!canCreate) {
+                WriteVerbose("c4db_getRemoteDBID: Remote ID not found for %.*s, and creation suppressed...", SPLAT(remoteAddress));
                 break;
-            } else if (creating) {
+            } 
+
+            if (creating) {
                 // Update or create the document, adding the identifier:
                 remoteID = 1;
                 Encoder enc;
@@ -255,6 +265,7 @@ C4RemoteID c4db_getRemoteDBID(C4Database *db, C4String remoteAddress, bool canCr
                         remoteID = max(remoteID, 1 + C4RemoteID(existingID));   // make sure new ID is unique
                     }
                 }
+
                 enc.writeKey(remoteAddress);                       // Add new entry
                 enc.writeUInt(remoteID);
                 enc.endDictionary();
@@ -264,15 +275,22 @@ C4RemoteID c4db_getRemoteDBID(C4Database *db, C4String remoteAddress, bool canCr
                 db->putRawDocument(toString(kC4InfoStore), slice(kRemoteDBURLsDoc), nullslice, body);
                 db->endTransaction(true);
                 inTransaction = false;
+                Log("c4db_getRemoteDBID: Created remote ID %" PRIu32 " for %.*s", remoteID, SPLAT(remoteAddress));
                 return remoteID;
             }
         }
         if (outError)
             *outError = c4error_make(LiteCoreDomain, kC4ErrorNotFound, {});
-        return C4RemoteID(0);
+
+        return C4RemoteID();
     });
-    if (inTransaction)
-        c4db_endTransaction(db, false, nullptr);
+
+    if (inTransaction) {
+        if(!c4db_endTransaction(db, false, nullptr)) {
+            WarnError("Failed to end transaction in c4db_getRemoteDBID");
+        }
+    }
+
     return remoteID;
 }
 
@@ -290,6 +308,8 @@ C4SliceResult c4db_getRemoteDBAddress(C4Database *db, C4RemoteID remoteID) C4API
                 }
             }
         }
+
+        Warn("c4db_getRemoteDBAddress: Address not found for remote ID %" PRIu32, remoteID);
         return C4SliceResult{};
     });
 }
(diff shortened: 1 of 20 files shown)


snej:
What's the PR about? There should be a description...

borrrden:
Good point, my tunnel vision strikes again.

borrrden:
Alright, the description is updated.

LiteCore/Support/Logging.hh
```diff
@@ -166,6 +166,7 @@ extern LogDomain DBLog, QueryLog, SyncLog, &ActorLog;
 #define Log(FMT, ...)                   LogToAt(kC4Cpp_DefaultLog, Info,    FMT, ##__VA_ARGS__)
 #define Warn(FMT, ...)                  LogToAt(kC4Cpp_DefaultLog, Warning, FMT, ##__VA_ARGS__)
 #define WarnError(FMT, ...)             LogToAt(kC4Cpp_DefaultLog, Error,   FMT, ##__VA_ARGS__)
+#define WriteVerbose(FMT, ...)          LogToAt(kC4Cpp_DefaultLog, Verbose, FMT, ##__VA_ARGS__)
 #define WriteDebug(FMT, ...)            LogToAt(kC4Cpp_DefaultLog, Debug,   FMT, ##__VA_ARGS__)
```
snej, line 170: We already have `LogVerbose` and `LogDebug`. Why are these new ones needed?

LiteCore/Support/c4ExceptionUtils.hh
```diff
@@ -43,11 +43,23 @@ namespace c4Internal {
...
             c4Internal::recordException(x, OUTERR); \
         }
 
     #define catchExceptions() \
-        catch (const std::exception &) { }
+        catch (const std::exception &x) { \
+            C4WarnError("Exception caught in %s: %s", __func__, x.what()); \
+        }
+
+    #define catchTripleDot() \
```
snej, line 55: I don't think this is needed. There are no exceptions in our code or the std lib that don't inherit from `std::exception`.

C/c4Document.cc
```diff
@@ -596,7 +638,17 @@ C4Document* c4doc_update(C4Document *doc,
 
 bool c4doc_removeRevisionBody(C4Document* doc) noexcept {
     auto idoc = asInternal(doc);
-    return idoc->mustBeInTransaction(NULL) && idoc->removeSelectedRevBody();
+    if(!idoc->mustBeInTransaction(nullptr)) {
+        WarnError("c4doc_removeRevisionBody: Called outside transaction");
```
snej, line 642: Wouldn't it be better to put this warning in `mustBeInTransaction`?

C/c4Document.cc
```diff
@@ -596,7 +638,17 @@ C4Document* c4doc_update(C4Document *doc,
...
+        return false;
+    }
+
+    if(!idoc->removeSelectedRevBody()) {
+        Warn("c4doc_removeRevisionBody: failed to remove");
```
snej, line 647: Not clear why this is worth a warning.

C/include/c4Base.h
```diff
@@ -142,6 +142,8 @@ void* c4base_retain(void *obj) C4API;
 void c4base_release(void *obj) C4API;
 
 // These types are reference counted and have c4xxx_retain / c4xxx_release functions:
+// NOTE that in debug builds retain and release cann throw exceptions despite being
+// marked noexcept.  Release builds are exception free.
```
snej, line 146: They can't, actually. The effect of `noexcept` is that, rather than allow an exception to be thrown, the C++ runtime will terminate the process. This basically turns exceptions into fatal assertion failures (which our terminate handler logs.)

Replicator/DatabaseCookies.cc
```diff
@@ -42,8 +42,11 @@ namespace litecore { namespace repl {
 
 
     void DatabaseCookies::saveChanges() {
-        if (!_store->changed())
+        if (!_store->changed()) {
+            WriteVerbose("Cookie Store unchanged, skipping save...");
```
snej, line 46: This doesn't seem worth logging.

Replicator/c4Replicator.cc
```diff
@@ -62,9 +62,15 @@ static uint16_t defaultPortForScheme(slice scheme) {
 bool c4repl_isValidDatabaseName(C4String dbName) C4API {
     slice name = dbName;
     // Same rules as Couchbase Lite 1.x and CouchDB
-    return name.size > 0 && name.size < 240
+    const auto valid = name.size > 0 && name.size < 240
         && islower(name[0])
         && !slice(name).findByteNotIn("abcdefghijklmnopqrstuvwxyz0123456789_$()+-/"_sl);
+
+    if(!valid) {
+        WarnError("%.*s is not a valid database name for replication", SPLAT(dbName));
```
snej, line 70: This seems the wrong place to log this. It's simply an informational function. There must be a function that calls this and, on false, causes an error; that would be the place to log.

Replicator/c4Replicator.cc
```diff
@@ -89,43 +95,65 @@ bool c4address_fromURL(C4String url, C4Address *address, C4String *dbName) C4API
     slice str = url;
 
     auto colon = str.findByteOrEnd(':');
-    if (!colon)
+    if (!colon) {
+        // Can't log URL yet unless we add a function to strip out a potential password
+        WarnError("c4address_fromURL: Invalid URL (missing colon)");
```
snej, line 100: This is overkill, logging every possible way a URL could be invalid. We could just have one place where we log the URL and say it's invalid. (Don't we already do that in the error we return from the replicator?)

Replicator/c4Socket.cc
```diff
@@ -37,6 +37,14 @@ using namespace fleece;
 using namespace litecore;
 using namespace litecore::net;
 
+#define catchAndClose() \
+    catch (const std::exception &x) { \
+        closeWithException(x); \
+    } catch (...) { \
```
snej, line 43: We have no non-`std::exception` exceptions, so this isn't needed.

Replicator/c4Socket.cc
```diff
@@ -145,16 +156,22 @@ namespace litecore { namespace repl {
         WebSocketImpl::connect();
         if (_factory.open) {
             net::Address c4addr(url());
-            _factory.open(this, &c4addr, options().data(), _factory.context);
+            try {
```
snej, line 159: I think the try/catch belongs in the factory implementations. They should not throw exceptions.

snej (changes requested):

C/c4Document.cc
borrrden: I am heavily erring on the side of verbosity here.  I got tired of finding places where it would have been useful to know why a function returned a non-expected value and so I've been going through and filling them in.

C/include/c4Base.h
borrrden: That is predicated on calling the C function that actually wires it up, which I don't think any of the platforms do unfortunately.  I'd like to see that be automatic....

Replicator/c4Socket.cc
borrrden: In general I am against putting trust in code that we don't control if it has the ability to cascade through our base like this....but I am willing to accept it under the argument of "This is not meant for public consumption and you do so at your own risk"

LiteCore/Support/Logging.hh
borrrden: WriteDebug was already there, I just made verbose available.  These write without specifying a log domain, as opposed to the others which accept the domain as their first argument.

LiteCore/Support/c4ExceptionUtils.hh
borrrden: I guess it would be better to just have the fatal crash trace on by default?

C/c4Document.cc
borrrden: Maybe, but it wouldn't indicate which function called it incorrectly.

Replicator/DatabaseCookies.cc
borrrden: Probably not, I do remember going a bit crazy toward the end.

Replicator/c4Socket.cc
borrrden: Interestingly some Windows errors manifest as exceptions that are not based on std::exception where on other platforms they would be signals.  This is the only way to catch such errors.  Whether or not that is useful, is another story.

C/include/c4Base.h
borrrden: I modified the comments

borrrden:
I resolved most of the comments but left a few open

Replicator/c4Replicator.cc
```diff
@@ -89,43 +91,58 @@ bool c4address_fromURL(C4String url, C4Address *address, C4String *dbName) C4API
...
     } else {
         address->hostname = nullslice;
     }
 
     colon = str.findByteOrEnd(':');
     auto pathStart = str.findByteOrEnd('/');
-    if (str.findByteOrEnd('@') < pathStart)
-        return false;                               // No usernames or passwords allowed!
+    if (str.findByteOrEnd('@') < pathStart) {
+        return false;
```
jayahariv, line 130: why not add log in this case `No usernames or passwords allowed!`?

jayahariv (approved):

PR #1099, opened 2021-01-05
CBL-1513: Don't leak internal stopping status

borrrden:
If adding a new internal status, it will need to be handled here as well.

### DIFF
--- a/Replicator/c4Replicator.hh
+++ b/Replicator/c4Replicator.hh
@@ -136,7 +136,25 @@ struct C4Replicator : public RefCounted,
 
     C4ReplicatorStatus status() {
         LOCK(_mutex);
-        return _status;
+        switch (_status.level) {
+        // CBL-1513: Any new approved statuses must be added to this list,
+        // or they will be forced to busy in order to prevent internal statuses
+        // from leaking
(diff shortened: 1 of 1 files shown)


jayahariv (approved):

snej (approved):

Replicator/c4Replicator.hh
```diff
@@ -136,7 +136,17 @@ struct C4Replicator : public RefCounted,
 
     C4ReplicatorStatus status() {
```
bmeike, line 137: I think that a much more robust thing to do would be to list, explicitly, the statuses that are returned, according to the doc, and to return them.
The default could be what kC4Stopping is now... or better yet, could throw an error, and kC4Stopping could be handled explicitly.

bmeike (approved):
I suggest returning visible states explicitly.  Just a suggestion, though.

Replicator/c4Replicator.hh
borrrden: That sounds reasonable, and new statuses would fail by being an incorrect but valid status instead of a potentially invalid one if they are not handled.  Sounds more futureproof.

PR #1100, opened 2021-01-05
Improved nullability annotations in C API

snej:
Now using Clang's `_Nullable` annotation, along with the pragmas
`clang assume_nonnull begin` and `clang assume_nonnull end`. This is
the same mechanism used in Objective-C APIs. Its advantage over
`__attribute((nonnull))` is that it does not imply that passing null
is undefined behavior, so it doesn't affect code optimization in ways
that make it impossible to add an explicit null check inside the
function.

Because of this way this stuff works, the headers now explicitly tag
_nullable_ pointers, instead of ones that cannot be null.

Pointer fields in structs also cannot be NULL except when tagged as
nullable.

This stuff is wrapped in macros, defined in c4Compat.h, for compatibility: in between
C4_ASSUME_NONNULL_BEGIN and C4_ASSUME_NONNULL_END, all pointer
declarations *implicitly* disallow NULL values, unless annotated with
C4NULLABLE (which must come after the `*`.)

An explicit C4NONNULL is occasionally necessary when there are
multiple levels of pointers.

### DIFF
--- a/C/include/c4Certificate.h
+++ b/C/include/c4Certificate.h
@@ -23,6 +23,8 @@
 
 #ifdef COUCHBASE_ENTERPRISE
 
+C4_ASSUME_NONNULL_BEGIN
+
 #ifdef __cplusplus
 extern "C" {
 #endif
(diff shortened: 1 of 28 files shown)


C/include/c4Certificate.h
```diff
@@ -23,6 +23,8 @@
 
 #ifdef COUCHBASE_ENTERPRISE
 
+C4_ASSUME_NONNULL_BEGIN
```
borrrden, line 26: The assume end is outside the COUCHBASE_ENTERPRISE ifdef, causing CE builds to fail

borrrden (changes requested):

borrrden (approved):
I'm a little skiddish about this because it seems to rule out the possibility of bringing it to other platforms later due to the odd pattern if begin and end.  However, that's currently only theoretical so I won't block on it.

PR #1101, opened 2021-01-06
Quiet unit test logging

snej:
Added QuietReporter, a new Catch Reporter subclass that suppresses console logging (except warnings/errors) until an assertion fails, then dumps the prior logs for that test.

Full logs are always written to binary files.

Use `-r quiet` on the command line to enable QuietReporter.

Tested on macOS, compiled with Clang, built with Xcode and CMake.

### DIFF
--- a/C/tests/c4Test.cc
+++ b/C/tests/c4Test.cc
@@ -17,6 +17,7 @@
 //
 
 #include "c4Test.hh"
+#include "TestsCommon.hh"
 #include "c4BlobStore.h"
 #include "c4Document+Fleece.h"
 #include "c4Private.h"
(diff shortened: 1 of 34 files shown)


C/tests/c4Test.cc
```diff
@@ -203,22 +203,6 @@ _versioning(kC4RevisionTrees)
 {
     static once_flag once;
     call_once(once, [] {
-        c4log_enableFatalExceptionBacktrace();
```
borrrden, line 206: Did you mean to get rid of this?  Without it there is no backtrace on crash.

snej: That's in QuietReporter.cc now, along with all the logging setup.

borrrden: I noticed that as well but what about the list reporter?

snej: I figured this one obsoletes it, except as the superclass. It's a bit awkward having the logging initialization go here, but it needs to so that it can set the log levels and find the log files.

borrrden (approved):

PR #1102, opened 2021-01-07
CBL-1515: Aggregate the platform variations of URL for Checkpoint

borrrden:
Some platforms were passing in an explicit port, even for default 80/443, while others were just leaving it blank. This led to discrepencies between the platform checkpoints, as a remote URL is used for calculating a replication checkpoint.  If a preloaded database is used on a platform that differs in this manner, then the previous checkpoint will fail to be detected.  This commit solves this by checking for both variations during the previous checkpoint search.

### DIFF
--- a/Replicator/Checkpointer.cc
+++ b/Replicator/Checkpointer.cc
@@ -241,12 +241,12 @@ namespace litecore { namespace repl {
         enc.writeString({&localUUID, sizeof(C4UUID)});
 
         alloc_slice rawURL(remoteDBIDString());
-        auto encodedURL = URLTransformer::Transform(rawURL, urlStrategy);
+        auto encodedURL = transform_url(rawURL, urlStrategy);
         if(!encodedURL) {
             return "";
         }
 
-        enc.writeString(encodedURL);;
+        enc.writeString(encodedURL);
         if (!channels.empty() || !docIDs.empty() || filter) {
             // Optional stuff:
             writeValueOrNull(enc, channels);
--- a/Replicator/Checkpointer.cc
+++ b/Replicator/Checkpointer.cc
@@ -205,7 +205,7 @@ namespace litecore { namespace repl {
             C4UUID privateID;
             if (!c4db_getUUIDs(db, nullptr, &privateID, err))
                 return nullslice;
-            _docID = docIDForUUID(privateID);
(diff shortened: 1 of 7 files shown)


jayahariv (approved):

Replicator/URLTransformer.hh
```diff
@@ -0,0 +1,71 @@
...
+    public:
+        /**
+         * Transforms the URL passed in input using the provided strategy
+         *
+         * @param input The URL to transform
+         * @param strategy The strategy to use
+         * @returns The transformed URL.  If the URL is not a candidate for AddPort or RemovePort
+         *          (e.g. not a valid URL, or not using a standard port) then nullslice is returned)
+         */
+        static alloc_slice Transform(const slice &input, URLTransformStrategy strategy);
```
snej, line 57: - Method names should be lowercase!
- `input` is a vague name for a parameter of a general-purpose type like `slice`. How about `inputURL`?
- Also, passing a slice as `const slice&` is unnecessary. The only reason for doing it with `alloc_slice` is to avoid a retain+release, but that isn't an issue with slice.

Replicator/Checkpointer.cc
```diff
@@ -240,7 +240,13 @@ namespace litecore { namespace repl {
...
         enc.writeString({&localUUID, sizeof(C4UUID)});
 
-        enc.writeString(remoteDBIDString());
+        alloc_slice rawURL(remoteDBIDString());
+        auto encodedURL = URLTransformer::Transform(rawURL, urlStrategy);
+        if(!encodedURL) {
+            return "";
+        }
+
+        enc.writeString(encodedURL);;
```
snej, line 249: Double semicolon

snej (changes requested):

snej:
The transformer class feels overengineered. I think it could be done more simply with one method and a switch statement, since the difference between the strategies just comes down to changing the port field.

Replicator/Checkpointer.cc
borrrden: 😱

borrrden:
I guess that's true.  The point I want to ensure is that it is testable.  I didn't want to bury it in the Checkpointer class.

snej (approved):

PR #1110, opened 2021-01-21
Fix query grammar for multi-joins. CBL-1560

jianminzhao:
Allow second JOIN to consume white spaces. Added test cases that would fail without this fix.


borrrden (changes requested):
Just to be consistent with the original ticket could you add or modify these tests to use LEFT JOIN instead of just JOIN?

jianminzhao:
@borrrden will do.

borrrden (approved):

snej (approved):

PR #1113, opened 2021-01-28
Make "AS" optional to lead the aliases for the column names. #CBL-1523

jianminzhao:
This requirement interferes with the current grammar for COLLATE suffix. Currently, COLLATE may lead multiple collation names. After the change, one may put multiple collation names inside parenthesis. The parenthesis may be omitted if it encloses only one collation name.

### DIFF
--- a/C/tests/c4QueryTest.cc
+++ b/C/tests/c4QueryTest.cc
@@ -412,15 +412,6 @@ N_WAY_TEST_CASE_METHOD(C4QueryTest, "C4Query dict literal", "[Query][C]") {
     CHECK(results[0] == "{d:1234.5,f:false,i:12345,id:\"0000001\",n:null,s:\"howdy\",t:true}");
 }
 
-N_WAY_TEST_CASE_METHOD(C4QueryTest, "C4Query N1QL parse error", "[Query][C][N1QL][!throws]") {
-    int errPos;
-    C4Error error;
-
-    C4Query *q = c4query_new2(db, kC4N1QLQuery, "SELECT foo asAlias, bar"_sl, &errPos, &error);
-    CHECK(q != nullptr);
-    c4query_release(q);
-}
-
 #pragma mark - FTS:
 
 
--- a/C/tests/c4QueryTest.cc
+++ b/C/tests/c4QueryTest.cc
@@ -415,15 +415,8 @@ N_WAY_TEST_CASE_METHOD(C4QueryTest, "C4Query dict literal", "[Query][C]") {
 N_WAY_TEST_CASE_METHOD(C4QueryTest, "C4Query N1QL parse error", "[Query][C][N1QL][!throws]") {
     int errPos;
     C4Error error;
-    {
-        ExpectingExceptions x;
-        CHECK(c4query_new2(db, kC4N1QLQuery, "SELECT foo bar"_sl, &errPos, &error) == nullptr);
-    }
-    CHECK(errPos == 11);
-    CHECK(error.domain == LiteCoreDomain);
-    CHECK(error.code == kC4ErrorInvalidQuery);
 
-    C4Query *q = c4query_new2(db, kC4N1QLQuery, "SELECT foo, bar"_sl, &errPos, &error);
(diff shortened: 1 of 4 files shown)


LiteCore/Query/N1QL_Parser/n1ql.leg
```diff
@@ -234,11 +234,31 @@ selectExpr =
...
+    (
+        c:collation _ !collation        { co = arrayWith(c); }
+      | '(' _
+        ( c:collation _                 { if (co.isNull()) {
+                                            co = arrayWith(c);
+                                          } else {
+                                            appendAny(co, c);
+                                          } }
+        )+ ')' _
+    )                                   { $$ = co; }
```
pasin, line 261: I believe the parser above would work. I'm trying to see if the parser code could process the multiple collations without putting them into an array first. Here is one idea, **but** I haven't tested so it might not work.

```
| x:baseExpr 
    (_ COLLATE                          { bool did_collateOp = false; }
        c:collation _ !collation.       { x = collateOp(x, c); }
            | '(' _
                ( c:collation _         { if (!did_collateOp) {
                                            x = collateOp(x, c);
                                            did_collateOp = true;
                                          } else {
                                            extendCollate(x, c2);
                                          } }
                )+ ')' _
    )?                                  { $$ = x; }
```

pasin: I see, only the var defined before colon can be seen through the inner blocks.

pasin (approved):

C/tests/c4QueryTest.cc
```diff
@@ -415,15 +415,8 @@ N_WAY_TEST_CASE_METHOD(C4QueryTest, "C4Query dict literal", "[Query][C]") {
...
     C4Error error;
-    {
-        ExpectingExceptions x;
-        CHECK(c4query_new2(db, kC4N1QLQuery, "SELECT foo bar"_sl, &errPos, &error) == nullptr);
-    }
-    CHECK(errPos == 11);
-    CHECK(error.domain == LiteCoreDomain);
-    CHECK(error.code == kC4ErrorInvalidQuery);
 
-    C4Query *q = c4query_new2(db, kC4N1QLQuery, "SELECT foo, bar"_sl, &errPos, &error);
```
borrrden, line 426: Did you remove this test on purpose?

jianminzhao: I did not remove it, but extended it to include the new rule allowing adding alias without AS. This case was in the preceding case that expect it to fail.

borrrden: Where did this case move to?  I see two test cases before, and only one after.  

SELECT foo bar (aka SELECT foo AS bar) -> SELECT foo asAlias
SELECT foo, bar (SELECT FOO (and), BAR) -> ???

C/tests/c4QueryTest.cc
```diff
@@ -415,15 +415,8 @@ N_WAY_TEST_CASE_METHOD(C4QueryTest, "C4Query dict literal", "[Query][C]") {
 N_WAY_TEST_CASE_METHOD(C4QueryTest, "C4Query N1QL parse error", "[Query][C][N1QL][!throws]") {
```
borrrden, line 415: There is no more error here so perhaps this test should be renamed.

C/tests/c4QueryTest.cc
jianminzhao: Basically, I modified the second test case to include the logic of the first test case. Do you prefer not to touch the existing case, but modify the first one to, instead of expecting failure, expect it to succeed? I can do that.

borrrden: It seems this test has become redundant really.  The same logic is tested in another test, and the purpose of this test is no longer to check parsing errors.

C/tests/c4QueryTest.cc
jianminzhao: I see. Let me just remove it altogether.

PR #1115, opened 2021-01-30
CBL-49: Fully support booleans in queries

borrrden:
Before some of them remained integers which could be interpreted as boolean results, but now all functions and operators which are logically thought of as boolean will return true or false instead of 0 or 1.

### DIFF
--- a/LiteCore/Query/QueryParser.cc
+++ b/LiteCore/Query/QueryParser.cc
@@ -33,6 +33,7 @@
 #include "NumConversion.hh"
 #include <utility>
 #include <algorithm>
+#include <unordered_set>
 
 using namespace std;
 using namespace fleece;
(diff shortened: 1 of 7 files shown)


jianminzhao (approved):

pasin:
@borrrden Can you help summarize how this would change the 3.3.2.1.3 Logical Operators in the API [Spec](https://docs.google.com/document/d/1FBN316lV0RXRibd71GKLGkgDZWSfKWhT-_2GZIKAoQY)?

LiteCore/Query/SQLiteFleeceUtil.cc
```diff
@@ -325,6 +326,44 @@ namespace litecore {
         }
     }
 
+    enhanced_bool_t booleanValue(sqlite3_context* ctx, sqlite3_value *arg) {
```
pasin, line 329: This is cool. Does this mean we can do N1QL boolean value as below?

* MISSING, NULL, and FALSE are FALSE
* numbers +0, -0, and NaN are FALSE
* empty strings, arrays, and objects are FALSE
* all other values are TRUE

LiteCore/Query/QueryParser.cc
```diff
@@ -44,6 +45,38 @@ namespace litecore {
...
+        bool isImplicitBool(const Value* op) {
+            if(!op) {
+                return false;
+            }
+
+            static const case_insensitive_set implicitBoolOps = {
+                "!="_sl, "="_sl, ">"_sl, "<"_sl, ">="_sl, "<="_sl,
+                "IS"_sl, "IS NOT"_sl, "NOT"_sl,
+                "BETWEEN"_sl, "AND"_sl, "OR"_sl,
+                "NOT IN"_sl, "EVERY"_sl, "ANY AND EVERY"_sl
```
pasin, line 74: * We also support SOME which is the same as ANY. 
* How about LIKE?

borrrden: LIKE actually translates to its own function that already returns a bool.  Not sure about SOME, I don't see it being mapped by the query parser to anything  I don't think it works right now.

LiteCore/Query/SQLiteFleeceUtil.cc
borrrden: We can, but we shouldn't because that is wrong....this boolean evaluation should match server.  

>   MISSING is MISSING.
NULL is NULL.
False is false.
Numbers +0, -0, and NaN are false.
Empty strings, arrays, and objects are false.
All other values are true.

pasin (approved):
After chatting with @borrrden in slack, I have better understanding about the change. I will update the spec accordingly.

PR #1117, opened 2021-02-09
N1ql functions: missingif and nullif. #CBL-1628

jianminzhao:


borrrden (approved):
Note that this continues the stance of "We are only going to accept floating point numbers as equal if they are byte for byte equal, regardless of the source of their generation"

LiteCore/Query/SQLiteN1QLFunctions.cc
```diff
@@ -340,15 +340,48 @@ namespace litecore {
...
+        }
+    
+        // Pre-conditions: none of argv0 or argv1 isNull() or isMissing()
+        bool isEqual(sqlite3_value* argv0, sqlite3_value* argv1) {
+            double d0;
+            bool ret = false;
+            if (isNumber(argv0, &d0)) {
+                double d1;
+                if (isNumber(argv1, &d1)) {
+                    ret = (d0 == d1);
```
snej, line 362: This can return false positives for very large integers, since doubles have only 53(?) bits of precision. You should compare two SQLITE_INTEGER values as int64s.

snej (changes requested):

LiteCore/Query/SQLiteN1QLFunctions.cc
jianminzhao: Yes, I thought of this too. I was hoping we will not deal with integers that is that large. (I noticed in sqlite3 api, sqlite3_value_int() and sqlite3_value_int64() actually return the same number (by casting).)
To be absolutely right, we either guarantee not to use integer larger than int32, or have to cross compare int64 with double using fmod. I can do the cross comparison (int64 vs int64, int64 vs double, double vs int64).

snej (approved):

PR #1120, opened 2021-02-12
CBL-1663: Fix infinite loop on socket POLLHUP

borrrden:
A zero return value for read or write from a socket can sometimes indicate that the socket has reached EOF, and in that case the polling should not continue.

### DIFF
--- a/Networking/WebSockets/BuiltInWebSocket.cc
+++ b/Networking/WebSockets/BuiltInWebSocket.cc
@@ -381,6 +381,18 @@ namespace litecore { namespace websocket {
                 closeWithError(_socket->error());
                 return;
             }
+            
+            // CBL-1633: Need to check for EOF here, or _socket->read returns
+            // 0 endlessly
+            if(n == 0 && _socket->atReadEOF()) {
+                if(_socket->atWriteEOF()) {
+                    closeSocket();
+                } else {
+                    close();
+                }
+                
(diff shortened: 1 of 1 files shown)


Sandychuang8 (approved):

Networking/WebSockets/BuiltInWebSocket.cc
```diff
@@ -435,6 +447,16 @@ namespace litecore { namespace websocket {
                     closeWithError(_socket->error());
                 return;
             }
+            
+            // CBL-1633: Need to check for EOF here, or _socket->write returns
+            // 0 endlessly
+            if(n == 0 && _socket->atWriteEOF()) {
+                if(_socket->atReadEOF()) {
+                    closeSocket();
+                }
```
pasin, line 456: Does it need to call close() if atReadEOF() returns false like it has been done in readFromSocket()?

borrrden: I don't think so because close will try to write, but we are already at writeEOF so we can't write anymore.

pasin: Got it.

pasin (approved):

snej (approved):

jianminzhao (approved):

PR #1121, opened 2021-02-17
Allow property to be implicitly scoped by the unique keyspace alias. …

jianminzhao:
…#CBL-1633

When there is unique keyspace alias, i.e., when not joining, properties do not have to start by the alias.

### DIFF
--- a/LiteCore/tests/N1QLParserTest.cc
+++ b/LiteCore/tests/N1QLParserTest.cc
@@ -122,14 +122,9 @@ TEST_CASE_METHOD(N1QLParserTest, "N1QL properties", "[Query][N1QL][C]") {
     CHECK(translate("SELECT DISTINCT custId FROM orders where test_id = 'agg_func' ORDER BY custId") ==
           "{'DISTINCT':true,'FROM':[{'AS':'orders'}],'ORDER_BY':[['.custId']],"
           "'WHAT':[['.custId']],'WHERE':['=',['.test_id'],'agg_func']}");
-    try {
-        // "custId" must be explicitly scoped by a keyspace alias if joining with another table.
-        translate("SELECT custId, other.custId FROM orders JOIN orders other ON orders.test_id = other.test_id ORDER BY custId");
-        CHECK(false);
-    } catch (const exception& exc) {
-        string error = "property 'custId.' does not begin with a declared 'AS' alias";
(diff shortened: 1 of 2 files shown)


snej (approved):

LiteCore/tests/N1QLParserTest.cc
```diff
@@ -117,6 +117,19 @@ TEST_CASE_METHOD(N1QLParserTest, "N1QL properties", "[Query][N1QL][C]") {
...
+    CHECK(translate("SELECT DISTINCT custId FROM orders where test_id = 'agg_func' ORDER BY custId") ==
+          "{'DISTINCT':true,'FROM':[{'AS':'orders'}],'ORDER_BY':[['.custId']],"
+          "'WHAT':[['.custId']],'WHERE':['=',['.test_id'],'agg_func']}");
+    try {
+        // "custId" must be explicitly scoped by a keyspace alias if joining with another table.
+        translate("SELECT custId, other.custId FROM orders JOIN orders other ON orders.test_id = other.test_id ORDER BY custId");
+        CHECK(false);
+    } catch (const exception& exc) {
+        string error = "property 'custId.' does not begin with a declared 'AS' alias";
+        CHECK(error == exc.what());
```
snej, line 131: FYI, the convention with CHECK / REQUIRE is to put the value to be tested first, then the expected value, like `CHECK(exc.what() == error);`. It doesn't really make a difference since both values get logged, but it's clearer to someone looking at a test failure.

snej (approved):

LiteCore/tests/N1QLParserTest.cc
jianminzhao: Understood, but the tricky thing is that the failure in translate() occurs at "string sql = parse(dict);", which simply throws. It passed the syntax check and succeeded in a valid JSON AST. When parsing that to SQL, it checks the AS alias and throws upon failure.

borrrden: There is actually a `CHECK_THROWS_WITH` function for checking that something throws with a given message.

jianminzhao: nice. go find it.

borrrden (approved):

pasin (approved):

PR #1125, opened 2021-03-01
More-reliable poll-waits in tests: CHECK_BEFORE(), etc. 

snej:
I recommend reviewing the two commits below separately. The second one has no actual effect on the code, it's just syntactic sugar.

Some tests are using delays to wait for events to occur, and either
the delay is too short to be reliable or it's too long and slows down
the test. Instead we should use loops with reasonably long timeouts
that exit as soon as the event occurs.

* I changed the relatively new `WaitUntil` function so it doesn't
  itself trigger test failure, just returns a boolean.
* Added CHECK_BEFORE() and REQUIRE_BEFORE() macros that act like
  CHECK and REQUIRE, for boolean conditions, but also take a maximum
  time interval to wait for the condition to become true.
* Updated tests that were using their own `this_thread::sleep_for`
  calls, to use these new tools instead.

With free bonus commit to adopt C++14 time-duration literal syntax ("200ms").

### DIFF
--- a/C/include/c4.hh
+++ b/C/include/c4.hh
@@ -103,6 +103,12 @@ namespace c4 {
         T* _obj;
     };
 
+
+    /// Convenience function for wrapping a new C4 object in a ref:
+    template <class T>
+    ref<T> make_ref(T *t) { return ref<T>(t); }
+
+
     /// Returns a description of a C4Error as a _temporary_ C string, for use in logging.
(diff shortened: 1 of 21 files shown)


borrrden (approved):
Beautiful!  I love it!

jianminzhao (approved):

PR #1126, opened 2021-03-03
We require db (alias) qualifiers for properties if there are multiple…

jianminzhao:
… aliases. However, the result alias should not contribute the decision of the alias count. #CBL-1668

### DIFF
--- a/LiteCore/Query/QueryParser.cc
+++ b/LiteCore/Query/QueryParser.cc
@@ -1343,12 +1343,22 @@ namespace litecore {
             iType = _aliases.find(alias);
         }
 
+        bool hasMultiDbAliases = false;
+        if (_aliases.size() > 1) {
+            int cnt = 0;
+            for (auto it = this->_aliases.begin(); it != this->_aliases.end(); ++it) {
+                if (it->second != kResultAlias) {
+                    if (++cnt == 2) {
+                        hasMultiDbAliases = true;
+                        break;
+                    }
+                }
+            }
+        }
         if (_propertiesUseSourcePrefix && !property.empty()) {
             // Interpret the first component of the property as a db alias:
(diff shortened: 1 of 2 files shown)


LiteCore/Query/QueryParser.cc
```diff
@@ -1346,7 +1346,9 @@ namespace litecore {
         if (_propertiesUseSourcePrefix && !property.empty()) {
             // Interpret the first component of the property as a db alias:
             require(property[0].isKey(), "Property path can't start with array index");
-            if (_aliases.size() > 1 || alias == _dbAlias) {
+            if (count_if(_aliases.begin(), _aliases.end(), [](decltype(_aliases)::value_type it) {
+                return  it.second != kResultAlias;
+            }) > 1 || alias == _dbAlias) {
```
borrrden, line 1351: Ideally this would have the "simple check" first so it becomes ever so slightly faster.  Also if a const ref is allowed for the "it" type that would be preferable (`const auto&` might work too)

borrrden (approved):
Style comments, functionally looks fine.

LiteCore/Query/QueryParser.cc
snej: Nice use of count_if, but the multi-line lambda inside an outer `if` feels too complex IMHO. I think it'd be clearer to break it up, with one statement assigning the `count_if` call to a variable, and then the `if` testing it.

And I agree with Jim's point, though right here it's less important since the query parser is not performance sensitive (though the query _runtime_ is.)

snej (changes requested):

LiteCore/Query/QueryParser.cc
jianminzhao: Good suggestion. Will refactor it.

snej (approved):

PR #1127, opened 2021-03-06
Add Meta() function to N1QL and corresponding Meta operator in JSON q…

jianminzhao:
…uery.

Current "meta.id" corresponds to "Meta().id" and "db.meta.id" correponds to "Meta(db).id". This PR is the first part of 1632. In the next PR, I will remove the curret syntax of meta properties and replace all tests of meta properties with the new new syntax.


borrrden (approved):

LiteCore/Query/QueryParser.cc
```diff
@@ -1164,30 +1166,37 @@ namespace litecore {
...
-            {"id",         "meta-id@@@"},
-            {"sequence",   "meta-sequence@@@"},
-            {"deleted",    "meta-deleted@@@"},
-            {"expiration", "meta-expiration@@@"},
-            {"revisionID", "meta-revisionID@@@"}
+            {"id",         "f1rK.b3xI2bA6" "id"},
+            {"sequence",   "f1rK.b3xI2bA6" "sequence"},
+            {"deleted",    "f1rK.b3xI2bA6" "deleted"},
+            {"expiration", "f1rK.b3xI2bA6" "expiration"},
+            {"revisionID", "f1rK.b3xI2bA6" "revisionID"}
```
borrrden, line 1173: Was there something wrong with the old way or does this add performance or something?

jianminzhao: not wrong with the old way. Basically I was looking for something that I can be sure that, ""f1rK.b3xI2bA6" was put there by me. Previously, I test "meta-key" at the beginning and "@@@" at the end. Now, I only test the beginning. Assumption here is no one will choose db alias starting by "f1rK.b3xI2bA6". 
The real improvement is that I avoided parsing JSON into a Value structure.

borrrden: While that assumption is probably fine any assumption that a user will not do something makes me nervous.  Is there a way to catch the case that this happens?  Perhaps even hard-coding in this value to be rejected as an alias?

jianminzhao: This is the story.
"Meta(alias)" will be translated to '["meta", "alias"]'. Users don't spell out all the meta keys. 
Then I played a trick, and translating it to a dictionary of '{"id": ["meta", "f1rK.b3xI2bA6idalias"], "deleted": ["meta", "f1rK.b3xI2bA6deletedalias"], ...}'. Then, by the current parser machine, the "meta" operator will be called again. At that time, I write to sql stream the function to get the meta key from sql table.
Apparently, I made an assumption here that users will never choose database alias like "f1rK.b3xI2bA6idalias". Basically, I overloaded operator, ["meta", ...]
To avoid the assumption, I can make a new internal operator, "meta-internal". This will make things crystal clear. Would you recommend this route?

LiteCore/Query/N1QL_Parser/n1ql.leg
```diff
@@ -315,7 +315,10 @@ propertyName
 
 
 function =
-    f:functionName e:parenExprs         { $$ = insertAny(e, 0, f.as<string>() + "()");}
+    "meta"i _ '(' _                    { f = op("meta");}
```
snej, line 318: Please use `META` instead. We try to declare the reserved words only once.

LiteCore/Query/QueryParser.cc
```diff
@@ -1007,6 +1009,36 @@ namespace litecore {
...
+            slice meta_key = operands[1]->asString();
+            if (meta_key == nullptr) {
+                break;
+            }
+            if (meta_key[0] == '.') {
+                meta_key.moveStart(1);
+            }
+            writeMetaPropertyGetter(meta_key.asString(), dbAlias.asString(), true);
+            return;
+        } while (false);
```
snej, line 1040: I find this `do{...break;...}while(false)` idiom confusing. I understand how it works, but IMHO it complicates the code. Could you factor this block out into a method, e.g. `bool handleMetaPropertyGetter(op, operands)`? Then the method can `return false` where you have `break`, and `return true` at the end. Then the insertion above turns into `if (handleMetaProperty(op, operands)) return;`.

LiteCore/Query/QueryParser.cc
```diff
@@ -1007,6 +1009,36 @@ namespace litecore {
...
+                }
+            }
+            slice meta_key = operands[1]->asString();
+            if (meta_key == nullptr) {
+                break;
+            }
+            if (meta_key[0] == '.') {
+                meta_key.moveStart(1);
+            }
+            writeMetaPropertyGetter(meta_key.asString(), dbAlias.asString(), true);
```
snej, line 1038: There doesn't seem any reason to convert these into `std::string`...

LiteCore/Query/QueryParser.cc
```diff
@@ -1132,6 +1164,102 @@ namespace litecore {
...
+            _checkedDeleted = true;     // note that the query has tested _deleted
+        } else if (metaKey == "revisionID") {
+            _sql << kVersionFnName << "(" << tablePrefix << "version" << ")";
+        } else if (metaKey == "sequence") {
+            writeMetaProperty(kValueFnName, tablePrefix, metaKey.c_str());
+            _checkedExpiration = true;
+        } else if (metaKey == "expiration") {
+            writeMetaProperty(kValueFnName, tablePrefix, metaKey.c_str());
+        }
+    }
```
snej, line 1262: What happens if `metaKey` isn't any of the above legal values? Shouldn't it call `fail`?

LiteCore/Query/QueryParser.cc
```diff
@@ -1132,6 +1164,102 @@ namespace litecore {
...
+            {"deleted",    "f1rK.b3xI2bA6" "deleted"},
+            {"expiration", "f1rK.b3xI2bA6" "expiration"},
+            {"revisionID", "f1rK.b3xI2bA6" "revisionID"}
+        };
+        const unsigned kMetaKeysCount = sizeof(metaKeys) / sizeof(metaKeys[0]);
+        const char* const kMetaCryptic = "f1rK.b3xI2bA6";
+
+        Retained<MutableDict> buildMetaDict(const string& dbAlias) {
+            static Retained<MutableDict> ret;
+            if (!ret) {
```
snej, line 1180: This is not thread-safe, especially since each caller to this method mutates the contents of the static dict. Why does it need to be static at all?

LiteCore/Query/QueryParser.cc
snej: This is _really_ convoluted and it's taking me a long time to figure out. As I understand it, you're handling `["META"]` by generating a Dict on the fly that contains META expressions to get each magic property, then feeding that back into the QueryParser to generate SQL.

Why not just emit a call to `dict_of`, using `writeMetaPropertyGetter` to write each of the values?

snej (changes requested):

LiteCore/Query/N1QL_Parser/n1ql.leg
jianminzhao: I tried to make a point that, like all other function names, "meta" is not a reserved word. In fact, function names, or, IDENTITY for that matter, cannot be reserved words. I feel that "meta" here is merely a function name. In fact, as the next step to remove meta property path, I planned to take META from reserved words.

LiteCore/Query/QueryParser.cc
jianminzhao: Taken. It would have to be private method, rather than a static/anonymous function.

LiteCore/Query/QueryParser.cc
jianminzhao: Right. But the second argument needs to be converted to string, according to the signature of the method. I will change the first one.

LiteCore/Query/QueryParser.cc
jianminzhao: I meant to let it go through to the main function, where it would fail, as trying to extract property from a dictionary in that the property does not exit: meta().not_a_mete_key.

LiteCore/Query/QueryParser.cc
jianminzhao: Oh, just wanted to save it for the sake of consistency. If threading is a concern, then, I cannot keep it. I am clear of the threading model here, but with concurrent threads, other parts would fail, such as _sql.

LiteCore/Query/QueryParser.cc
jianminzhao: In fact, this is only a optimization measure. "xyz" in meta().xyz does not match, we would compute meta() into a dictionary. Otherwise, the meta() won't be expanded.

LiteCore/Query/QueryParser.cc
jianminzhao: Yes, you're right: I meant to feed it back to QueryParser. This way, I can separate the dictionary part and the part to get the meta properties from the sqlite's table. And, I don't have to worry about how we turning a dictionary in query into dictionary in the result, and how to extract a field/property from the dictionary result.

LiteCore/Query/QueryParser.cc
```diff
@@ -1132,6 +1170,91 @@ namespace litecore {
...
+
+        string arg;
+        if (operands.count() > 0 && operands[0]->type() == kString) {
+            arg = operands[0]->asString();
+        }
+        
+        Path path {slice(arg+".id")};
+        const auto& dbIter = verifyDbAlias(path);
+        require(arg.empty() || arg == dbIter->first,
+                "database alias '%s' does not match a declared 'AS' alias", arg.c_str());
```
snej, line 1205: Looks like these lines could go inside the `if{}` block above, since they only apply if there is an operand.

snej (approved):
I have a minor note in a comment, but the commit is fine. It's up to you if you want to move those lines into the `if` block before committing :)

LiteCore/Query/QueryParser.cc
jianminzhao: Good advice and I will do it.

jianminzhao: Oh, wait, cannot do that. The above if{} is to get the optional argument. Regardless of the presence of the optional argument, we have to run the following piece of code.

PR #1128, opened 2021-03-08
CBL-1741: Skip FTS index tables when upgrading

borrrden:
They don't need to get the extra column, and currently fail anyway since they would need to be quoted


borrrden:
Filed CBL 1744 to track the intermittently failing test above.  It is not related to this PR

snej (approved):

PR #1129, opened 2021-03-09
fix build with clang 11

Dushistov:
After bug https://bugs.llvm.org/show_bug.cgi?id=24297
was fixed, construction like:
typedef enum name : unsigned int name;
becomes illegal.

evgeniy@15inch /tmp $ cat test.cc
typedef enum name : unsigned int name; enum name : unsigned int {
 X
};
evgeniy@15inch /tmp $
$ clang++ -std=c++17 -c test.cc
test.cc:1:19: error: non-defining declaration of enumeration with a fixed underlying type is only permitted as a standalone declaration; missing list of enumerators? [-Welaborated-enum-base]
typedef enum name : unsigned int name; enum name : unsigned int {
                  ^~~~~~~~~~~~~~
1 error generated.

### DIFF
--- a/C/c4Private.h
+++ b/C/c4Private.h
@@ -90,7 +90,7 @@ bool c4db_markSynced(C4Database *database,
 
 /** Flags produced by \ref c4db_findDocAncestors, the result of comparing a local document's
     revision(s) against the requested revID. */
-typedef C4_OPTIONS(uint8_t, C4FindDocAncestorsResultFlags) {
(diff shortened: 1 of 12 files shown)


borrrden:
I appreciate the PR, but unfortunately we will fix it in a different way.  Thanks though, this brought an important issue to the forefront!  I have filed [CBL-1749](https://issues.couchbase.com/browse/CBL-1749) to track.

PR #1136, opened 2021-03-16
Fix kDeleted is carried forward to the new rev when using version vector

pasin:
In VectorRecord::updateDocFlags, minus kDeleted from the current doc’s flag before applying with the new flags.

### DIFF
--- a/LiteCore/RevTrees/VectorRecord.cc
+++ b/LiteCore/RevTrees/VectorRecord.cc
@@ -404,8 +404,10 @@ namespace litecore {
     void VectorRecord::updateDocFlags() {
         // Take the local revision's flags, and add the Conflicted and Attachments flags
         // if any remote rev has them.
-        auto newDocFlags = _docFlags - DocumentFlags::kDeleted -
-            DocumentFlags::kConflicted - DocumentFlags::kHasAttachments;
+        auto newDocFlags = DocumentFlags::kNone;
+        if (_docFlags & DocumentFlags::kSynced)
+            newDocFlags |= DocumentFlags::kSynced;
+        
         newDocFlags = newDocFlags | _current.flags;
         for (Array::iterator i(_revisions); i; ++i) {
             Dict revDict = i.value().asDict();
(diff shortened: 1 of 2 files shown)


pasin:
Cancelled windows PR as it seems to hang.

LiteCore/RevTrees/VectorRecord.cc
```diff
@@ -404,7 +404,8 @@ namespace litecore {
     void VectorRecord::updateDocFlags() {
         // Take the local revision's flags, and add the Conflicted and Attachments flags
         // if any remote rev has them.
-        auto newDocFlags = _docFlags - DocumentFlags::kConflicted - DocumentFlags::kHasAttachments;
+        auto newDocFlags = _docFlags - DocumentFlags::kDeleted -
+            DocumentFlags::kConflicted - DocumentFlags::kHasAttachments;
```
borrrden, line 408: I really don't like this line.  I don't get how it is supposed to work in the first place.  At this point all you are doing is saving the Sync flag right?  So wouldn't this be more accurate as `_docFlags & DocumentFlags::kSynced` ?

borrrden (changes requested):

LiteCore/RevTrees/VectorRecord.cc
borrrden: (Or at least, `_docFlags & ~(DocumentFlags::kDeleted | DocumentFlags::kConflicted | DocumentFlags::kHasAttachments` )

pasin: First, I wasn't sure if will work as expected until I got to tested the behavior of using the minus operator on the enum. It produces the same result as your two suggestions. I will make change to `_docFlags & DocumentFlags::kSynced` which is more concise and understandable.

pasin: `_docFlags & DocumentFlags::kSynced` actually produces boolean result.

borrrden: Whoa...I didn't know that enums would overload the minus operator that's really wild!!  I thought this was some kind of hack gone wrong.  That changes my thoughts on the matter but still either way would be ok.

borrrden (approved):

snej (approved):

PR #1137, opened 2021-03-17
Implement toarray and toobject, #CBL-1662

jianminzhao:


LiteCore/Query/SQLiteN1QLFunctions.cc
```diff
@@ -1220,6 +1234,46 @@ namespace litecore {
...
+    //   v                    if v is an object
+    //   {}                   (the empty object), otherwise
+    static void toobject(sqlite3_context* ctx, int argc, sqlite3_value **argv) noexcept {
+        try {
+            if (isMissing(argv[0]) || isNull(argv[0]) || isObject(ctx, argv[0])) {
+                sqlite3_result_value(ctx, argv[0]);
+            } else {
+                Encoder enc;
+                enc.beginDictionary();
+                enc.endDictionary();
```
borrrden, line 1270: I wonder if we could have a const value of a pre-encoded empty dictionary instead of encoding it each time?  Although that's probably such a small gain that it won't be worth it.

borrrden (approved):

LiteCore/Query/SQLiteN1QLFunctions.cc
jianminzhao: Will do it.

PR #1138, opened 2021-03-18
Deprecate keyword, META, in N1QL, for designating the meta property i…

jianminzhao:
…n property path. #CBL-1632

We deprecate the special property paths for the meta properties, such as "meta.id", "meta.sequence", etc. in N1QL. Instead, we have a new function, meta(), that returns an object of the meta-keys and the corresponding values. To get a particular meta property, such as, "meta.id", one uses nested operator, "meta().id".
  The special syntax for meta key in JSON, such as "._id", remain valid. However, N1QL will no longer generate this style of referring to meta properties in Json.

### DIFF
--- a/LiteCore/tests/N1QLParserTest.cc
+++ b/LiteCore/tests/N1QLParserTest.cc
@@ -103,13 +103,15 @@ TEST_CASE_METHOD(N1QLParserTest, "N1QL properties", "[Query][N1QL][C]") {
     CHECK(translate("select `mr.grieves`.`hey`") == "{'WHAT':[['.mr\\\\.grieves.hey']]}");
     CHECK(translate("select `$type`") == "{'WHAT':[['.\\\\$type']]}");
 
-    CHECK(translate("select meta.id") == "{'WHAT':[['._id']]}");
-    CHECK(translate("select meta.sequence") == "{'WHAT':[['._sequence']]}");
-    CHECK(translate("select meta.deleted") == "{'WHAT':[['._deleted']]}");
-    CHECK(translate("select db.meta.id") == "{'WHAT':[['.db._id']]}");
-    CHECK(translate("select meta.bogus") == "");    // only specific meta properties allowed
-    CHECK(translate("select db.meta.bogus") == "");
-
+    CHECK(translate("select meta().id") == "{'WHAT':[['_.',['meta'],'.id']]}");
+    CHECK(translate("select meta().sequence") == "{'WHAT':[['_.',['meta'],'.sequence']]}");
+    CHECK(translate("select meta().deleted") == "{'WHAT':[['_.',['meta'],'.deleted']]}");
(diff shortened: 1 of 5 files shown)


LiteCore/tests/N1QLParserTest.cc
```diff
@@ -103,13 +103,15 @@ TEST_CASE_METHOD(N1QLParserTest, "N1QL properties", "[Query][N1QL][C]") {
...
-    CHECK(translate("select db.meta.bogus") == "");
-
+    CHECK(translate("select meta().id") == "{'WHAT':[['_.',['meta'],'.id']]}");
+    CHECK(translate("select meta().sequence") == "{'WHAT':[['_.',['meta'],'.sequence']]}");
+    CHECK(translate("select meta().deleted") == "{'WHAT':[['_.',['meta'],'.deleted']]}");
+    CHECK(translate("select meta(db).id from db") == "{'FROM':[{'AS':'db'}],'WHAT':[['_.',['meta','db'],'.id']]}");
+    {
+        ExpectingExceptions x;
+        CHECK_THROWS_WITH(translate("select meta().bogus"), "'bogus' is not a valid Meta key");
+        CHECK_THROWS_WITH(translate("select meta(db).bogus from db"), "'bogus' is not a valid Meta key");
```
borrrden, line 113: Is this a change in behavior with exceptions?

jianminzhao: Yes. In the deprecated method, we check the meta key in LEG parser. The error shows as empty string. In the new method, we check it in QueryParser. "meta().bogus" is valid w.r.t. LEG.

borrrden: I think it should be ok as long as we throw exceptions elsewhere.  I'm concerned about how they will present themselves.  They would need to make it up to the C API boundary and get trapped into a `C4Error`

jianminzhao: It behaves same as this one, 
    // "custId" is implicitly scoped by the unique alias, "orders".
    CHECK(translate("SELECT DISTINCT custId FROM orders where test_id = 'agg_func' ORDER BY custId") ==
          "{'DISTINCT':true,'FROM':[{'AS':'orders'}],'ORDER_BY':[['.custId']],"
          "'WHAT':[['.custId']],'WHERE':['=',['.test_id'],'agg_func']}");
    {
        ExpectingExceptions x;
        CHECK_THROWS_WITH(translate("SELECT custId, other.custId FROM orders JOIN orders other "
                                    "ON orders.test_id = other.test_id ORDER BY custId"),
                          "property 'custId.' does not begin with a declared 'AS' alias");
    }

jianminzhao: There are numerous occasions that QueryParser will throw. It's a question we would have to address.

borrrden: If the query parser already throws then it is likely that it does the right thing already.  I imagine it would bubble into a call to either `c4query_run` or `c4query_new` and be captured there.

borrrden (approved):

PR #1141, opened 2021-03-23
Tell clang to ignore a GCC pragma

bmeike:


borrrden:
Going to close this for now so that we can try again later when the issue gets rescheduled.

PR #1143, opened 2021-03-23
Export c4doc_getRevisionBody(), and fix bug therein

snej:
The bug is that the function would crash (throw an exception out of a noexcept method) if the doc's body wasn't loaded; but only in a version-vector doc. The unit test made this happen.


bmeike (approved):
I'll buy it.

PR #1144, opened 2021-03-24
CBL-1749: Redo C4_ENUM and C4_OPTIONS

borrrden:
clang has changed its way of doing compilation and no longer allows the previous usage of C4_ENUM, so update the definition to match what Apple did since that is what the implementation is based on.


borrrden:
Note: Fixes the problem addressed in #1129

bmeike:
I confirm that this change allows me to update to the latest NDK.  Thanks!

snej (approved):

PR #1146, opened 2021-03-25
N1QL should take interger of 64-bits. #CBL-1657

jianminzhao:
Also fixed a bug in fleece, setInt.


borrrden (approved):
I want to make sure @snej did not have any special intentions with the int cast in the Fleece change.  Barring that this is good.

pasin (approved):

PR #1147, opened 2021-03-27
 CBL-1714: [API] Provide platform independent equivalents of retryable POSIX codes

borrrden:
API CHANGE: If any platforms are sending platform specific errno, this needs to be changed to use the platform independent values in C4PosixErrorCode

Static analysis was performed to track down internal usages of errno and replace it with the new platform independent values.


pasin:
If I understand this correctly, we also need to to make the standardized posix error codes public at CBL API level?

pasin (approved):

LiteCore/Support/Error.cc
```diff
@@ -379,14 +450,58 @@ namespace litecore {
...
+                if(row->err == code) {
+                    return error(row->domain, row->code, _what(row->domain, row->code));
+                }
+            }
+
+            DebugAssert(false, "Invalid POSIX code for make_error: %d", code);
+            return error(error::UnexpectedError);
+        }
+
+        return error(domain, code, _what(domain, code));
```
jayahariv, line 465: why not only return this `return error::make_error(domain, code, _what(domain, code))` (without the above checks) since its already checked in the below function?

LiteCore/Support/Error.cc
```diff
@@ -379,14 +450,58 @@ namespace litecore {
...
+    }
+
+
+    error error::make_error(Domain domain, int code, const string& what) {
+        if(domain == POSIX && (code <= MinPosixErrorMinus1 || code >= MaxPosixErrorPlus1)) {
```
jayahariv, line 470: intentional?

```suggestion
        if(domain == POSIX && (code < MinPosixErrorMinus1 || code > MaxPosixErrorPlus1)) {
```

jayahariv (approved):

LiteCore/Support/Error.cc
borrrden: It needs to translate the domain and code first to get the correct error message out of `_what`

LiteCore/Support/Error.cc
borrrden: MinPosixErrorMinus1 is one before the first valid error code, and MaxPosixErrorPlus1 is one after, as indicated by their names.  Neither one is a valid error code.

jayahariv: right

Sandychuang8 (approved):

bmeike:
Sorry to be so late to this!

This PR seems to focus on an API change which would be irrelevant, because the JNI does appear to use `c4error_make`.  There's more, though, right?  We can now report errors to LiteCore in more detail, when we intercept them?

I also have the same question as Pasin's: These codes will go all the way to the client code?

bmeike (approved):
I can only comment on the concept, which is terrific.  Would love clarification on how the platforms will make use of this change, as noted in comments.

pasin:
> I also have the same question as Pasin's: These codes will go all the way to the client code?

Yes, it's defined in the spec here : https://docs.google.com/document/d/1vsTX68cxK23Q-btJotPQyOF81uytT6f67Q9l16_1EQw

LiteCore/Support/Error.cc
snej: This creates a weird dichotomy — when does one use the `error` constructor and when `make_error`? It seems like the only need for `make_error` is to convert platform to standard POSIX codes, so could we change this to `makePOSIXError` and use the constructor for everything else?

snej (commented):
It seems too messy that error objects sometimes get created with platform POSIX codes and sometimes our cross-platform codes. Can we make it so only the cross-platform codes are used, and have a special function that translates a platform error code to a cross-platform one?

borrrden:
For some reason I am unable to respond to this comment directly so:

> This creates a weird dichotomy — when does one use the error constructor and when make_error? It seems like the only need for make_error is to convert platform to standard POSIX codes, so could we change this to makePOSIXError and use the constructor for everything else?

I made the constructor private so the answer to this should be "always use make_error".  This has the side effect of protecting against any other changes we need to make later, but for this case `makePOSIXError` will work just as well.  This has the side effect of breaking whoever happens to be using regular errno now though.  Not just breaking, but the errors will all turn into what look like real errors but that make no sense for the situation since errno is in arbirtary order and errc is in alphabetical.  

> Since these POSIX constants match the ones in std::errc, why not just use that?

Because std::errc is not something that platforms like .NET are aware of.  Also I was hoping to make this change non breaking so that platforms that do go through the effort of mapping their codes to system specific POSIX codes don't immediately break.  There is also the annoyance that C4Error is just a freestanding struct and given that there is really no way to correctly interpret what is meant since errc overlaps errno unless we are forcing all system POSIX codes to be illegal and that they are actually the new ones.  If someone passed {POSIX, 2} would that be ENOENT or address_not_available?  With the offset numbering it is non-ambiguous.  Also, the platforms will need to be aware that in the case of "not found" they should use LiteCore instead of POSIX

> I don't like that these are in the same namespace (the error class) as the LiteCore specific codes

I could move them somewhere else for sure but I need something for the C# bindings script to work off of so it can't be just std::errc.

The TL;DR is "this is a way that works without breaking everything that exists now" and "let's always use make_error so that we can control what comes out just in case there are places that are still using errno by mistake" (It might be tricky to find because it would show up in some cases as just a nonsensical log message).

borrrden:
As an addendum we *could* use the errc mapping at the C++ level if we make the change non-compatible, but we'd still need a binding at the C level.

borrrden:
I've also just discovered that the values of std::errc vary just as wildly as the values of errno so it's become a nonstarter to use it directly...

borrrden:
Alright, this is the second attempt.  No more compatibility this will straight up just force the use of the new enums and any use of errno will be considered undefined behavior.

LiteCore/Support/Error.cc
```diff
@@ -63,8 +63,19 @@ namespace litecore {
         char buf[1024];
         buf[0] = '\x0';
         FormatMessageA(FORMAT_MESSAGE_FROM_SYSTEM | FORMAT_MESSAGE_IGNORE_INSERTS,
-			nullptr, err, MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT),
-			buf, sizeof(buf), nullptr);
+            nullptr, err, MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT),
+            buf, sizeof(buf), nullptr);
+
+        char* c = buf;
+        while(c) {
```
snej, line 70: That should be `while (*c)`. The current code will crash if the message doesn't contain a CR; is it untested, or do the tests just get lucky?

snej (commented):
This is a lot of stuff to add ... at this point it smells to me like it's a bad idea to expose POSIX-like error codes at all. 

Maybe it'd be better to just define our own error codes for errors that are relevant, and map POSIX errors to those. (Not that I'm saying we should abandon this PR; just something to think about.)

LiteCore/Support/Error.cc
borrrden: The tests got lucky!  I put this in because of an awkward newline in the messages.  Nice catch!

borrrden:
You are right that we don't necessarily need all of these codes to make this work for the purposes we need.  The minimal set of changes would be to change the network dependent and transient errors to something consistent between platforms.  Perhaps they could be added to the Network domain?

borrrden:
I've refactored this PR so now all the platforms need to do is the following:

Add the new network domain error codes to the bindings, if necessary, and use them in lieu of platform independent POSIX codes when a retry is desired.  Let me know if there are any codes that are not transient that would be nice to have as well and we can consider them.

C/include/c4Error.h
```diff
@@ -108,6 +112,17 @@ typedef C4_ENUM(int32_t, C4NetworkErrorCode) {
     kC4NetErrUnknown,                   // Unknown error
     kC4NetErrTLSCertRevoked,            // Peer's cert has been revoked
     kC4NetErrTLSCertNameMismatch,       // Peer's cert's Common Name doesn't match hostname
+    kC4NetErrNetworkReset,              // The network subsystem was reset [ENETRESET, retryable]
+    kC4NetErrConnectionAborted,         // The connection was aborted by the OS [ECONNABORTED, retryable]
+    kC4NetErrConnectionReset,           // The connection was reset by the other side [ECONNRESET, retryable]
```
Sandychuang8, line 117: Was ECONNRESET (Linux) not a retriable error before this PR?

borrrden: The retryable error list has not changed logically, these errors are just mirroring the POSIX errors in question.

PR #1148, opened 2021-03-29
Replace FTS MATCH operator with MATCH function, #CBL-1552

jianminzhao:


borrrden:
Note: This requires a behavioral change on the platform side wrt the JSON syntax change (for the QueryBuilder), so all platforms should be tagged.

LiteCore/Query/QueryParser.cc
```diff
@@ -1307,6 +1307,11 @@ namespace litecore {
         require(arity <= spec->maxArgs || spec->maxArgs >= 9,
                 "Too many arguments for function '%.*s'", SPLAT(op));
 
+        if (spec->name == "match"_sl) {
+            matchOp(op, operands);
+            return;
+        }
```
borrrden, line 1313: What makes this a special case that cannot be handled by the below if clause?

jianminzhao: Basically, the following code is to handle function op generically. Namely, it would write 
   'match' '(' argument-list ')'
and the function name, namely 'match', would be registered with sqlite. At "runtime," the function gets called, etc. "Match" as we use it here, is more for syntactic reason to take the form of function, but it is more like operator in JSON query, that we further digest it into SQL. We've already have it as operator. So, I just route to there.
In short, if we don't care of the json form, I don't have touch QueryParser.cc at all.

borrrden: If there is a way to avoid this I would prefer to do so, since it just seems like adding technical debt.  I guess the "best" method would be to rewrite matchOp as a match function and register it along with the other SQLite functions.  Leaving it as an operator in JSON would be 2nd best.

borrrden: I see, so the function will map to a SQLite operator.  We should keep it that way then.  Please put in a comment to explain why this is a special case, since I know at least I will surely forget why this is here.

borrrden (approved):

PR #1149, opened 2021-03-30
Fixed uncaught-exception crash in c4db_openNamed etc.

snej:
c4db_openNamed, c4db_exists, c4db_deleteNamed would crash with an uncaught C++ exception if the input database name or parentDirectory were a NULL slice. That's because the dbPath() subroutine would fail an assertion, and so throw an exception, but the call to dbPath() was not inside a try/catch block. Fixed.

### DIFF
--- a/C/c4Database.cc
+++ b/C/c4Database.cc
@@ -58,13 +58,13 @@ C4Database::~C4Database() {
 
 
 static FilePath dbPath(C4String name, C4String parentDir) {
-    Assert(name.buf != nullptr && parentDir.buf != nullptr);
+    Assert(name.buf != nullptr && parentDir.buf != nullptr);        // throws!
     return FilePath(string(slice(parentDir)), string(slice(name)))
                 .addingExtension(kC4DatabaseFilenameExtension);
 }
 
 
-static bool ensureConfigDirExists(const C4DatabaseConfig2 *config, C4Error *outError) {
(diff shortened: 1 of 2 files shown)


bmeike (approved):
Looks great to me...

C/c4Database.cc
```diff
@@ -115,11 +117,11 @@ C4Database* c4db_openNamed(C4String name,
...
                            C4Error *outError) C4API
 {
-    if (!ensureConfigDirExists(config, outError))
-        return nullptr;
-    FilePath path = dbPath(name, config->parentDirectory);
-    C4DatabaseConfig oldConfig = newToOldConfig(config);
-    return tryCatch<C4Database*>(outError, [=] {
+    return tryCatch<C4Database*>(outError, [=]() -> C4Database* {
+        if (!ensureConfigDirExists(config, outError))
+            return nullptr;
```
borrrden, line 122: Why was this moved inside the try catch if in the same PR you also added "noexcept" to the signature of ensureConfigDirExists?

snej: The important fix is that the call to `dbPath` is now inside the try/catch block.

I added `noexcept` to ensureConfigDirExists just to clarify that it can't throw an exception (because it's surrounded in a try/catch block.)

borrrden: Oh I didn't notice that it just wasn't there at all before as opposed to the other block where try catch was modified (copyNamed).  Nevermind.

borrrden (approved):

PR #1154, opened 2021-04-06
Make JSON query to look the same as the N1QL in terms of operator vs …

jianminzhao:
…function. #CBL-1823

['IS_VALUED()', ...] --> ['IS VALUED', ...]
['META', ...] --> ['META()', ...]


borrrden:
This is another change that is going to affect the platforms!  Is this necessary?

jianminzhao:
These two functions/operators are all new.

borrrden:
There is currently a way to get meta data in the JSON query syntax though.  Doesn't this change it?

jianminzhao:
No. I did not remove the existing method to address the meta properties, like "_id". @snej was suggesting to keep existing method intact for the time being. I guess there would be many meta things in the current query builder and we need to cooperate to remove old method.

borrrden (approved):

PR #1156, opened 2021-04-09
Return null-padded C4StringResults for error messages

snej:
The slice doesn't contain the 0 byte; rather, the 0 is one byte past the end. But it's a valid C string including that byte.

Also changed C4SliceResult to C4StringResult in some APIs, which has no effect on code but clarifies that these slices contain text, not binary data.

### DIFF
--- a/C/c4Base.cc
+++ b/C/c4Base.cc
@@ -79,7 +79,7 @@ static string getBuildInfo() {
 
 
 C4StringResult c4_getBuildInfo() C4API {
-    return sliceResult(getBuildInfo());
+    return stringResult(getBuildInfo());
 }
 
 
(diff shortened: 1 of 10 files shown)


borrrden:
@pasin I remember an awkward reason for making this fix but I can't remember what it was.  Did it have something to do with C?

pasin:
IIRC We had some discussion in slack about using CBL-C API that returns String as Slice but the Slice doesn't have null padding so it needs to be copied.

borrrden:
I'm interested in finding out where this is used.

jianminzhao:
Returning nullPaddedString() as C4StringResult for some APIs may be confusing. After consulting with Jens, I close it.

PR #1157, opened 2021-04-09
CBL-1723: Only change progress level via options if explicitly set

borrrden:
Otherwise it will mistakenly override an explicitly set value via c4repl_setProgressLevel


bmeike (approved):
👍🏻

PR #1161, opened 2021-04-15
Handle miscellaneous edge cases. #CBL-1840

jianminzhao:
1. Functions propagate MISSING and NULL: functions return MISSING if any argument is MISSING, otherwise they return NULL if any argument is NULL.
2. Functions return NULL if any argument is out of the domain, instead of emitting error. "lower([1,2])," for example, returns NULL instead of erring out the query process.
3. Removed "regex" from reserved words.
4. Division operator, "/", performs floating number division even if both operands are integers.
5. The column title of array path, such as "orderLines[0]" becomes "$n", instead of "*".

### DIFF
--- a/LiteCore/Query/QueryParser.cc
+++ b/LiteCore/Query/QueryParser.cc
@@ -745,12 +745,14 @@ namespace litecore {
 
     static string columnTitleFromProperty(const Path &property, bool useAlias) {
         if (property.empty())
-            return "*";
+            return "*"; // for the property ".", i.e. the entire doc
         string first(property[0].keyStr());
         if (first[0] == '_') {
             return first.substr(1);         // meta property
-        } else {
+        } else if (property[property.size()-1].keyStr()) {
             return string(property[property.size()-1].keyStr());
+        } else {
+            return {};
         }
     }
 
@@ -792,11 +794,9 @@ namespace litecore {
                     title = columnTitleFromProperty(Path(result->asString()), _propertiesUseSourcePrefix);
                 } else if (result->type() == kArray && expr[0]->asString().hasPrefix('.')) {
                     title = columnTitleFromProperty(propertyFromNode(result), _propertiesUseSourcePrefix);
-                } else {
-                    title = format("$%u", ++anonCount); // default for non-properties
                 }
                 if (title.empty())
-                    title = "*";        // for the property ".", i.e. the entire doc
+                    title = format("$%u", ++anonCount); // default for non-properties
             }
 
             // Make the title unique:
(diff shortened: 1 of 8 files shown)


LiteCore/Query/SQLiteN1QLFunctions.cc
```diff
@@ -639,16 +661,30 @@ namespace litecore {
 
     // length() returns the length in characters of a string.
     static void length(sqlite3_context* ctx, int argc, sqlite3_value **argv) noexcept {
+        if (sqlite3_value* mnArg = passMissingOrNull(1, argv); mnArg != nullptr) {
```
snej, line 664: Passing `argc` instead of `1` seems safer, since it won't break if we change the number of arguments of this function, or if copying/pasting code into a new function definition.

(This goes for many of the diffs below too)

snej (changes requested):

LiteCore/Query/SQLiteN1QLFunctions.cc
jianminzhao: Accepted: going to find all occurrences and change them to argc

LiteCore/Query/QueryParser.cc
```diff
@@ -745,12 +745,14 @@ namespace litecore {
 
     static string columnTitleFromProperty(const Path &property, bool useAlias) {
         if (property.empty())
-            return "*";
+            return "*"; // for the property ".", i.e. the entire doc
```
borrrden, line 748: This should be eventually translated to the name of the data source (i.e. `select * from orders` should be `orders` not `*`)  Has this been checked?

borrrden (changes requested):
Just put in the missing header on Linux and it should be ok.

LiteCore/Query/QueryParser.cc
jianminzhao: Not yet. I thought about it but balked for this talked. It is little tricky for Mobile N1QL, because we don't require the from clause. The default is "_doc", little ugly. Perhaps it could "default," or the like. Perhaps a separate ticket?

borrrden: I thought we decided to make the FROM clause mandatory.  It cannot just be "default", it should mirror what server does.

borrrden:
I removed myself from the reviewer list since I have nothing to add since last time.

LiteCore/Query/QueryParser.cc
jianminzhao: Not reflected in the code yet. This will invalidate many test cases.

snej (approved):
Thx!

LiteCore/Query/QueryParser.cc
snej: Especially important to do this once we support collections.
For now on master, add the default alias.

jianminzhao: Got it. Doing it now.

borrrden (approved):

PR #1164, opened 2021-04-21
Add two functions to n1ql, DIV, IDIV, #CBL-1858

jianminzhao:
DIV(x, y)  = (double)x / (double)y,
IDIV(x, y) = (int64_t)x / (int64_t)y
We now translate traditional "/" to the new function DIV, so "select 2/0" no longer return MISSING, instead of NULL, because we are taking control of return.

### DIFF
--- a/LiteCore/Query/SQLiteN1QLFunctions.cc
+++ b/LiteCore/Query/SQLiteN1QLFunctions.cc
@@ -1006,7 +1006,7 @@ namespace litecore {
     }
 
     static void fl_div(sqlite3_context* ctx, int argc, sqlite3_value **argv) {
-        if (sqlite3_value* mnArg = passMissingOrNull(1, argv); mnArg != nullptr) {
+        if (sqlite3_value* mnArg = passMissingOrNull(argc, argv); mnArg != nullptr) {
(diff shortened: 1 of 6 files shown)


LiteCore/Query/SQLiteN1QLFunctions.cc
```diff
@@ -1005,6 +1005,40 @@ namespace litecore {
         sqlite3_result_int(ctx, num > 0 ? 1 : (num < 0 ? -1 : 0) );
     }
 
+    static void fl_div(sqlite3_context* ctx, int argc, sqlite3_value **argv) {
+        if (sqlite3_value* mnArg = passMissingOrNull(1, argv); mnArg != nullptr) {
```
borrrden, line 1009: Follow suit with others and use argc here?  Unlikely to change, but still :)

borrrden: Actually isn't it supposed to be 2, not 1?

jianminzhao: Copy/paste mistake. Will change it, and another place.

borrrden (approved):

PR #1165, opened 2021-04-21
Fix MSVC build issue regarding copy elision

borrrden:
There is a bug in the MSVC 15.x compiler that Microsoft will not fix that prevents the use of guaranteed copy elision with deleted move constructors in some cases.  Use a workaround until upgrade to 16.x.

### DIFF
--- a/C/c4Certificate.cc
+++ b/C/c4Certificate.cc
@@ -28,16 +28,11 @@
 #include <vector>
 
 #if DEBUG
-#ifndef ENABLE_CERT_REQUEST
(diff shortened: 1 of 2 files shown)


borrrden:
Flaky tests are flaky....hard to keep track of which got fixed in which branch.  Need to merge the dev branch asap.

jianminzhao (approved):

pasin (approved):

snej (changes requested):
I'd prefer to use ifdefs to restrict this change to MSVC; also, could you add a comment describing the reason for the workaround?

PR #1166, opened 2021-04-22
Only add cert request support if asked for

borrrden:
Otherwise consuming libraries have an implicit dependency on LiteCoreREST that goes unrealized.


borrrden:
macOS failure:  @snej Continuous push from both sides is failing a lot  more often recently.  Is it something that needs some more `WAIT_UNTIL` attention?

UWP failure:  Addressed in another PR.

snej (approved):

PR #1169, opened 2021-04-27
mbedTLS optimization / size reduction

snej:
* The Xcode project was building mbedTLS with the Debug configuration,
  not RelWithDebInfo, even in Release builds. Oops.
* Turned off some mbedTLS flags that save about 80KB of code by
  disabling old/deprecated TLS versions and features we don't use.
  (I also tried turning off debug logging, which saves another 80KB,
  but I figured the logging can be useful for troubleshooting in the
  field so the savings aren't worth it.)


snej:
"unused local" warning/error from MSVC in an mbed function, caused by disabling MBEDTLS_SSL_PROTO_DTLS. I've pushed a fix to mbedTLS and updated the submodule.

snej:
macOS tests failed in "Continuous Push From Both Sides", though it's hard to read because the Catch logging is interspersed in stuff the replicator is spewing.

```
/Users/runner/work/couchbase-lite-core/couchbase-lite-core/Replicator/tests/ReplicatorLoopbackTest.cc:1307
/Users/runner/work/couchbase-lite-core/couchbase-lite-core/LiteCore/tests/../../Replicator/tests/ReplicatorLoopbackTest.hh:139: FAILED:
   CHECK( asVector(_docPushErrors) == asVector(_expectedDocPushErrors) )
 with expansion:
   {  } == { "doc" }
```

I've seen this a few times -- it's failing because the test managed to finish without the two sides conflicting with each other. I think this is a race condition that for some reason occurs more often with that latest replicator changes. I should keep it from failing the test.

borrrden (approved):

jianminzhao (approved):

borrrden:
@jianminzhao Can this be merged?

jianminzhao:
yes

PR #1170, opened 2021-04-28
Fix flaky "Continuous Push From Both Sides" test

snej:
The test expects there to be an error pushing its one document.
Most of the time there is, but if the two replicators interleave
their changes, the test can finish with no conflicts. For whatever
reason this has become more likely of late -- I like to think it's
because of my replicator and CRUD optimizations, but who knows.

This patch adds a new boolean flag to ReplicatorLoopbackTest that,
when set, makes it OK for there to be no document errors even if
the expected-errors sets are non-empty. This flag is set in the one
test that exhibits this problem.

### DIFF
--- a/Replicator/tests/ReplicatorLoopbackTest.hh
+++ b/Replicator/tests/ReplicatorLoopbackTest.hh
@@ -135,8 +135,10 @@ public:
         CHECK(_statusReceived.error.code == _expectedError.code);
         if (_expectedError.code)
             CHECK(_statusReceived.error.domain == _expectedError.domain);
-        CHECK(asVector(_docPullErrors) == asVector(_expectedDocPullErrors));
-        CHECK(asVector(_docPushErrors) == asVector(_expectedDocPushErrors));
+        if (!(_ignoreLackOfDocErrors &&_docPullErrors.empty()))
+            CHECK(asVector(_docPullErrors) == asVector(_expectedDocPullErrors));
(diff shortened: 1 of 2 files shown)


Replicator/tests/ReplicatorLoopbackTest.hh
```diff
@@ -135,8 +135,10 @@ public:
         CHECK(_statusReceived.error.code == _expectedError.code);
         if (_expectedError.code)
             CHECK(_statusReceived.error.domain == _expectedError.domain);
-        CHECK(asVector(_docPullErrors) == asVector(_expectedDocPullErrors));
-        CHECK(asVector(_docPushErrors) == asVector(_expectedDocPushErrors));
+        if (!(_ignoreLackOfDocErrors &&_docPullErrors.empty()))
+            CHECK(asVector(_docPullErrors) == asVector(_expectedDocPullErrors));
```
borrrden, line 139: This seems to me that there is now no reason to set the expected doc errors and ALSO set ignore errors since ignore errors will just not check at all.  It seems to me like it should check that the doc errors are either empty, or match the expected ones.

borrrden (changes requested):

Replicator/tests/ReplicatorLoopbackTest.hh
borrrden: Nevermind I misread this. I'm going to merge it since this test is a constant bother.

borrrden (approved):

PR #1171, opened 2021-05-17
The division operator, '/', in N1QL is given to SQLite3 to handle.

jianminzhao:
In 732eeef0ad9b5a8a, we changed it to DIV(), which is handled by our code and does floating number division. This commit reverts the change of 732eeef0ad9b5a8a with regard to '/'.


borrrden (approved):
This is fine as-is, but it might be more future proof to delegate to the `IDIV` function

pasin (approved):

pasin:
I think IDIV() behaves differently as it is always doing integer div but SQLite div will behave based on the parameter type.

PR #1172, opened 2021-05-18
Fixed assertion failure in c4db_openNamed etc. (CBL-1658)

snej:
Detect empty db name and fail with kC4ErrorInvalidParameter.
Formerly the empty string would be passed to the FilePath ctor, which
would interpret it as a directory, triggering a later assertion
failure.

This also applies to a couple other DB API calls that take database
names.

(Also fixed an unrelated issue I ran into during this: The `format`
parameter to C4Error::raise should be marked nullable.)

CBL-1658


borrrden (commented):
The output of the Linux tests are looking pretty ugly.  I can't imagine this being related but what in the world is going on?  I'd like to settle that first.

bmeike (approved):

PR #1174, opened 2021-05-19
maxAttempts, maxAttemptWaitTime. CBL-1941, CBL-1944, CBL-1943.

jianminzhao:
Added two replicator options,
    kC4ReplicatorOptionMaxAttempts        "maxAttempts" ///< Max number of attempts (int)
    kC4ReplicatorOptionMaxAttemptWaitTime "maxAttemptWaitTime" ///< Max delay betw attempts (secs)
They will replace "maxRetries" and "maxRetryInterval", respectively. maxAttempts == maxRetries + 1.

### DIFF
--- a/Replicator/c4RemoteReplicator.hh
+++ b/Replicator/c4RemoteReplicator.hh
@@ -27,13 +27,13 @@ namespace c4Internal {
     public:
 
         // Default maximum number of retry attempts before replications give up.
-        // These can be overridden by setting the option `kC4ReplicatorOptionMaxRetries`.
+        // These can be overridden by setting the option `kC4ReplicatorOptionMaxAttempts`, which is MaxRetryCount + 1.
         static constexpr unsigned kMaxOneShotRetryCount = 9;
         static constexpr unsigned kMaxContinuousRetryCount = UINT_MAX;
 
         // Longest possible retry delay, in seconds. The delay doubles on each failed retry
         // attempt, but pins to this value.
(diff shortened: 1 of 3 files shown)


Replicator/c4RemoteReplicator.hh
```diff
@@ -221,14 +221,20 @@ namespace c4Internal {
...
         }
 
 
         // Returns the maximum number of (failed) retry attempts.
         unsigned maxRetryCount() const {
             unsigned defaultCount = continuous() ? kMaxContinuousRetryCount : kMaxOneShotRetryCount;
-            return getIntProperty(kC4ReplicatorOptionMaxRetries, defaultCount);
+            unsigned ret = getIntProperty(kC4ReplicatorOptionMaxRetries, defaultCount);
+            ret = getIntProperty(kC4ReplicatorOptionMaxAttempts, ret + 1);
+            return ret == 0 ? defaultCount : ret - 1;
```
borrrden, line 237: Either very clever or a happy accident that a continuous replicator with neither of these properties set will end up as 0 due to overflow!

borrrden (approved):

Replicator/c4RemoteReplicator.hh
jianminzhao: Do you mean the situation when (ret + 1) overflow?

pasin: Good catch!

pasin (approved):

PR #1175, opened 2021-05-21
Add the capability of createIndex to take index expression in N1QL (i…

jianminzhao:
…n addition to JSON). #CBL-1939

### DIFF
--- a/LiteCore/Storage/KeyStore.cc
+++ b/LiteCore/Storage/KeyStore.cc
@@ -66,11 +66,11 @@ namespace litecore {
         return seq;
     }
 
-    bool KeyStore::createIndex2(slice name,
-                                slice expression,
-                                QueryLanguage queryLanguage,
-                                IndexSpec::Type type,
-                                const IndexSpec::Options* options)
+    bool KeyStore::createIndex(slice name,
+                               slice expression,
+                               QueryLanguage queryLanguage,
+                               IndexSpec::Type type,
+                               const IndexSpec::Options* options)
     {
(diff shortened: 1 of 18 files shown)


LiteCore/Storage/KeyStore.cc
```diff
@@ -66,12 +66,13 @@ namespace litecore {
         return seq;
     }
 
-    bool KeyStore::createIndex(slice name,
-                               slice expressionJSON,
-                               IndexSpec::Type type,
-                               const IndexSpec::Options* options)
+    bool KeyStore::createIndex2(slice name,
```
snej, line 69: I think I'd prefer this be named just `createIndex`, since C++ allows overloading.

snej (changes requested):
We need some more tests, too -- 
* of regular value indexes
* of using the C API.

LiteCore/Storage/KeyStore.cc
borrrden: Same for me!  This method already has some default parameters, the query language could be another so it doesn't even have to be overloaded if you add it at the end.  Overloading works too though.

borrrden (changes requested):

pasin (commented):
I don't have additional comments.

LiteCore/Storage/KeyStore.cc
jianminzhao: Changed accordingly.

snej (approved):

borrrden (approved):

PR #1178, opened 2021-06-09
Update access revocation implementation and fix issues

pasin:
* Updated the replicator protocol to CBMobile_3.
* Added enable auto-purge configuration.
* Fixed issues found while testing the implementation

### DIFF
--- a/Replicator/RevFinder.cc
+++ b/Replicator/RevFinder.cc
@@ -233,7 +233,7 @@ namespace litecore::repl {
                                                 _db->remoteDBID());
         // Look through the database response:
         unsigned itemsWritten = 0, requested = 0;
-        for (unsigned i = 0; i < nChanges; ++i) {
+        for (unsigned i = 0; i < changeIndexes.size(); ++i) {
             slice docID = docIDs[i], revID = revIDs[i];
             alloc_slice anc(std::move(ancestors[i]));
             C4FindDocAncestorsResultFlags status = anc ? (anc[0] - '0') : kRevsLocalIsOlder;
--- a/Replicator/RevFinder.cc
+++ b/Replicator/RevFinder.cc
@@ -217,6 +217,7 @@ namespace litecore::repl {
                 auto mode = (deletion < 4) ? RevocationMode::kRevokedAccess
                                            : RevocationMode::kRemovedFromChannel;
                 revoked.emplace_back(new RevToInsert(docID, mode));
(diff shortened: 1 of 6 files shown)


pasin:
PR Validation failures are not related to this PR.

Replicator/RevFinder.cc
```diff
@@ -217,6 +217,7 @@ namespace litecore::repl {
                 auto mode = (deletion < 4) ? RevocationMode::kRevokedAccess
                                            : RevocationMode::kRemovedFromChannel;
                 revoked.emplace_back(new RevToInsert(docID, mode));
+                sequences.push_back({RemoteSequence(change[0]), 0});
```
snej, line 220: This is wrong -- the `sequences` array is parallel with `docIDs` and `revIDs`. It isn't used for revoked revs.

snej, line 220: This line shouldn't be added -- the sequences array is parallel with docIDs and revIDs. It isn't used for revoked revs.

Replicator/RevFinder.cc
```diff
@@ -250,7 +251,7 @@ namespace litecore::repl {
                 if ((status & kRevsConflict) == kRevsConflict && passive()) {
                     // Passive puller refuses conflicts:
                     encoder.writeRaw("409");
-                    sequences[i].bodySize = 0;
+                    sequences[changeIndexes[i]].bodySize = 0;
```
snej, line 254: This change isn't needed, for the reason above ^

Replicator/RevFinder.cc
```diff
@@ -265,7 +266,7 @@ namespace litecore::repl {
             } else {
                 // I have an equal or newer revision; ignore this one:
                 // [Implicitly this appends a 0, but we're skipping trailing zeroes.]
-                sequences[i].bodySize = 0;
+                sequences[changeIndexes[i]].bodySize = 0;
```
snej, line 269: This change isn't needed, for the reason above ^

snej (changes requested):

Replicator/RevFinder.cc
pasin: The reason to add the sequence with body size = 0 here is to allow the checkpoint to be updated for the revoked revisions. Is there an alternative way for this?

pasin: From the Slack conversation with @snej, this solution is OK.

PR #1182, opened 2021-06-15
Fixed exception when BLIP logging set to Debug (CBL-2018)

snej:
With Debug logging, BLIPIO calls description() on ACK messages,
causing an exception when MessageOut::description() tried to read the
properties of an ACK message, which doesn't have any.

The fix is not to call _contents.getPropsAndBody() when the message's
type() is an ACK. To make this safer, I wrapped the call in a method
on MessageOut itself.

And to clean up a little technical debt, I changed the calling style
to return the two slices as a pair instead of as 'out' params.

The Fleece update is to fix a bug in slice_input that I discovered
while investigating, but which isn't actually triggered here.

Fixes CBL-2018

### DIFF
--- a/Networking/BLIP/MessageOut.cc
+++ b/Networking/BLIP/MessageOut.cc
@@ -129,8 +129,7 @@ namespace litecore { namespace blip {
 
 
     void MessageOut::dump(std::ostream& out, bool withBody) {
-        slice props, body;
-        _contents.getPropsAndBody(props, body);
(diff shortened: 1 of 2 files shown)


borrrden (approved):

borrrden:
Needs some Linux love though since it will treat unused values as errors.

snej:
I restarted the desktop Jenkins build and it passed.

PR #1183, opened 2021-06-15
Fix enable / disable auto-purge logic and add tests

pasin:
* Always use revocations=true when sending subChanges requests to get revoked/removed notification from SG.

* Ensure that documentEnded notification is sent whether the auto-purge is disabled or not.

* Call the pull filter (validation) if registered is called for the revoked rev except when the auto-purge is disabled as there is no benefit of calling the filter. Also, when calling the pull filter, generate an empty revision body to send to the filter.

* SG team confirms that the revision id will always be available for both revoked and remove notification in changes messages. Moreover, the revID is required by the pull filter. Therefore, always validate the revID along with the docID when processing the changes message.

* Add tests for testing both revoked and removed cases when the auto-purge is enable and disable.


pasin:
The PR validation failure is not related to this PR.

borrrden:
Why are we making these weird PRs that immediately fail Jenkins validation?

borrrden (approved):

pasin:
I fixed the build issue in the PR Validation by creating the corresponding branch in Lite-Core-EE Repo. The PR Validation is now failed while `Database Auto-Expiration` test on windows, which is not related to this PR. 

NOTE that this PR is the update on top of the original work from @snej, and eventually it will be merged into the dev branch or the master branch after the dev branch is merged (subject to discuss when it's ready to be merged).

pasin:
@borrrden I updated the regarding enabling/disabling revocations flag in subChanges request. Please check.

snej (approved):

borrrden:
What does this mean now for adding in listeners part way through?  It looks like now there is a disparity where the behavior is different based on whether you add the listener before start or after start...

pasin:
I have chatted with @borrrden, the behavior will be that if a user disables the auto-purge but add a document replication listener after starting the replicator (while the replicator is not IDLE), the user could miss some removed/revoked notification until the pull replicator sends the`subChanges` message again. The problem is sort of race condition but in a bigger scope. It's not perfect but at least it provides a way for SG to be performed if users don't care about the auto-purge feature at all. 

Alternatively, we need to change the API to provide a new option to `disable auto-purge with or without notification sent` OR have a new mechanism to tell SG to stop sending notification anytime. At this point, I'm not sure that it's worth pursuing any of these options.

borrrden (approved):

PR #1188, opened 2021-06-18
Skip the duplicated version vector tests. #CBL-2000

jianminzhao:
For all test cases in c4Test and ReplicatorLoopbackTest, each test case is duplicated for RevTree and VersionVector. Since VersionVector is not the release focus for now, we skip them for now. Following are the time differences on my Mac,
c++ tests: 303s vs 510s, and
c   tests: 106s vs 161s.

### DIFF
--- a/C/tests/c4Test.hh
+++ b/C/tests/c4Test.hh
@@ -165,6 +165,7 @@ class TransactionHelper {
 
 #pragma mark - C4TEST BASE CLASS:
 
+#define SkipVersionVectorTest 1
 
 /// Base test fixture class for C4 tests. Creates a new empty C4Database in its setUp method,
 /// and closes & deletes it in tearDown. Also checks for leaks of classes that are InstanceCounted.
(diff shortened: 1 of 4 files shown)


borrrden (changes requested):
Looks like the CE validation failed for valid reasons

C/tests/c4Test.hh
```diff
@@ -165,32 +165,40 @@ class TransactionHelper {
 
 #pragma mark - C4TEST BASE CLASS:
 
+#define SkipVersionVectorTest 1
```
snej, line 168: Could you please wrap this with `#ifndef SkipVersionVectorTest`? That way I can predefine it as 0 in a build setting to override this.

snej (changes requested):

C/tests/c4Test.hh
jianminzhao: sure

borrrden (approved):

borrrden (approved):

snej (approved):

PR #1189, opened 2021-06-18
In query allow "_" or db name as synonym for "_default"

snej:
As per our meeting yesterday (6/17/21). To be precise:

* In JSON, this applies to the value of `SELECT.FROM.COLLECTION`.
* In N1QL, this applies to the name after `FROM` (which translates to the above.)
* `_` is now a synonym for `_default`, i.e. the default collection
* The database name is also a synonym for `_default`, _unless_ there is a collection with that name, in which case it refers to that collection.

### Implementation notes:

- The logic above is added to `SQLiteDataFile::collectionTableName()`. 
- However, that class doesn't know the database name, only the path to the `db.sqlite3` file, so
- I added a `databaseName()` method to the `DataFile::Delegate` protocol.
- This is implemented primarily by `DatabaseImpl`, but also by `BackgroundDB` and `DataFileTestFixture`.
- The rest is unit tests.

### DIFF
--- a/C/tests/c4QueryTest.cc
+++ b/C/tests/c4QueryTest.cc
@@ -18,6 +18,7 @@
 
 #include "c4QueryTest.hh"
 #include "c4BlobStore.h"
+#include "c4Collection.h"
 #include "c4Observer.h"
 #include "StringUtil.hh"
 #include <thread>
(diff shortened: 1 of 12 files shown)


pasin:
LGTM. For the test failure, I'm wondering if it's the real issue or not. I have seen something similar to my previous PR validation as well. The issue was intermittent as restarting the build/test could make the issue go away.

pasin (approved):

PR #1190, opened 2021-06-18
Fixed: 1st param of MATCH() doesn't need a db alias

snej:
So `"FROM":[{"AS":"testdb"}],"WHERE":["MATCH()", "sentence", "words"]` is now legal again ... a regression had made it require `"testdb.sentence"`.

I improved QueryParser::FTSTableName to parse the FTS index name with the same logic as for regular property accesses, so it uses the same rules for when a DB alias is needed.


jianminzhao (approved):

borrrden (approved):

borrrden:
@jianminzhao This seems ready for merge

PR #1191, opened 2021-06-21
CBL-2019 : Support replication protocol version 2 and 3

pasin:
* To support both SG v2.x (use protocol BLIP3+CBMobile_2) and v3.0 (use protocol BLIP3+CBMobile_3), the replicator v3.0 will list both protocol versions in the "Sec-WebSocket-Protocol" request header using a command separated string format starting with the preferred protocol which is BLIP3+CBMobile_3.

* Replaced a constant kReplicatorProtocolName string with two static functions CompatibleProtocols() and ProtocolName(). The CompatibleProtocols() function is used by RESTSyncListener in EE repo to get a list of all supported protocols for accepting the proposed protocols by the replicator (client).

* Adjusted the logic in HTTPLogic and c4RemoteReplicator to support multiple protocols.

### DIFF
--- a/Replicator/Replicator.cc
+++ b/Replicator/Replicator.cc
@@ -56,20 +56,11 @@ namespace litecore { namespace repl {
         {{ WebSocketDomain, 503, 0 }, false, "The server is over capacity"_sl}
     };
 
-
-    std::vector<string> Replicator::CompatibleProtocols() {
-        static const char *ProtocolNames[] = {"+CBMobile_3", "+CBMobile_2"};
-        
-        std::vector<std::string> protocols;
-        for (int i = 0; i < sizeof(ProtocolNames)/sizeof(char*); i++)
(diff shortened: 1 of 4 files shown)


pasin:
Its related PR in EE Repo : https://github.com/couchbase/couchbase-lite-core-EE/pull/13

pasin:
The PR Validation Failure in Windows is not related to this PR.

Replicator/Replicator.cc
```diff
@@ -56,6 +57,24 @@ namespace litecore { namespace repl {
     };
 
 
+    std::vector<string> Replicator::CompatibleProtocols() {
```
borrrden, line 60: I don't think this needs to be dynamically constructed every time it is created.  I think a static C array of fully constructed names should be fine given how infrequently kWSProtocolName will change.  At least making the construction of this happen only once would be better.

pasin:
@borrrden I changed to use a constant array per suggestion.

borrrden (approved):

PR #1192, opened 2021-06-22
Better handling of document change/purge when loading revs

snej:
When a C4Document is initially loaded without its rev-tree, and the
tree needs to be loaded later, the error handling needed to be
improved. In some cases TreeDocument was just ignoring the failure to
load the rev-tree.

### DIFF
--- a/C/tests/c4DocumentTest.cc
+++ b/C/tests/c4DocumentTest.cc
@@ -701,6 +701,51 @@ N_WAY_TEST_CASE_METHOD(C4Test, "Document Delete then Update", "[Document][C]") {
 }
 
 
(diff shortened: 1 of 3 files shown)


borrrden (approved):
I don't think there is any way that the oddness of the conflict vs not found error will come up to the CBL API, but it does seem slightly confusing to get a conflict error from `c4doc_setRemoteAncestor` and `c4doc_resolveConflict`

snej:
It's MVCC. If you load a doc and call one of those functions, but by the time that function saves the doc it's been modified on disk, you get a conflict. Workaround: do it in a transaction, and don't do something dumb like intentionally purging the doc in your test (which is what those broken unit tests were doing.)

PR #1193, opened 2021-06-22
Reading doc extra shouldn't replace body

snej:
When a C4Document is loaded without its 'extra', and then a later
call causes 'extra' to be loaded, the KeyStore updates all the fields
of the Record including the 'body', swapping it out with a new
alloc_slice with the same data.

This means that after this call, the document's old body as used by
clients is no longer valid, causing nasty errors.

This commit changes Record's setters to ignore a change to 'body' or
'extra' that leaves the data the same.

### DIFF
--- a/C/tests/c4DocumentTest.cc
+++ b/C/tests/c4DocumentTest.cc
@@ -746,6 +746,30 @@ N_WAY_TEST_CASE_METHOD(C4Test, "LoadRevisions After Purge", "[Document][C]") {
 }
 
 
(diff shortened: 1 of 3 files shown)


jianminzhao (approved):

borrrden (approved):
Looks good for this case, but are there any known cases where a body is replaced with a new body of different content?  If so the same issue will occur if I understand correctly (the old body will become invalid).

snej:
No, there are not. It's always the same sequence.

PR #1198, opened 2021-06-30
CBL-2079 : Fix wrong key used in enableAutoPurge()

pasin:
Fixed a mistake that use kC4ReplicatorOptionDisableDeltas instead of kC4ReplicatorOptionAutoPurge.


Sandychuang8 (approved):
Looks like it's working. 
msg["revocations"] = "true"; is hit.

PR #1199, opened 2021-07-01
In the public query API, we require the FROM clause mandatory for the…

jianminzhao:
… N1QUL input, #CBL-2041

n1ql::parse() is not changed: the FROM clause is still optional. The mandatory FROM clause in the N1QL expression is only required in the public API, c4query_new2(). Precisely, it is enforced in SQLiteQuery.

### DIFF
--- a/LiteCore/Query/SQLiteQuery.cc
+++ b/LiteCore/Query/SQLiteQuery.cc
@@ -53,6 +53,17 @@ namespace litecore {
         kFTSOffsetsCol
     };
 
+    namespace {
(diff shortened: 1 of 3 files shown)


pasin (approved):

borrrden (commented):
Maybe I missed it but is there a test to ensure that excluding the FROM clause fails in some way?

jianminzhao:
you did not miss anything. I haven't put a test case to fail this new requirement. Let me add one.

borrrden (approved):

PR #1207, opened 2021-07-22
Catch failures to create the log files.

jianminzhao:
CBL-2077

### DIFF
--- a/LiteCore/Support/TestsCommon.cc
+++ b/LiteCore/Support/TestsCommon.cc
@@ -81,7 +81,7 @@ void InitTestLogging() {
             C4Error error;
             if (!c4log_writeToBinaryFile({kC4LogVerbose, slice(path), 16*1024, 1, false},
                                          &error)) {
-                C4WarnError("TestsCommon: Can't log to binary file");
+                C4WarnError("TestsCommon: Can't log to binary file, %s", error.description().c_str());
             }
             c4log_setCallbackLevel(kC4LogInfo);
         } else {
--- a/LiteCore/Support/TestsCommon.cc
+++ b/LiteCore/Support/TestsCommon.cc
@@ -81,7 +81,8 @@ void InitTestLogging() {
             C4Error error;
             if (!c4log_writeToBinaryFile({kC4LogVerbose, slice(path), 16*1024, 1, false},
                                          &error)) {
-                C4WarnError("TestsCommon: Can't log to binary file, %s", error.description().c_str());
+                C4WarnError("TestsCommon: Can't log to binary file, %.*s",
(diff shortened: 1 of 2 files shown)


jianminzhao:
Note to the reviewers. In the following snippet,
            if (!c4log_writeToBinaryFile({kC4LogVerbose, slice(path), 16*1024, 1, false},
                                         &error)) {
                C4WarnError("TestsCommon: Can't log to binary file, %s", error.description().c_str());
            }
there will be an error message like the following, 
07:59:09.140312| ERROR: TestsCommon: Can't log to binary file, LiteCore CantOpenFile, "File Logger fails to open files in directory /tmp/LiteCore_Tests_1626915543/binaryLogs/

pasin:
@jianminzhao PR Validation was failed. Can you check?

LiteCore/Support/TestsCommon.cc
```diff
@@ -81,7 +81,8 @@ void InitTestLogging() {
             C4Error error;
             if (!c4log_writeToBinaryFile({kC4LogVerbose, slice(path), 16*1024, 1, false},
                                          &error)) {
-                C4WarnError("TestsCommon: Can't log to binary file");
+                C4WarnError("TestsCommon: Can't log to binary file, %.*s",
```
pasin, line 84: I think we don't need "TestsCommon:".

jianminzhao: Okay, I'll take it out, but it is not added by this PR.

borrrden (commented):
There is another place that does this, in `rotateLog` and it feels like something should be done there (but not throwing an exception since it will surface in the method for writing log messages).  Any ideas on that?

jianminzhao:
Agree with @borrrden that "not throwing an exception since it will surface in the method for writing log messages" in rotateLog(), but the note the following.
1. When this type of errors occur, we cannot do much except log the situation. For instance, in TestsCommon, we simply log the error and go on.
2. The error we are concerned here is mostly due to that the directory does not exist. This situation is already caught in c4log_writeToBinaryFile(). rotateLog() merely creates new files in that directory. Adding a check here will only add values if we were able to open the file but not now. 
3. We only check the success of opening files, but there could be other failures. Since we don't have runtime measure to remedy the situation, the best we can do is to log the fact. Given that the c++ stream failure flags are sticky, we could check the flags as the files do flush. We may not know when the problem start to happen, but we know the file loggers are compromised.

pasin (commented):
Will there be any more changes coming in?

jianminzhao:
We have a problem to throw from inside c4log_xyz. I filed a bug, CBL-2212. We should revisit it after CBL-2212 is resolved.

PR #1210, opened 2021-07-26
Implemented PropertyEncryption [API]

snej:
API addition: Defined encryptable-property constants in c4Document+Fleece.h and c4Document.hh.

The PropertyEncryption class takes a document's properties as a Dict and applies property-level encryption or decryption.

Its implementation is in the EE repo: see PR https://github.com/couchbase/couchbase-lite-core-EE/pull/15. There is a stub source file here that either #includes it or defines a stub CE implementation that does nothing but return an error if an encryptable property is found.

### DIFF
--- a/Replicator/IncomingRev.cc
+++ b/Replicator/IncomingRev.cc
@@ -19,12 +19,14 @@
 #include "IncomingRev.hh"
 #include "Puller.hh"
 #include "DBAccess.hh"
+#include "PropertyEncryption.hh"
 #include "Increment.hh"
 #include "StringUtil.hh"
 #include "c4BlobStore.hh"
 #include "c4Document.hh"
 #include "Instrumentation.hh"
 #include "BLIP.hh"
+#include "fleece/Mutable.hh"
 #include <atomic>
 #include <deque>
 #include <set>
@@ -37,10 +39,6 @@ namespace litecore { namespace repl {
 
     // Docs with JSON bodies larger than this get parsed asynchronously (off the Puller thread)
     static constexpr size_t kMaxImmediateParseSize = 32 * 1024;
-
-    static inline bool jsonMightContainBlobs(slice json) {
-        return json.containsBytes("\"digest\""_sl);
-    }
 
     IncomingRev::IncomingRev(Puller *puller)
     :Worker(puller, "inc")
@@ -127,9 +125,13 @@ namespace litecore { namespace repl {
         if (_revMessage->noReply())
             _revMessage = nullptr;
 
+        _mayContainBlobs = jsonBody.containsBytes("\"digest\""_sl);
+        _mayContainEncryptedProperties = !_options.disablePropertyDecryption()
+                                         && MayContainPropertiesToDecrypt(jsonBody);
+
         // Decide whether to continue now (on the Puller thread) or asynchronously on my own:
         if (_options.pullValidator|| jsonBody.size > kMaxImmediateParseSize
-                                  || jsonMightContainBlobs(jsonBody))
+                                  || _mayContainBlobs || _mayContainEncryptedProperties)
             enqueue(FUNCTION_TO_QUEUE(IncomingRev::parseAndInsert), move(jsonBody));
         else
             parseAndInsert(move(jsonBody));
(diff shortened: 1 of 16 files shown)


pasin (approved):

Replicator/IncomingRev.cc
```diff
@@ -171,7 +173,7 @@ namespace litecore { namespace repl {
             if (!fleeceDoc)
                 err = C4Error::make(FleeceDomain, (int)encodeErr, "Incoming rev failed to encode"_sl);
 
-        } else if (_options.pullValidator || jsonMightContainBlobs(jsonBody)) {
+        } else if (_options.pullValidator || _mayContainBlobs || _mayContainEncryptedProperties) {
             // It's a delta, but we need the entire document body now because either it has to be
             // passed to the validation function, or it may contain new blobs to download.
```
borrrden, line 178: Comment here needs a small update for the new condition

Replicator/Pusher+Revs.cc
```diff
@@ -72,6 +74,22 @@ namespace litecore::repl {
...
+        MutableDict encryptedRoot;
+        if (root && MayContainPropertiesToEncrypt(doc->getRevisionBody())) {
+            logVerbose("Encrypting properties in doc '%.*s'", SPLAT(request->docID));
+            encryptedRoot = EncryptDocumentProperties(request->docID, root,
+                                                      _options.propertyEncryptor,
+                                                      _options.callbackContext,
+                                                      &c4err);
+            if (encryptedRoot)
+                root = encryptedRoot;
+            else if (c4err) {
```
borrrden, line 87: This means that `EncryptDocumentProperties` is guaranteed to throw an error right?  Is there a case in which `encryptedRoot` is `null` but there is no error?  I think that should also fail the document since no encryption took place.

Replicator/PropertyEncryption_stub.cc
```diff
@@ -0,0 +1,77 @@
...
+                                                  C4Error *outError) noexcept
+    {
+        // In CE, prevent any encryptable property from being accidentally pushed.
+        // This may happen if a database was created and used with EE, and sensitive data added,
+        // and then it's opened with a CE implementation.
+        if (outError)
+            *outError = {};
+        for (DeepIterator i(doc); i; ++i) {
+            if (i.key() == C4Document::kObjectTypeProperty) {
+                if (i.value().asString() == C4Document::kObjectType_Encryptable) {
```
borrrden, line 49: I believe this needs to check for the presence of an unencrypted value.  If there is no unencrypted value present then it can be sent normally (perhaps by returning the same value that was passed in if we are going to consider a null encryptedRoot to be invalid as suggested above?).

borrrden (changes requested):

Replicator/Pusher+Revs.cc
snej: The header answers these questions -- it's marked `noexcept`, and the `@return` line says it will return null if no encryption took place.

No encryption is not an error, since there might not be any actual encrypted properties -- MayContainPropertiesToDecrypt is a very quick check that can return false positives.

Replicator/PropertyEncryption_stub.cc
snej: I don't understand this comment. Checking for an unencrypted value is exactly what this function does. If there is none it returns null. It can't return the value passed in because it takes an immutable value and returns a mutable one.

pasin: I think @borrrden means if the value is already encrypted (containing kC4EncryptedCiphertextProperty instead of kC4EncryptableValueProperty), it is not an error (OK to push).

snej: That seemed reasonable, but when I looked at the code I realized that this never occurs. When the property is encrypted it's changed into the server schema, and vice versa. So there is never an `"@type":"EncryptedProperty"` object with a `ciphertext` property -- in fact the constant `kC4EncryptedCiphertextProperty` is never used anywhere and could be removed.

borrrden: Oh, yeah, that does seem to be the case!

borrrden (approved):

PR #1211, opened 2021-07-29
REST API, _replicate, CBL-2165

jianminzhao:
…llow the body of the POST include user name and password, and authType.

2. Fixed a hanging bug that hits one-shot replication.
3. Added 4 test cases: plain/auth x continuous/oneShot.
CBL-2165

### DIFF
--- a/REST/RESTListener+Replicate.cc
+++ b/REST/RESTListener+Replicate.cc
@@ -77,7 +77,21 @@ namespace litecore { namespace REST {
                     ((ReplicationTask*)context)->onReplStateChanged(status);
                 };
                 params.callbackContext = this;
-
+                if (_authType) {
+                    Encoder enc;
+                    enc.beginDict();
+                        enc.writeKey(C4STR(kC4ReplicatorOptionAuthentication));
+                        enc.beginDict();
+                            enc.writeKey(C4STR(kC4ReplicatorAuthType));
+                            enc.writeString(_authType);
+                            enc.writeKey(C4STR(kC4ReplicatorAuthUserName));
+                            enc.writeString(_user);
+                            enc.writeKey(C4STR(kC4ReplicatorAuthPassword));
+                            enc.writeString(_password);
+                        enc.endDict();
+                    enc.endDict();
+                    params.optionsDictFleece = AllocedDict(enc.finish()).data();
+                }
                 _repl = localDB->newReplicator(remoteAddress, remoteDbName, params);
                 _repl->start();
             } catch (...) {
@@ -190,13 +204,6 @@ namespace litecore { namespace REST {
         }
 
 
-        HTTPStatus wait() {
-            Lock lock(_mutex);
-            _cv.wait(lock, [this]{return finished();});
-            return _finalResult;
-        }
-
-
         void stop() override {
             Lock lock(_mutex);
             if (_repl) {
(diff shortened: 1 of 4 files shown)


REST/RESTListener+Replicate.cc
```diff
@@ -289,17 +311,19 @@ namespace litecore { namespace REST {
                          canceled ? "Stopped" : "No matching task");
             return;
         }
-
+        // Auth:
+        slice authType = params["authtype"].asString();
+        if (authType) {
```
borrrden, line 316: If I remember, only the basic auth type uses user and password.  If this is the only supported auth method in here then it should probably be compared accordingly.

REST/RESTListener+Replicate.cc
```diff
@@ -190,20 +204,26 @@ namespace litecore { namespace REST {
...
         }
+        
+
+        void setStopOnIdle(bool stopOnIdle) {
+            _stopOnIdle = stopOnIdle;
+        }
+
+        
+        void setAuth(slice authType, slice user, slice psw) {
+            _authType = authType;
```
borrrden, line 222: It doesn't seem like this is going to be of any use, since it looks to assume basic auth (as opposed to token)

REST/RESTListener+Replicate.cc
```diff
@@ -77,7 +77,21 @@ namespace litecore { namespace REST {
                     ((ReplicationTask*)context)->onReplStateChanged(status);
                 };
                 params.callbackContext = this;
-
+                if (_authType) {
```
borrrden, line 80: See other comment about auth type.

borrrden (changes requested):

REST/RESTListener+Replicate.cc
jianminzhao: Got it, will test it against "Basic". What to do with the else part: ignore and log a runtime warning?

REST/RESTListener+Replicate.cc
jianminzhao: Maybe keep it for the possible extension in the future?

REST/RESTListener+Replicate.cc
jianminzhao: yes, replied earlier. Asked advice of dealing with the else part.

REST/RESTListener+Replicate.cc
```diff
@@ -190,20 +204,26 @@ namespace litecore { namespace REST {
...
         void stop() override {
             Lock lock(_mutex);
             if (_repl) {
                 c4log(ListenerLog, kC4LogInfo, "Replicator task #%u stopping...", taskID());
                 _repl->stop();
             }
         }
+        
+
+        void setStopOnIdle(bool stopOnIdle) {
```
snej, line 216: We do not need a "stop on idle" mode. That's identical to a one-shot replication.

REST/RESTListener+Replicate.cc
snej: If `authType` is anything other than `"Basic"`, return a 401 status.

REST/RESTListener+Replicate.cc
snej: Return a 401 status.

REST/RESTListener+Replicate.cc
```diff
@@ -289,17 +311,19 @@ namespace litecore { namespace REST {
...
+            slice psw = params["password"].asString();
+            task->setAuth(authType, user, psw);
+        }
+        task->setStopOnIdle(params["stopOnIdle"].asBool());
         task->start(localDB, localName,
                     remoteAddress, remoteDbName,
                     pushMode, pullMode);
 
         HTTPStatus statusCode = HTTPStatus::OK;
-        if (!continuous) {
```
snej, line 298: One-shot replication requests need to be synchronous, because that's the way the REST API is defined. This code is necessary for that. Don't remove it.

snej (changes requested):

jianminzhao:
We will break this ticket into several ones.

PR #1212, opened 2021-07-30
REST _replicate, CBL-2208

jianminzhao:
1) Added capability of Basic auth support.
2) Added 4 test cases, {continuous, one-shot} x {auth, no-auth}.
However, One-shot is not working yet. They will be handled by separate bugs.

### DIFF
--- a/REST/RESTListener+Replicate.cc
+++ b/REST/RESTListener+Replicate.cc
@@ -77,18 +77,13 @@ namespace litecore { namespace REST {
                     ((ReplicationTask*)context)->onReplStateChanged(status);
                 };
                 params.callbackContext = this;
-                if (_authType) {
-                    if (!_authType.caseEquivalent("Basic")) {
-                        // Currently, we only support authtype of "Basic"
-                        // Give a warning here.
-                        c4log(ListenerLog, kC4LogWarning, "Sync-Gateway Server may only support \"Basic\" authtype. Your input is %*s", SPLAT(_authType));
-                    }
+                if (_user) {
                     Encoder enc;
                     enc.beginDict();
                         enc.writeKey(C4STR(kC4ReplicatorOptionAuthentication));
                         enc.beginDict();
                             enc.writeKey(C4STR(kC4ReplicatorAuthType));
-                            enc.writeString(_authType);
+                            enc.writeString("Basic");
                             enc.writeKey(C4STR(kC4ReplicatorAuthUserName));
                             enc.writeString(_user);
                             enc.writeKey(C4STR(kC4ReplicatorAuthPassword));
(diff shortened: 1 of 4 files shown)


REST/RESTListener+Replicate.cc
```diff
@@ -77,7 +77,26 @@ namespace litecore { namespace REST {
                     ((ReplicationTask*)context)->onReplStateChanged(status);
                 };
                 params.callbackContext = this;
-
+                if (_authType) {
+                    if (!_authType.caseEquivalent("Basic")) {
+                        // Currently, we only support authtype of "Basic"
+                        // Give a warning here.
+                        c4log(ListenerLog, kC4LogWarning, "Sync-Gateway Server may only support \"Basic\" authtype. Your input is %*s", SPLAT(_authType));
```
snej, line 84: This should return a 401 status.

REST/RESTListener+Replicate.cc
```diff
@@ -289,7 +317,13 @@ namespace litecore { namespace REST {
                          canceled ? "Stopped" : "No matching task");
             return;
         }
-
+        // Auth:
+        slice authType = params["authtype"].asString();
```
snej, line 321: This isn't compatible with the [CouchDB REST API](https://docs.couchdb.org/en/stable/api/server/common.html#api-server-replicate). Unfortunately the API docs don't describe how to provide credentials; it's been years since I worked with it but IIRC they're either encoded in the URL (which our URL parser doesn't handle), or else the `source` and `target` can be a dict instead of a string and the credentials are keys in that dict.

At this level I'm OK with not going by the spec. But don't include the `authtype` parameter, since all you're handling is Basic auth anyway. Just look for the presence of `user`.

snej (changes requested):

REST/RESTListener+Replicate.cc
jianminzhao: Okay, take it.

REST/RESTListener+Replicate.cc
```diff
@@ -77,18 +77,13 @@ namespace litecore { namespace REST {
...
-                        c4log(ListenerLog, kC4LogWarning, "Sync-Gateway Server may only support \"Basic\" authtype. Your input is %*s", SPLAT(_authType));
-                    }
+                if (_user) {
                     Encoder enc;
                     enc.beginDict();
                         enc.writeKey(C4STR(kC4ReplicatorOptionAuthentication));
                         enc.beginDict();
                             enc.writeKey(C4STR(kC4ReplicatorAuthType));
-                            enc.writeString(_authType);
+                            enc.writeString("Basic");
```
snej, line 86: Use `kC4AuthTypeBasic` constant.

snej (approved):
Looks good with the one change I noted ... no need to re-review, just add kC4AuthTypeBasic and push :)

PR #1213, opened 2021-07-31
Fix weird GCC 8 issue with constructors

borrrden:
GCC 7 and 9 are fine but 8 results in ambiguous constructors when constructing an alloc slice from a file path


C/c4BlobStore.cc
```diff
@@ -149,7 +149,7 @@ alloc_slice C4BlobStore::getFilePath(C4BlobKey key) const {
     else if (isEncrypted())
         error::_throw(error::WrongFormat);
     else
-        return alloc_slice(path);
```
jianminzhao, line 152: wonder if this works:
  alloc_slice tmp = alloc_slice(path);
  return tmp;

jianminzhao (approved):

PR #1215, opened 2021-08-03
Updated Fleece to fix leaky FLSliceResult -> alloc_slice conversions

snej:
Implicit conversion from FLSliceResult rvalue to alloc_slice was calling the
wrong constructor and not releasing the FLSliceResult, causing leaks.
(See the Fleece commit for details.)

The fix required minor changes to two places in LiteCore that converted an
FLSliceResult _lvalue_ to an alloc_slice; this now has to be done explicitly.

Fleece PR: couchbaselabs/fleece#106


borrrden (approved):
Be careful about merging this after merging the fleece PR though, it might need another submodule update.

PR #1217, opened 2021-08-04
Do not memcpy 0 bytes to avoid assertion. CBL-2169

jianminzhao:
This is the assertion that frowns:
  assert_precondition(off < size) in slice::operator[](size_t off)

### DIFF
--- a/Networking/WebSockets/WebSocketImpl.cc
+++ b/Networking/WebSockets/WebSocketImpl.cc
@@ -246,6 +246,8 @@ namespace litecore { namespace websocket {
         // Body:
         if (_curMessageLength + length > _curMessage.size)
             return false; // overflow!
+
+        // CBL-2169: addressing the 0-th element of 0-length slice will trigger assertion failure.
         if (length > 0) {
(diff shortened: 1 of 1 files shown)


Networking/WebSockets/WebSocketImpl.cc
```diff
@@ -246,8 +246,10 @@ namespace litecore { namespace websocket {
         // Body:
         if (_curMessageLength + length > _curMessage.size)
             return false; // overflow!
-        memcpy((void*)&_curMessage[_curMessageLength], data, length);
-        _curMessageLength += length;
+        if (length > 0) {
```
borrrden, line 249: The reason for this might become unclear later, so I would like to make it a habit of leaving a one line comment with a short reason, and the bug number 

e.g.
> CBL-2169: Attempting to read from a zero length slice throws an exception

borrrden (changes requested):

pasin (approved):

borrrden (approved):

PR #1223, opened 2021-08-11
MATCH's indexName param should be identifier, CBL-2206

jianminzhao:


LiteCore/Query/N1QL_Parser/n1ql_parser_internal.hh
```diff
@@ -282,7 +282,7 @@ static const char* kFunctions[] = {         // (copied from LiteCore's QueryPars
     "is_array",  "is_atom",  "is_boolean",  "is_number",  "is_object",  "is_string",  "typename",  "to_array",
     "to_atom",  "to_boolean",  "to_number",  "to_object",  "to_string",
     // FTS (not standard N1QL):
-    "match", "rank",
+    "rank",
```
pasin, line 285: I think we need to do the same for rank().

pasin (changes requested):

LiteCore/Query/N1QL_Parser/n1ql_parser_internal.hh
jianminzhao: Okay, will do it.

pasin (approved):

PR #1224, opened 2021-08-16
Handle client connections in separate threads. CBL-2210

jianminzhao:
We do it to avoid blocking the polling thread.

### DIFF
--- a/REST/Server.cc
+++ b/REST/Server.cc
@@ -29,6 +29,7 @@
 #include "c4ListenerInternal.hh"
 #include "PlatformCompat.hh"
 #include <mutex>
+#include <future>
 
 // TODO: Remove these pragmas when doc-comments in sockpp are fixed
 #pragma clang diagnostic push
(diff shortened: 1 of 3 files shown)


REST/Server.cc
```diff
@@ -186,7 +187,19 @@ namespace litecore { namespace REST {
...
             if (sock) {
                 sock.set_non_blocking(false);
-                handleConnection(move(sock));
+                // We are in the poller thread and go handle the client connection in a new thead to avoid
+                // blocking the polling thread.
+                // "this" is retained before this method is called. We need to retain it in the new thread.
+                // The calling thread must keep the ref-count before the new thread retains it.
+                std::promise<void> signalRetain;
+                std::future<void> waitHandle = signalRetain.get_future();
+                thread handleThread([signalRetain = move(signalRetain), sock = move(sock), this]() mutable {
```
snej, line 196: This seems overly complicated; I think you could just move the declaration of `selfRetain` before this line, then bind it as a value in the lambda. That way the Server is retained when the lambda is created, and until the thread exits.

snej (changes requested):

REST/Server.cc
jianminzhao: I see your point: instead of binding the promise in the lambda's capture, binding the selfRetain.
Thank you for the simplification.

snej (approved):

PR #1227, opened 2021-08-19
Reworked BuiltInWebSocket handling of EOF on read (CBL-2204)

snej:
Fix for cases where BuiltInWebSocket won't close the connection when the peer closes the socket unexpectedly:
- Peer calls shutdown
- BuiltInWebSocket::readFromSocket is called, reads 0 bytes, calls inherited close()
- WebSocketImpl::close sends a CLOSE message and waits for a reply.
- Five seconds later it times out. closeSocket() calls socket->close(), which invokes the `shutdown` system call and Poller::interrupt.
- Poller::interrupt tries to call the listeners on that FD, but there aren't any, so BuiltInWebSocket never finds out the socket is closed.

I added a new kDisconnect event for Poller, so the BIWS can be notified reliably when the socket closes.

Also, it doesn't make sense to send a CLOSE message in this situation (peer shutdown) since the peer can't possibly reply to it. Instead, WebSocketImpl should  treat it as a protocol violation to receive an EOF before the peer's sent a CLOSE. I added an `onEOF` method to signal this.

**NOTE:** Poller has two separate implementations of the low-level polling code. I only changed the non-Windows code; I didn't feel safe making the corresponding changes to ifdef'd-out code I can't compile or run. @borrrden, can you make those changes?

For CBL-2204

### DIFF
--- a/Networking/Poller.cc
+++ b/Networking/Poller.cc
@@ -212,8 +212,10 @@ namespace litecore { namespace net {
         }
 
         while(select(maxfd, &fds_read, &fds_write, &fds_err, nullptr) == SOCKET_ERROR) {
-            if(WSAGetLastError() != WSAEINTR)
+            if(WSAGetLastError() != WSAEINTR) {
+                LogError(WSLog, "Poller: poll() returned WSA error %d; stopping thread", WSAGetLastError());
                 return false;
+            }
         }
 
         _waiting = false;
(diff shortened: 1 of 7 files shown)


Networking/Poller.cc
```diff
@@ -154,6 +154,11 @@ namespace litecore { namespace net {
     }
 
 
+    void Poller::interrupt(int fd) {
```
snej, line 157: Note: I added an assertion here since passing a non-positive FD would have a bad effect. That necessitated splitting out the guts into `_interrupt` since Poller itself does need to pass negative numbers or zero.

Networking/WebSockets/BuiltInWebSocket.cc
```diff
@@ -103,9 +103,8 @@ namespace litecore { namespace websocket {
 
     void BuiltInWebSocket::closeSocket() {
         logVerbose("closeSocket");
-        if (_socket &&_socket->connected()) {
-            _socket->close();   // This will often be detected in readFromSocket(), which will close it
-            _socket->interrupt(); // But on Windows it usually is not, so interrupt it to poke it into doing the right thing
+        if (_socket) {
```
snej, line 106: Note: The `connected()` test and `interrupt()` were moved into `TCPSocket::close`, so the behavior remains the same.

Networking/WebSockets/BuiltInWebSocket.cc
```diff
@@ -448,16 +442,6 @@ namespace litecore { namespace websocket {
                 return;
             }
             
-            // CBL-1633: Need to check for EOF here, or _socket->write returns
-            // 0 endlessly
-            if(n == 0 && _socket->atWriteEOF()) {
```
snej, line 453: Note: If you look carefully at the code above this, you'll see that this block is unreachable (n cannot be zero); so I removed it.

snej:
Jenkins Android build failed due to "write error: No space left on device" ... restarting it.

borrrden (approved):

Networking/Poller.cc
```diff
@@ -154,6 +154,11 @@ namespace litecore { namespace net {
...
+        Assert(fd > 0);
+        _interrupt(fd);
+    }
+
     void Poller::stop() {
         interrupt(-1);
```
jianminzhao, line 163: -1 here?

snej: Yes. The code on the background thread interprets -1 as a request to stop the thread.

jianminzhao: Do you mean _interrupt(), since the above assertion?

snej: Oh, I see -- yes, line 163 should be calling `_interrupt`. I'll fix it.

jianminzhao (approved):

PR #1228, opened 2021-08-20
Small test improvements

snej:
* Xcode: Prevent exception breakpoint in two new tests
* HTTPTypes: `Method` enum was missing `HEAD`
* ReplicatorSGTest: Automatically skip SG 3 tests (when running with SG 2.x)

### DIFF
--- a/LiteCore/tests/LogEncoderTest.cc
+++ b/LiteCore/tests/LogEncoderTest.cc
@@ -274,6 +274,7 @@ TEST_CASE("Logging throw in c++", "[Log]") {
     string excMsg;
     const LogFileOptions prevOptions = LogDomain::currentLogFileOptions();
     try {
+        ExpectingExceptions x;
         LogDomain::writeEncodedLogsTo(fileOptions, "Hello");
     } catch (std::exception& exc) {
         excMsg = exc.what();
@@ -288,6 +289,7 @@ TEST_CASE("Logging throw in c4", "[Log]") {
     sprintf(folderName, "Log_Rollover_%" PRIms "/", now.count());
     FilePath tmpLogDir = TestFixture::sTempDir[folderName];
     // Note that we haven't created tmpLogDir.
+    ExpectingExceptions x;
     C4Error error;
     const LogFileOptions prevOptions = LogDomain::currentLogFileOptions();
     CHECK(!c4log_writeToBinaryFile({kC4LogVerbose, slice(tmpLogDir.path()), 16*1024, 1, false},
--- a/Replicator/tests/ReplicatorAPITest.hh
+++ b/Replicator/tests/ReplicatorAPITest.hh
@@ -398,6 +398,7 @@ public:
         if (r->run()) {
             *outStatus = r->status();
             *outError = {};
+            _serverName = r->header("Server");
             return r->body();
         } else {
             REQUIRE(r->error().code != 0);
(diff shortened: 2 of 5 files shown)


borrrden:
Jenkins machines are down, rendering the CI useless so we can only rely on GH actions for now.

borrrden (approved):

LiteCore/tests/LogEncoderTest.cc
```diff
@@ -288,6 +289,7 @@ TEST_CASE("Logging throw in c4", "[Log]") {
     sprintf(folderName, "Log_Rollover_%" PRIms "/", now.count());
     FilePath tmpLogDir = TestFixture::sTempDir[folderName];
     // Note that we haven't created tmpLogDir.
+    ExpectingExceptions x;
```
jianminzhao, line 292: Why do you use this? I tried to make c4log_writeToBinaryFile() not susceptible to the exception, so that the client code does not have to do something like above line.

LiteCore/tests/LogEncoderTest.cc
```diff
@@ -274,6 +274,7 @@ TEST_CASE("Logging throw in c++", "[Log]") {
     string excMsg;
     const LogFileOptions prevOptions = LogDomain::currentLogFileOptions();
     try {
+        ExpectingExceptions x;
```
jianminzhao, line 277: Is this line necessary? I tried to make logging functions self-contained, not relying this line.

Replicator/tests/ReplicatorAPITest.hh
```diff
@@ -434,6 +435,20 @@ public:
     }
 
 
+    bool requireSG3() {
+        // `_serverName` is set by sendRemoteRequest
+        if (!_serverName)
+            sendRemoteRequest("HEAD", "/");
+        REQUIRE(_serverName.hasPrefix("Couchbase Sync Gateway/"));
+        if (_serverName >= "Couchbase Sync Gateway/3") {
```
jianminzhao, line 443: Good until "/9".

LiteCore/tests/LogEncoderTest.cc
snej: The Xcode project has a C++ exception breakpoint, so it will pause at any `throw`. Whether or not it's caught doesn't matter; this is at the instant before the stack unwinds. This is a pain when running unit tests that happen to throw an exception.

To work around that, there's a global boolean that the breakpoint tests; if the flag is set the breakpoint is skipped. `ExpectingExceptions` is a tiny class that just sets that flag while it's in scope.

Replicator/tests/ReplicatorAPITest.hh
snej: Yup, we can update this line then. For now I don't see a need to do more complex parsing.

LiteCore/tests/LogEncoderTest.cc
jianminzhao: Is it this, 
struct ExpectingExceptions {
        ExpectingExceptions() {
            ++gC4ExpectExceptions;
            c4log_warnOnErrors(false);
        }
I had this in Logging.cc,
#define ERROR_THROW(ERROR_CODE, FMT, ...)                      \
    do {ScopedSetter setter(error::sWarnOnError, false);       \
        error::_throw(ERROR_CODE, FMT, ## __VA_ARGS__);} while (false)
It does the same thing as ExpectingExceptions, but doing it where the exception is thrown. If this line is needed here, I am afraid somewhere else there is thrown.

Replicator/tests/ReplicatorAPITest.hh
jianminzhao: Agreed

borrrden:
Can this branch be deleted?  I think this was addressed in another PR?

snej:
Yes, in https://github.com/couchbase/couchbase-lite-core/pull/1327 which was just merged.

jianminzhao:
I see, deleting branch in addition to closing the PR.

PR #1231, opened 2021-08-24
Update for new sdkmanager location

borrrden:
This is the result of the docker container update to JDK 11.  The sdkmanager in the previous location is not compatible with JDK 11, and frustratingly the new sdkmanager will download the old sdkmanager into the old location so we are forced to use the new one.


borrrden:
Weirdly PR validation failed with basically no output....I'll investigate that

borrrden:
Please squash this 🙏

jianminzhao (approved):

PR #1232, opened 2021-08-24
Handle bool in array_contains, DBL-2215

jianminzhao:
Before we compare equality, we convert N1QL bool to 1/0 and then do the integer comparison.

### DIFF
--- a/LiteCore/Query/SQLiteFleeceFunctions.cc
+++ b/LiteCore/Query/SQLiteFleeceFunctions.cc
@@ -371,12 +371,19 @@ namespace litecore {
             }
         }
 
+        auto comparable = [](valueType vtype1, valueType vtype2) {
+            // pre-condition: vtype2 != kBoolean
+            return   vtype1 == vtype2 ? true
+                   : vtype2 == kNumber ? vtype1 == kBoolean
+                   : false;
+        };
+
         // Now iterate the array/dict:
         bool found = false;
         if (collection->type() == kArray) {
             for (Array::iterator j(collection->asArray()); j; ++j) {
                 auto val = j.value();
-                if (val->type() == targetType && predicate(val, target)) {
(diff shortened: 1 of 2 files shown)


LiteCore/Query/SQLiteFleeceFunctions.cc
```diff
@@ -371,20 +371,27 @@ namespace litecore {
             }
         }
 
+        auto comparable = [](valueType vtype1, valueType vtype2) {
+            // pre-condition: vtype2 != kBoolean
+            return   vtype1 == vtype2 ? true
+                   : vtype2 == kNumber ? vtype1 == kBoolean
+                   : false;
```
borrrden, line 378: I find this hard to understand, I think it warrants a comment.  Why is it a requirement that vtype2 not be a boolean?  And what is this actually checking?  The ability to compare?

jianminzhao: vtype2 is targetType that is set according to arg. We don't check the subtype to determine whether it is a boolean or not, because in this context we will treat boolean as int. On the other hand, vtype1 derives from value in the array. As a Value, its type is finer, could be kBoolean. Basically, the vtype2 already converted to int, inside this function, we only need to convert vtype1 if it is kBoolean.

jianminzhao: The rule is, treat kBoolean as integer before doing the comparison.

jianminzhao: Basically, the pre-condition means that the 2nd argument has already been converted to integer if it is a kBoolean.

borrrden: That's the kind of thing that warrants a comment in the source.  I don't think I'd figure that out without at least an hour or two of analysis.

PR #1236, opened 2021-08-31
Upgrade db with indexes. CBL-2155

jianminzhao:
We need to re-create the indexes after upgrading the key table.


LiteCore/Database/DatabaseImpl+Upgrade.cc
```diff
@@ -54,6 +54,31 @@ namespace litecore {
...
+            : _db(db)
+            , _transaction(t)
+            , _specs{_db->defaultKeyStore().getIndexes()}
+        {
+            for (IndexSpec& spec: _specs) {
+                _db->defaultKeyStore().deleteIndex(spec.name, _transaction);
+            }
+        }
+
+        ~RestoreIndex() {
```
snej, line 73: We should avoid putting anything in a destructor that might throw an exception, since if that destructor is called as cleanup during exception handling ,and it throws another exception, C++ will terminate the process.

There isn't any need to restore the indexes on an exception, because the caller will just abort the transaction anyway.

You can just put the delete-indexes and restore-indexes code directly into upgradeDocumentVersioning.

LiteCore/Query/SQLiteKeyStore+Indexes.cc
```diff
@@ -101,14 +108,21 @@ namespace litecore {
...
             t.commit();
         } else {
             t.abort();
         }
     }
 
 
+    void SQLiteKeyStore::deleteIndex(slice name, ExclusiveTransaction &t)  {
+        auto spec = db().getIndex(name);
+        // pre-condition: !!spec
```
snej, line 121: Should return if `!spec`, as the existing deleteIndex method does.

snej (changes requested):

LiteCore/Database/DatabaseImpl+Upgrade.cc
jianminzhao: Good point. I will do as you suggesting.

snej (approved):

PR #1238, opened 2021-09-02
Support HTTP chunked-mode responses, fixing timeout error (CBL-2313)

snej:
The unit test "HTTPS Request to public host" has begin failing, not because of
a change in our code but because the couchbase.com server has started using
HTTP chunked mode to send the response. TCPStream doesn't recognize this
response type, so (because there's no Content-Length) it tries to read all data
till EOF. Meanwhile the server is using keep-alive, so after sending the
response it waits for another request on the socket. Deadlock! Which ends with
a timeout error.

Fixed by making TCPSocket recognize "Transfer-Encoding: chunked" and decoding
the output. While I was at it, I made it fail on other transfer encodings, and
made it only read till EOF if there is an explicit "Connection: close" header.

Note this is unlikely to affect CBL because the only HTTP requests it sends are
WebSocket connections, which don't have any regular response.

Fixes CBL-2313

PS: Additionally, this fixes three (3) bugs in the `readToDelimiter` method that somehow never came up before, but did when handling chunked responses.


jianminzhao (approved):

borrrden:
Broken Windows build needs investigation

borrrden (changes requested):
For reasons unknown to me, Windows is unhappy with `class CALLBACK`.  It can be named literally anything else (I think, I tried both `T` and `WRITEFN`) and this will compile.

snej:
OK, renamed it to `Callback` and updated the Fleece submodule.

borrrden:
Can you check the iOS failure?  I wonder why it's not present in the GHA Xcode build.

PR #1239, opened 2021-09-02
Making sure outErrorPos is initialized if there is an error. CBL-2217

jianminzhao:


C/c4Query.cc
```diff
@@ -60,10 +60,13 @@ Retained<C4Query> C4Query::newQuery(C4Collection *coll, C4QueryLanguage language
     } catch (Query::parseError &x) {
         if (outErrorPos) {
             *outErrorPos = x.errorPosition;
-            return nullptr;
-        } else {
-            throw;
         }
+        throw;
+    } catch (...) {
+        if (outErrorPos != nullptr) {
```
borrrden, line 66: Would prefer this to match the other clause as `if (outErrorPos)`

jianminzhao: Changed as suggested.

C/c4Query.cc
```diff
@@ -60,10 +60,13 @@ Retained<C4Query> C4Query::newQuery(C4Collection *coll, C4QueryLanguage language
     } catch (Query::parseError &x) {
         if (outErrorPos) {
             *outErrorPos = x.errorPosition;
-            return nullptr;
```
borrrden, line 63: Was this removal on purpose?

jianminzhao: yes, because I don't think we should change the logic here merely because being provided outErrorPos. The exception may contain other useful info, for example, why the syntax error, that the platform may deicide whatever to do with it.

borrrden (approved):

PR #1244, opened 2021-09-09
Adjust the SGW config to work with SGW3.0, CBL-2254

jianminzhao:


borrrden (approved):
I noticed that adding these removes the ability to start up with 2.8.x.  As long as that's ok.

JRascagneres (approved):
Looks fine to me. As Jim stated though this won't work for 2.8.x anymore. 
These options are also possible to pass in as command line flags if that suits you better? Otherwise this should work :)

jianminzhao:
@borrrden @pasin how about to change the current config.json to config_2.8.json? Or just overwrite it. If in some case we need to run sgw 2.8, we can always get the previous version from git. What is you opnion?

pasin (approved):

pasin:
IMO, as SGW 2.8 will be around for quite some time. It's good to have a backup for 2.8 config.

jianminzhao:
I added a tag to 2.8 config. "git show sgw2.8-config" will print the config.json for 2.8 sgw

PR #1248, opened 2021-09-15
CBL-2336: Call C4Query observers outside of the internal lock

borrrden:


borrrden:
@snej I am not quite sure how the lifetime of these objects is managed, but I don't see anywhere here that frees them so I am assuming it is ok to call the notify outside of the lock.

C/c4Query.cc
```diff
@@ -235,9 +235,23 @@ void C4Query::enableObserver(C4QueryObserverImpl *obs, bool enable) {
...
+        // in a deadlock, but on the other hand not calling it
+        // inside the lock could result in the callback coming back
+        // to mutate the collection while we are using it (which, 
+        // coincidentally, is why this deadlocks in the first place).
+        // So to counteract this, make a copy and iterate over that.
+        observers = _observers;
+    }
+
+    for(auto &obs : observers) {
         obs->notify(c4e, err);
```
jianminzhao, line 255: my concern of taking it out of the lock is whether it can be deleted from another thread. The original lock protects this situation. But really not sure. Defer to @snej

borrrden: I'm not sure even the original situation prevents deletion of the object because it's just a pointer to an externally controlled non ref-counted object.

jianminzhao: I also noticed the non ref-counted thing. The normal path as I can imagine would start from that observer obj is deleted. In the destructor of the observer, it enableObserver(false), which pulls itself from the list. But if it is locked by liveQuerierUpdated(), it has to wait for "updated" method to finish, then pulls itself out. Then returns to destructor. c.f. C4QueryObserverImpl::~C4QueryObserverImpl()

snej (approved):

snej:
As far as I can tell, this is the right thing to do.

PR #1252, opened 2021-09-23
Tagging DataFiles, CBL-2357

jianminzhao:
Tagging the DataFiles according to the how they are opened, as whether they are opened by externally or by CBL. The tagging will be used to provide better info to the user when deletion of database fails because of open connections.

### DIFF
--- a/C/c4Database.cc
+++ b/C/c4Database.cc
@@ -90,8 +90,17 @@ C4Database* c4db_open(C4Slice path,
                       const C4DatabaseConfig *configP,
                       C4Error *outError) noexcept
 {
+    return dbOpen(path, configP, DatabaseTagExternal, outError);
+}
+
+// The corresponding internally-faced API
+C4Database* dbOpen(C4String path,
+                   const C4DatabaseConfig *configP,
+                   C4DatabaseTag dbTag,
+                   C4Error *outError) noexcept
+{
     return tryCatch<C4Database*>(outError, [=] {
-        return retain(new C4Database(toString(path), *configP));
+            return retain(new C4Database(toString(path), *configP, dbTag));
     });
 }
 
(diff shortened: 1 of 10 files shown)


C/c4Database.cc
```diff
@@ -89,28 +89,55 @@ bool c4key_setPassword(C4EncryptionKey *key, C4String password, C4EncryptionAlgo
...
                       const C4DatabaseConfig *configP,
                       C4Error *outError) noexcept
+{
+    return dbOpen(path, configP, DatabaseTagExternal, outError);
+}
+
+// The corresponding internally-faced API
+C4Database* dbOpen(C4String path,
+                   const C4DatabaseConfig *configP,
+                   C4DatabaseTag dbTag,
```
snej, line 99: This seems like too much code change ... why not just add the C4DatabaseTag to the C4DatabaseConfig?

C/tests/c4DatabaseTest.cc
```diff
@@ -740,8 +740,8 @@ N_WAY_TEST_CASE_METHOD(C4DatabaseTest, "Test delete while database open", "[Data
     CHECK(timeTaken < 2s);
     CHECK(err.code == kC4ErrorBusy);
 
-    auto message = slice(c4error_getDescription(err));
-    CHECK(message == "Can't delete db file while the caller has open connections");
+    string message = string(slice(c4error_getDescription(err)));
+    CHECK(message.find("Can't delete db file while the caller has open connections") != string::npos);
```
snej, line 744: FYI, slice has a `find` method too, so no need to convert to string

LiteCore/Storage/DataFile.cc
```diff
@@ -235,10 +251,22 @@ namespace litecore {
...
+                        if (!dbtags.str().empty()) {
+                            dbtags << ", ";
+                        }
+                        dbtags << kDatabaseTags[df->databaseTag()];
+                    });
+                    ostringstream ss;
+                    ss << "Can't delete db file while other connections are open. The open connections are tagged "
+                    << dbtags.str() << ".";
+                    error::_throw(error::Busy, "Can't delete db file while other connections are open."
+                                               " The open connections are tagged %s.", ss.str().c_str());
```
snej, line 266: This ends up repeating the message twice, since `ss` already contains the entire message.
I would get rid of lines 262-264, then use `dbtags.str().c_str()` in the exception format.

snej (changes requested):

snej:
This commit would be a lot simpler if we just declared C4DatabaseTag in C4Database.h and added one as a field of C4DatabaseConfig.

C/c4Database.cc
jianminzhao: In that meeting, c.f. note of CBL-2357, we decided not to change the public APIs.
The goal was to make sure that c4db_openXs() are only called by the platform. Internally we can only call dbOpenXs(). This way, we will know what DataFiles are related to the applications. This distinction has business semantics (fail-fast), whereas other tags only contribute to the error message.

C/tests/c4DatabaseTest.cc
jianminzhao: I found an interesting thing here. Originally, Jim has it, auto message = slice(c4error_getDescription(err)). This caused sanitizer error. This is the main reason I changed it to string. Because of the time pressure, I did not spend too much to find out why.

LiteCore/Storage/DataFile.cc
jianminzhao: Yes, good catch. I meant to have the message, "Can't delete .... The open connections are tagged as x,y.z." But, that gets repeated. Will fix it.

C/c4Database.cc
jianminzhao: Also, if we make it a DatabaseConfig, the customer will be responsible to to fill up config correctly.

C/tests/c4DatabaseTest.cc
borrrden: Oh sorry, add the `_sl` suffix to the string there and I be that will fix the sanitizer error?

borrrden:
@snej I advocated for doing it without modifying the LiteCore API so that we wouldn't have to wrangle all the platforms into this.  If it's a whole lot simpler to do by adding a config property, though, then I guess everyone will have to be aware of it for both 2.8.x next (if applicable) and 3.0.

PR #1253, opened 2021-09-24
Tagging the opened database connections. CBL-2357

jianminzhao:
By default, a database connection is tagged "appOpened" as it is opened. These database connections are opened by the application, and, as such, the application must close all of them as it calls to delete the underlying database. Other tags are reserved for the connections opened by the Core library. The library is responsible to close them.
Also changed kOtherDBCloseTimeoutSecs from 3 seconds to 6 seconds for CBL-2383.

### DIFF
--- a/C/c4Private.h
+++ b/C/c4Private.h
@@ -55,6 +55,17 @@ bool c4log_getWarnOnErrors(void) C4API;
     exception occurs. */
 void c4log_enableFatalExceptionBacktrace(void) C4API;
 
+typedef C4_ENUM(uint32_t, C4DatabaseTag) {
+    DatabaseTagAppOpened,
+    DatabaseTagReplicator,
+    DatabaseTagBgDB,
+    DatabaseTagREST,
+    DatabaseTagOther
+};
+
+C4DatabaseTag c4db_getDatabaseTag(C4Database* db) C4API;
+void c4db_setDatabaseTag(C4Database* db, C4DatabaseTag dbTag) C4API;
+
 /** Locks a recursive mutex associated with the C4Database instance.
     Blocks if it's already locked. */
 void c4db_lock(C4Database *db) C4API;
--- a/C/c4Private.h
+++ b/C/c4Private.h
@@ -59,8 +59,7 @@ typedef C4_ENUM(uint32_t, C4DatabaseTag) {
     DatabaseTagAppOpened,
     DatabaseTagReplicator,
     DatabaseTagBgDB,
-    DatabaseTagREST,
-    DatabaseTagOther
+    DatabaseTagREST
 };
 
 C4DatabaseTag c4db_getDatabaseTag(C4Database* db) C4API;
--- a/C/tests/c4DatabaseTest.cc
+++ b/C/tests/c4DatabaseTest.cc
@@ -718,6 +718,8 @@ N_WAY_TEST_CASE_METHOD(C4DatabaseTest, "Database Config2 And ExtraInfo", "[Datab
     REQUIRE(c4db_deleteNamed(slice(db2Name), config.parentDirectory, &error));
 }
 
+//#define FAIL_FAST
+
 N_WAY_TEST_CASE_METHOD(C4DatabaseTest, "Test delete while database open", "[Database][C]") {
     // CBL-2357: Distinguish between internal and external database handles so that
     // external handles open during a delete will be a fast-fail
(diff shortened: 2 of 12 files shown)


Replicator/DBAccess.cc
```diff
@@ -63,6 +64,8 @@ namespace litecore { namespace repl {
                     if (!idb) {
                         logError("Couldn't open new db connection: %s", c4error_descriptionStr(error));
                         idb = c4db_retain(db);
+                    } else {
+                        c4db_setDatabaseTag(idb, DatabaseTagReplicator);
```
snej, line 68: As far as I can tell this is the only db that gets tagged Replicator. But this db is only used to insert pulled docs. So the main replicator database isn't tagged. It isn't created by DBAccess; IIRC it's created by C4Replicator and passed into the main Replicator.

snej (changes requested):
Looks good apart from not tagging the main replicator db.

Replicator/DBAccess.cc
snej: Look for the call to `db->openAgain()` -- that's what opens it.

Come to think of it, you should make openAgain propagate the tag to the new database.

jianminzhao: I grepped all c4db_open in the source and ignored places in files that has Test. 
If the app opens a db, with c4db_openX(), and we, for our own purpose, like the one here, openAgain(). Customers still owns the db they created, right?

jianminzhao: Sorry, only grepped c/cc files. There are some header doing it.
Will fix them

C/tests/c4DatabaseTest.cc
```diff
@@ -726,22 +728,31 @@ N_WAY_TEST_CASE_METHOD(C4DatabaseTest, "Test delete while database open", "[Data
...
     auto timeTaken = chrono::duration_cast<chrono::seconds>(end - start);
     CHECK(timeTaken < 2s);
+#endif
     CHECK(err.code == kC4ErrorBusy);
 
-    auto message = slice(c4error_getDescription(err));
-    CHECK(message == "Can't delete db file while the caller has open connections");
+    C4SliceResult message = c4error_getDescription(err);
+#ifdef FAIL_FAST
+    CHECK(slice(message) == "LiteCore error 16 \"Can't delete db file while the caller has open connections\"");
```
Sandychuang8, line 751: I remember the main goal to fail fast is to prevent the 3rd party developer to open a database with the same file?
Not sure why you print `Can't delete db file while the caller has open connections` It doesn't look right to me.

Replicator/DBAccess.cc
jianminzhao: Found two new places, c4IncomingReplicator.hh, c4RemoteReplicator.hh
There is still failed case (a new case) with FAIL-FAST.

C/tests/c4DatabaseTest.cc
snej: It's fine to open the same db file multiple times. The goal is to fail immediately if a developer tries to _delete_ a database that's still open in another connection. But even without that, we'll still fail, but after a delay.

jianminzhao: Jim came up with that message. He wrote the test case. I think "caller" here means the app developer. Connections mean database handles that the app obtains from calling c4db_open().

snej (approved):

C/c4Private.h
```diff
@@ -55,6 +55,17 @@ bool c4log_getWarnOnErrors(void) C4API;
     exception occurs. */
 void c4log_enableFatalExceptionBacktrace(void) C4API;
 
+typedef C4_ENUM(uint32_t, C4DatabaseTag) {
+    DatabaseTagAppOpened,
+    DatabaseTagReplicator,
+    DatabaseTagBgDB,
+    DatabaseTagREST,
+    DatabaseTagOther
```
pasin, line 63: I understand that the previous PR was using the other tag as the default tag for internal database open/reopen API. But as we have changed the API for setting the tag, do we still need the Other tag? I don't see it used anywhere as well.

jianminzhao: Yes, you're right. We reversed the logic in this regard. Now I don't need it. (In the first version, DataFile gets default internal tag, DatabaseTagOther. We reversed it and it takes AppOpened tag by default.)  Going to remove it.

Sandychuang8 (approved):

pasin (approved):

PR #1255, opened 2021-09-28
Fixing intermittent mismatching of replicator's status counts, CBL-1691

jianminzhao:


pasin:
I'm not familiar with all the codes here. I can't understand why getting the missing body size in _revReRequested could be incorrect? Can you explain a bit about the problem and the fix?

pasin:
I'm not familiar with all the codes here. I can't understand why getting the missing body size in _revReRequested could be incorrect? Can you explain a bit about the problem and the fix?

pasin (commented):
I'm not familiar with all the codes here. I do not understand why getting the missing body size in _revReRequested could be incorrect? Can you explain a bit about the problem and the fix?

borrrden (commented):
I strongly second Pasin's comment.  Why does this need a new testing parameter (numDocs) as well?

jianminzhao:
@borrrden numDocs is used as test param only in non-concurrent situation. In this particular test case, after 14 updates on the remote, local has rev 1, and remote has rev 15. SG tells local I have rev 15, then local will ask for all deltas from 2 to 15. For each doc, there is one trip of exchange. In the status statistics, it contributes to one total, one completed, one doc. I am not 100% sure, but been consistent for me. I actually is not 100% sure. @snej would you approve or disapprove it. I can remove that test. I put it there to demonstrate the contrast.
In concurrent case, if, between updates, there is enough time to update the local, from rev 1 to rev 2, to 3, etc, then each rev will contribute to the status as the doc count. Therefore, status.docCount > db.docCount. However, it progresses orderly, and we won't hit this bug.
Only if updates rapidly do we hit this bug, so rapidly that local cannot manage to update the database at the same speed. I don't know the details, but local may find it lacks the history from the remote to jump from the current rev as base to the target rev. In this case we will Re-request the rev. We hit the bug because we failed to count the re-request into the status.total, but status.completed is counted properly.

jianminzhao:
This commit attempts to fix the following problem.
* In IncomingRev::finish(), there are following two statements:
  1. _puller->revReRequest(this)
  2. _puller->revWasHandledthis)
* They queue two functions to _puller's mailbox. At the time queued statements get served, (1) will add to status.total and (2) will reset "this".
* They are queued in the logical order, the problem is they may be served out of order. When queued rapidly in sequence, we have in the queue,
   - reRequest(rev1)
   - reset(rev1)
   - reRequest(rev2)
   - reset(rev2)
* The problem is, there is an optimization as they are served. All reset functions are put in a batch queue. The effect is, as we serve rev1/reset, we will serve all reset functions in the mailbox. This is the actual order in execution:
  - reRrequest(rev1)
  - reset(rev1)
  - reset(rev2)
  - reRequest(rev2)
* Hence, rev2->reRequest is called after rev2->reset, which resets rev2 before rev2->reReRequest() is called.

After speaking with Jens, we fix it by getting the rev's bodySize a.k.a. total count at the time the function is queued. This is, instead of reRequest(rev), we have reRequest(rev->getMissingBodySize()).
Writing to here, I smell a racing condition. I will discuss it with Jens.

Replicator/Puller.cc
```diff
@@ -280,17 +283,17 @@ namespace litecore { namespace repl {
 
 
     void Puller::revReRequested(fleece::Retained<IncomingRev> inc) {
-        enqueue(FUNCTION_TO_QUEUE(Puller::_revReRequested), inc);
+        enqueue(FUNCTION_TO_QUEUE(Puller::_revReRequested), _missingSequences.bodySizeOfSequence(inc->remoteSequence()));
```
snej, line 286: You can't access `_missingSequences` in a public method since it's not necessarily called on the Puller's queue. Also, these public methods should be simple wrappers around `enqueue`, without any extra logic.

(We already talked about this on Slack, I'm just putting it here for completeness.)

snej (changes requested):

borrrden (approved):

Replicator/IncomingRev.cc
```diff
@@ -64,8 +64,9 @@ namespace litecore { namespace repl {
 
     // Read the 'rev' message, then parse either synchronously or asynchronously.
     // This runs on the caller's (Puller's) thread.
-    void IncomingRev::handleRev(blip::MessageIn *msg) {
+    void IncomingRev::handleRev(blip::MessageIn *msg, uint64_t bodySizeOfRemoteSequence) {
```
snej, line 67: Just `bodySize` is a better name. It does get stored in a map keyed by remote sequence ID, but that's not relevant at this point. It's just the (approximate) size in bytes of the revision body according to the peer.

Replicator/Puller.cc
```diff
@@ -280,17 +285,17 @@ namespace litecore { namespace repl {
 
 
     void Puller::revReRequested(fleece::Retained<IncomingRev> inc) {
```
snej, line 287: This is an async method, and all it should do is enqueue a call to the corresponding private method. It should not have other logic like pulling the body size out of the IncomingRev. Instead, add a `bodySize` parameter.

Replicator/IncomingRev.hh
```diff
@@ -48,6 +48,8 @@ namespace litecore { namespace repl {
 
         int progressNotificationLevel() const override;
 
+        uint64_t bodySizeOfRemoteSequence() const { return _bodySizeOfRemoteSequence; }
```
snej, line 51: This should not be exposed as a public method.

Replicator/tests/ReplicatorSGTest.cc
```diff
@@ -665,6 +665,118 @@ TEST_CASE_METHOD(ReplicatorSGTest, "Pull iTunes deltas from SG", "[.SyncServer][
           timeWithDelta, timeWithoutDelta, timeWithoutDelta/timeWithDelta);
 }
 
+
+TEST_CASE_METHOD(ReplicatorSGTest, "Replicator count balance", "[.SyncServer]") {
```
snej, line 669: ReplicatorSGTest is never automatically run, since it requires a specially set up SG server. So it's not a good place to put tests that should run in CI. Is there a way you can change this test so it can be moved to ReplicatorLoopbackTest or ReplicatorAPITest?

snej (changes requested):

Replicator/IncomingRev.cc
jianminzhao: Okay, will change it

Replicator/IncomingRev.hh
jianminzhao: The parent puller will have to call it. Make Puller a friend?

Replicator/Puller.cc
jianminzhao: Not totally getting it. I have changed Puller:_revReRequested() to take parameter of, instead of inc, bodySize, which is all that _revReRequested() needs to do its job. The bodySize is store in inc object, and I pull it out as I enqueue it rather than as it is executed in the Actor's thread.

Replicator/tests/ReplicatorSGTest.cc
jianminzhao: Urhn, I also need SG server to draw out the bug.

Replicator/IncomingRev.hh
borrrden: If it's called by the puller, wouldn't it be appropriate in the section marked "Called by the Puller" then?

jianminzhao: yes, I will do it.

pasin (approved):

Replicator/Puller.cc
snej: Actor objects have an idiom of a public method that calls `enqueue` to schedule a call to a private method.
In this idiom the public method is simply a wrapper. The two methods have _the same parameter list_, except as necessary to manage ref-counting. The public method has no other logic in it.

In this case you have changed the parameter type of the public method but not the private method, and you have hidden some logic in the public method. This is very unlike the idiom and will be confusing.

You should change the _public_ `revReRequested` method to take the bodySize and simply pass it through to the private method. Then the caller, IncomingRev, does the work of finding the body size. This also means you don't have to expose a public method to access the body size.

Replicator/IncomingRev.hh
snej: It doesn't need to be called by the Puller at all. The IncomingRev should pass the `_bodySize` to the Puller.

Replicator/Puller.cc
jianminzhao: Got it. Will follow the idiom

snej (approved):

PR #1257, opened 2021-10-01
2.8.7 vf pr, this PR (cherry) picks the resolutions of CBL-2410, CBL-2304

jianminzhao:


borrrden:
Please try not to use a title like 2.8.7 vf pr, it doesn't describe anything.   Write it as if the person looking at it has no context at all, because someone reading it later might be looking at this with no prior knowledge, or looking at it a year later to remember what happened.

borrrden (approved):
Don't squash this when merging so that the commit messages can remain intact and messages received onto the corresponding CBL tickets.

borrrden:
If you want the PR validation to pass you will need to rebase this PR, it is currently based on the commit before the HEAD of the VF branch.

PR #1264, opened 2021-10-08
Bump the fleece version for porting CBL-2358, CBL-2408 to Lithium. CB…

jianminzhao:
…L-2451


pasin:
@jianminzhao I think the fleece submodule in LiteCore's lithium branch should point to release/lithium branch instead. Could you please revert the change and have a new PR that points to the new fleece's release/lithium branch? Also before that you would need to cherry-pick the change for CBL-2358 and CBL-2408 from master to lithium branch.

PR #1265, opened 2021-10-08
CBL-2410: Don't tear down response timer in timeout situation

jianminzhao:
If onClose is reached with _timedOut being true, then that means not only did the timer already fire and tearing it down is extraneous at this point, but it is almost certainly the case that this method is reached synchronously from the timer callback via timer -> close/requestClose -> platform -> onClose.  This means that it is a hang situation in which the timer waits forever for its callback to finish, and the callback is waiting for the current method to finish.

Co-authored-by: Jim Borden <jim.borden@couchbase.com>

Cherry-picked from 4f8d51a on the branch of hotfix/2.8.7


borrrden:
Something very odd has happened to the macOS builds, please look and then re-add me as a reviewer.

PR #1266, opened 2021-10-09
CBL-2407: Tear down DBAccess when replicator stopped

jianminzhao:
As opposed to waiting until the replicator is completely out of memory.  This gives a more deterministic end of life to the object. Ported from aeb790f on branch hotfix/2.8.7

Fix instability caused by fix for 2407

Now with the DBAccess closing earlier, straggling actor messages could access them.  Switch this to throw an error instead that won't crash the process. Ported from fea753e on branch hotfix/2.8.7.


jianminzhao:
This commit is ported from two commits on the VF branch based on hydrogen. DBAccess had changed a lot from hydrogen to lithium. Please pay appropriate attention.

PR #1276, opened 2021-10-21
Clone the QueryEnumerator before passing it to the QueryObserver call…

jianminzhao:
…back. CBL-2460


pasin:
Seem like the test is hitting CBL-2459.

borrrden:
> 
> 
> Seem like the test is hitting [CBL-2459](https://issues.couchbase.com/browse/CBL-2459).

Is this an intermittent issue?  The test passed on GH actions

pasin:
There is a race. If the second observer is added after the first observer is notified, the second observer will not be notified until there is a change.

jianminzhao:
yes, I noticed this too. I did not mean to solve this problem (or not.a problem)

PR #1287, opened 2021-10-29
Cbl 2477: Tear down DBAccess on stopped instead of on release

jianminzhao:


jianminzhao:
This task is to port the fix in hotfix/2.8.7, filed as 2407, into Lithium. The code base has undergone non-trivial changes since Hydrogen(2.8.7); sheer cherry-picking can hardly do it. The idea is to release reference to the database as soon as the replicator reaches Stopped state, as supposed when all replicator's workers are released.

pasin (approved):

LiteCore/Support/access_lock.hh
```diff
@@ -74,6 +75,10 @@ namespace litecore {
             auto operator -> ()             {return &_ref;}
 
         private:
+            friend access_lock;
+            access(MUTEX &mut, REF ref, SENTRY* pSentry):access(mut, ref) {
```
borrrden, line 79: Let's change these to `SENTRY&` as @snej suggested

borrrden (approved):

PR #1289, opened 2021-11-08
More header docs on C4Observer

snej:
Also updated the Doxygen-generated docs in docs/C/html/


bmeike (approved):
Looks great to me.
Do you commit the generated docs?

snej:
Yes.

borrrden:
Although they are now available on our official docs page so having them here is less of a need (they used to show up in the couchbaselabs github page but this org doesn't have a GH page).

PR #1294, opened 2021-11-12
CBL-2519 allow N1QL expression to contain newline as whitespace.

jianminzhao:


LiteCore/Query/N1QL_Parser/n1ql.leg
```diff
@@ -461,16 +461,17 @@ _  = [ \t\r\n]*            # Optional whitespace
...
-    ctx->stream->get(buf, max_size);
+    ctx->stream->get(buf, max_size, kN1QLEND);
     return (int) ctx->stream->gcount();
 }
 
 
 FLMutableDict parse(const string &input, unsigned *errPos) {
     MutableDict result;
-    std::stringstream stream(input, std::ios_base::in);
+    std::stringstream stream(input+kN1QLEND, std::ios_base::in);
```
borrrden, line 474: If the terminator string is going to be null, appending it is not needed, and will only hurt performance

borrrden (changes requested):

borrrden:
Also be sure to link these Lithium tickets to the RC1 ticket!!  The RC1 ticket is falling behind.

LiteCore/Query/N1QL_Parser/n1ql.leg
jianminzhao: Are you saying "input" already contains  '\0' at the end? Or you're saying, if '\0' does not exist in the input, it should matter for our purpose?

borrrden (approved):

PR #1297, opened 2021-11-17
CBL-2553: N1QL Meta().<property> column names

jianminzhao:
Meta().id will assume "id" as the column name, instead of "$#".


borrrden (commented):
I remember `meta()` also taking an argument inside of the parenthesis.  I think that situation should also have tests.

jianminzhao:
yes, good point.

pasin (approved):

PR #1299, opened 2021-11-18
CBL-2562: Try to put down some long running race conditions

jianminzhao:
The replicator will close its database after it is stopped, but RevFinder continues trying to use it.  Wire up RevFinder so that it blocks the Replicator from stopping until RevFinder is finished.

### DIFF
--- a/Replicator/Puller.hh
+++ b/Replicator/Puller.hh
@@ -80,6 +80,12 @@ namespace litecore { namespace repl {
         bool _caughtUp {false};             // Got all historic sequences, now up to date
         bool _fatalError {false};           // Have I gotten a fatal error?
 
+#if __APPLE__
+        // This helps limit the number of threads used by GCD:
+        virtual actor::Mailbox* mailboxForChildren() override       {return &_revMailbox;}
+        actor::Mailbox _revMailbox;
+#endif
+
         RemoteSequenceSet _missingSequences; // Known sequences I need to pull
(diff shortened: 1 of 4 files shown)


borrrden (commented):
This is pretty much what I suggested, but I think you are waiting for something else so I will defer the review.

borrrden:
Test in Jenkins is hanging and needs investigation....

borrrden:
I think I fixed the hanging issue with the above commit.

borrrden (approved):

Replicator/Puller.hh
```diff
@@ -93,7 +93,7 @@ namespace litecore { namespace repl {
         actor::ActorBatcher<Puller,IncomingRev> _returningRevs;
         Retained<Inserter> _inserter;
         mutable Retained<RevFinder> _revFinder;
-        Status _revFinderStatus {};
+        Status _revFinderStatus {kC4Idle};
```
jianminzhao, line 96: This seems to have violated the usual transition rule: worker always starts being Stopped, and transitioned to Idle after leaving Stopped. I understand your purpose. How about _revFinderStatus.level != kC4Idle || _revFinderStatus.level != kC4Stopped?

borrrden: This is not a rule, this is only a rule on certain workers.  Workers start with the rule that they are either busy or idle.

jianminzhao: Okay, understood.

PR #1303, opened 2021-11-22
CBL-2446 fix unpaired beginTransaction/endTransaction.

jianminzhao:
Fix the bug and also prevent possible omission of endTransaction due to stack unwinding.


borrrden (changes requested):
We already have an implementation for this kind of this in the `Defer.hh` header so let's use that one instead

borrrden (approved):

PR #1307, opened 2021-11-29
CBL-2591 bump fleece to release/lithium

jianminzhao:


borrrden (approved):
It's better to not include the ticket name in the branch name, etc in this case because it's already included in fleece.  Since fleece is a tracked project we will already get notifications on the ticket when a build is made containing the fleece fix.  Now we will get double notifications and it will get noisy.

jianminzhao:
Do you mean not in the branch name or in the commit title? Are notifications based on branch name?

borrrden:
The notifications are based on the commit message.

PR #1311, opened 2021-12-02
Partially undo the fix for 2562 which seems to cause hanging

borrrden:
Only in MinSizeRel, and only on Windows.  The remaining parts of 844b7b3622b05939aa1614929ad1897502acacf1 seem to address the problem of crashing due to null replicator dereference.

### DIFF
--- a/LiteCore/Support/Error.cc
+++ b/LiteCore/Support/Error.cc
@@ -588,7 +588,7 @@ namespace litecore {
             return true;
         switch (domain) {
             case LiteCore:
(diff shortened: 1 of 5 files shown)


jianminzhao:
Seems it undid most of 2562, as the title suggests, won't it easier future readers to simply revert 2562 and add new commit?

borrrden:
All that achieves in my opinion is making two commits instead of one...

borrrden:
Ok, now the two commits flow the way you describe

jianminzhao (approved):

PR #1313, opened 2021-12-02
CBL-2308: Lookup ICU symbols at runtime

borrrden:
This avoids the annoyance of an obnoxious decision to purposely break the ICU API every single release, which makes portability of the library between Linux platforms a gigantic chore because the only two options are:  Distribute the library bundled with the massive ICU libs (not an option for smaller Linux devices) or make a build for every single platform where the only difference is which libicu is being linked with (and therefore which arbitrary suffix the API calls end in).


borrrden:
Note: This doesn't actually enable the feature, it is meant to first be enabled by CBL-C, while left alone for LiteCore.

LiteCore/Android/unicode/ndk_icu.c
```diff
@@ -158,7 +158,7 @@ static void init_icudata_version() {
     syms[10] = dlsym(handle_i18n, func_name);
 }
 
-UCollator* ucol_open(const char* loc, UErrorCode* status) {
+UCollator* lc_ucol_open(const char* loc, UErrorCode* status) {
```
jianminzhao, line 161: I don't find how/where icu_shim.h is included into this file.

borrrden: It's not.  It doesn't make a difference to the build.

jianminzhao: yes, It simply defines the function. But with header included, the complier may help ensure some integrity.

borrrden: I don't see how it will help ensure anything, but I'll put it in anyway.

jianminzhao (approved):

pasin (approved):

snej:
I don't understand what's going on in this patch, so I don't feel I can review it. If you want a review from me, could you explain how this works?

borrrden:
The short answer is "weak linking".  ICU makes frequent and willful changes to its major version, and each time it does every function in the entire API is renamed arbitrarily.  The side effect of this is that to work using the OS package manager's provided ICU library, a build has to be made for every single version of every single OS that we support, usually the only difference being the ICU version linked (and the similarly "mangled" names).  

This PR looks up the version of ICU at runtime and loads the symbols instead so that we can make fewer builds that are more portable between Linux variants.  The version is in the name of the library, and all the API calls have this version number appended.

LiteCore/Android/unicode/ndk_icu.c
```diff
@@ -169,7 +169,7 @@ UCollator* ucol_open(const char* loc, UErrorCode* status) {
   return ptr(loc, status);
 }
 
-void ucol_setAttribute(UCollator* coll, UColAttribute attr, UColAttributeValue value, UErrorCode* status) {
+void lc_ucol_setAttribute(UCollator* coll, UColAttribute attr, UColAttributeValue value, UErrorCode* status) {
```
snej, line 173: What's the reason for renaming all of these? (I assume `lc_` stands for LiteCore?)

LiteCore/Unix/icu_shim.c
```diff
@@ -0,0 +1,302 @@
...
+// the triples we support on Linux
+#if defined(__x86_64__)
+#define ARCH_FOLDER "x86_64-linux-gnu"
+#elif defined(__aarch64__)
+#define ARCH_FOLDER "aarch64-linux-gnu"
+#elif defined(__ARM_ARCH_7A__)
+#define ARCH_FOLDER "arm-linux-gnueabihf"
+#endif
+
+
```
snej, line 42: This file looks like a copy of ndk_icu.c, at least most of its code is the same. That's a lot of duplication. Could we do this with a single source file?

snej (changes requested):
I'm still confused.

Also, isn't the reason ICU changes its symbol names because it also (sometimes) changes the function signatures or semantics? It it safe for us to bypass this by just assuming the functions all take the parameters we expect?

borrrden:
> I'm still confused.
> 
> Also, isn't the reason ICU changes its symbol names because it also (sometimes) changes the function signatures or semantics? It it safe for us to bypass this by just assuming the functions all take the parameters we expect?

If that's the case then it would be a mistake for them to do things the way that they do.  The headers all contain macros that resolve a function call to the "real" function call (which is, for example, `foo()` -> `foo_66()`).  So far we have used this style of shim for nearly the entirety of 2.x for the Android implementation.  It may break one day (so far it hasn't), but since we have a concrete list of platforms we will know if it does.

LiteCore/Android/unicode/ndk_icu.c
borrrden: The ICU headers in the actual release define these function names as macros, so compilation was getting hairy with the existing names.  I wanted to make them distinct from that.

LiteCore/Unix/icu_shim.c
borrrden: There is a good chunk of its code that is the same.  I thought it was kind of nice having a straight up copy and paste for ndk_icu.c (which is specific to Android in the way that it looks up the shared library) with the license intact and this modified one for desktop platforms but there could be some refactoring done.  It will need to balance `#ifdef` with separate files as usual though.

snej (approved):

borrrden:
Had a look at the GH actions failure.  Looks like a network connectivity issue not related to the PR.

PR #1314, opened 2021-12-03
CBL-2614

jianminzhao:
1. Fix the order of _revMailbox initialization since passing "this" to revFinder can lead to it being used before _revMailbox is ready
2. Guard against a null deference in revfinder
3. Make NotOpen unremarkable so that in the event that the replicator is stopped with database operations still pending, the log does not get spammed with dozens of "Throwing LiteCore Error" messages (these will get swallowed anyway, and this behavior actually causes a long enough delay between the end of some tests in CppTests and the teardown to make the test fail due to trying to delete an in-use database). This fixes a ramificatino of cd06d35d4f2650d12da4313dce765e4c9c4c8a06.

### DIFF
--- a/Replicator/Puller.cc
+++ b/Replicator/Puller.cc
@@ -39,13 +39,15 @@ namespace litecore { namespace repl {
 
     Puller::Puller(Replicator *replicator)
     :Delegate(replicator, "Pull")
+#if __APPLE__
+    // This field must go before _revFinder because "this" is passed in "new RevFinder(replicator, this)," which will
+    // call this->mailboxForChildren() which depends on it.
+    ,_revMailbox(nullptr, "Puller revisions")
+#endif
     ,_inserter(new Inserter(replicator))
     ,_revFinder(new RevFinder(replicator, this))
     ,_provisionallyHandledRevs(this, "provisionallyHandledRevs", &Puller::_revsWereProvisionallyHandled)
     ,_returningRevs(this, "returningRevs", &Puller::_revsFinished)
-#if __APPLE__
-    ,_revMailbox(nullptr, "Puller revisions")
-#endif
     {
         _passive = _options.pull <= kC4Passive;
         registerHandler("rev",              &Puller::handleRev);
--- a/Replicator/Puller.cc
+++ b/Replicator/Puller.cc
@@ -40,8 +40,6 @@ namespace litecore { namespace repl {
     Puller::Puller(Replicator *replicator)
     :Delegate(replicator, "Pull")
 #if __APPLE__
-    // This field must go before _revFinder because "this" is passed in "new RevFinder(replicator, this)," which will
-    // call this->mailboxForChildren() which depends on it.
     ,_revMailbox(nullptr, "Puller revisions")
 #endif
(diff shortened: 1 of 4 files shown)


Replicator/Puller.cc
```diff
@@ -39,13 +39,15 @@ namespace litecore { namespace repl {
 
     Puller::Puller(Replicator *replicator)
     :Delegate(replicator, "Pull")
+#if __APPLE__
+    // This field must go before _revFinder because "this" is passed in "new RevFinder(replicator, this)," which will
```
snej, line 43: FYI, this order actually doesn't matter; I looked it up to make sure:

> The order of member initializers in the list is irrelevant  ... non-static data members are initialized in order of declaration in the class definition

So it's the order in the header that matters. But it's a good idea to have the same ordering in the implementation to avoid confusion!

This does mean that your explanatory comment should actually go in the declaration in the header instead.

Replicator/RevFinder.cc
```diff
@@ -281,9 +281,15 @@ namespace litecore::repl {
                     logDebug("    - Already have '%.*s' %.*s but need to mark it as remote ancestor",
                              SPLAT(docID), SPLAT(revID));
                     _db->setDocRemoteAncestor(docID, revID);
-                    if (!_passive && !_db->usingVersionVectors())
-                        replicator()->docRemoteAncestorChanged(alloc_slice(docID),
-                                                               alloc_slice(revID));
+                    if (!_passive && !_db->usingVersionVectors()) {
+                        auto repl = replicatorIfAny();
+                        if(repl) {
+                            replicator()->docRemoteAncestorChanged(alloc_slice(docID),
```
snej, line 287: Shouldn't this call `repl` instead of `replicator()`?

Replicator/RevFinder.cc
```diff
@@ -281,9 +281,15 @@ namespace litecore::repl {
...
+                                                                   alloc_slice(revID));
+                        } else {
+                            Warn("findRevs no longer has a replicator reference (replicator stopped?), ignoring docRemoteAncestorChange callback");
```
snej, line 290: Please try to limit lines to 100 chars.

snej (changes requested):

Replicator/Puller.cc
jianminzhao: Accepted.

snej:
Please resolve the other two issues I pointed out. Thanks!

Replicator/RevFinder.cc
jianminzhao: yes. limitation of manual porting. Thanks a lot. Otherwise it is going to be a random bomb.

Replicator/RevFinder.cc
jianminzhao: Accepted.

snej (approved):

borrrden (approved):
Before actually committing, please make sure that the first line of the final message is not just "CBL-2614".  It should contain a short description of the issue on the first line, since that is what is visible by default in github and git graphical programs.

jianminzhao:
Yes, will do it.

PR #1317, opened 2021-12-07
CBL-2374: Upgrade docs to 3.0 schema lazily

snej:
Version 3 changes the document schema in the database, adding an
'extra' column that stores the rev-tree, while the 'body' column just
contains the current revision's Fleece body.

Rewriting all the docs in the database to the new schema the first
time it's opened by 3.0 turns out to be too slow for very large
databases.

With this commit we skip the rewrite. Instead, the database reads
documents in either the v2 or v3 format. But docs are always saved in
v3, so over time the database will migrate, and its performance will
improve as less document data is loaded.

CBL-2374

### DIFF
--- a/LiteCore/RevTrees/RawRevTree.cc
+++ b/LiteCore/RevTrees/RawRevTree.cc
@@ -29,6 +29,29 @@ namespace litecore {
 #pragma pack()
 
 
+    bool RawRevision::isRevTree(slice raw_tree) {
+        // The data cannot be shorter than a single revision:
+        if (raw_tree.size < sizeof(RawRevision))
+            return false;
+        // Data must end with a 32-bit 0:
+        auto end = (const void*)offsetby(raw_tree.end(), -int(sizeof(uint32_t)));
+        uint32_t ending;
+        memcpy(&ending, end, sizeof(ending));
+        if (ending != 0)
+            return false;
+        // Check each revision:
+        const RawRevision *next;
+        for (auto rawRev = (const RawRevision*)raw_tree.buf; rawRev < end; rawRev = next) {
+            next = rawRev->next();
+            // Rev can't be too short for data, or extend past end of data:...
+            if (next > end || next <= (void*)&rawRev->revID[rawRev->revIDLen])
+                return false;
+        }
+        return true;
+    }
+
+
+
     std::deque<Rev> RawRevision::decodeTree(slice raw_tree,
                                             RevTree::RemoteRevMap &remoteMap,
                                             RevTree* owner,
--- a/C/c4Database.cc
+++ b/C/c4Database.cc
@@ -75,7 +75,7 @@ static C4DatabaseConfig newToOldConfig(const C4DatabaseConfig2 &config2) {
     return C4DatabaseConfig {
         config2.flags | kC4DB_AutoCompact,
         NULL,
-        (config2.flags & kC4DB_VersionVectors) ? kC4VectorVersioning : kC4TreeVersioning,
+        (config2.flags & kC4DB_VersionVectors) ? kC4VectorVersioning : kC4TreeVersioning_v2,
         config2.encryptionKey
     };
 }
--- a/C/tests/c4DatabaseTest.cc
+++ b/C/tests/c4DatabaseTest.cc
@@ -16,6 +16,7 @@
 #include "c4BlobStore.h"
 #include "c4Index.h"
 #include "c4IndexTypes.h"
+#include "c4Query.h"
 #include "FilePath.hh"
 #include "SecureRandomize.hh"
 #include <cmath>
@@ -881,7 +882,7 @@ static void testOpeningOlderDBFixture(const string & dbPath,
     C4Log("---- Opening copy of db %s with flags 0x%x", dbPath.c_str(), withFlags);
     C4DatabaseConfig2 config = {slice(TempDir()), withFlags};
     C4Error error;
-    C4Database *db;
+    c4::ref<C4Database> db;
     auto name = C4Test::copyFixtureDB(kVersionedFixturesSubDir + dbPath);
 
     if (expectedErrorCode == 0) {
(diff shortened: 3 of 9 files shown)


LiteCore/RevTrees/RawRevTree.cc
```diff
@@ -29,6 +29,29 @@ namespace litecore {
 #pragma pack()
 
 
+    bool RawRevision::isRevTree(slice raw_tree) {
+        // The data cannot be shorter than a single revision:
+        if (raw_tree.size < sizeof(RawRevision))
+            return false;
+        // Data must end with a 32-bit 0:
+        auto end = (const void*)offsetby(raw_tree.end(), -sizeof(uint32_t));
```
borrrden, line 37: "The result of a unary negative operator on an unsigned type is still unsigned" according to Windows.

jianminzhao:
Trying to understand the trick. As we open an v2 db, with this PR, will upgradeDocumentVersioning() be called or skipped?

snej:
Skipped. In `DatabaseImpl::checkDocumentVersioning`, `newVersioning` will now be `kC4TreeVersioning_v2`, not `kC4TreeVersioning`, so the method will exit early without upgrading.

borrrden:
Also just a small annoying point of order:  Since the linked ticket is for Lithium this should be a PR against release/lithium (with a cloned ticket later for master) or vice-versa.  If this is merged as-is, the linked ticket should have its fix version changed.

jianminzhao:
I imaged this, for that updating method does not seem to be changed. Thanks for clarifying it. So, when will v2 db be actually upgraded?

borrrden:
@jianminzhao See the original message:

> With this commit we skip the rewrite. Instead, the database reads
documents in either the v2 or v3 format. But docs are always saved in
v3, so over time the database will migrate, and its performance will
improve as less document data is loaded.

jianminzhao:
So, suppose we have a large db of v2, db_v2. Now we open it, perform an epsilon change -- add one record, say -- then, save. Will this bring us to the same story now? The flip of this question is, if we open it, query, no modifying, will the db stay in v2?

snej:
With this commit, `kC4TreeVersioning_v2` now means "the database _may_ contain v2 bodies". A newly created database will have `kC4TreeVersioning`.

It's possible to upgrade a database fully to `kC4TreeVersioning` (v3), but LiteCore doesn't expose any API to do it. I'm assuming that's not needed.

jianminzhao:
So a v2 db will be saved in v3 schema with empty "extra." Am I right?

borrrden:
> So, suppose we have a large db of v2, db_v2. Now we open it, perform an epsilon change -- add one record, say -- then, save. Will this bring us to the same story now? The flip of this question is, if we open it, query, no modifying, will the db stay in v2?

In the first scenario, only the saved document will be in the new format.  In the second nothing changes.  Each document will be updated one at a time the next time they are saved, and reads can handle either format.

jianminzhao:
I am trying to understand it. In v2, db schema does not have extra. As we open the v2 db in V3 Lite, we add column, "extra." Currently, as we open it first time, we decompose v2 body to v3 body + extra. This way, the code can assume uniform view of body+extra, but It takes toll from performance.
The new approach, won't touch the v2 body; the db lays out as, body: v2-body, extra: empty. No cost opening the db, but the code has to be ready for two types of bodies.
As we save new document, it will be laid out as, body: v3-body, extra: rev-tree
Am I right? As time goes, we will have many v2-bodies and v3-bodies mingled in the same db.
Question: what's the impact to the index?
Originally, we move the bodies from v2 to v3 body+extra, without touching the index, and it had problems so that we had to delete indexes -> move bodies -> recreate indexes. Is this no longer an issue in the new approach?

borrrden:
The reason for recreating the index like that is because having an index slows down writes a little and so when we are rewriting every document it takes quite a bit longer.  That won't be an issue with this approach since there is no more mass rewriting.  The indexes can simply remain in place.  Indexes don't really care what is actually stored in the data, since the information the index needs is stored separately.

jianminzhao:
I am not concerned about the performance here. The aforementioned problem is that if we had kept the index intact, we would have problem reading the properties of the document. Of course, we have different context here: we do not move the documents that are with the index together in the v2 db, and the index caused the problem when we moved the documents to new schema. I hope it wouldn't affect us if only new documents uses the new schema.

borrrden:
Part of the upgrade process, according to the code, still appears to be deleting and recreating the index.  So I'm unsure what you are referring to.  Perhaps you could provide a way to alter or create a test to prove what you say?

snej:
> Am I right? As time goes, we will have many v2-bodies and v3-bodies mingled in the same db.

That's right.

> Question: what's the impact to the index?
Originally, we move the bodies from v2 to v3 body+extra, without touching the index, and it had problems so that we had to delete indexes -> move bodies -> recreate indexes. Is this no longer an issue in the new approach?

tl;dr: This is no longer an issue because this PR upgrades the query code to understand both old and new `body` column values.

Long answer: The problem with keeping indexes during the conversion was that the query code did not understand the old layout and treated the `body` column as pure Fleece. When a SQLite row is updated, the indexer has to evaluate the index's expression on both the old and new row values. But during the upgrade, the old row value was in the old schema, which the query code couldn't understand, so it failed. The query code now understands the old schema.

It's still faster to drop and re-create the indexes during a full DB upgrade, which is why that code is still there. But the upgrade code won't be run in 3.0, not until we support version vectors.

jianminzhao:
Super. Very good explanation.

jianminzhao:
@snej Here is what I am doing, after yesterday's conversation with Jim. (Because of the combo dentist appointment, I virtually just started.)
- Based on the original test that tests upgrading db with indexes (of course, with this PR, it's a no-op).
- close db;
- modify a doc;
- add a new doc
- close db;
- check the index (as the test did currently). (We should have one updated doc (from v2) and a new doc (fresh v3).

snej:
@borrrden Jenkins tests failed with the process getting a SIGTERM. Does that mean Jenkins timed out the build?

borrrden:
Yes, the mac machine has been incredibly slow lately.  The iOS build took 32 minutes, leaving only 8 for the others.  The telltale signs of a timeout are a grey line icon instead of a red X in the blue ocean details.

jianminzhao (approved):

PR #1324, opened 2021-12-09
CBL-2583: Enhance proposeChanges message to accept revID in response

borrrden:
Combined with enhancements in the passive puller, and Sync Gateway, a pusher can now recover from a false conflict or unknown ancestor situation without the assistance of an active puller.  When active, the new format for a 409 response to a `proposeChanges` message will be `{"status": 409, "rev": "<current-rev-id>"}`

### DIFF
--- a/Replicator/Pusher.cc
+++ b/Replicator/Pusher.cc
@@ -459,22 +459,16 @@ namespace litecore { namespace repl {
                 if(receivedRevID && receivedRevID != rev->remoteAncestorRevID) {
                     // Remote ancestor received in proposeChanges response, so try with 
                     // this one instead
-                    try {
-                        // If receivedRevID is newer than rev->revID this will throw since locally we don't have
-                        // that rev yet.  So the only way this function succeeds is if receivedRevID is equal to
-                        // or an ancestor of rev->revID
-                        if(doc->selectCommonAncestorRevision(rev->revID, receivedRevID)) {
-                            logInfo("Remote reported different rev of '%.*s' (mine: %.*s theirs: %.*s); retrying push",
-                                SPLAT(rev->docID), SPLAT(rev->remoteAncestorRevID), SPLAT(receivedRevID));
-                            rev->remoteAncestorRevID = receivedRevID;
-                            return true;
-                        }
-                    } catch(error& e) {
-                        // Not Found error should be caught, because it means that receivedRevID is not
-                        // a part of the rev tree history
-                        if(e.domain != error::Domain::LiteCore || e.code != error::NotFound) {
-                            throw;
-                        }
+
+                    // If the first portion of this test passes, then the rev exists in the tree.
+                    // If the second portion passes, then receivedRevID is an ancestor of the 
+                    // current rev ID and it is usable for a retry.
+                    if(doc->selectRevision(receivedRevID, false) && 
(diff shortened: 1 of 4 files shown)


borrrden:
Jenkins failed because of time out.  Suggest increasing it....

Replicator/Pusher.cc
```diff
@@ -436,15 +450,40 @@ namespace litecore { namespace repl {
...
             Retained<C4Document> doc = _db->getDoc(rev->docID, kDocGetAll);
             if (doc && C4Document::equalRevIDs(doc->revID(), rev->revID)) {
+                if(receivedRevID && receivedRevID != rev->remoteAncestorRevID) {
+                    // Remote ancestor received in proposeChanges response, so try with 
+                    // this one instead
+                    try {
+                        // If receivedRevID is newer than rev->revID this will throw since locally we don't have
+                        // that rev yet.  So the only way this function succeeds is if receivedRevID is equal to
+                        // or an ancestor of rev->revID
+                        if(doc->selectCommonAncestorRevision(rev->revID, receivedRevID)) {
```
snej, line 466: I don't like the idea of throwing an exception in normal operation. Why not preflight by testing `if (doc->selectRevision(receivedRevID))` instead?

Replicator/RevFinder.cc
```diff
@@ -141,7 +142,7 @@ namespace litecore::repl {
 
                 auto &encoder = response.jsonBody();
                 encoder.beginArray();
-                int requested = proposed ? findProposedRevs(changes, encoder, sequences)
+                int requested = proposed ? findProposedRevs(changes, encoder, req->boolProperty(Pusher::kConflictIncludesRevProperty), sequences)
```
snej, line 145: This line looks way longer than 100 chars

snej (changes requested):

Replicator/Pusher.cc
```diff
@@ -225,6 +225,10 @@ namespace litecore { namespace repl {
     // Sends a "changes" or "proposeChanges" message.
     void Pusher::sendChanges(RevToSendList &changes) {
         MessageBuilder req(_proposeChanges ? "proposeChanges"_sl : "changes"_sl);
+        if(_proposeChanges) {
+            req[kConflictIncludesRevProperty] = "true"_sl;
```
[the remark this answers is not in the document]
borrrden: The property name was based on the spec I wrote and send out in an email for feedback.  If it changes now it will need a Sync Gateway patch as well (for either of these changes).  Do you still want to go forward with that?

Replicator/Pusher.cc
borrrden: I like this idea much better.  I didn't like the idea of doing a try catch inside a try catch either....

Replicator/RevFinder.cc
borrrden: Oh!  I wasn't yet aware of this style guideline, I'll keep that in mind.

Replicator/Pusher.cc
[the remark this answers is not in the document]
snej: No, oh well! I'm sorry I didn't catch this earlier.

snej (approved):

PR #1325, opened 2021-12-10
CBL-2638: Upgrade docs to 3.0 schema lazily

jianminzhao:
Version 3 changes the document schema in the database, adding an
'extra' column that stores the rev-tree, while the 'body' column just
contains the current revision's Fleece body.

Rewriting all the docs in the database to the new schema the first
time it's opened by 3.0 turns out to be too slow for very large
databases.

With this commit we skip the rewrite. Instead, the database reads
documents in either the v2 or v3 format. But docs are always saved in
v3, so over time the database will migrate, and its performance will
improve as less document data is loaded.

Cherry-picked from CBL 2374.


borrrden (approved):
Shouldn't be a problem but wait for the PR validation just in case there are any hidden surprises.

PR #1326, opened 2021-12-10
CBL-2639 augmenting a test case, Database upgrade from 2.8.

jianminzhao:
We augment it to test the logic of CBL 2638.

### DIFF
--- a/C/tests/c4DatabaseTest.cc
+++ b/C/tests/c4DatabaseTest.cc
@@ -12,12 +12,15 @@
 
 #include "c4Test.hh"
 #include "c4Private.h"
+#include "c4Collection.h"
 #include "c4DocEnumerator.h"
 #include "c4BlobStore.h"
 #include "c4Index.h"
 #include "c4IndexTypes.h"
 #include "FilePath.hh"
+#include "fleece/Mutable.hh"
 #include "SecureRandomize.hh"
+#include <array>
 #include <cmath>
 #include <errno.h>
 #include <iostream>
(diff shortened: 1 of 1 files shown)


C/tests/c4DatabaseTest.cc
```diff
@@ -947,19 +950,133 @@ TEST_CASE("Database Upgrade From 2.8", "[Database][Upgrade][C]") {
...
+    c4::ref<C4Document> doc3;
+    {
+        TransactionHelper t(db);
+        doc3 = c4coll_putDoc(c4db_getDefaultCollection(db), &rq, nullptr, &err);
+    }
+    CHECK(doc3);
+    CHECK(slice(doc3->revID).asString().find("1-") == 0);
+    // Make sure the db now contains 3 documents, and re-check the index.
+    CHECK(c4db_getDocumentCount(db) == 3);
+    checkIndex(db);
```
borrrden, line 1079: This index check adds nothing to the test.

C/tests/c4DatabaseTest.cc
```diff
@@ -947,19 +950,133 @@ TEST_CASE("Database Upgrade From 2.8", "[Database][Upgrade][C]") {
...
+
+    REQUIRE(c4db_close(db, WITH_ERROR()));
+
+    // Reopen the database and add a new document
+
+    db = c4db_openNamed(name, &config, WITH_ERROR(&err));
+    CHECK(db);
+    CHECK(c4db_getDocumentCount(db) == 2);
+
+    // Add a new document
```
borrrden, line 1059: What value does this last portion add to the test?  After modifying the document above there is already a v2 and v3 document in the database.

C/tests/c4DatabaseTest.cc
```diff
@@ -947,19 +950,133 @@ TEST_CASE("Database Upgrade From 2.8", "[Database][Upgrade][C]") {
...
+    rq.docID = doc1ID;
+    rq.allocedBody = fl_dict2;
+    C4String history[1] = {revID1};
+    rq.history = history;
+    rq.historyCount = 1;
+    rq.save = true;
+    {
+        TransactionHelper t(db);
+        doc1 = c4coll_putDoc(c4db_getDefaultCollection(db), &rq, nullptr, &err);
+    }
```
borrrden, line 1044: This is reinventing `C4Test::createRev`

C/tests/c4DatabaseTest.cc
```diff
@@ -947,19 +950,133 @@ TEST_CASE("Database Upgrade From 2.8", "[Database][Upgrade][C]") {
...
+    CHECK(doc1);
+    CHECK(slice(doc1->revID).asString().find("2-") == 0);
```
borrrden, line 1046: I don't think this is relevant to the test.  The test isn't here to test everything about the way that the DB works.

C/tests/c4DatabaseTest.cc
```diff
@@ -947,19 +950,133 @@ TEST_CASE("Database Upgrade From 2.8", "[Database][Upgrade][C]") {
...
+    // Make sure the db still contains 2 documents, and re-check the index.
+    CHECK(c4db_getDocumentCount(db) == 2);
+    checkIndex(db);
```
borrrden, line 1049: There is no value in checking the index existence multiple times.  There was never a case in which the index vanished.

borrrden (changes requested):
Huge scope creep here.  There is no need to test every aspect of the library's functionality here.  The issue at hand is:  Can we modify / add new documents after the upgrade and will a query still return the correct results?

C/tests/c4DatabaseTest.cc
jianminzhao: That's the old test (index test), which would fail fail if we did not re-index before cbl 2638.

jianminzhao: Wanted to make sure the index is still correct after A. we modify one document, namely adding a rev to existing document, and B. add a new document.

borrrden: Logically there is no difference to this test between modifying a document and adding a document.  Checking the index multiple times seems redundant.  Checking it once after a mutation should be sufficient.

borrrden (commented):
What has changed since my last review?

jianminzhao:
"That's the test plan. Would you agree?"

borrrden:
I left 4 comments, and the last in each is still my comment.  What do you have to say about those?

jianminzhao:
The purpose of the test: there are three checks on index, checkIndex().
* The first checkIndex() existed before CBL-2638, prior to which we called DatabaseImpl::upgradeDocumentVersioning() whenever we read in v2 database. With this method, checkIndex() would fail if we did not touch the index. checkIndex() passed as we re-created the index after moving all the kv records.
* CBL-2638 does away with DatabaseImpl::upgradeDocumentVersioning() . Therefore, the first check should automatically pass.
* Now that we don't migrate the old records (v2 records), the second IndexCheck() assures that it passes after an old document is added a new revision.
* The third checkIndex() assures that it passes after a new document is added to the database.

borrrden:
I assert that mutating an existing document and adding a new document are testing the same thing in terms of this test, unless you can point out a specific difference.

jianminzhao:
It logically sounds correct since they all amount to creating a new record with the new schema, with one insert and the other update. I will remove the third checkIndex.

C/tests/c4DatabaseTest.cc
```diff
@@ -12,12 +12,15 @@
 
 #include "c4Test.hh"
 #include "c4Private.h"
+#include "c4Collection.h"
```
borrrden, line 15: I think this header is no longer needed?

C/tests/c4DatabaseTest.cc
```diff
@@ -947,19 +950,93 @@ TEST_CASE("Database Upgrade From 2.8", "[Database][Upgrade][C]") {
...
+                if (i >= keys.size()) {
+                    return;
+                }
+                CHECK(it.keyString().asString() == keys[i].asString());
+                CHECK(it.value().asString().asString() == values[i].asString());
+                ++i;
+            }
+        };
+
+    checkDict(rev1Dict, {"firstName"_sl, "lastName"_sl}, {"fName"_sl, "lName"_sl});
```
borrrden, line 1009: This is testing pure KV reading, and is also extraneous to the test.  Reading from the DB does not involve the index.  The other upgrade tests will simply check the doc IDs and previously this test didn't check reading at all.

borrrden (changes requested):

borrrden (commented):
This looks good but I don't want to merge this into this branch now that our internal RC has happened.  It will trigger another build and cause confusion.

C/tests/c4DatabaseTest.cc
```diff
@@ -947,19 +949,86 @@ TEST_CASE("Database Upgrade From 2.8", "[Database][Upgrade][C]") {
...
+    CHECK(c4enum_next(e, ERROR_INFO(&err)));
+
+    c4::ref<C4Document> doc1 = c4enum_getDocument(e, &err);
+    CHECK(doc1);
+    alloc_slice doc1ID = doc1->docID;
+    alloc_slice revID1 = doc1->revID;
+
+    // Get the revision body
+
+    Doc fl_docRev1(alloc_slice(c4doc_getRevisionBody(doc1)), kFLTrusted, c4db_getFLSharedKeys(db));
```
snej, line 991: It's better to use `c4doc_getProperties`

C/tests/c4DatabaseTest.cc
```diff
@@ -947,19 +949,86 @@ TEST_CASE("Database Upgrade From 2.8", "[Database][Upgrade][C]") {
...
+        FLEncoder_WriteValue(enc, dict);
+        FLError flError = kFLNoError;
+        auto ret = FLEncoder_Finish(enc, &flError);
+        CHECK(flError == kFLNoError);
+        return ret;
+    };
+
+    // Encode the modified dict:
+    FLSliceResult fl_dict2 = encodeDict(db, mdict2);
+
```
snej, line 1023: Please use the utility methods in C4Test to update the document. I don't see a need to write all this code just to test updating a document,

snej (changes requested):

PR #1330, opened 2021-12-15
CBL-2675: Puller needs to consider the kSynced flag

jianminzhao:
The kSynced flag is a shortcut flag on a document that indicates that the latest revision has been pushed to the remote, but that this information has not yet been recorded into the body of the document (i.e. the document has not been saved since).  The pusher uses a method which consolidates this information when deciding which revision to send as the ancestor in a proposeChanges message.  The puller also needs to consolidate this information when deciding whether or not a change is properly set as the remote ancestor when a changes message comes from the other side because otherwise it will sometimes falsely believe it already has things correctly set by scanning the document body, but not considering the kSynced flag.
This fix is ported here from release/lithium, 46fea9da79548a9c1488269253bc96b906e50f82.

### DIFF
--- a/LiteCore/Database/TreeDocument.cc
+++ b/LiteCore/Database/TreeDocument.cc
@@ -584,7 +584,7 @@ namespace c4Internal {
             revMap[docIDs[i]] = revIDs[i];
         stringstream result;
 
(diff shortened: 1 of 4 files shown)


borrrden (approved):

borrrden:
After merge I'll run this against the one-off test program for this issue to verify.

PR #1332, opened 2021-12-16
CBL-2678 : Fix LiveQuerier and C4Query leak

pasin:
* Called liveQuerierStopped() regardless of _query being initialized or not so that the LiveQuerier and C4Query will be cleaned up afters stopped.

* Made sure that _stop() will be called only once by checking whether the _stopping has already be set to true in stop().


borrrden (approved):
I'm assuming that the failing CBL-C tests were run with this.

pasin:
Yes, I did run it with all C tests.

PR #1335, opened 2021-12-17
Fixed query crashes in upgraded databases (CBL-2680, CBL-2682)

snej:
* Fixed the test for whether a slice contains a serialized RevTree.
  When I wrote it I'd forgotten that in a replicated doc, the remote
  rev mapping follows the rev-tree, and that broke the test.
* When grabbing the rev data out of a serialized rev tree, it may be
  at an odd address, which breaks Fleece. I had to reinstitute the old
  workaround that copies the data to a new heap block.
* SQLiteFleeceEach may also read a direct doc body and needs the same
  logic as QueryFleeceScope. (It can't directly use a QueryFleeceScope
  because it doesn't have access to a `sqlite3_context*`.)
* `fl_root` also needs the same transformation of doc bodies.

CBL-2680
CBL-2682

### DIFF
--- a/LiteCore/Query/SQLiteFleeceEach.cc
+++ b/LiteCore/Query/SQLiteFleeceEach.cc
@@ -75,6 +75,7 @@ private:
     // Instance data:
     FleeceVTab* _vtab;                  // The virtual table
     optional<Scope> _scope;             // Fleece document
+    bool _scopeDataIsCopied {false};    // If true, _scope.data() is a malloc'ed block
     alloc_slice _rootPath;              // The path string within the data, if any
     const Value *_container;            // The object being iterated (target of the path)
     valueType _containerType;           // The value type of _container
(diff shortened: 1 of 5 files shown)


borrrden:
External manual validation with the N1QL test suite showed 20 errors before this PR due to either "datatype mismatch" or "corrupt fleece data".  Post-PR the errors are gone and 244 queries are successful.

jianminzhao (approved):

jianminzhao:
@borrrden how long does it take to run the external test suite? If this test suite had been par of PR validation, we would have found it before RC1.

borrrden:
It takes 40 seconds to run, but only by pure chance did it have the ability to catch the problem this time.  It is not meant to catch problems like this, but rather validate that N1QL still gives conformant answers to the equivalent server query.

PR #1337, opened 2021-12-20
LiveQuery fix for CBL-2692 and CBL-2693

pasin:
CBL-2692 is cloned from Lithium's CBL-2673.
CBL-2693 is cloned from Lithium's CBL-2678.


borrrden:
In this case these are supposed to be cloned and get their own ticket numbers.  These tickets are already resolved.

pasin:
OK

jianminzhao (approved):

PR #1342, opened 2022-01-04
CBL-2639: add a test case for upgrading db with index to v3.

jianminzhao:


jianminzhao:
Hi @snej , happy new year! Remember that I had to regenerate the index when upgrading v2 db to v3? With your recent fix, I think it is no longer needed. The above PR is a test case for that:
1. I reverted the commit that delete and regenerate the index when moving all the v2 records to v3 schema. 
2. Tested it that (1) works, in the case of version vector.
3. Tested we don't have to do (1), namely not having to touch the existent v2 records, in the case revision tree.

snej (approved):

PR #1344, opened 2022-01-06
Type-safe C4SequenceNumber and C4Timestamp

snej:
When `LITECORE_CPP_API` is defined as 1 — which is true when you include c4Base.hh, and I’m now predefining it in the LiteCore build — the type `C4SequenceNumber` is now a C++ `enum class` instead of a simple typedef of `uint64_t`. So is the internal `sequence_t`, which is now a typedef of `C4SequenceNumber`.

This means it’s no longer interchangeable with an integer at all. You have to cast it to convert to or from an integer.

There are operator overloads that allow:
* C4SequenceNumber +/- uint64_t —> C4SequenceNumber
* C4SequenceNumber - C4SequenceNumber —> uint64_t
* ++C4SequenceNumber, --C4SequenceNumber, C4SequenceNumber++, C4SequenceNumber--

There’s also a C++ literal suffix `_seq` that when appended to an integer literal makes it a C4SequenceNumber, e.g. "`0_seq`".

A second commit gives `C4Timestamp` (aka `expiration_t`) the same treatment, except you can't increment or decrement it because that would make little sense.

### DIFF
--- a/C/Cpp_include/c4EnumUtil.hh
+++ b/C/Cpp_include/c4EnumUtil.hh
@@ -1,0 +1,38 @@
+//
+// c4EnumUtil.hh
+//
+// Copyright 2021-Present Couchbase, Inc.
+//
+// Use of this software is governed by the Business Source License included
+// in the file licenses/BSL-Couchbase.txt.  As of the Change Date specified
+// in that file, in accordance with the Business Source License, use of this
+// software will be governed by the Apache License, Version 2.0, included in
+// the file licenses/APL2.txt.
+//
+
+#pragma once
+#include <type_traits>
+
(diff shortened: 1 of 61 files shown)


jianminzhao (approved):

C/Cpp_include/c4EnumUtil.hh
```diff
@@ -0,0 +1,32 @@
+//
+// c4EnumUtil.hh
+//
+// Copyright © 2022 Couchbase. All rights reserved.
+//
```
borrrden, line 5: Copyright needs fixing

snej: done

borrrden (approved):

PR #1346, opened 2022-01-10
CBL-2575: catch db exception properly in Replicator.cc

jianminzhao:
Fixing Repicator::isDocumentPending and Replicator::pendingDocumentIDs.

### DIFF
--- a/Replicator/Replicator.cc
+++ b/Replicator/Replicator.cc
@@ -664,9 +664,19 @@ namespace litecore { namespace repl {
         if(!db) {
             return false;
         }
-        
-        _checkpointer.pendingDocumentIDs(db->useLocked(), callback);
-        return true;
+
+        try {
+            db->useLocked([this, callback](const Retained<C4Database>& db) {
+                _checkpointer.pendingDocumentIDs(db, callback);
+            });
+            return true;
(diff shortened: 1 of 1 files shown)


Replicator/Replicator.cc
```diff
@@ -664,9 +664,19 @@ namespace litecore { namespace repl {
         if(!db) {
             return false;
         }
-        
-        _checkpointer.pendingDocumentIDs(db->useLocked(), callback);
-        return true;
+
+        try {
+            db->useLocked([this, callback](const Retained<C4Database>& db) {
```
snej, line 669: It should be OK to keep the shorter `db->useLocked()` mode ... it unlocks the db if there's an exception.
(If you found that it didn't, that's an important bug; let me know!)

snej (approved):

Replicator/Replicator.cc
jianminzhao: Use db->useLocked() will throw. However, I want to protect the situation when at the time calling db->useLocked(), it's in good state but when we use the db from the return, it is subject to other thread nulling it. By using this version, I will protect it because I am holding the mutex during entire time the db is used.

PR #1349, opened 2022-01-13
Fixed query crashes in upgraded databases

jianminzhao:
* Fixed the test for whether a slice contains a serialized RevTree.
  When I wrote it I'd forgotten that in a replicated doc, the remote
  rev mapping follows the rev-tree, and that broke the test.
* When grabbing the rev data out of a serialized rev tree, it may be
  at an odd address, which breaks Fleece. I had to reinstitute the old
  workaround that copies the data to a new heap block.
* SQLiteFleeceEach may also read a direct doc body and needs the same
  logic as QueryFleeceScope. (It can't directly use a QueryFleeceScope
  because it doesn't have access to a `sqlite3_context*`.)
* `fl_root` also needs the same transformation of doc bodies.

### DIFF
--- a/LiteCore/Query/SQLiteFleeceUtil.cc
+++ b/LiteCore/Query/SQLiteFleeceUtil.cc
@@ -32,19 +32,27 @@ namespace litecore {
     const char* const kFleeceValuePointerType = "FleeceValue";
 
 
-    static inline slice argAsDocBody(sqlite3_context* ctx, sqlite3_value *arg) {
+    slice valueAsDocBody(sqlite3_value *arg, bool &outCopied) {
         auto type = sqlite3_value_type(arg);
         if (_usuallyFalse(type == SQLITE_NULL))
             return nullslice;             // No 'body' column; may be deleted doc
         DebugAssert(type == SQLITE_BLOB);
         DebugAssert(sqlite3_value_subtype(arg) == 0);
-        auto body = valueAsSlice(arg);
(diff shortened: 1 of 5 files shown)


jianminzhao:
This is automatically done by cherry-pick from lithium,  03524fed93bd96

LiteCore/Query/SQLiteFleeceUtil.cc
```diff
@@ -32,19 +32,27 @@ namespace litecore {
...
+    slice valueAsDocBody(sqlite3_value *arg, bool &outCopied) {
         auto type = sqlite3_value_type(arg);
         if (_usuallyFalse(type == SQLITE_NULL))
             return nullslice;             // No 'body' column; may be deleted doc
         DebugAssert(type == SQLITE_BLOB);
         DebugAssert(sqlite3_value_subtype(arg) == 0);
-        auto body = valueAsSlice(arg);
-        if (RawRevision::isRevTree(body)) {
+        auto fleece = valueAsSlice(arg);
+        outCopied = false;
```
pasin, line 42: Will there be another cherry pick to fix this, right?

pasin (approved):

LiteCore/Query/SQLiteFleeceUtil.cc
jianminzhao: yes.

PR #1351, opened 2022-01-14
Don't include c4Base.hh in Base.hh

snej:
It breaks the Tools build by forcing the LiteCore C++ API.
Instead include c4Base.h.
Requires a few changes in the LiteCore C++ Tests build.


jianminzhao (approved):

borrrden (approved):
Look forward to the tools builds recovering!

borrrden:
FYI EE needs some of this same treatment it seems.

PR #1357, opened 2022-01-19
Update check_deps.py script 

jayahariv:
* include `-path` arg, so we can reuse it for other projects.
* changes reflected on workflow xml file.
* CBL-2504

### DIFF
--- a/.github/workflows/check_deps.yml
+++ b/.github/workflows/check_deps.yml
@@ -26,4 +26,4 @@ jobs:
       working-directory: ${{github.workspace}}/build_cmake/scripts
       run: |
         pip install termcolor
-        ./check_deps.py -branch $GITHUB_BASE_REF -path ${{github.workspace}}
+        ./check_deps.py -branch $GITHUB_BASE_REF -path ${{github.workspace}}/
--- a/.github/workflows/check_deps.yml
+++ b/.github/workflows/check_deps.yml
@@ -26,4 +26,4 @@ jobs:
       working-directory: ${{github.workspace}}/build_cmake/scripts
       run: |
         pip install termcolor
-        ./check_deps.py -branch $GITHUB_BASE_REF -path ${{github.workspace}}/
+        ./check_deps.py -branch $GITHUB_BASE_REF -path ${{github.workspace}}
--- a/.github/workflows/check_deps.yml
+++ b/.github/workflows/check_deps.yml
@@ -26,4 +26,4 @@ jobs:
       working-directory: ${{github.workspace}}/build_cmake/scripts
       run: |
         pip install termcolor
-        ./check_deps.py $GITHUB_BASE_REF
+        ./check_deps.py -branch $GITHUB_BASE_REF -path ${{github.workspace}}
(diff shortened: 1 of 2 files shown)


.github/workflows/check_deps.yml
```diff
@@ -26,4 +26,4 @@ jobs:
       working-directory: ${{github.workspace}}/build_cmake/scripts
       run: |
         pip install termcolor
-        ./check_deps.py $GITHUB_BASE_REF
+        ./check_deps.py -branch $GITHUB_BASE_REF -path ${{github.workspace}}
```
jayahariv, line 29:
```suggestion
        ./check_deps.py -branch $GITHUB_BASE_REF -path "${{github.workspace}}/"
```

build_cmake/scripts/check_deps.py
```diff
@@ -123,8 +123,8 @@ def main(dir: str, branch: str):
 
 if __name__ == '__main__':
     parser = argparse.ArgumentParser(description='Validate submodules for PR')
-    parser.add_argument('branch', type=str, help="The branch to check for in non-couchbasedep repos")
+    parser.add_argument('-branch', type=str, help="The branch to check for in non-couchbasedep repos")
+    parser.add_argument('-path', type=str, help="The path to the parent repo")
```
jayahariv, line 127:
```suggestion
    parser.add_argument('-path', type=str, help="The path to the parent repo, ends with `/`")
```

build_cmake/scripts/check_deps.py
```diff
@@ -123,8 +123,8 @@ def main(dir: str, branch: str):
...
     parser = argparse.ArgumentParser(description='Validate submodules for PR')
-    parser.add_argument('branch', type=str, help="The branch to check for in non-couchbasedep repos")
+    parser.add_argument('-branch', type=str, help="The branch to check for in non-couchbasedep repos")
+    parser.add_argument('-path', type=str, help="The path to the parent repo, ends with `/`")
 
     args = parser.parse_args()
-    input_dir = os.path.dirname(os.path.realpath(__file__)) + os.sep + ".." + os.sep + ".." + os.sep
-    sys.exit(main(input_dir, args.branch))
\ No newline at end of file
+    sys.exit(main(args.path, args.branch))
```
pasin, line 130: As '/' is required, can we add it here such as `os.path.realpath(args.path) + os.sep`?

build_cmake/scripts/check_deps.py
```diff
@@ -123,8 +123,9 @@ def main(dir: str, branch: str):
 
 if __name__ == '__main__':
     parser = argparse.ArgumentParser(description='Validate submodules for PR')
-    parser.add_argument('branch', type=str, help="The branch to check for in non-couchbasedep repos")
+    parser.add_argument('-branch', type=str, help="The branch to check for in non-couchbasedep repos")
+    parser.add_argument('-path', type=str, help="The path to the parent repo to check")
```
borrrden, line 127: I'd rather have these both as positional arguments, or at the very least have them both marked as `required=True` because now as it stands you can omit either one as a valid invocation (`-` and `--` arguments are considered optional)

borrrden (changes requested):

build_cmake/scripts/check_deps.py
jayahariv: Will make it as positional arg

borrrden (approved):

PR #1358, opened 2022-01-19
Feature/signing

snej:
Implemented support for signing documents, actually Fleece Dicts, actually any Values. The signature is a Dict that could be stored in the document or elsewhere. This is based on my old [Signed Documents](https://github.com/couchbase/couchbase-lite-ios/wiki/Signed-Documents) spec, but modified slightly.

Signing can use the existing PublicKey/PrivateKey classes; unfortunately these are RSA and generate really big signatures, like 256 bytes binary or 341 bytes base64-encoded.

I brought in the Monocypher crypto library to get support for Ed25519 keys, which are more or less the state of the art in elliptic curves. The keys are only 32 bytes and signatures are 64 bytes, with the same security as RSA-2048. The math is faster than RSA, too.

I also overhauled SecureDigest to add support for SHA-256, which was needed for implementing RSA signatures.


borrrden:
Is this not something mbed is capable of?  Because this will mean two different crypto implementations in the library.

snej:
mbedCrypto has been promising to add Ed25519 for at least three years. I found a couple issues about it in their repo, and at least twice someone has submitted a PR. It looks like they're partway through, but they promised it in a release last year and haven't delivered yet. (And that would be in a 3.x release, while we're still on 2.x.)

The actual implementation of the algorithms is a small amount of code; the complexity seems to come from integrating them into all the other stuff in mbed. I've been using Monocypher for a while and it's a nice solid library. Small (2000 LOC, and Ed25519 only uses a fraction of that) and it's gone through security audit.

borrrden:
In that case, we should set it up as an official couchbasedeps fork and put it onto a release tag so that the check submodules script will verify it correctly.

PR #1361, opened 2022-01-24
CBL-2750 : fixing bugs that failed tests in CBLs, cf ticket cbl 2734

jianminzhao:
Incoporated those test cases into the Core.

### DIFF
--- a/C/tests/c4DatabaseTest.cc
+++ b/C/tests/c4DatabaseTest.cc
@@ -599,6 +599,37 @@ N_WAY_TEST_CASE_METHOD(C4DatabaseTest, "Database Expired Multiple Instances", "[
     c4db_release(db2);
 }
 
+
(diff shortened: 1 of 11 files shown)


borrrden:
Perhaps put a message about what exactly you fixed into the description?

borrrden (commented):
I don't think I have enough background about the problem to be able to review the solution.

jianminzhao:
I did "git commit --amend", modified the commit messages, then, "git push origin +HEAD"
I did not change the source code at all. It triggered validation build, but not the commit message in the PR. Here is the new commit message in git,
    CBL-2750 : fixing bugs that failed tests in CBLs, cf ticket cbl 2734
    Incoporated those test cases into the Core. Specifically, we fixed folowing cases
    (1) cross join
    (2) select * from db; // where there is only one database alias ("from db" is parsed to from _default as db")
    (3) select meta().expiration from a doc that is just deleted, whereby the expiration was set before the doc is deleted.
I will make sure this message appear in the final squash merge.

snej (approved):

PR #1363, opened 2022-01-25
Improved format-args security (-Wformat-literals, etc)

snej:
- In Xcode, enabled -Wformat -Wformat-nonliteral -Wformat-security
- Fixed a couple of places that were accidentally passing non-literal strings to format args.
- Declared more functions as being __printflike to fix warnings.
- Updated Fleece, where I made similar changes.

>Note: When the 2nd parameter of `__printflike` is 0, like `__printflike(n,0)`, it means that the format string's args come from a `va_list` parameter instead of "`...`". The reason to declare this is that it suppresses a non-literal format string warning in the _implementation_ of the function, where the format string and `va_list` are passed to a function like `vsprintf`. (I haven't seen this documented anywhere; I just found this usage of `__printflike` in C library headers, and adopting it fixed the warnings/errors I was getting in functions like `C4Error::vmakeError`.)

### DIFF
--- a/C/c4Document.cc
+++ b/C/c4Document.cc
@@ -41,9 +41,18 @@ C4Document::C4Document(C4Collection *collection, alloc_slice docID_)
 :_collection(asInternal(collection))
 ,_docID(move(docID_))
 {
+
+// GCC will (probably falsely) flag the below as a null directive
+// (I guess since docID_.buf could potentially be null?)
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wformat-overflow"
+
     // Quick sanity test of the docID, but no need to scan for valid UTF-8 since we're not inserting.
     if (_docID.size < 1 || _docID.size > kMaxDocIDLength)
         error::_throw(error::BadDocID, "Invalid docID \"%.*s\"", SPLAT(docID_));
+
+#pragma GCC diagnostic pop
+
 #if DEBUG
     // Make sure that C4Document and C4Document_C line up so that the same object can serve as both:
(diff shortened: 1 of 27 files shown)


Networking/Poller.cc
```diff
@@ -294,8 +294,10 @@ namespace litecore { namespace net {
                 if (fd == _interruptReadFD) {
                     // This is an interrupt -- read the byte from the pipe:
                     int message;
-                    ::read(_interruptReadFD, &message, sizeof(message));
-                    if (message < 0) {
+                    auto nread = ::read(_interruptReadFD, &message, sizeof(message));
+                    if(_usuallyFalse(nread == -1)) {
+                        result = false;
+                    } else if (message < 0) {
```
borrrden, line 300: @snej The above is the only logical change based on warning reactions, but do you think this is valid?  Or should I return true instead?

C/c4Document.cc
```diff
@@ -41,9 +41,18 @@ C4Document::C4Document(C4Collection *collection, alloc_slice docID_)
 :_collection(asInternal(collection))
 ,_docID(move(docID_))
 {
+
+// GCC will (probably falsely) flag the below as a null directive
```
snej, line 45: Oh! GCC noticed that `docID_` was moved from (on line 42) so it's known to be null.
The real fix is to change `docID_` to `_docID` on line 52; that should take care of the warning too.

Networking/Poller.cc
snej: I think it's valid.

borrrden (approved):

PR #1364, opened 2022-01-26
CBL-2701: SecItemDelete failures

jayahariv:
* Reference objects returned from SecTrustCopyCertificateChain, is not able to SecItemDelete, so we are getting it again from Keychain to delete.
* For non-chain certs, we are deleting the already returned certificate itself.

### DIFF
--- a/Crypto/PublicKey+Apple.mm
+++ b/Crypto/PublicKey+Apple.mm
@@ -699,13 +699,26 @@ namespace litecore { namespace crypto {
                     if (getChildCertCount(copiedRef) < 2) {
                         // Get public key hash from kSecAttrApplicationLabel attribute:
                         SecKeyRef publicKeyRef = SecCertificateCopyKey(copiedRef);
-                        NSDictionary* attrs = CFBridgingRelease(SecKeyCopyAttributes(publicKeyRef));
-                        NSData* publicKeyHash = [attrs objectForKey: (id)kSecAttrApplicationLabel];
+                        NSDictionary* pubKeyAttrs = CFBridgingRelease(SecKeyCopyAttributes(publicKeyRef));
+                        NSData* publicKeyHash = [pubKeyAttrs objectForKey: (id)kSecAttrApplicationLabel];
                         CFRelease(publicKeyRef);
+                        
+                        // For certs, primary key: issuer + serial-num + cert-type
+                        NSDictionary* attrs = CFBridgingRelease(findInKeychain(@{
+                            (id)kSecClass:              (id)kSecClassCertificate,
+                            (id)kSecValueRef:           (__bridge id)copiedRef,
+                            (id)kSecReturnAttributes:   @YES
+                        }));
+                        NSString* issuer = [attrs objectForKey: (id)kSecAttrIssuer];
+                        Assert(issuer);
+                        NSString* serialNum = [attrs objectForKey: (id)kSecAttrSerialNumber];
+                        Assert(serialNum);
                         
                         NSDictionary* params = @{
                             (id)kSecClass:                  (__bridge id)kSecClassCertificate,
-                            (id)kSecAttrPublicKeyHash:      publicKeyHash
+                            (id)kSecAttrPublicKeyHash:      publicKeyHash,
+                            (id)kSecAttrIssuer:             issuer,
+                            (id)kSecAttrSerialNumber:       serialNum,
                         };
                         checkOSStatus(SecItemDelete((CFDictionaryRef)params),
                                       "SecItemDelete",
--- a/Crypto/PublicKey+Apple.mm
+++ b/Crypto/PublicKey+Apple.mm
@@ -595,10 +595,8 @@ namespace litecore { namespace crypto {
             OSStatus err;
             if (@available(iOS 12.0, macos 10.14, *)) {
                 CFErrorRef cferr;
-                if (!SecTrustEvaluateWithError(trustRef, &cferr)) {
-                    auto error = (__bridge NSError*)cferr;
-                    LogTo(TLSLogDomain, "SecTrustEvaluateWithError %s", error.description.UTF8String);
-                }
+                if (!SecTrustEvaluateWithError(trustRef, &cferr))
+                    warnCFError(cferr, "SecTrustEvaluateWithError");
                 err = SecTrustGetTrustResult(trustRef, &result);
             } else {
 #if TARGET_OS_MACCATALYST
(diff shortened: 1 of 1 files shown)


Crypto/PublicKey+Apple.mm
```diff
@@ -682,16 +686,35 @@ virtual int _sign(int/*mbedtls_md_type_t*/ mbedDigestAlgorithm,
...
+            }
 #if __MAC_OS_X_VERSION_MAX_ALLOWED >= 120000 || __IPHONE_OS_VERSION_MAX_REQUIRED >= 150000
             if (@available(macOS 12.0, iOS 15.0, *)) {
                 CFArrayRef certs = SecTrustCopyCertificateChain(trustRef);
                 for (CFIndex i = count - 1; i >= 0; i--) {
-                    SecCertificateRef ref = (SecCertificateRef)CFArrayGetValueAtIndex(certs, i);
-                    if (getChildCertCount(ref) < 2) {
+                    SecCertificateRef copiedRef = (SecCertificateRef)CFArrayGetValueAtIndex(certs, i);
+                    if (getChildCertCount(copiedRef) < 2) {
+                        NSData* certData = CFBridgingRelease(findInKeychain(@{
```
jayahariv, line 705:
```suggestion
                        // SecRef returned is not able to SecItemDelete, so we fetch directly. 
                        NSData* certData = CFBridgingRelease(findInKeychain(@{
```

Crypto/PublicKey+Apple.mm
```diff
@@ -595,8 +595,10 @@ virtual int _sign(int/*mbedtls_md_type_t*/ mbedDigestAlgorithm,
             OSStatus err;
             if (@available(iOS 12.0, macos 10.14, *)) {
                 CFErrorRef cferr;
-                if (!SecTrustEvaluateWithError(trustRef, &cferr))
-                    warnCFError(cferr, "SecTrustEvaluateWithError");
+                if (!SecTrustEvaluateWithError(trustRef, &cferr)) {
+                    auto error = (__bridge NSError*)cferr;
+                    LogTo(TLSLogDomain, "SecTrustEvaluateWithError %s", error.description.UTF8String);
```
pasin, line 600: This will not log to CBL log?

Crypto/PublicKey+Apple.mm
```diff
@@ -682,16 +686,36 @@ virtual int _sign(int/*mbedtls_md_type_t*/ mbedDigestAlgorithm,
...
                 for (CFIndex i = count - 1; i >= 0; i--) {
-                    SecCertificateRef ref = (SecCertificateRef)CFArrayGetValueAtIndex(certs, i);
-                    if (getChildCertCount(ref) < 2) {
+                    SecCertificateRef copiedRef = (SecCertificateRef)CFArrayGetValueAtIndex(certs, i);
+                    if (getChildCertCount(copiedRef) < 2) {
+                        // CertRef returned from CopyCertificateChain won't get deleted, so we fetch directly. 
+                        NSData* certData = CFBridgingRelease(findInKeychain(@{
+                            (id)kSecClass:              (id)kSecClassCertificate,
+                            (id)kSecValueRef:           (__bridge id)copiedRef,
+                            (id)kSecAttrLabel:          label,
```
pasin, line 709: This might not be correct as the chain (parent) cert will not have the label. I think only value ref is enough?

Crypto/PublicKey+Apple.mm
```diff
@@ -682,16 +686,36 @@ virtual int _sign(int/*mbedtls_md_type_t*/ mbedDigestAlgorithm,
...
+                            (id)kSecReturnData:         @YES
+                        }));
+                        
                         NSDictionary* params = @{
                             (id)kSecClass:              (id)kSecClassCertificate,
-                            (id)kSecValueRef:           (__bridge id)ref
+                            (id)kSecAttrLabel:          label,
```
pasin, line 715: The label is not needed as we have use the whole certData? Also the parent cert will not have label.

Crypto/PublicKey+Apple.mm
jayahariv: No, it will not log. I guess the TLSLogDomain is not getting logged as part of Sync. I will check it into more details.

jayahariv:
Will update this PR soon, since I faced with couple of flaky failures, looking into them.

Crypto/PublicKey+Apple.mm
jayahariv: when I removed the label, the SecDeleteItem fails with `-61`, 
```
Couldn't remove a private key from the Keychain (SecItemDelete returned -61)
```

Crypto/PublicKey+Apple.mm
jayahariv: Spend couple of hours to investigate but didn't find the root cause why not logging, will further check in separate task CBL-2778

Crypto/PublicKey+Apple.mm
```diff
@@ -699,13 +699,26 @@ virtual int _sign(int/*mbedtls_md_type_t*/ mbedDigestAlgorithm,
...
+                        Assert(issuer);
+                        NSString* serialNum = [attrs objectForKey: (id)kSecAttrSerialNumber];
+                        Assert(serialNum);
+                        
                         NSDictionary* params = @{
                             (id)kSecClass:                  (__bridge id)kSecClassCertificate,
-                            (id)kSecAttrPublicKeyHash:      publicKeyHash
+                            (id)kSecAttrPublicKeyHash:      publicKeyHash,
+                            (id)kSecAttrIssuer:             issuer,
+                            (id)kSecAttrSerialNumber:       serialNum,
```
pasin, line 721: Can you include kSecAttrCertificateType as well?

pasin (changes requested):

Crypto/PublicKey+Apple.mm
```diff
@@ -682,15 +681,47 @@ virtual int _sign(int/*mbedtls_md_type_t*/ mbedDigestAlgorithm,
...
+                        NSString* issuer = [attrs objectForKey: (id)kSecAttrIssuer];
+                        Assert(issuer);
+                        NSString* serialNum = [attrs objectForKey: (id)kSecAttrSerialNumber];
+                        Assert(serialNum);
+                        NSNumber* certType = [attrs objectForKey: (id)kSecAttrCertificateType];
+                        Assert(certType != nil);
+                        
+                        NSDictionary* params = @{
+                            (id)kSecClass:                  (__bridge id)kSecClassCertificate,
+                            (id)kSecAttrPublicKeyHash:      publicKeyHash,
```
pasin, line 721: Can you remove public key hash from the query as the primary key is primary key: issuer + serial-num + cert-type?

Crypto/PublicKey+Apple.mm
```diff
@@ -682,15 +681,47 @@ virtual int _sign(int/*mbedtls_md_type_t*/ mbedDigestAlgorithm,
...
+                    SecCertificateRef copiedRef = (SecCertificateRef)CFArrayGetValueAtIndex(certs, i);
+                    if (getChildCertCount(copiedRef) < 2) {
+                        // Get public key hash from kSecAttrApplicationLabel attribute:
+                        SecKeyRef publicKeyRef = SecCertificateCopyKey(copiedRef);
+                        NSDictionary* pubKeyAttrs = CFBridgingRelease(SecKeyCopyAttributes(publicKeyRef));
+                        NSData* publicKeyHash = [pubKeyAttrs objectForKey: (id)kSecAttrApplicationLabel];
+                        CFRelease(publicKeyRef);
+                        
+                        // For certs, primary key: issuer + serial-num + cert-type
+                        NSDictionary* attrs = CFBridgingRelease(findInKeychain(@{
```
pasin, line 707: Can you add a note  about why we need to do this in order to delete the certs?

pasin (approved):
I have two comments. The rest looks good.

PR #1366, opened 2022-01-28
CBL-2750 : fixing the following two failures discovered in CBLs, c.f.…

jianminzhao:
… ticket cbl 2734:

(1) replication of deleted documents that were deleted after being pushed.
(2) Queries that may involve live documents as well as deleted documents.

### DIFF
--- a/LiteCore/Storage/KeyStore.hh
+++ b/LiteCore/Storage/KeyStore.hh
@@ -122,7 +122,8 @@ namespace litecore {
 
         void setKV(Record&, ExclusiveTransaction&);
 
-        virtual bool del(slice key, ExclusiveTransaction&, sequence_t replacingSequence ={}) =0;
+        virtual bool del(slice key, ExclusiveTransaction&, sequence_t replacingSequence ={},
+                         uint64_t replacingSubsequence =-1) =0;
         bool del(const Record &rec, ExclusiveTransaction &t)                 {return del(rec.key(), t);}
 
         /** Sets a flag of a record, without having to read/write the Record. */
--- a/LiteCore/Storage/KeyStore.hh
+++ b/LiteCore/Storage/KeyStore.hh
@@ -123,7 +123,7 @@ namespace litecore {
         void setKV(Record&, ExclusiveTransaction&);
 
         virtual bool del(slice key, ExclusiveTransaction&, sequence_t replacingSequence ={},
-                         uint64_t replacingSubsequence =-1) =0;
+                         std::optional<uint64_t> replacingSubsequence =std::nullopt) =0;
         bool del(const Record &rec, ExclusiveTransaction &t)                 {return del(rec.key(), t);}
 
         /** Sets a flag of a record, without having to read/write the Record. */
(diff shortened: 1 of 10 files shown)


jianminzhao:
Remarks: the test failure of query is due to require(_delegate.tableExists("all_default"). The "tableExits" checking is done first time the "from" entry is parsed and _aliases is constructed. Later the entry gets modified in lookForDeleted() when we change the table to all_ or del_. Maybe at this time we check the existence.
From another perspective, the new table name is given by _delegate.collectionTableName(), should the delegate be responsible for the existence of the new table given valid collection?
Since all_coll" is a "temp view", sqlite query for table or view existence yields false.

LiteCore/Storage/SQLiteDataFile.cc
```diff
@@ -669,6 +669,10 @@ namespace litecore {
 
 
     bool SQLiteDataFile::tableExists(const string &name) const {
+        if (((string_view)name).substr(0, 4) == "all_") {
+            // "all_xxx" is a TEMP VIEW that is not visible to getSchma.
+            return true;
```
snej, line 674: This will return false positives. Instead replace "all_" with "kv_" and then fall through to calling `getSchema`.

snej (changes requested):

LiteCore/Storage/SQLiteDataFile.cc
jianminzhao: Yes, agreed. Or, perhaps all_ must be used in the context of deleted. What do think falling through with  kv_del_ ?

LiteCore/Storage/KeyStore.hh
```diff
@@ -122,7 +122,8 @@ namespace litecore {
 
         void setKV(Record&, ExclusiveTransaction&);
 
-        virtual bool del(slice key, ExclusiveTransaction&, sequence_t replacingSequence ={}) =0;
+        virtual bool del(slice key, ExclusiveTransaction&, sequence_t replacingSequence ={},
+                         uint64_t replacingSubsequence =-1) =0;
```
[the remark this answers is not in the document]
jianminzhao: I will use optional, (:- only gives compiler a little more work.

LiteCore/Storage/SQLiteDataFile.cc
snej: You can query for all deleted/non-deleted docs, even if there aren't any. So the deleted key-store shouldn't be required right?

jianminzhao: According to current code, we use all_ only if DeletionStatus == kLiveAndDeletedDocs, and all_ is a view that combines kv_ and kv_del_, right? In actual code consideration, we should already have checked kv_ because kv_del_ or all_ are the result of replacing aliasInfo created in the first pass, where tabelExists() already called.

snej: OK.

snej (approved):

PR #1369, opened 2022-02-04
CBD-4587, create lite-core artifacts to publish to latestbuild.

mhocouchbase:
CBD-4587, create lite-core artifacts to publish to latestbuild.
migrated unix & windows build scripts over from build repo.

-Ming Ho

### DIFF
--- a/jenkins/build_server_android.sh
+++ b/jenkins/build_server_android.sh
@@ -75,6 +75,10 @@ if [[ "${ANDROID_ARCH}" == "x86_64" ]] || [[ "${ANDROID_ARCH}" == "arm64-v8a" ]]
     ARCH_VERSION="21"
 fi
 
+#create artifacts dir for publishing to latestbuild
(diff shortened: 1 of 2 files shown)


borrrden:
There is already a Windows build script in here, so I don't think that .bat version is needed.  Is it still being used on the build side?

mhocouchbase:
Yes, the jenkins job still uses the .bat file.  Let me try .ps1 to make sure it works.
Having 2nd thought on the file names.  Why "build_server" was used?  "do-<OS>build" seems to be a better representation of the purpose.

borrrden:
In that case I think I'd like to jump on here and make it so.  I don't like .bat files, they are basically the perl of the Windows world (hard to read and pretty much outdated).  Also the reason for the name is so fewer people would be confused when they tried to use it and realized nothing worked.  "do build" in this context sounds like anyone who checks out the repo can use it.  I called it "build server" because its the script that runs "on the build server"

mhocouchbase:
Removed build_server_win.bat from this PR.

borrrden (approved):

jianminzhao (approved):

PR #1371, opened 2022-02-05
Make the build server Windows script parallel

borrrden:


jenkins/build_server_win.ps1
```diff
@@ -150,16 +154,21 @@ function Run-UnitTest() {
     Pop-Location
 }
 
-foreach ($arch in @("Win32", "Win64", "ARM")) {
+$ArtifactsDir = "$env:WORKSPACE\artifacts\$($ShaVersion.substring(0, 2))\$ShaVersion"
```
mhocouchbase, line 157: does $ArtifactsDir need to be created first or Copy-Item will automatically create it?

mhocouchbase (commented):
need to add "--parallel $Parallel" to cmake to build in Parallel.

mhocouchbase:
The naming convention for debug zips in this script look more consistent than what was in the bat script.  Need to make sure the downstream consumer is aware of the name change.
%PRODUCT%-%VERSION%-%SHA_VERSION%-%OS%-win32-debug.zip
%PRODUCT%-%VERSION%-%SHA_VERSION%-%OS%-debug-arm.zip
%PRODUCT%-%VERSION%-%SHA_VERSION%-%OS%-win64-winstore-debug.zip

jenkins/build_server_win.ps1
borrrden: It is created on the line below this one actually

borrrden:
What is the `$Parallel` value?

borrrden:
Noted on the names

mhocouchbase (approved):

PR #1375, opened 2022-02-09
CBL-2751 : do not react to WebSocketImpl::onClose() if it does not pl…

jianminzhao:
…ay the role of client.

### DIFF
--- a/Networking/WebSockets/WebSocketImpl.cc
+++ b/Networking/WebSockets/WebSocketImpl.cc
@@ -454,6 +454,18 @@ namespace litecore { namespace websocket {
     // Called when the underlying socket closes.
     void WebSocketImpl::onClose(CloseStatus status) {
         {
+            if (!hasDelegate()) {
+                // CBL-2751
+                // The delegate is assigned when WebSocket::connect(Delegate*) is called,
+                // which is an act of client.
+                // Our assertions are:
+                // (1) client calls WebSocket::connect(Delegate*) to start
+                // (2) therefore, hasDelegate() == true
+                // (3) Otherwise, if hasDelegate() == false, it must not be a client and
+                //     we therefore do not react to this call.
+                return;
+            }
+
             lock_guard<mutex> lock(_mutex);
 
             if (_closed)
--- a/Networking/WebSockets/WebSocketImpl.cc
+++ b/Networking/WebSockets/WebSocketImpl.cc
@@ -85,6 +85,9 @@ namespace litecore { namespace websocket {
     }
 
 
+#define CHECK_DELEGATE { if (!hasDelegate()) return; }
+
+
     void WebSocketImpl::connect() {
         logInfo("Connecting...");
         startResponseTimer(chrono::seconds(kConnectTimeoutSecs));
@@ -92,11 +95,13 @@ namespace litecore { namespace websocket {
 
 
     void WebSocketImpl::gotHTTPResponse(int status, const websocket::Headers &headersFleece) {
+        CHECK_DELEGATE
         logInfo("Got HTTP response (status %d)", status);
         delegate().onWebSocketGotHTTPResponse(status, headersFleece);
     }
 
     void WebSocketImpl::onConnect() {
+        CHECK_DELEGATE
         if(_closed) {
             // If the WebSocket has already been closed, which only happens in rare cases
             // such as stopping a Replicator during the connecting phase, then don't continue...
@@ -164,6 +169,7 @@ namespace litecore { namespace websocket {
 
 
     void WebSocketImpl::onWriteComplete(size_t size) {
+        CHECK_DELEGATE
         bool notify, disconnect;
         {
             lock_guard<mutex> lock(_mutex);
@@ -299,6 +305,7 @@ namespace litecore { namespace websocket {
 
 
     void WebSocketImpl::deliverMessageToDelegate(slice data, bool binary) {
+        CHECK_DELEGATE
         logVerbose("Received %zu-byte message", data.size);
         _deliveredBytes += data.size;
         Retained<Message> message(new MessageImpl(this, data, true));
@@ -373,6 +380,7 @@ namespace litecore { namespace websocket {
 
     // Initiates a request to close the connection cleanly.
     void WebSocketImpl::close(int status, fleece::slice message) {
+        CHECK_DELEGATE
         if(!_didConnect && _framing) {
             // The web socket is being requested to close before it's even connected, so just
             // shortcut to the callback and make sure that onConnect does nothing now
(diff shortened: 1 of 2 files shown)


Networking/WebSockets/WebSocketImpl.cc
```diff
@@ -454,6 +454,18 @@ namespace litecore { namespace websocket {
     // Called when the underlying socket closes.
     void WebSocketImpl::onClose(CloseStatus status) {
         {
+            if (!hasDelegate()) {
```
borrrden, line 457: This conditional does not match the description in the commit, and requires comment analysis.  I feel like this should be `if(role() != Role::Client)` which more clearly identifies the information we are after.

borrrden (changes requested):
I think this logic should be introduced into a few more places, not just the result of `c4socket_closed`.  There should probably be some tests which make these calls as well, to ensure that they do not crash.

Networking/WebSockets/WebSocketImpl.cc
jianminzhao: I use this as a test because I found server role may also take a factory as argument, c.f.
C4Socket* C4Socket::fromNative(const C4SocketFactory &factory,
                               void *nativeHandle,
                               const C4Address &address)
{
    // Note: This should be wrapped in `retain()` since `C4SocketImpl` is ref-counted,
    // but doing so would cause client code to leak. Instead I added a warning to the doc-comment.
    return new C4SocketImpl(address.toURL(), Role::Server, {}, &factory, nativeHandle);
}
This makes me this, "roll() == Role::Server" may also call factory method acting like a client in the dialog of client v server. If it calls factory.open(), then it would get a delegate and expect onX methods.
The common factor is, WebSocket::connect(Delegate*). As far as I can see from the source code, it seems that role being Role::Server cannot bar it from calling factory.open, etc. Otherwise, why is factory is given to C4Socket of server role? I can change the test to Role() != Role::Client if I am sure that in this case, "factory" passed in the constructor shall not be used.

As for why I only put the test onClose, that's because I find it's not easy for the client code, Java in particular, to call other onX methods because it needs to pass the opaque C4Socket* handle which is passed to client by Core as argument of factory::open. Only onClose is little confusing. It is called in, for instance, in Java's C4Socket::release(), which thinks I am closed because I am reaching end of my life. Other onX methods have clear semantics. In other words, I would think it as a program error leading to UB. It seems to be least disruptive now.

borrrden: Server side (passive), by definition, does not open connections from itself.  It sits there and listens for, and accepts incoming connections.  The factory is given to the server side socket so that it can know how to correctly write bytes and acknowledge receives.  Opening and closing connections is the responsibility of the active (client) side.

It may be that `c4socket_closed` is the only dangerous method, but it's worth checking out the path of the others like `c4socket_opened` to see if they lead to any delegate calls.  I don't think whether or not a platform calls them is important, what is important is that they are able to be called without arbitrarily causing a native crash.

jianminzhao: Okay, I will add a DBAssert to connect(Delegate*) to assert Role() == client. Then we can just check Role() == client in the client callback methods.
There are 6 methods in WebSocketImp that use delegate(). They all need be guarded against misuse. Very ugly. In the future, we probably should do what @bmeike suggested to subclass them into two, client & server.

jianminzhao: "Server side (passive), by definition, does not open connections from itself. " -- this statement is incorrect. The following LoopbackWebSocket is created with role == server as follows,
        _replServer = new Replicator(dbServer,
                                     new LoopbackWebSocket(alloc_slice("ws://cli/"_sl), Role::Server, kLatency),
                                     *this, optsRef2);
but hits my assertion:
    void WebSocket::connect(Delegate *delegate) {
//        DebugAssert(role() == Role::Client);
        if (role() != Role::Client) {
            printf("hey got you.\n");
        }
We have to think the role as replicator's role, as supposed to the connection's role.

borrrden: In that case there is no choice but in my opinion it's now even messier than you suggested because we shouldn't be thinking about what a higher level component is doing when working in a lower level component....

borrrden (commented):
I'm getting bad vibes from this situation and would like Jens to have a look and weigh in.

snej:
Thoughts:
* LoopbackWebSocket is weird. It was originally written just for testing. If it would make the overall logic cleaner & more consistent to have it set up the connection differently, we should consider that.
* Maybe make a NoOpDelegate class and set that as the delegate when it would otherwise be null? Then callers wouldn't have to check.

Networking/WebSockets/WebSocketImpl.cc
```diff
@@ -85,18 +85,23 @@ namespace litecore { namespace websocket {
...
 
 
     void WebSocketImpl::gotHTTPResponse(int status, const websocket::Headers &headersFleece) {
+        CHECK_DELEGATE
         logInfo("Got HTTP response (status %d)", status);
         delegate().onWebSocketGotHTTPResponse(status, headersFleece);
     }
 
     void WebSocketImpl::onConnect() {
+        CHECK_DELEGATE
```
snej, line 104: This method does a lot of stuff besides call the delegate. Why would you make it return and do nothing, when the delegate is null?

Networking/WebSockets/WebSocketImpl.cc
```diff
@@ -453,6 +461,15 @@ namespace litecore { namespace websocket {
 
     // Called when the underlying socket closes.
     void WebSocketImpl::onClose(CloseStatus status) {
+        CHECK_DELEGATE
+        // CBL-2751
+        // The delegate is assigned when WebSocket::connect(Delegate*) is called,
+        // which is an act of client.
```
snej, line 467: That's not true. Both client and server side call connect(). And any active WebSocket has to have a delegate, because incoming messages are sent to it.

You can find the call hierarchy with Xcode, for example:
* WebSocket::connect
* BLIPIO::_start
* BLIPIO::start
* blip::Connection::start
* Replicator::_start
* Replicator::start
* C4ReplicatorImpl::_start
* C4ReplicatorImpl::start
* RESTSyncListener::handleSync

snej (changes requested):

Networking/WebSockets/WebSocketImpl.cc
jianminzhao: Thanks, they are so helpful. Your's saying, if not active, the argument of WebSocket::connect(Delegate*) is null. Am I right?

jianminzhao: But there is an assertion asserts none-null.

PR #1376, opened 2022-02-10
Add in a script to automate fetching prebuilt LiteCore

borrrden:


jianminzhao (approved):

scripts/fetch_litecore.py
```diff
@@ -0,0 +1,309 @@
...
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(description='Fetch a specific prebuilt LiteCore')
+    parser.add_argument('--variants', nargs='+', type=str, help='A space separated list of variants to download', required=True)
+    parser.add_argument('--debug', action='store_true', help='If specified, download debug variants')
+    parser.add_argument('--dry-run', action='store_true', help='Check for existience of indicated artifacts, but do not perform download')
+    parser.add_argument('--sha', type=str, help="The SHA to download.  If not provided, calculated based on the provided CE and (optionally) EE repos.  Required if CE not specified.")
+    parser.add_argument('--ce', type=str, help="The path to the CE LiteCore repo.  Required if SHA not specified.")
+    parser.add_argument('--ee', type=str, help="The path to the EE LiteCore repo")
+    parser.add_argument('--path', type=str, help="The path in which the platform specific extensions to this script are defined (platform_fetch.py).  If a relative path is passed, it will be relative to fetch_litecore.py.  By default it is the current working directory.",
```
bmeike, line 288: Would love either a -q --quiet or a -v verbose flag that allowed me to silence all but essential output.

scripts/fetch_litecore.py
```diff
@@ -0,0 +1,309 @@
...
+        sys.path.insert(0, str(path.absolute()))
+        global subdirectory_for_variant
+        from platform_fetch import subdirectory_for_variant
+        global has_platform
+        has_platform = True
+    except ImportError:
+        print("!!! Unable to import platform extensions, falling back to <os>/<abi> for subdirectory...")
+
+def variant_to_pair(variant: str) -> Sequence[str]:
+    """Given a variant, splits it into an OS and ABI pair
```
bmeike, line 182: It took me a while to convince myself that this is, actually, ok.  I think the comment is wrong, though.  It is not converting to an OS and ABI pair... it is normalizing, very specifically, to LatestBuilds path names.

Once I realized that, though, it makes perfect sense.

I wonder if, instead of expressing this as code, it would make sense as a set of RE mappings.  That might, just possibly, be easier to understand, maintain and extend.

Just a thought.

bmeike (approved):
This looks functional to me.  If there are any problems with it, I will post them as bugs.
I would like the changes I've suggested, in the options/parameter flags but no biggie.

bmeike:
One more comment....
In your documentation, it might be a good idea to mention that the arguments to `subdirectory_for_variant` are *post* conversion, e.g., ["windows", "x86_64"], not ["win64", "..."]

bmeike:
Also, in your documentation, it might be a good idea to tell what packages I need to install, to get this to work.
FWIW, doing those installs is going to be kind of a pain, during each new Jenkins build.  Maybe we can get the build team to install them?

bmeike:
This is a really excellent thing.  I am totally looking forward to throwing out the tens of lines of flaky bash/cmd code that it will replace.

It will, easily, save the time you've put into it.

Nice work.

borrrden:
> it might be a good idea to tell what packages I need to install, to get this to work.

For example?  I am completely unaware of any packages that are needed but it might be because they are already installed by chance on my machine.  My goal is to not need any.

> This script's heavy-handedness wrt os/abi names worries me.

I am not sure what you mean by this.  What would be the non heavy handed version?  

> Am I the only one that uses "macos" instead of "macosx"

Not sure, macosx is what the name is currently on nexus and that carried over to latestbuilds

borrrden:
> I was pretty surprised that I got all of the support libs put into the same directory as LiteCore

I was a bit worried about this, but the fact that they are in the archive is not anything new.  If you have a suggestion or would like some additional extension methods to filter then that's something that could be possible.

scripts/fetch_litecore.py
```diff
@@ -0,0 +1,309 @@
...
+    args = parser.parse_args()
+
+    if not args.sha and not args.ce:
+        print("!!! Neither SHA nor CE repo path defined, aborting...")
+        parser.print_usage()
+        exit(-1)
+
+    full_path = resolve_platform_path(args.path)
+    import_platform_extensions(full_path)
+
```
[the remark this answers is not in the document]
borrrden: Yes, I'm not sure why you would want to do that anyway.  It's not like you can build them both using the same process.

scripts/fetch_litecore.py
borrrden: What would you consider essential?  I guess the output of a dry run should not be silenced but other than that there is not really a need for any output.

scripts/fetch_litecore.py
borrrden: The RE would be pretty messy since the Windows ABI names are inconsistent and some platforms have no ABI at all (e.g. iOS is a fat binary with multiple architectures).  I wanted to standardize them to the official names since it will always be "wrong" for someone (i.e. Microsoft uses x64, Android uses x86_64).  For the OS name I wasn't sure what to do so I kept the latestbuilds name intact.  If something else would be better than I'd like to hear it.

scripts/fetch_litecore.py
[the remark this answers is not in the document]
borrrden: Nevermind, my approach has changed for this.

bmeike:
> For example? I am completely unaware of any packages that are needed 

I had to install `Git`

> What would be the non heavy handed version

Right...   What I meant by that is that the script seemed to me to introduce an entirely new middle language for naming artifacts, neither the one used by latestbuilds nor the one used by (my) platform.  What I hoped to communicate was that I think *NOT* doing that should be a goal.  I think this should accept the latestbuilds language, for sure, and platform languages where possible and not introduce anything new.

> macosx is what the name is currently on nexus and that carried over to latestbuilds

Let's use it, then.  I'll adjust this on my side.

> I was pretty surprised that I got all of the support libs put into the same directory as LiteCore

I can adapt to this, on my side, too.  I don't think that target destination by file-name, or something, belongs in this tool.

borrrden:
Ah I see, I thought "packages" meant `pip` (python) packages.  Yes you need git installed as a side effect of using `PyGit` or whatever it is called to get the commit hashes of repos.  I'd be surprised if build machines didn't have git installed on them.  

As far as naming goes, I plucked the naming directly from what the artifacts are named by the build job, and accordingly how they currently exist and are being pulled from Nexus.  You can see an example of this in your own [current fetch script](https://github.com/couchbase/couchbase-lite-java-common/blob/184d173ccad51cff04d8bf0e14b5f1d37e7efcef/tools/fetch_litecore.sh#L72)

bmeike (approved):
Good grief!!!  How have I not approved this???

Sandychuang8 (approved):
Looks good but I haven't had chance to try it out...

PR #1379, opened 2022-02-15
CBL-1361 : logging the configs of Listener and Replicator at LogInfo.

jianminzhao:


REST/c4Listener.cc
```diff
@@ -38,6 +38,40 @@ string C4Listener::URLNameFromPath(slice pathSlice) {
     return Listener::databaseNameFromPath(FilePath(pathSlice, ""));
 }
 
+namespace {
+    std::stringstream& operator<<(std::stringstream& ss, const C4ListenerConfig& config) {
+        ss << "{" << "apis: " << (config.apis == kC4RESTAPI ? "REST" : "Sync") << ", "
+        << "networkInterface: " << std::string(config.networkInterface) << ", "
```
pasin, line 44: if `config.networkInterface` is nullslice, should we put null or "NULL" here?

REST/c4Listener.cc
```diff
@@ -38,6 +38,40 @@ string C4Listener::URLNameFromPath(slice pathSlice) {
...
+            << "certificate: " << (config.tlsConfig->certificate == nullptr ? "NULL" : "***") << ", "
+            << "requireClientCerts: " << config.tlsConfig->requireClientCerts << ", "
+            << "rootClientCerts: " << (config.tlsConfig->rootClientCerts == nullptr ? "NULL" : "***") << ", "
+            << "certAuthCallback: " << (config.tlsConfig->certAuthCallback == nullptr ? "NULL" : "***") << ", "
+            << "tlsCallbackContext: " << (config.tlsConfig->tlsCallbackContext == nullptr ? "NULL" : "***");
+        }
+        ss << "}, "
+        << "httpAuthCallback: " << (config.httpAuthCallback == nullptr ? "NULL" : "***") << ", "
+        << "callbackContext: " << (config.callbackContext == nullptr ? "NULL" : "***") << ", "
+        << "directory: " << std::string(config.directory) << ", ";
```
pasin, line 61: If config.directory is nullslice, I think we should put NULL here.

jianminzhao: same above

REST/c4Listener.cc
```diff
@@ -38,6 +38,45 @@ string C4Listener::URLNameFromPath(slice pathSlice) {
     return Listener::databaseNameFromPath(FilePath(pathSlice, ""));
 }
 
+namespace {
+    bool isEmptyString(const C4String& c4str) {
+        return !c4str || c4str.size == 0;
+    }
```
borrrden, line 44: Instead of this, why not have a `toString` method that gets rid of the repeated tertiary operator?

jianminzhao: toString may yield empty string. That's what the previous version does. I think Pasin wants NULL for empty string.
By the way, this is not strict json string value whereby empty string is "", and here empty string is  , really empty.

borrrden (commented):
Could someone explain what the value is in adding a bunch of "***" to the log?

REST/c4Listener.cc
borrrden: Right but if you are making a method here anyway, why not have it return the string `"NULL"` here?

jianminzhao:
One purpose "***" can serve, I think, is that the caller did supply something.

borrrden:
Yes, I considered that but have we ever had a case where we weren't sure if the caller supplied us something or not?  I know this logging was requested from outside parties but there was never any details about what exactly they were looking to get out of it.

jianminzhao:
Taking for examples, certAuthCallback and tlsCallbackContext, they are void*, the actual numbers may not be too much useful unless taking together with other bugs where actual pointers are logged, but perhaps the presence of them can lead to different code paths, etc. I am open to what we log. @pasin ?

REST/c4Listener.cc
pasin: I think it's better to separate b/w NULL and empty string. If empty string, the log will print "" and if NULL, the log will print NULL. Also the NULL representation of C4String is { nullptr, 0 }?

jianminzhao: Let me note: we are not doing json, where a string value is quoted both its binary representation of printed form; we print it for human consumption. So, we will print a C4String in 3 shape:
--------
directory: this/path
directory: ""
directory: NULL
--------
sounds good?

pasin: **BIG and CLEAR**

jianminzhao: What is printed:
01:15:08.076671| [Listener]: Listener config: {apis: REST, networkInterface: NULL, port: 0, tlsConfig: {}, httpAuthCallback: NULL, callbackContext: NULL, directory: "", allowCreateDBs: 0, allowDeleteDBs: 0}

pasin (approved):

PR #1385, opened 2022-02-18
CBL-2808 : Ignore EWOULDBLOCK when write in non-block mode

pasin:
* The problem was as the replicator’s connection kept getting closed when connecting to SG with TLS. From the provided log file, the error that caused the connection to get closed is the POSIX 35 (EWOULDBLOCK) error (on MacOS platform). The EWOULDBLOCK errors in the log were found during both READ and WRITE to the socket stream. There were more error during READ than WRITE.

* The user who reported the problem used CBL-C 3.0.0 Beta 2 on MacOS, which doesn't contain the EWOULDBLOCK fix for READ. After providing the user with the RC2 binary, the problems during READ were gone, however, the problem during WRITE were remained.

* In Non-Block mode, the EWOULDBLOCK should not be considered as an error and should be ignored. This PR made the TCPSocket::write(vector<slice> &ioByteRanges) ignored the EWOULDBLOCK error. Noted that the other write() methods have already done that.

* Created a utility function (called checkReadWriteStreamError()) for checking the non-blocking mode and EWOULDBLOCK error before calling checkStreamError() as this logic has been done in multiple places.

* In BuiltInWebSocket::writeToSocket(), made sure to call awaitWriteable() when the socket's write() function returns zero (e.g. when getting EWOULDBLOCK error).

* I have tested this fix manually by modifying the TCPSoket to randomly cause an EWOULDBLOCK error before actually write the data. The TCPSocket could ignore the error as expected, and the replicator can continue and finish the replication.

* NOTE: This fix will be included in CBL-C 3.0.1 release.

### DIFF
--- a/Networking/TCPSocket.cc
+++ b/Networking/TCPSocket.cc
@@ -223,9 +223,8 @@ namespace litecore { namespace net {
             return 0;
         ssize_t written = _socket->write(data.buf, data.size);
         if (written < 0) {
-            if (_nonBlocking && socketToPosixErrCode(_socket->last_error()) == EWOULDBLOCK)
+            if (checkReadWriteStreamError())
                 return 0;
-            checkStreamError();
         } else if (written == 0) {
             _eofOnWrite = true;
         }
@@ -238,9 +237,8 @@ namespace litecore { namespace net {
             return 0;
         ssize_t written = _socket->write_n(data.buf, data.size);
         if (written < 0) {
-            if (_nonBlocking && socketToPosixErrCode(_socket->last_error()) == EWOULDBLOCK)
+            if (checkReadWriteStreamError())
                 return 0;
-            checkStreamError();
         }
         return written;
     }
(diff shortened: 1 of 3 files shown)


borrrden (approved):

Networking/WebSockets/BuiltInWebSocket.cc
```diff
@@ -438,6 +438,8 @@ namespace litecore { namespace websocket {
             if (_usuallyFalse(n <= 0)) {
                 if (n < 0)
                     closeWithError(_socket->error());
+                else
+                    awaitWriteable();
```
jianminzhao, line 442: Looks like this is the only semantic change of this entire PR, right?

pasin: Two main changes:
1. [TCPSocket] Ignored EWOULDBLOCK error in write().
2. [BuiltInWebSocket] Wait for the writable again when that happens.

jianminzhao (approved):

Networking/TCPSocket.cc
```diff
@@ -622,4 +620,16 @@ namespace litecore { namespace net {
             setError(NetworkDomain, mbedToNetworkErrCode(err), slice(msgbuf));
         }
     }
+
+
+    bool TCPSocket::checkReadWriteStreamError() {
```
snej, line 625: The name of this function seems backwards -- it returns true if there's _not_ an error. I'd prefer that the return value be flipped.

snej (changes requested):

Networking/TCPSocket.cc
pasin: I just merged the PR so I will submit a new PR to fix this.

Issue #1387, opened 2022-02-22
Can not compile on Linux/amd64


Dushistov:
I tried to compile current master ( 817f1f979ebfe5877b22b32589795216a0620ace ) on Linux/amd64/gcc 11.2 .
And got two compilation problems:

1) invalid 'static_cast' from type 'std::__cxx11::basic_stringstream<char>' to type 'std::stringstream&' in `#define CONCAT(EXPR)   (static_cast<std::stringstream&>(std::stringstream() << EXPR)).str()`. The problem can be fixed like this:
   
```
 #ifndef _LIBCPP_VERSION
-    #define CONCAT(EXPR)   (static_cast<std::stringstream&>(std::stringstream() << EXPR)).str()
+    #define CONCAT(EXPR)   (static_cast<const std::stringstream&>(std::stringstream() << EXPR)).str()
```
but may be it should be `#ifdef _MSC_VER`, because of only MSVC can compile such code https://godbolt.org/z/fh7ecs3h6 ,
clang/gcc gives compilation failure on such code, plus `git blame` shows that this code was introduced to fix "Windows build".

2) 
```
couchbase-lite-core/C/tests/../../vendor/fleece/vendor/catch/catch.hpp:10823:58: error: call to non-'constexpr' function 'long int sysconf(int)'
10823 |     static constexpr std::size_t sigStackSize = 32768 >= MINSIGSTKSZ ? 32768 : MINSIGSTKSZ;
```
   This is known issue (https://github.com/catchorg/Catch2/issues/2178 ) and was fixed in `catch` year ago,
   but it is hard to tell what exact version of catch uses couchabse-lite-core, so hard to cherry-pick fix.

borrrden:
Really wild.  We compile this code constantly with older GCC so it's always annoying when this happens.  godbolt shows your code snippet compiling just fine with 9.3.  Filed as [CBL-2873](https://issues.couchbase.com/browse/CBL-2873)

PR #1392, opened 2022-03-04
fix build of StringUtil.hh with gcc 11.2.0

Dushistov:
I veriefied that with old compilers all works fine,
except msvc 15, because of it is not availble in goldbot:
gcc 7.1 is ok: https://godbolt.org/z/h364dc9M8
clang 5.0 is ok: https://godbolt.org/z/vzd8qeedY
clang 5.0 + libc++ is ok: https://godbolt.org/z/YjbbcxebG
msvc 19.4 is ok: https://godbolt.org/z/Kobh61Tb4
comment #1387

build of tests still broken, because of catch dependency,
but at least without tests all works fine


borrrden:
I checked the latest msvc 15 and it has no issue with this.

borrrden:
P.S. according to the headers we use Catch 2.13.3 if that helps

jianminzhao (approved):

PR #1393, opened 2022-03-04
CBL-2670 : fix a bug in BLIPIO that caused loopback test result flaky.

jianminzhao:


Networking/BLIP/BLIPConnection.cc
```diff
@@ -283,7 +281,9 @@ namespace litecore { namespace blip {
         void requeue(MessageOut *msg, bool andWrite =false) {
             DebugAssert(!_outbox.contains(msg));
             auto i = _outbox.end();
-            if (msg->urgent() && _outbox.size() > 1) {
+            // Only allow unnumbered msg to be inserted.
+            // Numbered msg must be appended
+            if (msg->_number == 0 && msg->urgent() && _outbox.size() > 1) {
```
snej, line 286: This is incorrect. Please try without it; the other change below should fix the bug.

snej (changes requested):

Networking/BLIP/BLIPConnection.cc
jianminzhao: If a numbered msg is inserted, say, m1 m3 m2, then at the time it is pulled out from the outbox, we will have out-of-order and triggering "too high" exception.

borrrden:
This warrants some explanation as to how this was found instead of an empty commit description.

jianminzhao:
Author's notes: Messages travels through BLIPIO via FIFO queue, outbox. At the place where we pull messages out of the queue, we assert that messages follows consecutive order, 1,2,3... However, at the place we send the message to the queue, we may insert it before the last item. If the messages are already numbered, it would violate the assertion.
Since the message number is used to multiplex purpose, the initial assignment of the number is arbitrary. So, instead of assigning the number before push it to the queue, we assign the number at the place after we pull it out of the queue and put it the wire. For those messages that are already numbered, I cannot change the number and avoid inserting them.
Of course, if "urgent" is more important, which is why we insert it, there is no absolute reason why we cannot send msg 3 before msg 2, we just need to give up the assertion.

Networking/BLIP/BLIPConnection.cc
snej: Only the first frame of a message has to have the consecutive numbering.

snej (changes requested):
Please remove the changes to the queuing method.

snej:
> At the place where we pull messages out of the queue, we assert that messages follows consecutive order, 1,2,3

No. The assertion happens on the receiving end when it reads a frame. That has nothing to do with the outbox queue, which is used for sending.

Only the first frame of each message needs to have the next sequential number. When sending subsequent frames, it reuses the existing number. That's the purpose of the number: to let the receiver associate multiple frames of the same message.

snej (approved):

PR #1394, opened 2022-03-04
Connected client and Async

snej:
Infrastructure:
* `Result<T>` holds either a `T` or a `C4Error`. Similar to types found in Swift, Kotlin, Rust, ...
* `Async<T>` is a promise/future of a `T`. (Actually of a `Result<T>`, since it can fail.) Integrates with Actors.

Connected Client:
* `ConnectedClient` opens a WebSocket (just like Replicator) and uses the BLIP replicator API to make CRUD requests and observe changes. Results are returned as `Async` values.
* This required a few additions to the replicator API, so I've extended Pusher and Puller to support them so we can run unit tests.


snej:
Jenkins iOS build is failing with `clang: error: unknown argument: '-fsanitize-ignorelist=/Users/mobile/..../couchbase-lite-core/Xcode/sanitizer-ignore-list.txt'`
```

I'm guessing the Xcode on the build machine has a version of Clang that's old enough not to understand this flag.

**Comment by pasin** [comment]
@snej Is this ready to review?

File: Networking/BLIP/BLIPProtocol.hh [hunk 1]
```diff
@@ -11,36 +11,61 @@
...
         kMoreComing = 0x40,     // Used only in frames, not in messages
     };
 
 
-    typedef uint64_t MessageNo;
+    using messageno_t = uint64_t;
+
+    /// A message number. Each peer numbers messages it sends sequentially starting at 1.
+    /// Each peer's message numbers are independent.
+    enum class MessageNo : messageno_t { };
```
callumbirks, line 47: This is something I've seen before but don't understand, what's the reasoning for declaring this an `enum class`?

callumbirks: Additionally, what's the motivation in the first place to make `MessageNo` its own class rather than simply using u64?

Networking/BLIP/docs/Async.md
callumbirks, line 1: I _love_ this document. Very excited to get these changes into the codebase.

callumbirks (commented):
LGTM, as far as I can tell from attempting to read through. I'm certain there's things I could have missed, but that's what tests are for, and it looks like we're pretty well covered with those.
I need to update this branch to the latest master, and prepare to merge it in. What else do we need before that happens? I imagine changes for VV support? Anything else?

Replicator/c4Socket+Internal.hh
```diff
@@ -22,7 +22,7 @@ namespace litecore { namespace repl {
     // Main factory function to create a WebSocket.
     fleece::Retained<websocket::WebSocket> CreateWebSocket(websocket::URL,
                                                            fleece::alloc_slice options,
-                                                           C4Database* NONNULL,
+                                                           C4Database*,
```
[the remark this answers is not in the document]
callumbirks: Actually, this question is irrelevant because it's replaced with shared_ptr<DBAccess> in master

PR #1397, opened 2022-03-09
CBL-2764 : fix a bug of "heap-use-after-free".

jianminzhao:
From the report of Mac's Address Sanitizer, we ran into a situation whereby, concurrently,
A. as the holder of the last reference of a Query object, LiveQuerierDelegate, is deleted, the Query object goes to the deletion process;
B. as DataFile is deleted, the close method is called on the same Query object.
Currently, DataFile keeps plain pointers of Query objects, which are registered/unregistered as they are constructed/destructed. We fix the problem by letting DataFile to hold Retained<Query>s, so that the query objects would be deleted either in the DataFile's thread or in the LiveQuerierDelegate's thread.
There is a technical obstacle: the constructor of Query may throw. If it throws, the ref count will be out of balance. We solve this problem by holding the exception and handle it in the higher level.

### DIFF
--- a/C/Cpp_include/c4Query.hh
+++ b/C/Cpp_include/c4Query.hh
@@ -57,6 +57,7 @@ public:
     void setParameters(slice parameters);
 
     alloc_slice fullTextMatched(const C4FullTextMatch&);
+    std::exception_ptr current_exception();
 
     // Running the query:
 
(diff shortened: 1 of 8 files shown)


snej (changes requested):
You can’t have SQLiteDataFile retaining the queries it registers. That means any query you create will never be freed until the database is closed. That’s exactly why I used raw pointers: they’re weak references.

Adding a mutex seems like the right solution to this problem.

snej:
[How about this fix?](https://github.com/couchbase/couchbase-lite-core/files/8226115/query.patch.txt)

I think it's safe, since none of the code run while holding the mutex grabs any other mutexes.

PR #1401, opened 2022-03-16
CBL-2670 : raise the log level to debug for BLIPLog

jianminzhao:
This bug is not easy to show. It occurs intermittently in Jenkins' macosx build. The purpose of this commit is to have detailed logs when the test fails.


jianminzhao:
Author's notes: the change won't affect the console output if tests succeed.

borrrden (approved):
Changing to debug won't do anything more than verbose for release builds, FYI.

PR #1404, opened 2022-03-16
use quiet reporter in Jenkins build.

jianminzhao:


borrrden:
What about https://github.com/couchbase/couchbase-lite-core/blob/master/jenkins/build_server_win.ps1 ?

borrrden (changes requested):
You can do away with the environment variable, and also make the same changes to the Windows build script.

borrrden (approved):

PR #1405, opened 2022-03-16
CBL-2671 : raise the log level to debug for BLIPMessages (#1401)

jianminzhao:
This bug is not easy to show. It occurs intermittently in Jenkins' macosx build. The purpose of this commit is to have detailed logs when the test fails.


borrrden (approved):

pasin:
Do we have a ticket to un-raise the log level especially before release?

PR #1409, opened 2022-03-17
CBL-2670 : ensure message order for LoopbackWebSocket.

jianminzhao:
LoopbackWebSocket implements messsaging within the same application by threaded queues. For Mac, in particular, we use GCD's dispatch_after in order to simulate network latency. Since dispatch_after takes absolute time in the future, as opposed to a delta delay, the order may be displaced if two future times are numerically equal. To guarantee the ordering, we take advantage of the ordering guarantee of dispatch_async with the serial queue: we queue the messages with dispatch_async, and actually send the message with dispatch_after.

### DIFF
--- a/Networking/BLIP/LoopbackProvider.hh
+++ b/Networking/BLIP/LoopbackProvider.hh
@@ -38,7 +38,7 @@ namespace litecore { namespace websocket {
         class Driver;
     private:
         Retained<Driver> _driver;
-        actor::delay_t _latency;
+        const actor::delay_t _latency;
 
     public:
 
@@ -320,7 +320,7 @@ namespace litecore { namespace websocket {
             friend class LoopbackWebSocket;
 
             Retained<LoopbackWebSocket> _webSocket;
-            actor::delay_t _latency {0.0};
+            const actor::delay_t _latency {0.0};
             Retained<LoopbackWebSocket> _peer;
             websocket::Headers _responseHeaders;
             std::atomic<size_t> _bufferedBytes {0};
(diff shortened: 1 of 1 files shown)


snej (changes requested):
This fix is correct, but it's only needed in the case where latency is nonzero, i.e. unit tests. Unfortunately it will affect performance in actual (non-test) use, since two async calls are issued instead of one.

I suggest having the `received` method use the old behavior when latency==0 and your new behavior when it's nonzero.

jianminzhao:
Good point. Will do it.

borrrden:
I don't understand how this works.  What if you enqueue a message with a delay of 10 seconds, and then enqueue another one with a delay of 5?

Won't this guarantee now that they are in the wrong order?  The will enqueue in the correct order but dequeue in the wrong order.

snej (approved):

borrrden:
I don't understand how this works.  Won't this guarantee that things come in the wrong order in the following case?

Add a message for 10 seconds later
Add a message for 5 seconds later

5 seconds later, the enqueueAfter fires and processes the message meant for 10 seconds later.

jianminzhao:
@borrrden you have a valid point by reading the math locally. However, this is simulating the web socket's latency. By the protocol of web sockets, messages must be received in the order sent.
You are right, latency is a member field of LoopbackWebSocket object. It is supposed to be constant. I probably should emphasize by qualifying it with const.

borrrden (approved):
Ok with constant latency the route to failure is removed

PR #1411, opened 2022-03-18
Connected client LiteCore APIs

jayahariv:
* Connected client APIs for create a new connected client and get document

### DIFF
--- a/C/Cpp_include/c4ConnectedClient.hh
+++ b/C/Cpp_include/c4ConnectedClient.hh
@@ -19,11 +19,16 @@ C4_ASSUME_NONNULL_BEGIN
 
 struct C4ConnectedClient  : public fleece::RefCounted,
                             public fleece::InstanceCountedIn<C4Database>,
-                            C4Base  {
+                            C4Base
+{
     
     static Retained<C4ConnectedClient> newClient(const C4ConnectedClientParameters &params);
                                 
     virtual litecore::actor::Async<C4DocResponse> getDoc(C4Slice, C4Slice, C4Slice, bool) noexcept=0;
+
+    virtual void start() noexcept=0;
+
+    virtual void stop() noexcept=0;
 };
 
 C4_ASSUME_NONNULL_END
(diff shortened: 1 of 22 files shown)


C/Cpp_include/c4ConnectedClient.hh
```diff
@@ -0,0 +1,37 @@
...
+// Use of this software is governed by the Business Source License included
+// in the file licenses/BSL-Couchbase.txt.  As of the Change Date specified
+// in that file, in accordance with the Business Source License, use of this
+// software will be governed by the Apache License, Version 2.0, included in
+// the file licenses/APL2.txt.
+//
+
+#pragma once
+#include "c4Base.hh"
+#include "Async.hh"
```
snej, line 15: This is interesting ... I hadn't considered exposing Async as part of the LiteCore C++ API; I considered it internal like Actor. But it would definitely simplify this API. And Async.hh has few dependencies. I'll think about this.

snej (changes requested):
This is looking good overall! Reading through this, I'm realizing how much LiteCore code you had to figure out. 🤓

Besides little nits, the only troublesome thing is the `_delegate = nullptr` issue, which has been known to cause pain in other contexts. We should ask @borrrden what he thinks.

Replicator/ConnectedClient/ConnectedClient.cc
```diff
@@ -90,8 +100,10 @@ namespace litecore::client {
     void ConnectedClient::onHTTPResponse(int status, const websocket::Headers &headers) {
         asCurrentActor([=] {
             logVerbose("Got HTTP response from server, status %d", status);
+            LOCK(_mutex);
```
snej, line 103: The scope of the lock shouldn't extend past the delegate call; wrap curly braces around these 3 lines.

Replicator/ConnectedClient/ConnectedClient.cc
```diff
@@ -141,7 +153,8 @@ namespace litecore::client {
                 }
                 gotError(C4Error::make(domain, code, status.message));
             }
-
+            
+            LOCK(_mutex);
```
snej, line 157: The scope of the lock shouldn't extend past the delegate call; wrap curly braces around these 3 lines.

snej (changes requested):

snej (approved):

PR #1415, opened 2022-03-22
faster start of replicator

KraemerDEM:
When starting a replicator there are 3 queries done on the local database which are not using the correct index.
This leads to a very long time until the first reaction from the replicator occurs.
I had a 3GB sqlite database file on an IPad which took several minutes(!) to complete these queries. This PR ensures the appropriate indexes are used and the queries are completed almost instantly.

### DIFF
--- a/LiteCore/Storage/SQLiteEnumerator.cc
+++ b/LiteCore/Storage/SQLiteEnumerator.cc
@@ -81,7 +81,9 @@ namespace litecore {
         sql << (options.contentOption >= kEntireBody     ? ", extra" : ", length(extra)");
         sql << (mayHaveExpiration() ? ", expiration" : ", 0");
         sql << " FROM kv_" << name();
-        
+        if (options.onlyConflicts)
+            sql << " INDEXED BY kv_" << name() << "_conflicts";
+
         bool writeAnd = false;
         if (bySequence) {
             sql << " WHERE sequence > ?";
--- a/LiteCore/Storage/SQLiteKeyStore.cc
+++ b/LiteCore/Storage/SQLiteKeyStore.cc
@@ -584,7 +584,7 @@ namespace litecore {
     expiration_t SQLiteKeyStore::nextExpiration() {
         expiration_t next = expiration_t::None;
         if (mayHaveExpiration()) {
-            auto &stmt = compileCached("SELECT min(expiration) FROM kv_@");
+            auto &stmt = compileCached("SELECT min(expiration) FROM kv_@ where expiration NOT NULL");
             UsingStatement u(stmt);
             if (!stmt.executeStep())
                 return next;


LiteCore/Storage/SQLiteEnumerator.cc
```diff
@@ -81,7 +81,9 @@ namespace litecore {
         sql << (options.contentOption >= kEntireBody     ? ", extra" : ", length(extra)");
         sql << (mayHaveExpiration() ? ", expiration" : ", 0");
         sql << " FROM kv_" << name();
-        
+        if (options.onlyConflicts)
+            sql << " INDEXED BY kv_" << name() << "_conflicts";
```
snej, line 85: Interesting; the query doesn't use this index otherwise?

But I'm put off by this very emphatic statement from the SQLite docs:

> The INDEXED BY clause is _not_ intended for use in tuning the performance of a query. The intent of the INDEXED BY clause is to raise a run-time error if a schema change, such as dropping or creating an index, causes the query plan for a time-sensitive query to change. The INDEXED BY clause is designed to help detect undesirable query plan changes during regression testing. Application developers are admonished to omit all use of INDEXED BY during application design, implementation, testing, and tuning. If INDEXED BY is to be used at all, it should be inserted at the very end of the development process when "locking down" a design.

Is there any other way to get the query to use this index?

LiteCore/Storage/SQLiteKeyStore.cc
```diff
@@ -584,7 +584,7 @@ namespace litecore {
     expiration_t SQLiteKeyStore::nextExpiration() {
         expiration_t next = expiration_t::None;
         if (mayHaveExpiration()) {
-            auto &stmt = compileCached("SELECT min(expiration) FROM kv_@");
+            auto &stmt = compileCached("SELECT min(expiration) FROM kv_@ where expiration NOT NULL");
```
snej, line 587: Good catch -- the index definition says `NOT NULL` so the query has to include it too.

LiteCore/Storage/SQLiteEnumerator.cc
KraemerDEM: The query is this:
`SELECT sequence, flags, key, version, length(body), length(extra), expiration FROM kv_default WHERE (flags & 2) != 0 ORDER BY key`
It seems the ORDER BY clause causes SQLite to prefer an Index on "key".
If I remove the order by clause, SQLite uses kv_default_conflicts.
Maybe adding "key" to the fields of the  kv_default_conflicts index will help?
IMHO SQLite is not very good at choosing an index if there are multiple matching indexes to choose from.

snej: The query shouldn't have an ORDER BY clause. It isn't necessary, and the `Replicator::_findExistingConflicts` calls `DBAccess::unresolvedDocsEnumerator`) with a `false` parameter for `orderByID`, so the query's `kC4Unsorted` flag is set, which should disable ORDER BY.

snej: Oh, the sort gets changed to `kAscending` in `BothKeyStore::newEnumeratorImpl`, because the `includeDeleted` flag is set, so `BothKeyStore` needs to merge results from the live and dead docs KeyStores.

SQLite's ["unary +" operator](https://sqlite.org/optoverview.html#uplus) tells the optimizer not to use an index for a term. Unfortunately the docs say it only works in the WHERE clause.

Another option would be to turn off the includeDeleted option for this query. Can a deleted document ever be conflicted? I've forgotten this part of the replicator logic.

borrrden:
What is the plan for this?  Should we accept just the undisputed half?

LiteCore/Storage/SQLiteKeyStore.cc
jianminzhao: This issue has been found in a separate ticket. It's fixed in release/lithium, 1e2cbb4a1

jianminzhao:
@KraemerDEM , as Jens pointed it out, it should not need to use "ORDER BY key." How did you get it to use ORDER BY?

jianminzhao:
@KraemerDEM, as @jens pointed out, the query should not use "ORDER BY key". How did you get it to use "ORDER BY" ?

KraemerDEM:
Well, as @snej also pointed out, it's because in BothKeyStore::newEnumeratorImpl the sort order is changed as options.includeDeleted is set?

jianminzhao:
. @KraemerDEM  BothKeyStore starts in release/3.1 branch. I assume you were using release/lithium. Which version of CBL are you using?

KraemerDEM:
@jianminzhao I was using the git master branch at that time.

PR #1417, opened 2022-03-23
Fix Linux build of perf test with prebuilt LiteCore

borrrden:
And make sure that per-platform metrics are output


borrrden:
I've been sneaking these in but this one I wanted a PR for because there was a risk of breaking compilation for non-Windows.  Seems ok though.

jianminzhao (approved):

PR #1419, opened 2022-03-24
CBL-2942 : Fix crash when disable query observer twice

pasin:
* When enableObserver() is called twice to disable, there is a race b/w resetting _bgQuery in C4Query::liveQuerierStopped() called by an actor thread and calling _bgQuerier.stop() in C4Query::enableObserver(). As a result, _bgQuerier.stop() would could cause a crash as _bgQuerier could be null.

* Added a mutex lock in liveQuerierStopped() to prevent the race.

### DIFF
--- a/C/c4Query.cc
+++ b/C/c4Query.cc
@@ -306,8 +306,17 @@ void C4Query::liveQuerierStopped() {
     // CBL-2673: Wait until _bgQuerier is done with its async stuff before freeing it
     // and its delegate, otherwise a race could cause a liveQuerierUpdated call to
     // a garbage delegate.
-    _bgQuerier = nullptr;
-    _bgQuerierDelegate = nullptr;
+    //
+    // Retain and Release the C4Query object to make sure that the object is not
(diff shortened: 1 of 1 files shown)


borrrden (approved):

C/c4Query.cc
```diff
@@ -306,8 +306,17 @@ void C4Query::liveQuerierStopped() {
...
+    // Retain and Release the C4Query object to make sure that the object is not
+    // freed until the end as setting _bgQuerierDelegate (retaining the C4QueryObject)
+    // could trigger the freeing chain including the _mutex itself.
+    c4query_retain(this);
+    {
+        LOCK(_mutex);
+        _bgQuerier = nullptr;
+        _bgQuerierDelegate = nullptr;
+    }
+    c4query_release(this);
```
jianminzhao, line 319: Whoever calls this should hold a reference count, right? Somewhere else seems to have a problem

pasin: Not sure what you mean but the release here is matched with the retain above. If nobody retains the query object, it will be freed.

PR #1421, opened 2022-03-25
SequenceTracker: reset external every time

### LINKED ISSUE #1413: c4dbobs_getChanges return external = true for local change
I saw the following behavior with current couchbase-lite-core master (c9c67c4dcf35f6dfcd8421544ca3e71f709ba3f2)
1. Client A edit document **X**
2. Client B received notification about change in document **X** via `c4dbobs_getChanges`
    with `outExternal` = true
3. Client B edit document **X**, and receive `outExternal` = true from `c4dbobs_getChanges`

where `outExternal` is paramter of `c4dbobs_getChanges`
```
    uint32_t c4dbobs_getChanges(C4CollectionObserver *observer,
                                C4CollectionChange outChanges[C4NONNULL],
                                uint32_t maxChanges,
                                bool *outExternal) C4API;
```

So if Client A not connected or not edit document, `external` flag setted correctly for local change on client B side,
but after the first external edit, `c4dbobs_getChanges` starts setted `outExternal` to `true` even for local changes.

I fixed it like this:
```
@@ -258,6 +258,8 @@ namespace litecore {
         if (!inTransaction()) {
             entry->committedSequence = sequence;
             entry->external = true; // it must have come from addExternalTransaction()
+        } else {
+            entry->external = false;
         }
 
         // Notify document notifiers:
```

though not sure is this right fix.

Dushistov:
fixes #1413 , to prevent local change wrongly marked as external,
because of there was external change in this document before.

Not sure about `committedSequence`. Do I need to reset this variable too? 
Because of `committedSequence` is setted in the same way as `external`:
```
        if (!inTransaction()) {
            entry->committedSequence = sequence;
            entry->external = true; // it must have come from addExternalTransaction()
        }
```

### DIFF
--- a/LiteCore/Database/SequenceTracker.cc
+++ b/LiteCore/Database/SequenceTracker.cc
@@ -247,6 +247,7 @@ namespace litecore {
             entry->sequence = sequence;
             entry->bodySize = shortBodySize;
             entry->flags = flags;
+            entry->external = false;
         } else {
             // or create a new entry at the end:
             _changes.emplace_back(docID, revID, sequence, shortBodySize, flags);


LiteCore/Database/SequenceTracker.cc
```diff
@@ -255,6 +255,7 @@ namespace litecore {
             entry = &*change;
         }
 
+        entry->external = false;
```
snej, line 258: This should be on line 250.

snej:
> Not sure about committedSequence. Do I need to reset this variable too?

No, I believe it's correct as it is.

LiteCore/Database/SequenceTracker.cc
Dushistov: Done

snej (approved):

PR #1422, opened 2022-03-25
CBL-2877 : use random string for the revision ID in lieu of SHA1 dige…

jianminzhao:
…st of the content.

### DIFF
--- a/LiteCore/Database/TreeDocument.cc
+++ b/LiteCore/Database/TreeDocument.cc
@@ -591,9 +591,7 @@ namespace litecore {
             if (!body)
                 return false;
 
-            DatabaseImpl* db = _collection->dbImpl();
-            revidBuffer encodedNewRevID = generateDocRevID(body, _selected.revID, deletion,
-                db->dataFile()->documentKeys());
+            revidBuffer encodedNewRevID = generateDocRevID(body, _selected.revID, deletion);
 
             C4ErrorCode errorCode = {};
             int httpStatus;
@@ -660,27 +658,25 @@ namespace litecore {
             };
             AuxDoc doc(body, sk);
 
-            bool ret = false;
             for (DeepIterator i(doc.asDict()); i; ++i) {
                 const Dict* dict = i.value()->asDict();
                 if (dict) {
                     const Value* objType = dict->get(C4Document::kObjectTypeProperty);
                     if (objType && objType->asString() == C4Document::kObjectType_Encryptable) {
-                        ret = true;
-                        break;
+                        return true;
                     }
                 }
             }
-            return ret;
+            return false;
 #endif
         }
 
-        static revidBuffer generateDocRevID(slice body, slice parentRevID, bool deleted, SharedKeys* sk) {
+        revidBuffer generateDocRevID(slice body, slice parentRevID, bool deleted) {
             // Get SHA-1 digest of (length-prefixed) parent rev ID, deletion flag, and revision body:
             uint8_t revLen = (uint8_t)min((unsigned long)parentRevID.size, 255ul);
             uint8_t delByte = deleted;
             SHA1 digest;
-            if (hasEncryptables(body, sk)) {
+            if (hasEncryptables(body, _collection->dbImpl()->dataFile()->documentKeys())) {
                 mutable_slice mslice(digest.asSlice());
                 SecureRandomize(mslice);
             } else {
--- a/LiteCore/Database/TreeDocument.cc
+++ b/LiteCore/Database/TreeDocument.cc
@@ -20,11 +20,13 @@
 #include "Record.hh"
 #include "RawRevTree.hh"
 #include "RevTreeRecord.hh"
+#include "DeepIterator.hh"
 #include "Delimiter.hh"
 #include "Error.hh"
 #include "StringUtil.hh"
 #include "SecureRandomize.hh"
 #include "SecureDigest.hh"
+#include "SecureSymmetricCrypto.hh"
 #include "FleeceImpl.hh"
 #include "slice_stream.hh"
 #include <ctime>
@@ -589,7 +591,9 @@ namespace litecore {
             if (!body)
                 return false;
 
-            revidBuffer encodedNewRevID = generateDocRevID(body, _selected.revID, deletion);
+            DatabaseImpl* db = _collection->dbImpl();
+            revidBuffer encodedNewRevID = generateDocRevID(body, _selected.revID, deletion,
+                db->dataFile()->documentKeys());
 
             C4ErrorCode errorCode = {};
             int httpStatus;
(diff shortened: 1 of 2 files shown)


LiteCore/Database/TreeDocument.cc
```diff
@@ -640,14 +644,51 @@ namespace litecore {
...
-        static revidBuffer generateDocRevID(slice body, slice parentRevID, bool deleted) {
+        static revidBuffer generateDocRevID(slice body, slice parentRevID, bool deleted, SharedKeys* sk) {
             // Get SHA-1 digest of (length-prefixed) parent rev ID, deletion flag, and revision body:
             uint8_t revLen = (uint8_t)min((unsigned long)parentRevID.size, 255ul);
             uint8_t delByte = deleted;
-            SHA1 digest = (SHA1Builder() << revLen << slice(parentRevID.buf, revLen)
-                                         << delByte << body)
-                           .finish();
+            SHA1 digest;
+            if (hasEncryptables(body, sk)) {
```
pasin, line 683: CBL can provide this info similar to kRevHasAttachments. Want to discuss if we would like CBL to provide or as this is a security issue, we want to do it in LiteCore to avoid bugs from CBL side?

borrrden (commented):
I am also curious about what Pasin mentioned

jianminzhao:
Not clear of Pasin's concern. Take it to slack.

jianminzhao:
I had slacked with Pasin about his comment. Let me venture a summary here.
This is the only place where we calculate the revID, 
  bool TreeDocument::putNewRevision(const C4DocPutRequest &rq, C4Error *outError) override
 can we require the caller to provide the info as whether the revision contains encryptable in the argument, rq.flags, to save the cost of walking through the properties to detect the encryptable?
This is motivated by the fact that we already walked through the properties in CBL as we encode the body, CBLDocument::encodeBody(),
"""alloc_slice CBLDocument::encodeBody(CBLDatabase* db,
                                    C4Database* c4db,
                                    bool releaseNewBlob,
                                    C4RevisionFlags &outRevFlags) const
{
    auto c4doc = _c4doc.useLocked();
    // Save new blobs and check encryptables in arrays:
    bool hasBlobs = saveBlobsAndCheckEncryptables(db, releaseNewBlob);
    outRevFlags = hasBlobs ? kRevHasAttachments : 0;
"""
Perhaps we can add a flag here, kRevHasEncruptable.
@pasin - is my summary accurate?
@snej @borrrden - what do you think?

pasin:
Yes, the summary is accurate.

LiteCore/Database/TreeDocument.cc
```diff
@@ -589,7 +591,9 @@ namespace litecore {
             if (!body)
                 return false;
 
-            revidBuffer encodedNewRevID = generateDocRevID(body, _selected.revID, deletion);
+            DatabaseImpl* db = _collection->dbImpl();
+            revidBuffer encodedNewRevID = generateDocRevID(body, _selected.revID, deletion,
```
snej, line 595: Instead of adding a SharedKeys parameter to `generateDocRevID`, just make that method non-static, so it can get it itself. (It still makes sense for it to take the revID as a parameter, IMO.)

LiteCore/Database/TreeDocument.cc
```diff
@@ -640,14 +644,51 @@ namespace litecore {
...
+        static bool hasEncryptables(slice body, SharedKeys* sk) {
+#ifndef COUCHBASE_ENTERPRISE
+            return false;
+#else
+            const Value* v = Value::fromTrustedData(body);
+            if (v == nullptr) {
+                return false;
+            }
+
+            struct AuxDoc : public Doc {
```
snej, line 656: I don't see a need to subclass Doc; you're just passing different parameters to the constructor.

LiteCore/Database/TreeDocument.cc
```diff
@@ -640,14 +644,51 @@ namespace litecore {
...
+                AuxDoc(slice body, SharedKeys* sk)
+                : Doc(alloc_slice(body), Doc::kTrusted, sk)
+                {}
+            };
+            AuxDoc doc(body, sk);
+
+            bool ret = false;
+            for (DeepIterator i(doc.asDict()); i; ++i) {
```
snej, line 664: Can't you use `v->asDict()` instead and avoid having to create the Doc?

LiteCore/Database/TreeDocument.cc
```diff
@@ -640,14 +644,51 @@ namespace litecore {
...
+                const Dict* dict = i.value()->asDict();
+                if (dict) {
+                    const Value* objType = dict->get(C4Document::kObjectTypeProperty);
+                    if (objType && objType->asString() == C4Document::kObjectType_Encryptable) {
+                        ret = true;
```
snej, line 669: Simpler just to `return true`

snej (changes requested):

snej:
> Perhaps we can add a flag here, kRevHasEncryptable

That would be good, except there are no available bits in Rev::Flags, and that field is stored as a byte in the binary revision encoding, so expanding it would break binary compatibility.

I'm thinking we could repurpose kRevHasAttachments to mean "has attachments or encryptables", since they are both embedded dicts requiring special processing... That would likely require some changes to code that uses the flag.

snej:
I think for now, leave the deep scan as in this PR. But we should look into using a flag or something to avoid having to make multiple passes.

This is another case where I really wish we had performance tests so we could tell whether this is a noticeable regression.

LiteCore/Database/TreeDocument.cc
jianminzhao: @snej I just found out that we have duplicated code of generating rev id in 
   alloc_slice VectorRecord::generateRevID(Dict body, revid parentRevID, DocumentFlags flags)
Do I need to take care of it here?

snej: No. Version-vector revIDs are not affected by the content of the document, so they don't have this issue that we're fixing.

LiteCore/Database/TreeDocument.cc
jianminzhao: The reason I subclass it is because it is in fleece::impl and here fleece::impl::Doc has destructor being protected.

LiteCore/Database/TreeDocument.cc
jianminzhao: This is where I don't fully understand it and took the safe path. In this function, I am not sure whether body has already registered with the scope. Has it? If yes, then I don't need to resort to Doc, I think. If not, don't I need to wrap it into a Doc with SharedKeys?

jianminzhao: Looks like it has not been registered with the scope for the program throws if I use v->asDict(), not wrapping it to a Doc with shared keys.

LiteCore/Database/TreeDocument.cc
snej: That's because Doc is refcounted. You never allocate a RefCounted object on the stack, only on the heap, and use a Retained value to manage the reference.

The protected destructor is explicitly meant to keep people from trying to do what you were doing.

LiteCore/Database/TreeDocument.cc
snej: That's unfortunate but I can see why it's happening (for historical reasons.) It's better to use a fleece::impl::Scope, then, because it doesn't require copying the doc body to an alloc_slice.

(Scope is not ref-counted, so you can declare one directly, btw.)

borrrden (approved):
Since there are issues using the flag route, I will approve this but let's keep our eyes on the next round of perf test results to see if we notice anything.

pasin (approved):

snej (approved):

PR #1425, opened 2022-03-28
CBL-2942 : Move retain/release on query object

pasin:
According to code review, move retain/release on the query object to the place that uses the query object.

### DIFF
--- a/C/c4Query.cc
+++ b/C/c4Query.cc
@@ -215,7 +215,15 @@ public:
     }
 
     void liveQuerierStopped() override {
+        // There is circular retain ref between LiveQuerierDelegate and C4Query object.
+        // The circular ref will be broken when _query->liveQuerierStopped() is called
+        // as the LiveQuerierDelegate object will be set to NULL and be freed.
+        // However, if nobody else has retained the _query object, the _query object would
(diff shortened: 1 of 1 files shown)


jianminzhao (approved):

C/c4Query.cc
```diff
@@ -215,7 +215,15 @@ class C4Query::LiveQuerierDelegate : public LiveQuerier::Delegate {
     }
 
     void liveQuerierStopped() override {
+        // There is circular retain ref between LiveQuerierDelegate and C4Query object.
+        // The circular ref will be broken when _query->liveQuerierStopped() is called
+        // as the LiveQuerierDelegate object will be set to NULL and be freed.
+        // However, if nobody else has retained the _query object, the _query object would
+        // be freed as well. To ensure that _query lives beyound the liveQuerierStopped() call,
+        // addtional retain and release on the _query object are added here.
+        c4query_retain(_query);
```
jianminzhao, line 224: Windows and Linus failed. I know what's going on. It's because the delegate object is already deleted  after query->liveQuerierStopped(). Can be fixed by
    Retained<C4Query> local = _query;
    local->liveQuerierStopped();

PR #1429, opened 2022-03-31
Update Android NDK to 23b

borrrden:
In the process it was determined that we need to move to a stock CMake and Ninja, so fetch those from cbdeps

### DIFF
--- a/LiteCore/Android/getifaddrs.h
+++ b/LiteCore/Android/getifaddrs.h
@@ -51,8 +51,8 @@ __BEGIN_DECLS
 #define ifa_broadaddr ifa_ifu.ifu_broadaddr
 #define ifa_dstaddr ifa_ifu.ifu_dstaddr*/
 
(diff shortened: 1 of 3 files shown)


borrrden:
macOS test failure unrelated to this PR

bmeike:
Is there a branch I can check out, to see if this affects me at all.  I'm pretty sure it does not, but I'd like to be able to  verify that.

.... and sorry to be so long about looking at this.  I didn't realize until today that it was waiting on me.

borrrden:
A branch you can check out, yes, but I am guessing that is not what you mean?  Do you mean that you want a prebuilt artifact to test with?  The branch name is at the top of this PR:

`feature/cbl-2943_bump_ndk`

bmeike:
Yeah... I could work with just the branch... but if you have artifacts that would let me get to it more quickly...

bmeike:
I am about to commit a change that will have Android building with exactly the same tool chain.  This should be good to go.

bmeike (approved):
👍🏻

PR #1432, opened 2022-04-02
CBL-2679: Implement an info API for environment information 

borrrden:
Such as what LiteCore thinks the current timezone offset is and what locales it supports for collation


borrrden:
Please squash into just the first commit message 😄

jianminzhao (approved):

PR #1439, opened 2022-04-11
CBL-3007 : disable test 'Rest root level' for Win32.

jianminzhao:


borrrden (approved):
This could get messy in the future.  For now it is ok but you could also use ifdef to simply not compile this test.

PR #1443, opened 2022-04-15
Fix SequenceTracker bug with docs observed before a db change [CBL 3013]

snej:
SequenceTracker::_documentChanged was incorrectly ignoring a change
to a document with an existing-but-idle Entry when there were no db
change notifiers.

In practice this meant that if you created a doc-change notifier for
a doc that hadn't been modified yet in this db instance, and then
updated the document, the change wouldn't be recorded in the
SequenceTracker so other db instances wouldn't find out about it.

The offending lines were last edited in 2016(!), so I am not sure
exactly why they were there. I think they date back to a time when
there was only a single SequenceTracker shared by all DB instances on
a file, in which case that behavior would be a valid optimization.

It should be safe to backport this fix to any earlier version of
CBL, back to 2.0.

CBL-3013

### DIFF
--- a/LiteCore/Database/SequenceTracker.cc
+++ b/LiteCore/Database/SequenceTracker.cc
@@ -224,6 +224,8 @@ namespace litecore {
                                            uint64_t bodySize,
                                            RevisionFlags flags)
     {
+        logDebug("documentChanged('%.*s', %.*s, %llu, size=%llu, flags=%hhx",
+                 SPLAT(docID), SPLAT(revID), sequence, bodySize, flags);
         auto shortBodySize = (uint32_t)min(bodySize, (uint64_t)UINT32_MAX);
         bool listChanged = true;
         Entry *entry;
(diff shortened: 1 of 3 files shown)


LiteCore/Database/SequenceTracker.cc
```diff
@@ -224,24 +224,22 @@ namespace litecore {
...
+        logDebug("documentChanged('%.*s', %.*s, %llu, size=%llu, flags=%hhx",
+                 SPLAT(docID), SPLAT(revID), sequence, bodySize, flags);
         auto shortBodySize = (uint32_t)min(bodySize, (uint64_t)UINT32_MAX);
         bool listChanged = true;
         Entry *entry;
         auto i = _byDocID.find(docID);
         if (i != _byDocID.end()) {
             // Move existing entry to the end of the list:
             entry = &*i->second;
-            if (entry->isIdle() && !hasDBChangeNotifiers()) {
```
borrrden, line 234: Is `hasDBChangeNotifiers` used anywhere else?  If not can it be removed?

snej: Good point; the answers are No and Yes. I've removed it now.

pasin (approved):
Also verified with the test using the replicator.

PR #1447, opened 2022-04-20
CBL-2936 : update mbedTLS to 2.28.0

jianminzhao:


pasin (approved):
Seem like there are a couple of minor things got updated as part of the upgrade as well.

PR #1454, opened 2022-04-22
Proper fleece encoding for uint64 query result

### LINKED ISSUE #1445: std::numeric_limits<uint64_t>::max() is not FLValue_IsUnsigned
In code bellow I create document with `std::numeric_limits<uint64_t>::max()` as field value,
save it to database and then using query I get it back.
All works fine, except that `FLValue_IsUnsigned` return false for result of query,
but  FLValue_AsUnsigned  return right value (18446744073709551615).
Debugger show that `Value::IsUnsigned` return false, because of tag() == fleece::impl::internal::kShortIntTag .


```c++
#include <c4Database.h>                                                                                                                                                       
#include <c4Query.h>                                                                                                                                                          
#include <c4Document+Fleece.h>                                                                                                                                                
#include <fleece/FLEncoder.h>                                                                                                                                                 
#include <cstdlib>                                                                                                                                                            
#include <cassert>                                                                                                                                                            
#include <limits>                                                                                                                                                             
                                                                                                                                                                                
int main() {                                                                                                                                                                  
    C4Error err;                                                                                                                                                                
    C4DatabaseConfig2 cfg = {                                                                                                                                                   
      FLStr("/tmp"),                                                                                                                                                            
      kC4DB_Create,                                                                                                                                                             
      {kC4EncryptionNone}                                                                                                                                                       
    };                                                                                                                                                                          
    auto db = c4db_openNamed(FLStr("demo"), &cfg, &err);                                                                                                                        
    assert(db != nullptr);                                                                                                                                                      
                                                                                                                                                                                
    {                                                                                                                                                                           
        const auto docId = FLStr("document");                                                                                                                                   
        auto doc = c4doc_get(db, docId, false, &err);                                                                                                                           
        assert(doc != nullptr);                                                                                                                                                       
        bool ret = c4db_beginTransaction(db, &err);                                                                                                                             
        assert(ret);                                                                                                                                                            
                                                                                                                                                                                
        auto enc = c4db_getSharedFleeceEncoder(db);                                                                                                                             
        assert(enc);                                                                                                                                                            
        FLEncoder_ConvertJSON(enc, FLStr("{\"ulong\": 18446744073709551615}"));                                                                                                 
        FLError fl_err;                                                                                                                                                         
        auto body = FLEncoder_Finish(enc, &fl_err);                                                                                                                             
        assert(body.size != 0);                                                                                                                                                 
        doc = c4doc_update(doc, {body.buf, body.size}, 0, &err);                                                                                                                
        assert(doc != nullptr);                                                                                                                                                 
        ret = c4db_endTransaction(db, true, &err);                                                                                                                              
        assert(ret);                                                                                                                                                            
    }
                                                                                                                                                                               
    auto query = c4query_new2(                                                                                                                                                  
        db, kC4N1QLQuery,                                                                                                                                                       
        FLStr("SELECT ulong FROM _default"),                                                                                                                                    
        nullptr, &err);                                                                                                                                                         
    assert(query != nullptr);                                                                                                                                                   
    auto en = c4query_run(query, nullptr, kFLSliceNull, &err);                                                                                                                  
    assert(en != nullptr);                                                                                                                                                      
    bool res = c4queryenum_next(en, &err);                                                                                                                                      
    assert(res);                                                                                                                                                                
    auto val = FLArrayIterator_GetValueAt(&en->columns, 0);                                                                                                                     
    printf("val is number %d, unsigned %d, is int %d\n", int(FLValue_GetType(val) == kFLNumber),                                                                                
           int(FLValue_IsUnsigned(val)), int(FLValue_IsInteger(val)));                                                                                                          
    assert(FLValue_AsUnsigned(val) == std::numeric_limits<uint64_t>::max());                                                                                                    
    printf("ulong value CORRECT\n");                                                                                                                                            
  } 
```

Dushistov:
fixes #1445

### DIFF
--- a/LiteCore/tests/QueryTest.cc
+++ b/LiteCore/tests/QueryTest.cc
@@ -12,10 +12,12 @@
 
 #include "QueryTest.hh"
 #include "SQLiteDataFile.hh"
+#include <cstdint>
 #include <ctime>
 #include <cfloat>
 #include <cinttypes>
 #include <chrono>
+#include <limits>
 #include <numeric>
 #include "date/date.h"
 #include "ParseDate.hh"
@@ -333,6 +335,44 @@ N_WAY_TEST_CASE_METHOD(QueryTest, "Query boolean", "[Query]") {
     int i = 1;
     while (e->next()) {
         CHECK(e->columns()[0]->asString().asString() == stringWithFormat("rec-%03d", i++));
+    }
+}
+
+N_WAY_TEST_CASE_METHOD(QueryTest, "Query uint64", "[Query]") {
+    {
+        ExclusiveTransaction t(store->dataFile());
+        const int64_t ivals[] = { std::numeric_limits<int64_t>::min(),
+                                  std::numeric_limits<int64_t>::max() };
+        const uint64_t uvals[] = { 0, std::numeric_limits<uint64_t>::max(), };
+        for(int i = 0; i < 4; i++) {
+            string docID = stringWithFormat("rec-%03d", i + 1);
+            writeDoc(slice(docID), DocumentFlags::kNone, t, [=](Encoder &enc) {
+                enc.writeKey("value");
+                if (i < 2)
+                    enc.writeInt(ivals[i]);
+                else
+                    enc.writeUInt(uvals[i - 2]);
+            });
+        }
+        t.commit();
+    }
+
+    // Check the data type of the returned values:
+    Retained<Query> query = store->compileQuery(json5( "{WHAT: ['.value']}"));
+    Retained<QueryEnumerator> e = query->createEnumerator();
+    REQUIRE(e->getRowCount() == 4);
+    int row = 0;
+    while (e->next()) {
+        auto type = e->columns()[0]->type();
+        CHECK(type == kNumber);
+        auto isInt = e->columns()[0]->isInteger();
+        CHECK(isInt);
+        auto isUnsigned = e->columns()[0]->isUnsigned();
+        if (row < 3)
+            CHECK(!isUnsigned);
+        else
+            CHECK(isUnsigned);
+        ++row;
     }
 }
 
--- a/LiteCore/Query/SQLiteFleeceFunctions.cc
+++ b/LiteCore/Query/SQLiteFleeceFunctions.cc
@@ -230,15 +230,23 @@ namespace litecore {
                     }
                     break;
                 }
-                case SQLITE_INTEGER:
-                    if (sqlite3_value_subtype(arg) == kFleeceIntBoolean) {
+                case SQLITE_INTEGER: {
+                    auto subtype = sqlite3_value_subtype(arg);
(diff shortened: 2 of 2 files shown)


LiteCore/tests/QueryTest.cc
```diff
@@ -336,6 +338,44 @@ N_WAY_TEST_CASE_METHOD(QueryTest, "Query boolean", "[Query]") {
...
+    Retained<QueryEnumerator> e = query->createEnumerator();
+    REQUIRE(e->getRowCount() == 4);
+    int row = 0;
+    while (e->next()) {
+        auto type = e->columns()[0]->type();
+        CHECK(type == kNumber);
+        auto isInt = e->columns()[0]->isInteger();
+        CHECK(isInt);
+        auto isUnsigned = e->columns()[0]->isUnsigned();
+        if (row < 3)
```
jianminzhao, line 371: Should this be "row < 2" ?

jianminzhao (approved):

LiteCore/tests/QueryTest.cc
Dushistov: No, this is special case < 2048:
```
 void Encoder::writeUInt(uint64_t i) {writeInt(i, (i < 2048),               true);} 
 void Encoder::writeInt(uint64_t i, bool isSmall, bool isUnsigned) {                                                                                                         
        if (isSmall) {                                                                                                                                                          
            new (placeItem()) Value(kShortIntTag, (i >> 8) & 0x0F, i & 0xFF);   
        } else
```
so during fleece encoding we lost all information about that it was unsigned.

jianminzhao: I see. I read the relevant code over the weekend and I think, first of all, your fix in fleece function is definitely right. Second, whether to encode in fleece as signed or unsigned is an implementation decision with the constraint of being correct representing the number. With 64-bit int, any number whose absolute value is less than or equal to numeric_limit<int64_t>::max can be represented either in signed or unsigned formats. SQLite stopped here; they only use signed int. LiteCore uses extra bit to allow greater numbers. That's why, even though you call enc.wirteUInt, we may still encode it in singed way because its scale can fit in our signed representation. The actual cutoff now is 12 bits signed (absolute value fits in 11 bits.)
Therefore, the meaningful checks would be:
* for negative number, its fleece representation must be signed. Or
      ivals[r] >= 0 || !col->isUnsiigned()
* for positive number, its fleece representation unsigned if it cannot fit into 63 bits. Or
     col->asInt() >= 0 || col->isUnsigned()
That is, for negative numbers, isUnsigned() == false regardless the magnitude. For positive numbers, even you call enc->writeUInt(), we may still represent it in signed format provided its magnitude allows faithful representation as numbers.
Therefore, I would suggest that test focus on whether Fleece correctly represent the number correctly, not to use the knowledge based on the above code of small integer.

PR #1455, opened 2022-04-22
CBL-3040 : do not apply the rule of result alias if the property is e…

jianminzhao:
…xplicitly led by the database alias.

### DIFF
--- a/LiteCore/Query/QueryParser.cc
+++ b/LiteCore/Query/QueryParser.cc
@@ -1548,9 +1548,9 @@ namespace litecore {
 
     // Writes a call to a Fleece SQL function, including the closing ")".
     void QueryParser::writePropertyGetter(slice fn, Path &&property, const Value *param) {
-        size_t property_size_in = property.size();
+        size_t propertySizeIn = property.size();
         auto &&iType = verifyDbAlias(property);
-        bool property_starts_with_explicit_alias = (property.size() + 1 == property_size_in);
+        bool propertyStartsWithExplicitAlias = (property.size() + 1 == propertySizeIn);
         const string &alias = iType->first;
         aliasType type = iType->second.type;
         string tablePrefix = alias.empty() ? "" : quotedIdentifierString(alias) + ".";
@@ -1564,7 +1564,7 @@ namespace litecore {
         // CBL-3040. We should not apply the following rule of result alias if the
         // property starts with a database collection alias explicitly. In this case,
         // the following name is the proerty name in the collection.
-        if (!property_starts_with_explicit_alias) {
+        if (!propertyStartsWithExplicitAlias) {
             // Check out the case the property starts with the result alias.
             auto resultAliasIter = _aliases.end();
             if (!property.empty()) {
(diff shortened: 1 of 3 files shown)


jianminzhao:
Author's notes: I decided not to dwell on the SQL standard that column aliases shall be only available in the ORDER BY clause.
First, albeit the standard, SqLite now only precludes the use of column aliases in the results. It allows them, for example, in the WHERE clause.
Second, the real problem of the bug is that we dropped db alias before sending to the function that processes the property. If a property name is led by the database alias, it should be interpreted as a property name in the db table aliased by the alias.
I take back the idea of using the context of resultOp, because if users use it, it's user error. SQLite will detect it.
This commit not only fix the problem as reported in the ticket, but also potential problem of using it wrongly with the WHERE clause. I have a test case for it.

snej (approved):

LiteCore/Query/QueryParser.cc
```diff
@@ -1548,7 +1548,9 @@ namespace litecore {
 
     // Writes a call to a Fleece SQL function, including the closing ")".
     void QueryParser::writePropertyGetter(slice fn, Path &&property, const Value *param) {
+        size_t property_size_in = property.size();
         auto &&iType = verifyDbAlias(property);
+        bool property_starts_with_explicit_alias = (property.size() + 1 == property_size_in);
```
borrrden, line 1553: This is a confusing line that I think warrants a comment.  I had to look through the implementation of verifyDbAlias to see why this would ever be true.

borrrden (approved):

LiteCore/Query/QueryParser.cc
```diff
@@ -1559,34 +1561,39 @@ namespace litecore {
...
-        // Check out the case the property starts with the result alias.
-        auto resultAliasIter = _aliases.end();
-        if (!property.empty()) {
-            resultAliasIter = _aliases.find(property[0].keyStr().asString());
-            if (resultAliasIter != _aliases.end() && resultAliasIter->second.type != kResultAlias) {
-                resultAliasIter = _aliases.end();
+        // CBL-3040. We should not apply the following rule of result alias if the
+        // property starts with a database collection alias explicitly. In this case,
+        // the following name is the proerty name in the collection.
+        if (!property_starts_with_explicit_alias) {
```
pasin, line 1567: I think I probably don't understand the problem. 

What will be the behavior of the following giving that the doc has `c1` and `label1` property?

`SELECT c1 as label1, label1 FROM _`

pasin: OK, I think I understand it now. The above SQL will not valid. To make it valid, it needs the datasource such as 

`SELECT c1 as label1, mydb.label1 FROM _ AS mydb`

pasin (approved):
Just small format things.

LiteCore/Query/QueryParser.cc
jianminzhao: Right

jianminzhao: Before this PR, the above will be parsed as if "SELECT c1 as label1, label1 FROM _" and rejected by Sqlite.

pasin: Very clear now. Thanks!

PR #1458, opened 2022-04-28
CBL-3050 : Account for a race between pendingDocs and terminate

jianminzhao:
This commit is ported from f457d9bc74af276516d868b61a48b81b5d717e5c based on the logic, because the source has been reorganized significantly.


borrrden (approved):
I feel like there was more to this, with some sort of guard value in the use function, but let's see how this does.

PR #1461, opened 2022-05-02
CBL-3056 : LiteCore crashes in pendingDocumentIDs

jianminzhao:
After we ported CBL-2407 from 2.8.7 to Lithium, we had to apply some patch up to pass unit tests in Jenkins. The patch-up turns out to be crutial not only for Lithium, but also for Hydrogen as discovered by Platforms' tests.

### DIFF
--- a/Replicator/DBAccess.cc
+++ b/Replicator/DBAccess.cc
@@ -58,7 +58,6 @@ namespace litecore { namespace repl {
     access_lock<C4Database*>& DBAccess::insertionDB() {
         if (!_insertionDB) {
             use([&](C4Database *db) {
-                AssertDBOpen(db);
                 if (!_insertionDB) {
                     C4Error error;
                     C4Database *idb = c4db_openAgain(db, &error);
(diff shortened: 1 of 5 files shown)


Replicator/DBAccess.cc
```diff
@@ -82,28 +81,30 @@ namespace litecore { namespace repl {
...
+        }
         _timer.stop();
-        use([&](C4Database *&db) {
+        use([this](C4Database *&db) {
             // Any use of the class after this will result in a crash that 
             // should be easily identifiable, so forgo asserting if the pointer 
             // is null in other areas.
             c4db_release(db);
             db = nullptr;
+            this->_sentry = &DBAccess::AssertDBOpen;
```
pasin, line 94: We don't do c4db_release(db) in the current master branch. Can you double check if it's correct to do the release here?

pasin (commented):
I have one comment. The rest looks good to me.

Replicator/DBAccess.cc
jianminzhao: In 2.8, DBAccess is access_lock<C4Database*>, whereas in 3.0, it is access_lock<Retained<C4Database>>. Therefore, we need to manually do c4db_release in 2.8

pasin (approved):

PR #1463, opened 2022-05-04
CBL-3073 : Database could be corrupted after being copied in linux pl…

jianminzhao:
…atform

Fixing the Linux version of copyfile, which uses sys/sendfile with which "a successful call to sendfile() may write fewer bytes than requested; the caller should be prepared to retry the call if there were unsent bytes."

### DIFF
--- a/LiteCore/Support/FilePath.cc
+++ b/LiteCore/Support/FilePath.cc
@@ -86,7 +86,8 @@ static int copyfile(const char* from, const char* to)
     ssize_t bytes = 0;
     int noProgressRetry = 2;
     while (bytes < expected) {
-        bytes = sendfile(write_fd, read_fd, &offset, expected - bytes);
+        expected -= bytes;
+        bytes = sendfile(write_fd, read_fd, &offset, expected);
         if (bytes < 0) {
             int e = errno;
             close(read_fd);
--- a/LiteCore/Support/FilePath.cc
+++ b/LiteCore/Support/FilePath.cc
@@ -54,6 +54,7 @@
 
 using namespace std;
 using namespace fleece;
+using namespace litecore;
 
 #ifdef __linux__
 static int copyfile(const char* from, const char* to)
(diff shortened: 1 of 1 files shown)


LiteCore/Support/FilePath.cc
```diff
@@ -80,13 +81,29 @@ static int copyfile(const char* from, const char* to)
...
-        close(read_fd);
-        close(write_fd);
-        errno = e;
-        return -1;
+
+    size_t expected = stat_buf.st_size;
+    ssize_t bytes = 0;
+    int noProgressRetry = 2;
+    while (bytes < expected) {
+        bytes = sendfile(write_fd, read_fd, &offset, expected - bytes);
```
borrrden, line 89: Shouldn’t this be += ?

jianminzhao: No, because I subtracted "exptected" by "bytes"

jianminzhao: Oh, actually, there is a problem as explain it.

pasin (approved):

LiteCore/Support/FilePath.cc
```diff
@@ -80,13 +81,30 @@ static int copyfile(const char* from, const char* to)
...
-        close(read_fd);
-        close(write_fd);
-        errno = e;
-        return -1;
+
+    size_t expected = stat_buf.st_size;
+    ssize_t bytes = 0;
+    int noProgressRetry = 2;
+    while (bytes < expected) {
+        expected -= bytes;
```
snej, line 89: There's no reason to do this before the first copy; why not move it to the end of the loop, i.e. after `noProgressRetry = 2;`?

LiteCore/Support/FilePath.cc
```diff
@@ -80,13 +81,30 @@ static int copyfile(const char* from, const char* to)
...
+        bytes = sendfile(write_fd, read_fd, &offset, expected);
+        if (bytes < 0) {
+            int e = errno;
+            close(read_fd);
+            close(write_fd);
+            errno = e;
+            return -1;
+        } else if (bytes == 0) {
```
snej, line 97: This might be overdesign. If sendfile can't copy any bytes I would consider it an error and give up, even if it didn't return an error.

LiteCore/Support/FilePath.cc
jianminzhao: This is because I wanted to avoid subtraction of unsigned without comparison, like 
while (expected > 0) {
 bytes = ...
 expected -= bytes;
}
although in theory we should have Assert(bytes <= expected).

LiteCore/Support/FilePath.cc
jianminzhao: I agreed. I changed it in porting it to lithium,
        } else if (bytes == 0) {
            // zero bytes are read. Do we want to try again? Well, let's consider it as an error
            Warn("sys/sendfile makes no progress copying %s to %s and we bail out as failure.", from, to);
       ...
I will add you to the reviewer in the lithium PR.

PR #1465, opened 2022-05-05
CBL-3074 : Database could be corrupted after being copied in linux pl…

jianminzhao:
…… (#1463)

Fixing the Linux version of copyfile, which uses sys/sendfile with which "a successful call to sendfile() may write fewer bytes than requested; the caller should be prepared to retry the call if there were unsent bytes."
Cheryy-picked from 19cbfc0cfa6bedc9f41184d2b992af3ddb1b3d35

### DIFF
--- a/LiteCore/Support/FilePath.cc
+++ b/LiteCore/Support/FilePath.cc
@@ -48,6 +48,7 @@
 
 using namespace std;
 using namespace fleece;
+using namespace litecore;
 
 #ifdef __linux__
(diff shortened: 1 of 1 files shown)


borrrden (approved):

pasin (approved):

jianminzhao:
@snej suggested to treat 0 bytes as error, and I agree.

borrrden (approved):

PR #1468, opened 2022-05-10
Add GHA check to scan for black duck dependency changes

borrrden:
To help move to a system where signature scanning is not used, expliclty listed dependencies are needed and this script will run on a PR to ensure that if a dependency was changed that its corresponding manifest entry was updated, and vice versa.


borrrden:
The key point here is the new github action "check blackduck manifest".  From now on when you update a downstream dep, you also need to change the jenkins/couchbase-lite-core-blackduck-manifest.yaml accordingly to the new version.  The GHA will check that you did so (note the potential for false positives if a bugfix is made, but let's evaluate how noisy that gets)

borrrden:
(please squash this when merging, only the first message is important)

jianminzhao (approved):

PR #1471, opened 2022-05-10
CBL-2773 : use WeakHolder for websocket::Delegate.

jianminzhao:
Following is a chain of strong ownership:
  Replicator -> Connection -> BLIPIO -> WebSocket
Currently, the reverse relationship from WebSocket to BLIPIO is established by plain pointer, for good reasons. However, this creates a problem if the websocket outlives BLIPIO for a fleeting moment of time. In this commit, we use WeakHolder for the delegate.


snej:
It seems like this change is equivalent to making `WebSocket::_delegate` an access_lock. I don't think we need a new ref-counted class for this with a new API to call through to the delegate...?

jianminzhao:
It's true we access_lock can be used here, but the emphasis is different. With access_lock, do we do this,
- access_lock<Delegate*> weakDele;
- in owner's destructor, weakDele.useLocked([](Delegate*& d) { d = nullptr; });
- where used, we have to check each time, weakDele.useLocked([](Delegate* d) {
        if (d != 0) { // do thing}
   }
I think WeakHolder bake the above logic into one little semantics, relieving the care & error from the users. There is even no way to leak the plain pointer and only the owner may invalidate the pointer.
What do you think?

borrrden:
What about putting this functionality directly into access_lock?

jianminzhao:
The following will work, if we want to encapsulate the common code for the client,
-----------------------------------------------
template<typename T>
class WeakHolder : private access_lock<T*> {
public:
    WeakHolder(T* pointer)
        : access_lock<T*>(pointer)
    {
        DebugAssert(pointer != nullptr);
    }

    // Only the original owner may rescind the pointer.
    // After rescind(), the held pointer becomes nullptr and invoke() returns false.
    void rescind(T* owner) {
        useLocked([owner](T*& pointer) {
            if (pointer == owner) {
                pointer = nullptr;
            }
        });
    }

    template<typename Callable, typename ... Args>
    bool invoke(Callable fun, Args&& ... args) {
        useLocked([&](T* pointer) {
            if (pointer != nullptr) {
                fun(pointer, std::forward<Args>(args)...);
                return true;
            } else {
                return false;
            }
        });
    }
};
--------------------------------------------
but it's not any shorter than current WeakHolder.

borrrden:
I was in favor of simply adding these method to access lock rather than making a subclass.  That would shorten things quite a bit.

jianminzhao:
Most of our current uses of access_lcck do not check null pointer; it's not designed to guard against "weakness.." Besides, the pattern that this WeakHolder intends to address is quite popular. With a dedicated class, I hope to make the code more fluent.
I don't have to make WeakHolder ref-counted. I can use shared_ptr to hold it, but I thought RefCounted is much more economic in this case than the general-purpose shared_ptr.

borrrden (approved):
I don't have a problem with it as-is

snej (approved):

pasin (approved):

borrrden (approved):

PR #1475, opened 2022-05-26
CBL-3148: Add c4listener_shareCollection API

borrrden:
:new: API added c4listener_shareCollection (and c++ equivalent)
:new: API added c4listener_unshareCollection (and c++ equivalent)

Note that this API is additive to c4listener_shareDB, and not a replacement.  See the interface document for more details.

### DIFF
--- a/REST/Listener.cc
+++ b/REST/Listener.cc
@@ -104,7 +104,7 @@ namespace litecore { namespace REST {
         return false;
     }
 
-    bool Listener::registerCollection(const string& name, C4CollectionSpec collection) {
+    bool Listener::registerCollection(const string& name, CollectionSpec collection) {
         lock_guard<mutex> lock(_mutex);
         auto i = _databases.find(name);
         if (i == _databases.end())
@@ -112,7 +112,7 @@ namespace litecore { namespace REST {
 
         auto j = _allowedCollections.find(name);
         if(j == _allowedCollections.end()) {
-            vector<C4CollectionSpec> collections({collection});
+            vector<CollectionSpec> collections({collection});
             _allowedCollections.emplace(name, collections);
         } else {
             j->second.push_back(collection);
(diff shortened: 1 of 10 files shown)


borrrden:
I am going to add tests as another commit, to keep the review focused on the implementation.

REST/Listener.cc
```diff
@@ -98,6 +104,40 @@ namespace litecore { namespace REST {
...
+            _allowedCollections.emplace(name, collections);
+        } else {
+            j->second.push_back(collection);
+        }
+        
+        return true;
+    }
+
+
+    bool Listener::unregisterCollection(const string& name, C4CollectionSpec collection) {
```
pasin, line 125: Use CollectionSpec for consistent with other C++ code? It also implements == operator.

REST/Listener.hh
```diff
@@ -54,6 +55,13 @@ namespace litecore { namespace REST {
 
         bool unregisterDatabase(C4Database *db);
 
+         /** Adds a collection to be used by Sync Listener on a given database shared via name.  
+             Only collections that are registered will be replicated.  
+             If none are registered, the default collection will be replicated */
+        bool registerCollection(const std::string& name, C4CollectionSpec collection);
```
pasin, line 61: Use CollectionSpec for consistency with other C++ API.

REST/RESTListener+Handlers.cc
```diff
@@ -76,10 +76,10 @@ namespace litecore { namespace REST {
 #pragma mark - DATABASE HANDLERS:
 
     
-    void RESTListener::handleGetDatabase(RequestResponse &rq, C4Database *db) {
-        auto docCount = db->getDocumentCount();  // TODO: Collection-aware
-        auto lastSequence = db->getLastSequence();
-        C4UUID uuid = db->getPublicUUID();
+    void RESTListener::handleGetDatabase(RequestResponse &rq, C4Collection *coll) {
```
pasin, line 79: Does it require to have scope and collection name in the result?

borrrden: I don't think so, since the user is requesting a specific collection via URL

pasin (approved):

PR #1478, opened 2022-05-27
CBL-3799: More REST listener collection work

snej:
- GET /db includes a `scopes` property containing all the scopes. Every scope contains its collections, every collection contains its `doc_count` and `update_seq`, and all of them were coming from St. Ives.
- Implemented PUT /db.scope.coll and DELETE /db.scope.coll .
- C4ListenerConfig controls whether collections can be created/deleted.
- More error handling when parsing the scope/collection from a request.
- Unit tests

Also a separate commit that fixes a C4Database crash that was exposed by the other commit.


C/include/c4ListenerTypes.h
```diff
@@ -79,6 +79,8 @@ typedef struct C4ListenerConfig {
     C4String directory;                     ///< Directory where newly-PUT databases will be created
     bool allowCreateDBs;                    ///< If true, "PUT /db" is allowed
     bool allowDeleteDBs;                    ///< If true, "DELETE /db" is allowed
+    bool allowCreateCollections;            ///< If true, "PUT /db.scope.coll" is allowed
```
borrrden, line 82: There isn't any precedence for this (the Sync Gateway REST API does not allow this) so I'm wondering what the goal of this is.

snej: Well, part of what a collections API does is let you create and delete them. It's a straightforward extension of databases. IIRC there are some SG performance tests that use cblite+REST to sync with SG, and if those tests want to use collections they might well need this.

borrrden: This makes me feel quite awkward because there is a growing set of complex functionality that gets tested on every build, and will block progress if it is not correct, but is not a supported part of the Couchbase Lite binary.  If possible I think it should be removed and built elsewhere if the only thing it is used for is to shoehorn it into a testing framework.

snej: The SG REST spec doesn't talk about whether the `/db/` endpoint supports collections. I added [a comment asking about that](https://docs.google.com/document/d/1Jlt6AJeu6lRw7QralS1deizC4sN0Ng1n4ISyt41i1hg/edit?disco=AAAAaR9NckY). I think it has to, otherwise there's no way to use GET to find the doc-count or current sequence of a collection.

borrrden: The GET endpoint supports it, but in SG since you don't directly deal with data the way you do with CBL the delete endpoint is limited to database (or rather, the entry in the SG config).  There is no way to delete or add a collection via SG REST (I asked this to confirm)

snej: You're right, SG was never able to create or delete buckets. But CBL can of course manage databases, and the REST API supports that, so I think we should allow the REST API to manage collections too.

snej:
The Jenkins tests failed because I forgot to push the LiteCore-EE repo changes. Retrying...

borrrden (approved):
It's not going to cause issues, but this is only going to make the distributed library bigger for no reason, I think.  This should be conditionally compiled only for the cblite tool in my opinion.

PR #1479, opened 2022-05-31
CBL-3149: Make all observers collection aware

borrrden:
⚠️ Breaking change adds collection spec to callbacks registered via c4docobs_create (and c4docobs_createWithCollection)
⚠️ c4dbobs_getChanges signature removes external, and moves it to a new return type that contains the original return type as numChanges, along with external and collection

Added collection based revision creation methods for C4Tests, adjusted signature of other tests based on the above changes

### DIFF
--- a/C/Cpp_include/c4Collection.hh
+++ b/C/Cpp_include/c4Collection.hh
@@ -104,6 +104,7 @@ struct C4Collection : public fleece::RefCounted, C4Base, fleece::InstanceCounted
 
     using CollectionObserverCallback = std::function<void(C4CollectionObserver*)>;
     using DocumentObserverCallback = std::function<void(C4DocumentObserver*,
+                                                        C4Collection*,
                                                         slice docID,
                                                         C4SequenceNumber)>;
 
--- a/C/Cpp_include/c4Observer.hh
+++ b/C/Cpp_include/c4Observer.hh
@@ -55,9 +55,8 @@ struct C4CollectionObserver : public fleece::InstanceCounted, C4Base {
         \note The usual way to use this method is to allocate a reasonably sized buffer, maybe
                 100 changes, and keep calling getChanges passing in the entire buffer, until
                 it returns 0 to indicate no more changes. */
-    virtual uint32_t getChanges(Change outChanges[C4NONNULL],
-                                uint32_t maxChanges,
-                                bool *outExternal) =0;
+    virtual C4CollectionObservation getChanges(Change outChanges[C4NONNULL],
+                                               uint32_t maxChanges) =0;
 };
 
 
--- a/C/include/c4Observer.h
+++ b/C/include/c4Observer.h
@@ -80,13 +80,11 @@ C4API_BEGIN_DECLS
                             written.
         @param maxChanges  The maximum number of changes to return, i.e. the size of the caller's
                             outChanges buffer.
-        @param outExternal  Will be set to true if the changes were made by a different C4Database.
-        @return  The number of changes written to `outChanges`. If this is less than `maxChanges`,
-                            the end has been reached and the observer is reset. */
-    CBL_CORE_API uint32_t c4dbobs_getChanges(C4CollectionObserver *observer,
-                                C4CollectionChange outChanges[C4NONNULL],
-                                uint32_t maxChanges,
-                                bool *outExternal) C4API;
+        @return  Common information about the changes contained in outChanges (number of changes, 
+                 external vs non-external, and the relevant collection) */
+    CBL_CORE_API C4CollectionObservation c4dbobs_getChanges(C4CollectionObserver *observer,
+                                         C4CollectionChange outChanges[C4NONNULL],
+                                         uint32_t maxChanges) C4API;
 
     /** Releases the memory used by the `C4CollectionChange` structs (to hold the docID and revID
         strings.) This must be called after \ref c4dbobs_getChanges().
@@ -108,6 +106,7 @@ C4API_BEGIN_DECLS
         @param sequence  The sequence number of the change.
         @param context  user-defined parameter given when registering the callback. */
     typedef void (*C4DocumentObserverCallback)(C4DocumentObserver* observer,
+                                               C4Collection* collection,
                                                C4String docID,
                                                C4SequenceNumber sequence,
                                                void * C4NULLABLE context);
--- a/LiteCore/Database/SequenceTracker.hh
+++ b/LiteCore/Database/SequenceTracker.hh
@@ -20,7 +20,6 @@
 
 namespace litecore {
     class CollectionChangeNotifier;
-    class DatabaseImpl;
     class DocChangeNotifier;
 
     
--- a/C/Cpp_include/c4Database.hh
+++ b/C/Cpp_include/c4Database.hh
@@ -98,6 +98,7 @@ public:
     /// You can pass a `CollectionSpec` parameter as simply a collection name slice, implying the
     /// default scope; or as `{collectionname, scopename}`.
     struct CollectionSpec : public C4CollectionSpec {
+        CollectionSpec()                             :C4CollectionSpec{kC4DefaultCollectionName, kC4DefaultScopeID} { }
         CollectionSpec(const C4CollectionSpec &spec) :C4CollectionSpec(spec) { }
         CollectionSpec(FLString name, FLString scope):C4CollectionSpec{name, scope} { }
         CollectionSpec(FLString name)                :C4CollectionSpec{name, kC4DefaultScopeID} { }
--- a/C/c4CAPI.cc
+++ b/C/c4CAPI.cc
@@ -1259,20 +1259,20 @@ C4DatabaseObserver* c4dbobs_createOnCollection(C4Collection* coll,
 }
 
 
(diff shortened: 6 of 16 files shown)


C/Cpp_include/c4Observer.hh
```diff
@@ -42,6 +44,7 @@ struct C4CollectionObserver : public fleece::InstanceCounted, C4Base {
 
     /// Metadata of a change recorded by C4CollectionObserver. (Equivalent to C4CollectionChange.)
     struct Change {
+        CollectionSpec collection;      ///< The collection the document belongs to
```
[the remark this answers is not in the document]
borrrden: It's a debatable point, but in the current API we are, just in case they want to assign a common implementation that varies slightly based on the collection.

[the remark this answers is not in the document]
jianminzhao: In this case, the client code could supply the "context" in 
c4docobs_create(C4Database *db,
                                    C4Slice docID,
                                    C4DocumentObserverCallback callback,
                                    void *context) noexcept
which will be passed back to the client in callback.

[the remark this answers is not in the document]
borrrden: Usually the context is already being used for storing some kind of other reference (Mapping to a Java or .NET object is common).  This does not translate all the way up to the user.

C/include/c4Observer.h
```diff
@@ -107,18 +107,21 @@ C4API_BEGIN_DECLS
         @param docID  The ID of the document that changed.
         @param sequence  The sequence number of the change.
         @param context  user-defined parameter given when registering the callback. */
-    typedef void (*C4DocumentObserverCallback)(C4DocumentObserver* observer,
-                                               C4String docID,
-                                               C4SequenceNumber sequence,
-                                               void * C4NULLABLE context);
+    typedef void (*C4CollectionDocObserverCallback)(C4DocumentObserver* observer,
+                                                   C4CollectionSpec collection,
```
jianminzhao, line 111: If we expose the collection of C4DocumentObserver, then we don't need to provide collection as argument in this callback. I this is the identity: observer->collection == collection. That is, there is a redundancy here.

LiteCore/Database/SequenceTracker.cc
```diff
@@ -83,18 +98,18 @@ namespace litecore {
...
-        :docID(d), revID(r), sequence(s), bodySize(bs), flags(flags), idle(false), external(false)
+        Entry(CollectionSpec c, const alloc_slice &d, alloc_slice r, sequence_t s, uint32_t bs, RevisionFlags flags)
+        :collection(c), docID(d), revID(r), sequence(s), bodySize(bs), flags(flags), idle(false), external(false)
         {
             DebugAssert(docID != nullslice);
         }
 
         explicit Entry(const alloc_slice &d)
-        :Entry(d, nullslice, 0_seq, 0, {})
+        :Entry(kC4DefaultCollectionSpec, d, nullslice, 0_seq, 0, {})
```
jianminzhao, line 108: SequenceTracker is owned by CollectionImpl. I feel we don't need extra collection info in the Entry, if we store the collectionSpec in the SequenceTracker.

C/Cpp_include/c4Observer.hh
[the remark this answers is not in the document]
jianminzhao: Users probably may store the collection info in the context. Each collection commands a unique context object. This can save a lot of complexity in the LiteCore.

LiteCore/Database/SequenceTracker.cc
borrrden: Is there a path up to the callback in that case?  Entries are used to generate document ended callback events.

C/include/c4Observer.h
jianminzhao: Why not? A getter of CollectonSpec, which is a pure value, is simple and safe, removing the redundancy.

LiteCore/Database/SequenceTracker.cc
```diff
@@ -183,7 +198,7 @@ namespace litecore {
                 nextEntry = next(entry);
                 if (!entry->isPlaceholder()) {
                     // moves entry!
-                    _documentChanged(entry->docID, entry->revID,
+                    _documentChanged(entry->collection, entry->docID, entry->revID,
```
jianminzhao, line 201: Here, for example, if we store the collection in SequenceTracker itself, "entry->collection" can be replaced by this->collection. Logically, collection is welded into the Tracker, const/immutable.

LiteCore/Database/SequenceTracker.cc
```diff
@@ -475,9 +501,9 @@ namespace litecore {
 #pragma mark - DOC CHANGE NOTIFIER:
 
 
-    DocChangeNotifier::DocChangeNotifier(SequenceTracker &t, slice docID, Callback cb)
+    DocChangeNotifier::DocChangeNotifier(SequenceTracker &t, CollectionSpec collection, slice docID, Callback cb)
```
jianminzhao, line 504: same here, the invariant is t->collection == collection. We don't need the collection argument if it is readily fetched from SequenceTracker.

C/include/c4Observer.h
borrrden: It will require the addition of c4docobs_getCollectionSpec.  A simple getter is not usable for the C API.

C/Cpp_include/c4Observer.hh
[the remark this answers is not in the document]
snej: Why a CollectionSpec instead of a C4Collection?

snej (changes requested):

C/Cpp_include/c4Observer.hh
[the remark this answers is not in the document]
borrrden: I think the reasoning was that it is easier to get a collection object on the platform side given its information rather than a pointer which would need to be linked later to a platform collection object.

C/Cpp_include/c4Collection.hh
```diff
@@ -104,6 +104,7 @@ struct C4Collection : public fleece::RefCounted, C4Base, fleece::InstanceCounted
 
     using CollectionObserverCallback = std::function<void(C4CollectionObserver*)>;
     using DocumentObserverCallback = std::function<void(C4DocumentObserver*,
+                                                        C4CollectionSpec,
```
snej, line 107: It seems better to pass a C4Collection pointer than a C4CollectionSpec. It's smaller, and the platform code can probably map it to its own thing more efficiently than having to look up two strings.

snej (changes requested):

C/Cpp_include/c4Collection.hh
borrrden: @pasin Let's see what the platform owners think about which is easier

borrrden:
I made quite a few changes and the PR became much smaller and simpler.

jianminzhao (approved):

snej (changes requested):

C/Cpp_include/c4Observer.hh
```diff
@@ -13,6 +13,7 @@
 #pragma once
 #include "c4Base.hh"
 #include "c4Collection.hh"
+#include "c4Database.hh"
```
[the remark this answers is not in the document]
borrrden: Looks like not anymore, it would have previously been to get CollectionSpec

LiteCore/Database/SequenceTracker.hh
```diff
@@ -146,9 +146,9 @@ namespace litecore {
     /** Tracks changes to a single document and calls a client callback. */
     class DocChangeNotifier {
     public:
-        using Callback = std::function<void(DocChangeNotifier&, slice docID, sequence_t)>;
+        using Callback = std::function<void(DocChangeNotifier&, C4Collection*, slice docID, sequence_t)>;
 
-        DocChangeNotifier(SequenceTracker &t, slice docID, Callback cb);
+        DocChangeNotifier(SequenceTracker &t, CollectionImpl*, slice docID, Callback cb);
```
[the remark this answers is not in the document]
borrrden: Oh, excellent suggestion!  That's done now.

borrrden:
Gah I always forget that CppTests also does this, will push up the changes to revert them shortly

LiteCore/Database/SequenceTracker.cc
```diff
@@ -59,6 +59,7 @@ Then document A changes, but no notifications are sent:
 
 namespace litecore {
     using namespace std;
+    using namespace fleece;
```
snej, line 62: This line must be unnecessary, since the file compiled without it...

LiteCore/Database/SequenceTracker.hh
```diff
@@ -20,7 +20,7 @@
 
 namespace litecore {
     class CollectionChangeNotifier;
-    class DatabaseImpl;
+    class CollectionImpl;
```
snej, line 23: This change must be unnecessary, since the file compiled without it. And in fact `class DatabaseImpl` must be unnecessary too if you could change it without breaking anything, so this line can be deleted.

snej (approved):
Looks good! The changes to SequenceTracker are unnecessary, but I'll leave it up to you if you want to take the time to remove them...

PR #1480, opened 2022-06-01
 CBL-3196 CBL-3198: Track collection removal across database instances 

borrrden:
Also, as part of improving the collection test CBL-3194 to rewrite the collection tests in the C API instead and move them to C4Tests

### DIFF
--- a/C/c4Database.cc
+++ b/C/c4Database.cc
@@ -178,6 +178,14 @@ static C4DatabaseConfig newToOldConfig(const C4DatabaseConfig2 &config2) {
     SQLiteDataFile::shutdown();
 }
 
+C4Collection* C4Database::getDefaultCollectionSafe() const {
+    if(!_defaultCollection) {        
+        error::_throw(error::NotOpen);
+    }
+
+    return _defaultCollection;
+}
+
 
 C4Collection* C4Database::getDefaultCollection() const {
     // Make a distinction: If the DB is open and the default collection is deleted
@@ -217,6 +225,9 @@ Retained<C4Query> C4Database::newQuery(C4QueryLanguage language, slice expr, int
 
 
 #pragma mark - INDEXES:
+
+
+
 
 
 void C4Database::createIndex(slice indexName,
--- a/C/c4Database.cc
+++ b/C/c4Database.cc
@@ -178,14 +178,6 @@ static C4DatabaseConfig newToOldConfig(const C4DatabaseConfig2 &config2) {
     SQLiteDataFile::shutdown();
 }
 
-C4Collection* C4Database::getDefaultCollectionSafe() const {
-    if(!_defaultCollection) {        
-        error::_throw(error::NotOpen);
-    }
-
-    return _defaultCollection;
-}
-
 
 C4Collection* C4Database::getDefaultCollection() const {
     // Make a distinction: If the DB is open and the default collection is deleted
@@ -225,9 +217,6 @@ Retained<C4Query> C4Database::newQuery(C4QueryLanguage language, slice expr, int
 
 
 #pragma mark - INDEXES:
-
-
-
 
 
 void C4Database::createIndex(slice indexName,
--- a/C/tests/c4CollectionTest.cc
+++ b/C/tests/c4CollectionTest.cc
@@ -30,24 +30,38 @@ public:
 
     C4CollectionTest(int testOption) :C4Test(testOption) { }
 
-    string getCollectionNames(slice inScope) {
+    string getNames(FLMutableArray source) {
         stringstream result;
         delimiter delim(", ");
-        db->forEachCollection(inScope, [&](C4CollectionSpec spec) {
-            CHECK(spec.scope == inScope);
-            result << delim << string_view(slice(spec.name));
-        });
+        FLArrayIterator i;
+        FLArrayIterator_Begin(source, &i);
+        if(FLArrayIterator_GetCount(&i) == 0) {
+            return "";
+        }
+
+        do {
+            string next = (string)(FLValue_AsString(FLArrayIterator_GetValue(&i)));
+            result << delim << next;
+        } while(FLArrayIterator_Next(&i));
+
         return result.str();
+    }
+
+    string getCollectionNames(slice inScope) {
+        auto names = c4db_collectionNames(db, inScope, ERROR_INFO());
+        REQUIRE(names);
+        auto result = getNames(names);
+        FLMutableArray_Release(names);
+        return result;
     }
 
 
     string getScopeNames() {
-        stringstream result;
-        delimiter delim(", ");
-        db->forEachScope([&](slice name) {
-            result << delim << string_view(name);
-        });
-        return result.str();
+        auto names = c4db_scopeNames(db, ERROR_INFO());
+        REQUIRE(names);
+        auto result = getNames(names);
+        FLMutableArray_Release(names);
+        return result;
     }
 
 
@@ -63,10 +77,9 @@ public:
             rq.historyCount = 1;
             rq.body = kFleeceBody;
             rq.save = true;
-            auto doc = coll->putDocument(rq, nullptr, ERROR_INFO());
+            auto doc = c4coll_putDoc(coll, &rq, nullptr, ERROR_INFO());
             REQUIRE(doc);
-            CHECK(doc->collection() == coll);
-            CHECK(doc->database() == db);
+            c4doc_release(doc);
         }
     }
 };
@@ -79,22 +92,22 @@ static constexpr C4CollectionSpec Guitars = { GuitarsName, kC4DefaultScopeID };
 N_WAY_TEST_CASE_METHOD(C4CollectionTest, "Default Collection", "[Database][Collection][C]") {
     CHECK(getScopeNames() == "_default");
     CHECK(getCollectionNames(kC4DefaultScopeID) == "_default");
-    CHECK(db->hasCollection({kC4DefaultCollectionName, kC4DefaultScopeID}));
+    CHECK(c4db_hasCollection(db, kC4DefaultCollectionSpec));
+    CHECK(c4db_hasScope(db, kC4DefaultScopeID));
 
-    C4Collection* dflt = db->getDefaultCollection();
+    C4Collection* dflt = c4db_getDefaultCollection(db);
     REQUIRE(dflt);
-    CHECK(dflt == db->getDefaultCollection());              // Must be idempotent!
-    CHECK(dflt == db->getCollection({kC4DefaultCollectionName, kC4DefaultScopeID}));
-    CHECK(dflt == db->createCollection({kC4DefaultCollectionName, kC4DefaultScopeID}));
+    CHECK(dflt == c4db_getDefaultCollection(db));              // Must be idempotent!
+    CHECK(dflt == c4db_getCollection(db, kC4DefaultCollectionSpec, ERROR_INFO()));
+    CHECK(dflt == c4db_createCollection(db, kC4DefaultCollectionSpec, ERROR_INFO()));
+    
+    auto spec = c4coll_getSpec(dflt);
+    CHECK(spec.name == kC4DefaultCollectionName);
+    CHECK(spec.scope == kC4DefaultScopeID);
 
-    CHECK(dflt->getName() == kC4DefaultCollectionName);
-    CHECK(dflt->getScope() == kC4DefaultCollectionName);
-    CHECK(dflt->getSpec().name == kC4DefaultCollectionName);
-    CHECK(dflt->getSpec().scope == kC4DefaultCollectionName);
-
-    CHECK(dflt->getDatabase() == db);
-    CHECK(dflt->getDocumentCount() == 0);
-    CHECK(dflt->getLastSequence() == 0_seq);
+    CHECK(c4coll_getDatabase(dflt) == db);
+    CHECK(c4coll_getDocumentCount(dflt) == 0);
+    CHECK(c4coll_getLastSequence(dflt) == 0_seq);
     // The existing c4Database tests exercise the C4Collection API for the default collection,
     // since they call the c4Database C functions which are wrappers that indirect through
     // db->getDefaultCollection().
@@ -102,14 +115,18 @@ N_WAY_TEST_CASE_METHOD(C4CollectionTest, "Default Collection", "[Database][Colle
     CHECK(getCollectionNames(kC4DefaultScopeID) == "_default");
 
     // It is, surprisingly, legal to delete the default collection:
-    db->deleteCollection(kC4DefaultCollectionName);
-    CHECK(db->getDefaultCollection() == nullptr);
-    CHECK(db->getCollection(kC4DefaultCollectionName) == nullptr);
+    REQUIRE(c4db_deleteCollection(db, kC4DefaultCollectionSpec, ERROR_INFO()));
+    CHECK(c4db_getDefaultCollection(db) == nullptr);
+    CHECK(c4db_getCollection(db, kC4DefaultCollectionSpec, ERROR_INFO()) == nullptr);
     CHECK(getCollectionNames(kC4DefaultScopeID) == "");
+    
     // But you can't recreate it:
-    C4ExpectException(LiteCoreDomain, kC4ErrorInvalidParameter, [&]{
-        db->createCollection(kC4DefaultCollectionName);
-    });
+    c4log_warnOnErrors(false);
+    C4Error err;
+    CHECK(!c4db_createCollection(db, kC4DefaultCollectionSpec, &err));
+    CHECK(err.domain == LiteCoreDomain);
+    CHECK(err.code == kC4ErrorInvalidParameter);
+    c4log_warnOnErrors(true);
 
     // However, the default scope still exists
     CHECK(c4db_hasScope(db, kC4DefaultScopeID));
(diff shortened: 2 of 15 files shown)


LiteCore/Database/DatabaseImpl.cc
```diff
@@ -762,6 +768,17 @@ namespace litecore {
     }
 
 
+    void DatabaseImpl::collectionRemoved(slice scope, slice name) {
+        LOCK(_collectionsMutex);
+        auto c = _collections.find({name, scope});
+        if(c != _collections.end()) {
+            CollectionImpl* coll = (CollectionImpl *)c->second.get();
```
snej, line 775: Use `asImpl(Coll)` to access it as a CollectionImpl.

LiteCore/Database/DatabaseImpl.cc
```diff
@@ -762,6 +768,17 @@ namespace litecore {
...
+            coll->close();
```
snej, line 776: C4CollectionImpl is not thread-safe, and calling `close()` from another thread can cause a number of Bad Things, if the regular thread is using that state:
- `_documentFactory` could be freed while another thread is in the middle of calling it to create a document.
- `_sequenceTracker` could be freed while another thread is calling it while trying to start a transaction.
- `_housekeeper` could be released and freed while another thread is updating doc expiration or expiring docs.

You could add a mutex to CollectionImpl to handle this, but you'd have to add a lot of lock calls.

This is why I was preferring not to do this kind of notification -- instead either just handling the no-such-table errors from SQLite, or preflighting.

snej (changes requested):

borrrden:
On the point of C4Tests I would tend to agree.  The reason for moving these tests is because of their location and wording.  They are in a C4Tests named file inside the C4Tests folder, but were being compiled into CppTests.  I can work on making C4Tests do less, but still I'd need to hear some ideas about how to ensure the integrity of the C API as well (using it without accidentally using something that is not exported, for example)

LiteCore/Database/DatabaseImpl.cc
```diff
@@ -625,12 +629,25 @@ namespace litecore {
...
+                if(!canCreate) {
+                    if(!_dataFile->keyStoreExists(keyStoreName)) {
+                        // This is an invalid collection, and it has not been recreated elsewhere.
+                        // Furthermore, canCreate is false so don't return a collection object
+                        return nullptr;
+                    } 
+                }
+
+                // After some more validation, we can reset valid to true later
+                collectionPtr = asInternal(i->second.get());
```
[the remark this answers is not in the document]
borrrden: I would like to do it this way but the issue is that removing from the collection object has the strong potential to immediately free all the memory inside of it, since we don't expect the caller to retain the result of a collection.  That means that the memory inside becomes invalid and turns into a bomb.  If we keep the object alive for at least as long as the database, then that issue is avoided and we only need to make sure that sensible errors are returned.  At least that is what I am told.  @snej Could you elaborate a little more on the race you found here?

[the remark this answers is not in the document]
jianminzhao: But there is an asymmetry: we still remove the collection object from _collections of the db instance (this db instance) from which the collection is deleted, c.f. DatabaseImpl::deleteCollection, we only keep it in other db instance, c.f. DatabaseImpl::collectionRemoved. If this is a bump, it could blow in this db instance. Right?

LiteCore/Database/DatabaseImpl.cc
```diff
@@ -642,30 +659,46 @@ namespace litecore {
...
+            _collections.insert({CollectionSpec(collection->getSpec()),
+                                 move(collection)});
+
+            if (isInTransaction())
+                collectionPtr->transactionBegan();
+        }
+
+        
+
+        collectionPtr->setValid(true);
```
jianminzhao, line 678: As I noted, we only  setValid(false) for other db instance. We deleted underlying sqlite table. Now, we reuse the collection object on logically completely different collection sharing the collection name.

LiteCore/Storage/SQLiteKeyStore.cc
```diff
@@ -46,6 +46,7 @@ namespace litecore {
 
     void SQLiteDataFile::deleteKeyStore(const std::string &name) {
         exec("DROP TABLE IF EXISTS \"kv_" + name + "\"");
+        DataFile::deleteKeyStore(name);
```
jianminzhao, line 49: Here "name" is actually the name of the BothKeyStore. It is equal to BothKeyStore::liveKeyStore. So, it drops the table associated with the liveKeyStore. deadKeyStore is not dropped.

LiteCore/Database/DatabaseImpl.cc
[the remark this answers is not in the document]
borrrden: Yes, we do this because if you are using the object on one thread like you are supposed to, you will not run into any races.  The race comes because sending this signal to other database instances is a violation of the threading model.

LiteCore/Database/DatabaseImpl.cc
borrrden: I don't see the issue with that.  If you recreate the table, then everything is ok again.  We reset to valid, and the table is there.

LiteCore/Storage/SQLiteKeyStore.cc
borrrden: That seems like an existing issue that was not introduced here.

borrrden:
FYI we could probably solve all this by having database's being dumb to the underlying collection change and simply keep invalid collections around that will give errors when they are attempted to be used.

LiteCore/Database/DatabaseImpl.cc
jianminzhao: Perhaps we have different mental model regarding delete and recreate the collection. In my model, deleted and recreated collections have nothing to do with each other except they share the same name. This is something confuses me.

jianminzhao: Perhaps we have different mental model regarding delete and recreate the collection. In my model, deleted and recreated collections have nothing to do with each other except they share the same name. This is something confuses me.

LiteCore/Storage/SQLiteKeyStore.cc
jianminzhao: That's correct.

jianminzhao: That's correct.

jianminzhao:
I totally agree.

jianminzhao:
I totally agree.

LiteCore/Database/DatabaseImpl.cc
borrrden: A collection is nothing more than a logical wrapper around a table, like a pointer.  So if you restore the underlying data then in my mind you've restored the contents of the wrapper, then there is no significant difference between that and creating a brand new object just to point to the same table name again.  However, I suppose that if the old one were removed at the time of the new one's creation that would not present any threading issues so that might be a bit more clean.  That is to say, if you ask a given database for a collection, and it finds one that is invalid, it will try to retrieve it again from disk and if it finds it then instead of resurrecting the old one, it removes it and creates a new one (if the consumer obeys the threading model then all of this will happen on one thread, so it's ok, as opposed to removing it from the external notification which is running on another thread probably).  Is that in line with what you are thinking?

jianminzhao: I think collection observers are associated with the collection object. Now, if I delete the collection, what does it mean to the observers? The observers now don't have the object to observe. At this time, they exist to observe what ever in the future a collection with the same name is created. Does the observer observe this collection, or any collection of that name? If any collection by that name, then is it problem to erase the collection from _collections of the active db instance?

LiteCore/Storage/SQLiteKeyStore.cc
borrrden: Fixed this here anyway

LiteCore/Database/DatabaseImpl.cc
```diff
@@ -626,11 +626,28 @@ namespace litecore {
...
+                        // Furthermore, canCreate is false so don't return a collection object
+                        return nullptr;
+                    } 
+                }
+
+                // If we reach here, either canCreate is true and we are about to recreate
+                // a previously deleted collection, or canCreate is false and the deleted collection
+                // was created by another instance already.  Either way, it is time to remove the
+                // old invalid entry now that it is safe to do so.
+                asInternal(i->second)->close();
```
jianminzhao, line 649: If won't ever resurrect the collection object, is it still needed to. have a _valid flag? Could C4Collection::invalidate simply assign _database to NULL?

C/tests/c4CollectionTest.cc
```diff
@@ -190,12 +190,25 @@ N_WAY_TEST_CASE_METHOD(C4CollectionTest, "Collection Lifecycle Multi-DB", "[Data
...
     Retained<C4Collection> guitars2 = c4db_getCollection(db2, Guitars, ERROR_INFO());
     REQUIRE(guitars2);
 
+    CHECK(c4coll_getDocumentCount(guitars) == 100);
+    CHECK(c4coll_getDocumentCount(guitars2) == 100);
+    CHECK(c4coll_getLastSequence(guitars) == 100_seq);
+    CHECK(c4coll_getLastSequence(guitars2) == 100_seq);
+
+    // First delete the collection on the second DB instances
     REQUIRE(c4db_deleteCollection(db2, Guitars, ERROR_INFO()));
```
jianminzhao, line 212: The asymmetry between guitars and guitars2 as regard to Retained: "guitars2" must be retained or the next statement will blow up, but "guitars" does not have to be Retained only because it is the not derived from db2 by which the collection is deleted. It is awkward to describe. I wonder if you would consider to also keep it in the _collections for the db instance that has the collection be deleted? It won't harm to keep it in _collections as we already have the code to deal with it.

LiteCore/Database/DatabaseImpl.cc
borrrden: That might be true, I will think about that.

LiteCore/Database/DatabaseImpl.cc
```diff
@@ -642,30 +659,46 @@ namespace litecore {
...
+                _defaultCollection = nullptr;
+
+            t.commit(); 
         }
-        _dataFile->deleteKeyStore(collectionNameToKeyStoreName(spec));
-        if (isDefaultCollection(spec))
-            _defaultCollection = nullptr;
 
-        t.commit();
+        _dataFile->forOtherDataFiles([spec](DataFile* df) {
```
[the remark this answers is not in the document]
jianminzhao: After today's meeting I feel more strongly to move the above statement inside the transaction.

C/tests/c4CollectionTest.cc
borrrden: I don't consider this essential since we have a clear warning about this in the header, but on the other hand "why not?" comes to mind.

LiteCore/Database/DatabaseImpl.cc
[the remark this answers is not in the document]
borrrden: I think I see the problem you are trying to point out, but a transaction is not going to help.

The problem would be if a collection is deleted on one thread and created on another in between the collection invalidation and the datafile reset.  The creation would not recreate the table because of the existence flag.  A transaction is not going to stop this logic though because it has nothing to do with the database.  A transaction would force the CREATE TABLE statement to wait, but it's not going to do anything about the existence flag.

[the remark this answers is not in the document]
borrrden: I have an idea though, I will switch the order.  Inform the external files FIRST that the collection is removed, so that they may invalidate.  Then, as the final step, delete the actual table.  If this happens in a transaction it will correctly block any new table creation.

C/c4Database.cc
```diff
@@ -191,36 +199,39 @@ C4Database::C4Database(std::string name, std::string dir, const C4DatabaseConfig
...
 Retained<C4Query> C4Database::newQuery(C4QueryLanguage language, slice expr, int *errPos) const {
-    return C4Query::newQuery(_defaultCollection, language, expr, errPos);
+    return C4Query::newQuery(getDefaultCollectionSafe(), language, expr, errPos);
 }
 
 
 #pragma mark - INDEXES:
 
 
+
```
jianminzhao, line 209: extra spaces are intentional?

jianminzhao (approved):

C/c4Database.cc
borrrden: No, I'm not sure how these got in

borrrden:
I moved a chunk of this PR into another PR (#1490) so that one should go in first and then this one rebased on top of it.

LiteCore/Database/DatabaseImpl.cc
```diff
@@ -658,19 +670,25 @@ namespace litecore {
         // as happens if you call `deleteCollection(coll->spec())`:
         string keyStoreName = collectionNameToKeyStoreName(spec);
         bool isDefault = isDefaultCollection(spec);
-
         Transaction t(this);
 
+        _dataFile->forOtherDataFiles([&keyStoreName](DataFile* df) {
```
snej, line 675: Strictly speaking, you shouldn't notify until after the `commit()` (or maybe just before?), because the transaction might yet fail if `deleteKeyStore` fails.

LiteCore/Storage/DataFile.cc
```diff
@@ -324,6 +324,19 @@ namespace litecore {
         }
     }
 
+    void DataFile::resetKeyStore(const string &name) {
+        logDebug("reset KVS '%s'", name.c_str());
+        auto i = _keyStores.find(name);
```
snej, line 329: I hate to say it, but … this isn't thread-safe 😩 I guess we need a mutex to protect `_keyStores`?

snej (changes requested):

LiteCore/Database/DatabaseImpl.cc
borrrden: I see the logic there.  I think the benefit is enough here because the worst that happens if the call fails is that the next time a collection on this table is used, it will do a no-op create (CREATE IF NOT EXISTS).  Otherwise you have a situation where the table is gone, but the other database instances are not informed yet.  A creation during this window would cause issues with nonexistent tables.

LiteCore/Storage/DataFile.cc
borrrden: Are there other non thread safe methods on here ? At a higher level this method is protected by the collections mutex, or are there other situations you are thinking of?

LiteCore/Database/DatabaseImpl.cc
jianminzhao: Interesting I had similar train of thoughts. In my abandoned PR, I actually tried to set the (existence) flags only when I know when it's going to commit. As I was reviewing this PR, I was really thinking it and reasoned same path. If the flag is mistakenly nonexistent, it's not fatal because, as Jim pointed out, we create new table with CREATE IF NOT EXISTS. The reverse error is detrimental: the flag says committed, but table actually missing.

LiteCore/Storage/SQLiteKeyStore.cc
```diff
@@ -87,7 +83,7 @@ namespace litecore {
                                 "  version BLOB,"
                                 "  body BLOB,"
                                 "  extra BLOB)"));
-        _existence = db().inTransaction() ? kUncommitted : kCommitted;
+        _uncommitedTable = db().inTransaction();
```
jianminzhao, line 86: This corresponds to "uncommitted create table," The other part is "uncommitted delete table." Do we need to worry about it? Perhaps not, just mention it. We may probably tolerate the bias: a good Collection object is guaranteed to be good, but the bad (invalid) collection object may not be bad. User should try to get the collection object again.

borrrden: No, the reason to worry about uncommitted create table is in the case of a transaction in which a table is created and several cached statements are made.  If the transaction is aborted, those cached statements also need to be discarded since they are cached at the SQLite level and refer to a now nonexistent table.

LiteCore/Storage/SQLiteKeyStore.cc
```diff
@@ -99,14 +95,10 @@ namespace litecore {
 
 
     void SQLiteKeyStore::reopen() {
-        if (_existence == kNonexistent)
+        if (!db().keyStoreExists(tableName()))
```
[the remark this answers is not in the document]
borrrden: It is there to protect the uncommitted table logic.  If the table already exists, uncommitted table should remain false regardless of whether or not a transaction is occurring.

LiteCore/Storage/SQLiteKeyStore.cc
```diff
@@ -97,7 +95,7 @@ namespace litecore {
 
 
     void SQLiteKeyStore::reopen() {
-        if (_existence == kNonexistent)
+        if (!db().keyStoreExists(_name))
```
jianminzhao, line 98: I meant the above if-test, whether it is necessary?

LiteCore/Storage/SQLiteKeyStore.cc
jianminzhao: Likewise, if delete-table gets aborted, we will have collection object marked invalid and not reverted given that the transaction is aborted.

LiteCore/Storage/SQLiteKeyStore.cc
borrrden: I know what you mean, and my answer was based on that.  Once the createTable method is entered there is no way to determine how to correctly set the _uncommittedTable boolean.  The only way is to check if it is already existing.

jianminzhao: I see, you wanted to be sure that the following create table actually creates one.

LiteCore/Storage/SQLiteKeyStore.cc
jianminzhao: I was talking about "uncommitted delete table."

LiteCore/Storage/SQLiteKeyStore.cc
borrrden: Reasonably sure, yes.  Since it is a check and create, there is still room for racing but the penalty for losing the race is small (losing cached statements)

LiteCore/Storage/SQLiteKeyStore.cc
borrrden: This part is not related directly to collections.  I am not worried about an aborted transaction causing collections to be falsely marked invalid since the only thing that happens is that the next time around the collection objects are recreated.  There was nothing in the logic before to handle this case and I don't see a reason to add it.

snej (approved):
Whew! Looks good.

PR #1485, opened 2022-06-07
CBL-3218: Enable F_BARRIERFSYNC on macOS / iOS

borrrden:
Note to function with EE, the EE commit must be at least 44e79198


borrrden:
Note the blackduck manifest scan will fail, but it can be considered a false positive.  The version of the underlying software we use has not changed, it was just slightly altered with a patch fix.

snej (approved):

PR #1486, opened 2022-06-07
CBL-3092 : BLIPConnection crashes on the delegate.

jianminzhao:
The ConnectionDelegate is a plain pointer to its owner, Replicator. It may be deleted as it is used by the BLIPConnection. We use WeakHolder to solve the problem.

### DIFF
--- a/Replicator/Replicator.cc
+++ b/Replicator/Replicator.cc
@@ -64,7 +64,7 @@ namespace litecore { namespace repl {
                            websocket::WebSocket *webSocket,
                            Delegate &delegate,
                            Options options)
-    :Worker(new Connection(webSocket, options.properties, *this),
+    :Worker(new Connection(webSocket, options.properties, {}),
             nullptr,
             options,
             make_shared<DBAccess>(db, options.properties["disable_blob_support"_sl].asBool()),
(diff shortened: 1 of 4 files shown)


Replicator/Replicator.cc
```diff
@@ -119,7 +125,7 @@ namespace litecore { namespace repl {
             Assert(_connectionState == Connection::kClosed);
             Signpost::begin(Signpost::replication, uintptr_t(this));
             _connectionState = Connection::kConnecting;
-            connection().start();
+            connection().start(_weakConnectionDelegateThis);
```
pasin, line 128: Maybe setting the delegate in the constructor so the start() doesn't need to change the signature?

jianminzhao: It is due to a technical reason: WeakHolder is a member of Replicator, and I cannot pass a member in the constructor of the parent class.

borrrden: What about adding a setDelegate method instead, that you can call in the Replicator constructor?

borrrden (changes requested):

Replicator/Replicator.cc
jianminzhao: I thought about it, too, but, perhaps, randomly, I feel start is also good place to set it: the delegate is not meaningful until the replication is started. It also is consistent with WebSocket::connect(Delegate), i.e., you need a delegate only after connect.

Replicator/Replicator.cc
```diff
@@ -106,6 +107,11 @@ namespace litecore { namespace repl {
     }
 
 
+    Replicator::~Replicator() {
+        _weakConnectionDelegateThis->rescind(this);
```
borrrden, line 111: More of a comment on WeakHolder but it seems this would be easy to forget.  It would be cleaner if there were a WeakHolder parent of some kind, that can issue weak children, which are auto invalidated when the parent destructs.

jianminzhao: It's good comment, something like Weakable interface. I would do it next time. My goal this time is to not expose anything beyond the delegate interface. If I put this structure in Weakable parent class, we will have two non-abstract inheritance. I think Scala's Trait is a good language construct for it.

borrrden (approved):

PR #1489, opened 2022-06-09
Set the create flag when rekeying

borrrden:
This allows a temporary blob store to be created for the purposes of reencrypting the blobs

### DIFF
--- a/LiteCore/Database/DatabaseImpl.cc
+++ b/LiteCore/Database/DatabaseImpl.cc
@@ -252,7 +252,7 @@ namespace litecore {
         // Create a new BlobStore and copy/rekey the blobs into it:
         filePath().subdirectoryNamed("Attachments_temp").delRecursive();
         auto &blobStore = getBlobStore();
-        auto newStore = createBlobStore("Attachments_temp", *newKey);
+        auto newStore = createBlobStore("Attachments_temp", *newKey, true);
         try {
             blobStore.copyBlobsTo(*newStore);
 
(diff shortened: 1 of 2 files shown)


LiteCore/Database/DatabaseImpl.cc
```diff
@@ -353,8 +356,8 @@ namespace litecore {
     unique_ptr<C4BlobStore> DatabaseImpl::createBlobStore(const string &dirname,
                                                           C4EncryptionKey encryptionKey) const
     {
-	// Split path into a separate variable to workaround GCC 8 constructor resolution issue
-	alloc_slice path = filePath().subdirectoryNamed(dirname);
+	    // Split path into a separate variable to workaround GCC 8 constructor resolution issue
```
snej, line 359: Indentation is wrong

LiteCore/Database/DatabaseImpl.cc
```diff
@@ -252,7 +252,10 @@ namespace litecore {
         // Create a new BlobStore and copy/rekey the blobs into it:
         filePath().subdirectoryNamed("Attachments_temp").delRecursive();
         auto &blobStore = getBlobStore();
+        auto previous = _config.flags;
+        _config.flags |= kC4DB_Create; // Temporarily allow creation, since we need a temporary key store
```
snej, line 256: This is kind of a kludge and will leave the flags incorrect if `createBlobStore` fails. 🧐 
How about adding a flags parameter to `createBlobStore` instead?

snej (changes requested):

LiteCore/Database/DatabaseImpl.cc
borrrden: Good point, although adding it to the signature is like saying "create it" or "REALLY create it" as the method logic.

jianminzhao (approved):

PR #1490, opened 2022-06-10
Add error info to getDefaultCollection and add more closed checks

borrrden:
:warning: C4Error parameter added to c4db_getDefaultCollection

Without these close checks, use after close of a DB simply hard crashes.  Now a test will ensure that it returns caller visible error information, and allows a distinction between such things as "default collection was deleted" and "default collection was requested from a closed database"

### DIFF
--- a/C/Cpp_include/c4Database.hh
+++ b/C/Cpp_include/c4Database.hh
@@ -107,7 +107,7 @@ public:
     };
 
     /// Returns the default collection, whose name is "_default" (`kC4DefaultCollectionName`).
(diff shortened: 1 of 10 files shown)


jianminzhao:
In general, most functions will hit CheckOpen() eventually, for instance, setCookies(). getDefaultCollectionSafe, I think is important to prevent runtime crashing, I wonder if we want to protect all uses of unopened database  objects.

borrrden:
Yes, I think going forward it will be important to do that because the users now have another object in between them and the database (collection).  However, guaranteeing that is outside the scope of this PR.

jianminzhao (approved):

PR #1492, opened 2022-06-11
CBL-3220 : ReplicatorOptions to include collection

jianminzhao:
Convert C4ReplicatorParameters to repl::Options.

### DIFF
--- a/Replicator/ReplicatorOptions.hh
+++ b/Replicator/ReplicatorOptions.hh
@@ -15,6 +15,7 @@
 #include "fleece/RefCounted.hh"
 #include "fleece/Fleece.hh"
 #include "fleece/Expert.hh"  // for AllocedDict
+#include <iterator>
 
 namespace litecore { namespace repl {
 
@@ -41,16 +42,24 @@ namespace litecore { namespace repl {
 
         //---- Constructors/factories:
 
+        inline void setCollectionOptions();
+        inline void setCollectionOptions(C4ReplicatorParameters params);
+        inline void setCollectionOptions(const Options& opt);
+
         Options(Mode push_, Mode pull_)
         :push(push_), pull(pull_)
-        { }
+        {
+            setCollectionOptions();
+        }
 
         template <class SLICE>
         Options(Mode push_, Mode pull_, SLICE propertiesFleece)
         :push(push_)
         ,pull(pull_)
         ,properties(propertiesFleece)
-        { }
+        {
+            setCollectionOptions();
+        }
 
         explicit Options(C4ReplicatorParameters params)
         :push(params.push)
@@ -61,7 +70,9 @@ namespace litecore { namespace repl {
         ,propertyEncryptor(params.propertyEncryptor)
         ,propertyDecryptor(params.propertyDecryptor)
         ,callbackContext(params.callbackContext)
-        { }
+        {
+            setCollectionOptions(params);
+        }
 
         Options(const Options &opt)     // copy ctor, required because std::atomic doesn't have one
         :push(opt.push)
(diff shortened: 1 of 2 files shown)


Replicator/ReplicatorOptions.hh
```diff
@@ -158,6 +176,127 @@ namespace litecore { namespace repl {
...
+            return collectionOpts.size();
+        }
+
+//        void deprecate() {
+//            *const_cast<Mode*>(&push) = kC4Disabled;
+//            *const_cast<Mode*>(&pull) = kC4Disabled;
+//            pushFilter = nullptr;
+//            pullValidator = nullptr;
+//            callbackContext = nullptr;
+//        }
```
borrrden, line 257: Don't include commented code like this

Replicator/ReplicatorOptions.hh
```diff
@@ -158,6 +176,127 @@ namespace litecore { namespace repl {
...
+                addScope = false;
+            }
+            if (addScope) {
+                std::copy((uint8_t*)spec.scope.buf, (uint8_t*)spec.scope.buf + spec.scope.size,
+                          std::inserter(path, path.end()));
+                path.push_back('/');
+            }
+            std::copy((uint8_t*)spec.name.buf, (uint8_t*)spec.name.buf + spec.name.size,
+                      std::inserter(path, path.end()));
+            return {&path[0], path.size()};
```
borrrden, line 195: What is the purpose of this logic?

borrrden (changes requested):

Replicator/ReplicatorOptions.hh
jianminzhao: I would use it to partially test the code flow, and I will need it before I completely plumb up everything.
I can save it and delete it from the PR.

borrrden:
I'm still unclear about what all this path conversion stuff is doing.  I assume it will be used later because it looks unused now.

jianminzhao:
Do you question whether I need to save collectionSpec that is in C4ReplicationCollection in CollectonOptions at all, or whether I should convert it to collectionPath?  Or do you ask why I need function collectionPathToSpec? This function is not needed now yet, just put it there for the sake of completeness.

Replicator/ReplicatorOptions.hh
```diff
@@ -158,6 +175,120 @@ namespace litecore { namespace repl {
...
         explicit operator std::string() const;
+
+        // Collection Options:
+
+        static alloc_slice collectionSpecToPath(C4CollectionSpec spec, bool omitDefaultScope=true) {
+            bool addScope = true;
+            if (FLSlice_Compare(spec.scope, kC4DefaultScopeID) == 0 && omitDefaultScope) {
+                addScope = false;
+            }
+            size_t size = addScope ? spec.scope.size + 1 : 0;
```
borrrden, line 186: This could be set in the above if clause to avoid rechecking the bool.

Replicator/ReplicatorOptions.hh
```diff
@@ -158,6 +175,120 @@ namespace litecore { namespace repl {
...
+            size += spec.name.size;
```
borrrden, line 187: Why not start with the name size, and add on the scope + 1 if needed?

Replicator/ReplicatorOptions.hh
```diff
@@ -158,6 +175,120 @@ namespace litecore { namespace repl {
         bool boolProperty(slice property) const   {return properties[property].asBool();}
 
         explicit operator std::string() const;
+
+        // Collection Options:
+
+        static alloc_slice collectionSpecToPath(C4CollectionSpec spec, bool omitDefaultScope=true) {
```
borrrden, line 181: Perhaps add a comment about the reasoning behind this involving the BLIP protocol.

borrrden (changes requested):

borrrden (approved):

PR #1494, opened 2022-06-14
CBL-3257 : c4db_scopeNames should include name of the default scope e…

jianminzhao:
…ven if it does not contain collections.

### DIFF
--- a/C/tests/c4CollectionTest.cc
+++ b/C/tests/c4CollectionTest.cc
@@ -111,8 +111,18 @@ N_WAY_TEST_CASE_METHOD(C4CollectionTest, "Default Collection", "[Database][Colle
         db->createCollection(kC4DefaultCollectionName);
     });
 
-    // However, the default scope still exists
+    // However, the default scope still exists,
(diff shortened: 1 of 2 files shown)


C/tests/c4CollectionTest.cc
```diff
@@ -111,8 +111,18 @@ N_WAY_TEST_CASE_METHOD(C4CollectionTest, "Default Collection", "[Database][Colle
         db->createCollection(kC4DefaultCollectionName);
     });
 
-    // However, the default scope still exists
+    // However, the default scope still exists,
     CHECK(c4db_hasScope(db, kC4DefaultScopeID));
+    // and even there is no collections at all.
+    unsigned collectionCount = 0;
+    db->forEachCollection([&](C4CollectionSpec spec) {
```
borrrden, line 118: I'd like to ask that you keep C++ methods like this out.  I am trying to keep these tests written in C so that they can move into the C4Test suite in my PR

jianminzhao: Got it.

borrrden (approved):

PR #1498, opened 2022-06-17
CBL-3221, CBL-3261 : use collection-aware repl::Options in C4Replicator

jianminzhao:
- Changed Options::push and Option::pull to functions, Options::pushOf(unsigned collectionIndex) and Options::pullOf(unsigned collectionIndex)
- Changed Worker::pasive() to Worker::passive(unsigned collectionIndex)

This PR is one of a series of PRs that will gradually approach to replicating of multiple collections. For each PR in the series, I will try to have all tests pass. In the end, all current tests become special cases whereby the collection list consisting but one collection.

### DIFF
--- a/Replicator/ChangesFeed.cc
+++ b/Replicator/ChangesFeed.cc
@@ -44,8 +44,7 @@ namespace litecore { namespace repl {
     ,_options(options)
     ,_db(db)
     ,_checkpointer(checkpointer)
-    ,_continuous(_options->push == kC4Continuous)
-    ,_passive(_options->push <= kC4Passive)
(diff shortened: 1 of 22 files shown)


borrrden:
For messaging, perhaps a summary of the reasoning for the changes, rather than just a simple bullet list of signature changes.

borrrden (approved):
I have to guess at the overall intent here, but it looks like this provides a facility to work with collection specific data while for the time being keeping the functionality exactly the same (non collection) as now.

jianminzhao:
yes, exactly.

PR #1502, opened 2022-06-28
CBL-3298: Make c4db_hasScope multi handle aware

borrrden:
If searching the cache fails, enumerate all the key store names and search for a scope that way.

### DIFF
--- a/LiteCore/Database/DatabaseImpl.cc
+++ b/LiteCore/Database/DatabaseImpl.cc
@@ -620,6 +620,15 @@ namespace litecore {
             }
         }
 
+        // CBL-3298: Final fallback to detect scopes added in another handle
+        auto allStores = _dataFile->allKeyStoreNames();
+        for (const auto& store : allStores) {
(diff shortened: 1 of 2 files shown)


LiteCore/Database/DatabaseImpl.cc
```diff
@@ -620,6 +620,15 @@ namespace litecore {
             }
         }
 
+        // CBL-3298: Final fallback to detect scopes added in another handle
```
jianminzhao, line 623: We seem to have an asymmetry here: got deleteCollection, we make sure to invalidate other db instance, but for createCollection, we do not do it and that's why we resort to the following fallback. What do you think?

borrrden: I agree it is asymmetric but it doesn't really bother me.  We have the same situation in hasCollection and the only way to avoid it is to add another delegate call to sync all handles on add.  I don't think it adds enough benefit vs falling back to checking the disk if the cache is outdated.  That's essentially what most caches do anyway.

jianminzhao (approved):

PR #1503, opened 2022-06-28
CBL-3306: Fix bug where expiration only works on default collection

borrrden:
The fix is simple, but reveals another bug where the delete statement doesn't quote the table name and thus fails in most cases since non default collections contains one or more periods.

### DIFF
--- a/LiteCore/Storage/SQLiteKeyStore.cc
+++ b/LiteCore/Storage/SQLiteKeyStore.cc
@@ -590,7 +590,7 @@ namespace litecore {
             }
         }
         if (!none) {
-            expired = db().exec(format("DELETE FROM kv_%s WHERE expiration <= %" PRId64,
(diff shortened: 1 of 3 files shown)


jianminzhao (approved):

LiteCore/Storage/SQLiteKeyStore.cc
```diff
@@ -590,7 +590,7 @@ namespace litecore {
             }
         }
         if (!none) {
-            expired = db().exec(format("DELETE FROM kv_%s WHERE expiration <= %" PRId64,
+            expired = db().exec(format("DELETE FROM \"kv_%s\" WHERE expiration <= %" PRId64,
```
jianminzhao, line 593: There is a method, quotedTableName() in SQLiteKeyStore.

jianminzhao (approved):

PR #1504, opened 2022-06-28
CBL-3284 CBL-3285: Account for collection in changes / proposeChanges

borrrden:
I decided to put the collection index inside of Pusher (and later Puller) rather than inside Worker since Replicator is also a Worker and collection index is not helpful inside the Replicator since it handles all of them by creating the child workers in the first place (future work)


borrrden:
Will do this again when master has the relevant API changes.

PR #1505, opened 2022-06-29
CBL-3349 : getPendingDocIDs and isDocumentPending to have collection

jianminzhao:
…as argument


borrrden (changes requested):
The commit message needs to have warnings about public API change

borrrden:
Would recommend something describing the change rather than "API change happened." because these warnings are meant to be all aggregated into a staging/master commit and will lose context otherwise.

PR #1506, opened 2022-06-29
CBL-3333: Use size_t for types representing array size

borrrden:
:warning: C4ReplicatorParameters::collectionCount changed from unsigned to size_t (size change on 64-bit platforms)

### DIFF
--- a/C/include/c4ReplicatorTypes.h
+++ b/C/include/c4ReplicatorTypes.h
@@ -226,7 +226,7 @@ C4API_BEGIN_DECLS
         // If collections == nullptr, we will use the deprecated fields to build
         // the internal config for one collection being the default collection.
         C4ReplicationCollection             *collections;
-        unsigned                            collectionCount;
+        size_t                               collectionCount;
     } C4ReplicatorParameters;
 
 


C/include/c4ReplicatorTypes.h
```diff
@@ -226,7 +226,7 @@ C4API_BEGIN_DECLS
         // If collections == nullptr, we will use the deprecated fields to build
         // the internal config for one collection being the default collection.
         C4ReplicationCollection             *collections;
-        unsigned                            collectionCount;
+        size_t                               collectionCount;
```
jianminzhao, line 229: Is it necessary? Does it mean my CollectionIndex also should be size_t? That many collections?

borrrden: It's not about the number of collections, but rather the pattern we use whereby size_t represents size of things.  Furthermore it is much more well defined than a simple unsigned, which can be different sizes depending on the compiler and operating system and make it difficult to create bindings to managed languages.  That's my stance anyway, but this is an existing issue filed by @pasin so I will let him also add his justification.

pasin: > Furthermore it is much more well defined than a simple unsigned, which can be different sizes depending on the compiler and operating system and make it difficult to create bindings to managed languages. 

This is the accurate reason.

pasin (approved):

PR #1507, opened 2022-06-30
CBL-3087: Correct bug in array_count that counts null values

borrrden:
The documentation says that it is supposed to consider only non-null values


jianminzhao:
Something else smells weird: why is it not routed to fl_array_count?

borrrden:
I don't have the answer to that, and I consider it outside the scope of fixing the bug.

jianminzhao:
I found it, fl_array_count if for literal array, and fl_count if for property that is an array. In this case, "name" is a property.

jianminzhao (approved):

PR #1508, opened 2022-06-30
Vectorize pushers and pullers by collections.

jianminzhao:
CBL-3288 : Replicator to work with collections
CBL-3290 : Pusher to work with collections
CBL-3292 : Puller to work with collections
CBL-3289 : Checkpointer to work with collections
CBL-3291 : ReplicatorChangesFeed to work collections
CBL-3293 : RevFinder to work with collections
CBL-3294 : IncomingRev to work collections
CBL-3295 : Inserter to work collections

Note: there are 4 broken tests, which are tagged [.broken], to be fixed, and a number "TBD:"s to be dealt with.

### DIFF
--- a/REST/tests/SyncListenerTest.cc
+++ b/REST/tests/SyncListenerTest.cc
@@ -242,7 +242,7 @@ TEST_CASE_METHOD(C4SyncListenerTest, "P2P Sync connection count", "[Listener][C]
 }
 
 
-TEST_CASE_METHOD(C4SyncListenerTest, "P2P ReadOnly Sync", "[Push][Pull][Listener][C]") {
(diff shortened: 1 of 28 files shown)


borrrden:
> Yes, unfortunately, this does not work with Windows

It's not defined behavior, this is not a Windows problem.  Windows just does you the courtesy of telling you about it.  I think there is still a way to do it from the back (via `base()` etc) but I don't see how this is a worthwhile optimization.  What guarantee is there that all of the items to remove exist in the back of the collection?  

> I think enqueue method will retain/release the actor/worker around calling it.

It does, but in this case the message being enqueued is `_childChangedStatus`.  The *parent* is being retained, but the child (i.e. the first argument to the function) is not.  The child is what is being referenced after release and there should be a guarantee that the child is kept alive until after this function is done.

borrrden (approved):

PR #1510, opened 2022-07-07
CBL-3284 CBL-3285: Account for collection in changes / proposeChanges

borrrden:


Replicator/Pusher.cc
```diff
@@ -227,6 +227,10 @@ namespace litecore { namespace repl {
             req[kConflictIncludesRevProperty] = "true"_sl;
         }
 
+        if (collectionIndex() != kNotCollectionIndex) {
```
jianminzhao, line 230: It should be Assert. If false, a lot of things won't work.

jianminzhao:
I think you also need to deal with the passive part, right? Being server, the handlers are:
        registerHandler("changes",          &RevFinder::handleChanges);
        registerHandler("proposeChanges",   &RevFinder::handleChanges);

jianminzhao (changes requested):

Replicator/Pusher.cc
borrrden: I don't understand.  This property is not required to function for the BLIP message.  If omitted, it means "only the default collection".  What will collectionIndex be set to for legacy calls?

borrrden:
Yes, I forgot about the passive part.  I will include those shortly.

Replicator/Pusher.cc
borrrden: I see, so what you are saying is not "this should be an assert" but rather "this check is not needed" ?

Replicator/Worker.hh
```diff
@@ -123,6 +130,8 @@ namespace litecore { namespace repl {
         virtual void afterEvent() override;
         virtual void caughtException(const std::exception &x) override;
 
+        fleece::Retained<C4Collection> collection();
```
jianminzhao, line 133: Perhaps add a comment that this method will fire assert if called on Replicator, whose collectionIndex, from Worker's perspective, is kNotCollectionIndex

Replicator/RevFinder.cc
```diff
@@ -83,6 +83,8 @@ namespace litecore::repl {
 
     // Actually handle a "changes" (or "proposeChanges") message:
     void RevFinder::handleChangesNow(MessageIn *req) {
+        Assert(req->intProperty(kCollectionProperty, kNotCollectionIndex) == collectionIndex());
```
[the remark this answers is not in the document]
jianminzhao: I may change the handler to be static method, so one handler for one request, and based on the collection property to dispatch to member function.

Replicator/Worker.hh
borrrden: It will only do so in debug mode so I want to avoid that specific wording, see updated commit though.

Replicator/Pusher.cc
```diff
@@ -227,10 +227,7 @@ namespace litecore { namespace repl {
             req[kConflictIncludesRevProperty] = "true"_sl;
         }
 
-        if (collectionIndex() != kNotCollectionIndex) {
-            req[kCollectionProperty] = collectionIndex();
-        }
-
+        req[kCollectionProperty] = collectionIndex();
```
pasin, line 230: I'm wondering if this will confuse SG for the case of replicating only default collection which uses getRemoteCheckpoint() instead of getCollections().

Replicator/RevFinder.cc
```diff
@@ -83,7 +83,9 @@ namespace litecore::repl {
 
     // Actually handle a "changes" (or "proposeChanges") message:
     void RevFinder::handleChangesNow(MessageIn *req) {
-        Assert(req->intProperty(kCollectionProperty, kNotCollectionIndex) == collectionIndex());
+        if (req->intProperty(kCollectionProperty, kNotCollectionIndex) == collectionIndex()) {
+            return; // TODO: Assert once refactored to static handler
+        }
```
pasin, line 88: If we support P2P b/w CBL 3.0 and helium, if the message doesn't have index, should it refer (safely) to a default collection somehow?

Replicator/Replicator.cc
```diff
@@ -899,6 +899,7 @@ namespace litecore { namespace repl {
         _getCollectionsRequested = true;
         
         for (int i = 0; i < _collections.size(); i++) {
+            _sessionCollections.push_back(_collections[i].get());
```
pasin, line 902: I forgot the case of using default collection only, which will call getRemoteCheckpoint(). I guess when calling getRemoteCheckpoint() first time, we need to populate sessionCollections as well?

pasin: Similarly in handleGetCheckpoint(). Seem like we need a design about how to handle both collection-index and legacy BLIP messages.

borrrden: Right, this will also depend on your other comment whether we need to completely drop the collection index in this case or not.

Replicator/Pusher.cc
borrrden: That was one of the considerations, but I believe it will simply be ignored instead.  I asked for confirmation on this.

Replicator/RevFinder.cc
borrrden: Side comment: The if logic is backwards from what it should be!!

Replicator/Replicator.cc
jianminzhao, line 902: Why is this? Let me refer to static index as the index from _options. The active replicator always use the static index, right? And the static index will become protocol index. 
For the passive replicator, the static index does not have any bearing on the protocol. It only define the set that it is only willing to work with the client, ie collections of getCollections must be a subset. The request sends the mapping from index i to client spec, and the server must find the static index that corresponds to that client spec.
Right?

borrrden: This is for ease of use.  Now all we have to do to map is take the received collection index, and use it to retrieve the correct collection rather than say "if passive, use this and if active use that"

jianminzhao: Do we still populate all the workers, pusher/pullers, in the constructor of Replicator based on _options?

borrrden:
This PR has been superceded

PR #1511, opened 2022-07-08
CBL-3278 : Implement BLIP getCollections() and handleGetCollections()

pasin:
* Implemented `getCollections()` which follows the same logic from `getRemoteCheckpoint()`, and implemented handleGetCollections().

* As getCollections() will be called only when the replicator is started, `getCollections()` doesn’t include the `refresh` parameter like `getRemoteCheckpoint()` does.

* When there is a NULL checkpoint returned from the getCollections() request, the `getCollections()` will return the error as `LiteCoreDomain / kC4ErrorNotFound`. 

* Added the Checkpoint’s `readDict()` function which is used by `getCollections()` to read each checkpoint dictionary. Also, the original `readJSON()` now calls `readDict()`.

* Changed ReplicatorOptions’s `collectionPathToSpec()` to take `slice` instead of `alloc_slice` parameter so that there is no needs to create `alloc_slice` from `slice` to pass to the function.

**NOTE :**

* getCollections() **will need to be hooked up in the Replicator.cc** when ready. Basically it will be called from the `start()` and `_onConnect()` function when the _collections doesn't contain only the default collection. Otherwise, the `getRemoteCheckpoint()` should be used.

* In `saveCheckpointNow()` which is per collection, I understand that `getRemoteCheckpoint()` (not `getCollections()`) will be used to fetch the remote checkpoint if the result of setCheckpoint() request is 409.

* I have done sanity tests to verify the data format of both request and response from the getCollections.

### DIFF
--- a/Replicator/Checkpoint.cc
+++ b/Replicator/Checkpoint.cc
@@ -74,26 +74,22 @@ namespace litecore { namespace repl {
 
 
     void Checkpoint::readJSON(slice json) {
+        Doc root;
         if (json) {
-            Doc root = Doc::fromJSON(json, nullptr);
+            root = Doc::fromJSON(json, nullptr);
             if (!root)
                 LogError(SyncLog, "Unparseable checkpoint: %.*s", SPLAT(json));
-            else {
-                readDict(root);
-                return;
-            }
         }
-        resetLocal();
-        _remote = {};
+        readDict(root);
     }
+
 
     void Checkpoint::readDict(Dict root) {
         resetLocal();
         _remote = {};
-        if (!root) {
-            LogError(SyncLog, "Invalid checkpoint dictionary");
+        
+        if (!root)
             return;
-        }
         
         _remote = RemoteSequence(root["remote"_sl]);
 
--- a/Replicator/Checkpoint.cc
+++ b/Replicator/Checkpoint.cc
@@ -74,32 +74,44 @@ namespace litecore { namespace repl {
 
 
     void Checkpoint::readJSON(slice json) {
-        resetLocal();
-        _remote = {};
         if (json) {
             Doc root = Doc::fromJSON(json, nullptr);
-            if (!root) {
+            if (!root)
                 LogError(SyncLog, "Unparseable checkpoint: %.*s", SPLAT(json));
+            else {
+                readDict(root);
                 return;
             }
-            _remote = RemoteSequence(root["remote"_sl]);
+        }
+        resetLocal();
+        _remote = {};
+    }
+
+    void Checkpoint::readDict(Dict root) {
+        resetLocal();
+        _remote = {};
+        if (!root) {
+            LogError(SyncLog, "Invalid checkpoint dictionary");
+            return;
+        }
+        
+        _remote = RemoteSequence(root["remote"_sl]);
 
 #ifdef SPARSE_CHECKPOINTS
-            // New properties for sparse checkpoint:
-            Array pending = root["localCompleted"].asArray();
-            if (pending) {
-                for (Array::iterator i(pending); i; ++i) {
-                    auto first = C4SequenceNumber(i->asUnsigned());
-                    auto last = C4SequenceNumber((++i)->asUnsigned());
-                    if (last >= first)
-                        _completed.add(first, last + 1);
-                }
-            } else
-#endif
-            {
-                auto minSequence = (C4SequenceNumber) root["local"_sl].asInt();
-                _completed.add(0_seq, minSequence + 1);
+        // New properties for sparse checkpoint:
+        Array pending = root["localCompleted"].asArray();
+        if (pending) {
+            for (Array::iterator i(pending); i; ++i) {
+                auto first = C4SequenceNumber(i->asUnsigned());
+                auto last = C4SequenceNumber((++i)->asUnsigned());
+                if (last >= first)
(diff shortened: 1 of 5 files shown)


Replicator/Checkpoint.cc
```diff
@@ -74,32 +74,44 @@ namespace litecore { namespace repl {
 
 
     void Checkpoint::readJSON(slice json) {
-        resetLocal();
-        _remote = {};
```
borrrden, line 78: Why remove this when every path does it anyway?

Replicator/Checkpoint.cc
```diff
@@ -74,32 +74,44 @@ namespace litecore { namespace repl {
...
         if (json) {
             Doc root = Doc::fromJSON(json, nullptr);
-            if (!root) {
+            if (!root)
```
borrrden, line 79: This reversal of flow logic makes things hard to follow.  The former was easier:  If fail, return early.  This is "If success, return early, else log error and continue"

Replicator/Replicator.cc
```diff
@@ -980,4 +1100,78 @@ namespace litecore { namespace repl {
...
+            alloc_slice body, revID;
+            int status = 0;
+            try {
+                if (!Checkpointer::getPeerCheckpoint(_db->useLocked(), checkpointID, body, revID)) {
+                    enc.writeValue(Dict::emptyDict());
+                    continue;
+                }
+            } catch (...) {
+                C4Error::warnCurrentException("Replicator::handleGetCollections");
+                status = 502;
```
borrrden, line 1160: Why 502 for unknown exception?

borrrden (changes requested):

Replicator/Replicator.cc
pasin: It's the same error returned in GetRemoteCheckPoint(). I do not know the reason why it's 502 instead 400 or something else.

Replicator/Checkpoint.cc
pasin: It is my intention to avoid calling resetLocal() and reseting _remote twice. Agreed that it's hard to understand so it's not worth it. I will revert it back

pasin:
**Note:** 
From slack, SG has proposed the change to the format of scope and collection from [scope]/<collection> to [scope].<collection>. This change will not be handled in this PR. This PR has used Options::collectionSpecToPath and Options::collectionPathToSpec so I expect that the change will be done there.

Replicator/Replicator.cc
```diff
@@ -980,4 +1100,78 @@ namespace litecore { namespace repl {
...
+        }
+        
+        if (checkpointIDs.count() != collections.count()) {
+            request->respondWithError({"BLIP"_sl, 400, "Invalid getCollections message: mismatched checkpoint_ids and collections"_sl});
+            return;
+        }
+        
+        unordered_set<C4Database::CollectionSpec> specs;
+        for (auto &coll : _collections) {
+            specs.insert(coll->getSpec());
```
jianminzhao, line 1131: Perhaps we should add a check that the collection has at least one Passive element (push/pull).

pasin: Is this required? Do we allow to mix passive and active configuration? If we do not allow that this should be validated when the Replicator is created so that we don't have mixes.

pasin: From slack, this situation could happen within LiteCore's tests.

Replicator/Replicator.cc
```diff
@@ -783,6 +785,124 @@ namespace litecore { namespace repl {
...
+    void Replicator::getCollections() {
+        if (_getCollectionsRequested)
+            return;     // already in progress
+        
+        for (int i = 0; i < _collections.size(); i++) {
+            if (!_remoteCheckpointDocID[i])
+                _remoteCheckpointDocID[i] = _checkpointer[i]->initialCheckpointID();
+            
+            // Note: is there a case that _remoteCheckpointDocID[i] is nullslice?
+            if (!_remoteCheckpointDocID[i] || _connectionState != Connection::kConnected)
```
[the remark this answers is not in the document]
pasin: From slack, we will move it up from the loop.

jianminzhao (approved):

Replicator/Replicator.cc
```diff
@@ -1126,11 +1128,18 @@ namespace litecore { namespace repl {
             return;
         }
         
-        unordered_set<C4Database::CollectionSpec> specs;
-        for (auto &coll : _collections) {
-            specs.insert(coll->getSpec());
+        unordered_map<C4Database::CollectionSpec, C4Collection*> passiveCollections;
+        for (int i = 0; i < _collections.size(); i++) {
+            if (_options->pushOf(i) == kC4Passive || _options->pullOf(i) == kC4Passive) {
```
borrrden, line 1133: I still refuse to believe that we mix active and passive.  This should be an assertion, not a check that makes the code less readable and guards in multiple places against an invalid situation.

borrrden (changes requested):

borrrden (approved):

Replicator/Replicator.cc
pasin: From slack discussion, there shouldn't be mix and match b/w active and passive collections. The assertion will be done when the Replicator is created. The filter will be remove from here.

PR #1512, opened 2022-07-11
CBL-3388 : Lump all collection-based vectors into one vector of structs

jianminzhao:


Replicator/Replicator.cc
```diff
@@ -116,39 +105,39 @@ namespace litecore { namespace repl {
...
+                sub.pushStatus = kC4Stopped;
             }
             if (_options->pullOf(i) != kC4Disabled) {
-                _pullers[i] = new Puller(this, i);
-                _pullStatusV[i] = kC4Busy;
-                if (_checkpointer[i].get() == nullptr) {
-                    _checkpointer[i].reset(new Checkpointer(_options, webSocket->url(), _collections[i]));
+                sub.puller = new Puller(this, i);
+                sub.pullStatus = kC4Busy;
+                if (sub.checkpointer.get() == nullptr) {
```
pasin, line 120: This can be done as `if (!sub.checkpointer)` as well.

Replicator/Replicator.cc
```diff
@@ -311,45 +299,48 @@ namespace litecore { namespace repl {
 
     string Replicator::statusVString() const {
         std::stringstream ss;
-        for (CollectionIndex i = 0; i < _pushStatusV.size(); ++i) {
+        for (CollectionIndex i = 0; i < _subRepls.size(); ++i) {
```
pasin, line 302: Please be aware that if this function is also used by the passive replicator, it may not work as expected.

Replicator/Replicator.cc
```diff
@@ -418,17 +406,17 @@ namespace litecore { namespace repl {
...
-        for (CollectionIndex i = 0; i < _collections.size(); ++i) {
-            if (_pullStatusV[i].error.code)
-                onError(_pullStatusV[i].error);
-            else if (_pushStatusV[i].error.code)
-                onError(_pushStatusV[i].error);
+        for (CollectionIndex i = 0; i < _subRepls.size(); ++i) {
+            if (_subRepls[i].pullStatus.error.code)
+                onError(_subRepls[i].pullStatus.error);
+            else if (_subRepls[i].pushStatus.error.code)
+                onError(_subRepls[i].pushStatus.error);
```
pasin, line 413: What will happen when there are multiple onError() calls?

pasin (approved):
I don't see any issues with the changes when comparing with the original code and logic. I do have one question about onError() but that might not be directly related to this PR.

Replicator/Replicator.cc
jianminzhao: Yes, thanks for pointing it out. I may try to unify them when I come to the passive basics. Make an index mapping from, active collections to configured collection. For Active replicator, it would be i -> i. For Passive replicator, it will be non trivial mapping.

Replicator/Replicator.cc
jianminzhao: I thought about it as well, but we may occur before the collections,
       if (_pullStatus.error.code)
            onError(_pullStatus.error);
        else if (_pushStatus.error.code)
            onError(_pushStatus.error);
  although not from different collection, but may come from puller and pusher. Now, it simply multiply by number of collections.
I guess, the first onError will throw if it is non-transient error.

PR #1513, opened 2022-07-13
CBL-3405 : removing the TBDs that left collectionIndex with 0 as temp…

jianminzhao:
…orary measure.

Also added repl::Options::sanitize() to enusre a Replicator is either active or passive. That is, it cannot have one collection doing active replication and the other doing passive replication.

### DIFF
--- a/Replicator/ChangesFeed.cc
+++ b/Replicator/ChangesFeed.cc
@@ -47,7 +47,7 @@ namespace litecore { namespace repl {
     ,_skipDeleted(_options->skipDeleted())
     {
         DebugAssert(_checkpointer);
-        auto i = _options->collectionSpecToIndex[_checkpointer->collection()->getSpec()];
+        auto i = _options->collectionSpecToIndex().at(_checkpointer->collection()->getSpec());
         _continuous = _options->pushOf((CollectionIndex)i) == kC4Continuous;
         filterByDocIDs(_options->docIDs());
     }
@@ -58,7 +58,7 @@ namespace litecore { namespace repl {
 
     string ChangesFeed::loggingClassName() const  {
         string className = Logging::loggingClassName();
-        if (!_options->isActive)
+        if (!_options->isActive())
             toLowercase(className);
         return className;
     }
@@ -75,7 +75,7 @@ namespace litecore { namespace repl {
                 combined->insert(move(docID));
         }
         _docIDs = move(combined);
-        if (!_options->isActive)
+        if (!_options->isActive())
             logInfo("Peer requested filtering to %zu docIDs", _docIDs->size());
     }
 
@@ -102,7 +102,7 @@ namespace litecore { namespace repl {
             getHistoricalChanges(changes, limit);
         changes.lastSequence = _maxSequence;
 
-        if (_options->isActive && changes.lastSequence >= changes.firstSequence) {
+        if (_options->isActive() && changes.lastSequence >= changes.firstSequence) {
             _checkpointer->addPendingSequences(changes.revs,
                                                changes.firstSequence, changes.lastSequence);
         }
@@ -240,7 +240,7 @@ namespace litecore { namespace repl {
         if (info.expiration > C4Timestamp::None && info.expiration < c4_now()) {
             logVerbose("'%.*s' is expired; not pushing it", SPLAT(info.docID));
             return nullptr;             // skip rev: expired
-        } else if (_options->isActive && _checkpointer->isSequenceCompleted(info.sequence)) {
+        } else if (_options->isActive() && _checkpointer->isSequenceCompleted(info.sequence)) {
             return nullptr;             // skip rev: checkpoint says we already pushed it before
         } else if (_docIDs != nullptr
                     && _docIDs->find(slice(info.docID).asString()) == _docIDs->end()) {
@@ -336,7 +336,7 @@ namespace litecore { namespace repl {
         if (foreignAncestor && !_usingVersionVectors
                     && C4Document::getRevIDGeneration(foreignAncestor)
                         >= C4Document::getRevIDGeneration(doc->revID())) {
-            if (!_options->isActive) {
+            if (!_options->isActive()) {
                 C4Error error = C4Error::make(WebSocketDomain, 409,
                                      "conflicts with newer server revision"_sl);
                 _delegate.failedToGetChange(rev, error, false);
--- a/Replicator/ChangesFeed.cc
+++ b/Replicator/ChangesFeed.cc
@@ -48,7 +48,7 @@ namespace litecore { namespace repl {
     {
         DebugAssert(_checkpointer);
         auto i = _options->collectionSpecToIndex().at(_checkpointer->collection()->getSpec());
-        _continuous = _options->pushOf((CollectionIndex)i) == kC4Continuous;
+        _continuous = _options->push((CollectionIndex)i) == kC4Continuous;
         filterByDocIDs(_options->docIDs());
     }
 
--- a/Replicator/ChangesFeed.cc
+++ b/Replicator/ChangesFeed.cc
@@ -44,9 +44,11 @@ namespace litecore { namespace repl {
     ,_options(options)
     ,_db(db)
     ,_checkpointer(checkpointer)
-    ,_continuous(_options->pushOf(0) == kC4Continuous) // TBD: replacing 0 with actual collection index.
     ,_skipDeleted(_options->skipDeleted())
     {
+        DebugAssert(_checkpointer);
+        auto i = _options->collectionSpecToIndex[_checkpointer->collection()->getSpec()];
+        _continuous = _options->pushOf((CollectionIndex)i) == kC4Continuous;
         filterByDocIDs(_options->docIDs());
     }
 
@@ -56,7 +58,7 @@ namespace litecore { namespace repl {
 
     string ChangesFeed::loggingClassName() const  {
         string className = Logging::loggingClassName();
-        if (passive())
+        if (!_options->isActive)
             toLowercase(className);
         return className;
     }
@@ -73,7 +75,7 @@ namespace litecore { namespace repl {
                 combined->insert(move(docID));
         }
         _docIDs = move(combined);
-        if (passive())
+        if (!_options->isActive)
             logInfo("Peer requested filtering to %zu docIDs", _docIDs->size());
     }
 
@@ -100,7 +102,7 @@ namespace litecore { namespace repl {
             getHistoricalChanges(changes, limit);
         changes.lastSequence = _maxSequence;
 
-        if (!passive() && _checkpointer && changes.lastSequence >= changes.firstSequence) {
+        if (_options->isActive && changes.lastSequence >= changes.firstSequence) {
             _checkpointer->addPendingSequences(changes.revs,
                                                changes.firstSequence, changes.lastSequence);
         }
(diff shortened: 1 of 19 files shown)


Replicator/ReplicatorOptions.hh
```diff
@@ -315,26 +321,60 @@ namespace litecore { namespace repl {
...
             }
-            if (c.push != kC4Passive || c.pull != kC4Passive) {
-                isAllPassive = false;
+        }
+        isActive = actiCount > 0;
+        
+        if (collectionOpts.size() == 1) {
+            auto spec = collectionPathToSpec(collectionOpts[0].collectionPath);
+            if (spec == kC4DefaultCollectionSpec) {
+                collectionAware = false;
```
pasin, line 363: This collectionAware is only valid for active replicator. I don't know now that any handler for passive replicators will need to use this flag or not.

Replicator/Worker.hh
```diff
@@ -133,13 +133,21 @@ namespace litecore { namespace repl {
...
+            if (!_options->isActive) {
+                return false;
+            }
             auto collIndex = collectionIndex();
             if (collIndex == kNotCollectionIndex) {
-                //TBD: this is a Replicator. what should it be?
-                collIndex = 0;
+                for (CollectionIndex i = 0; i < _options->collectionCount(); ++i) {
+                    if (_options->pushOf(i) == kC4Continuous) {
+                        return true;
```
pasin, line 143: I don't think we have case that one collection is single shot and another is continuous. Should this be validated in sanitize() as well?

pasin (changes requested):

Replicator/ChangesFeed.cc
```diff
@@ -100,7 +102,7 @@ namespace litecore { namespace repl {
             getHistoricalChanges(changes, limit);
         changes.lastSequence = _maxSequence;
 
-        if (!passive() && _checkpointer && changes.lastSequence >= changes.firstSequence) {
+        if (_options->isActive && changes.lastSequence >= changes.firstSequence) {
```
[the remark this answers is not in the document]
jianminzhao: yes. I already ASSERT it.

Replicator/ChangesFeed.cc
```diff
@@ -238,13 +240,12 @@ namespace litecore { namespace repl {
         if (info.expiration > C4Timestamp::None && info.expiration < c4_now()) {
             logVerbose("'%.*s' is expired; not pushing it", SPLAT(info.docID));
             return nullptr;             // skip rev: expired
-        } else if (!passive() && _checkpointer && _checkpointer->isSequenceCompleted(info.sequence)) {
+        } else if (_options->isActive && _checkpointer->isSequenceCompleted(info.sequence)) {
```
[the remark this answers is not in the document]
jianminzhao: same answer

Replicator/ReplicatorOptions.hh
pasin: From slack: @jianminzhao told me that collectionAware do not used by passive replicator.

borrrden: I think this property is misnamed anyway.  passive will not need to use this flag, but rather it will need to search for the default collection in the list of collections if necessary, which is something I am working on.  What this flag is really indicating is "legacy mode or not"

Replicator/Replicator.cc
```diff
@@ -926,7 +942,9 @@ namespace litecore { namespace repl {
         Assert(json);
 
         MessageBuilder msg("setCheckpoint"_sl);
-        msg["collection"_sl] = coll;
+        if (_options->collectionAware) {
+            msg["collection"_sl] = coll;
```
borrrden, line 946: Need a note to switch to a constant here since this property will be widely used (I defined one in my PR)

Replicator/ReplicatorOptions.hh
```diff
@@ -36,20 +39,24 @@ namespace litecore { namespace repl {
         PropertyDecryptor       propertyDecryptor       {nullptr};
         void*                   callbackContext         {nullptr};
         std::atomic<C4ReplicatorProgressLevel> progressLevel {kC4ReplProgressOverall};
-        bool                    collectionAware         {true};
-
+        mutable bool            collectionAware         {true};
```
borrrden, line 42: I think this should be a method, rather than a forcibly mutable item in an otherwise mutable struct.  The logic can be the same logic used to calculate this method.

Replicator/ReplicatorOptions.hh
```diff
@@ -36,20 +39,24 @@ namespace litecore { namespace repl {
...
+        mutable bool            isActive                {true};
```
borrrden, line 43: I also think this can be a method rather than an item.

Replicator/ReplicatorOptions.hh
```diff
@@ -36,20 +39,24 @@ namespace litecore { namespace repl {
...
+        mutable std::unordered_map<C4CollectionSpec, size_t> collectionSpecToIndex;
```
borrrden, line 44: Why does this need to be explicitly marked mutable?  Doing this should be an exceptional case (mutating inside of const function for some reason)

borrrden (changes requested):

Replicator/ReplicatorOptions.hh
jianminzhao: See above. Once adopted by Replicator and its workers, it is designated as constant, and this map is is computed inside Replicator::Replicator(). Actually, I am still working on it, CBL-3406, I find I have access to the non-const options in Replicator's constructor, and can use that to avoid mutable. 
One style I mean to adhere to for this struct is to have direct access to the members as much as possible.

Replicator/ReplicatorOptions.hh
jianminzhao: As I weighed it against "legacy mode," I found legacy mode may not stick as we move on. As time goes, we have new legacy things. This is a particular legacy when we don't have the notion of collection, and hence, whenever collection is required in a context, default collection is substituted.
As for passive replicator, we deal with the legacy based on the request. If the request does not specify "collection," default collection is assumed and it is equivalent to contemporary (vs legacy) ['default']. Then, of course, we need to find whether the config is ready for this collection. In my current task, I added a mapping from collectionSpec to index in Options.

Replicator/ReplicatorOptions.hh
pasin: Another option is to add isActive() and isCollectionAware() in the Replicator class. If doing that, isCollectionAware() can be computed for both Active and Passive replicator.

Replicator/ReplicatorOptions.hh
```diff
@@ -315,26 +329,89 @@ namespace litecore { namespace repl {
             back.pullFilter = collOpts.pullFilter;
             back.callbackContext = collOpts.callbackContext;
         }
-        trimDisabledCollections();
     }
 
-    inline void Options::trimDisabledCollections() {
+    inline void Options::verify() const {
```
pasin, line 334: One thing that feels little weird is that without calling verify() some status (e.g. _isActive and _collectionAware) is not valid.

Replicator/ReplicatorOptions.hh
```diff
@@ -36,20 +39,27 @@ namespace litecore { namespace repl {
         PropertyDecryptor       propertyDecryptor       {nullptr};
         void*                   callbackContext         {nullptr};
         std::atomic<C4ReplicatorProgressLevel> progressLevel {kC4ReplProgressOverall};
-        bool                    collectionAware         {true};
 
+        bool collectionAware() const {return _collectionAware;}
+        bool isActive()        const {return _isActive;}
+        const std::unordered_map<C4CollectionSpec, size_t>& collectionSpecToIndex() const {
```
pasin, line 45: This is debatable. IMO, _collectionSpecToIndex is an internal structure and also it is more convenient if proving the functions such as 

```
pushOf(C4CollectionSpec) -> bool / optional(bool)
pullOf(C4CollectionSpec) -> bool / optional(bool)
collectionOption(C4CollecitonSpec) -> CollectionOptions / optional(CollectionOptions)
```

pasin: This is not a big deal. I feel a bit weird with the function named pushOf() or pullOf(). Maybe isPush() or isPull()?

Replicator/ReplicatorOptions.hh
pasin: Then it would need a big note, or it should be moved out to the Replicator, as its value will not be valid or computed before calling verify() function (I also commented about this in another place).

borrrden (commented):
I would be ok with this as it is with some notes about the places that are a bit suspicious so that they could be fixed up later.

Replicator/Replicator.cc
jianminzhao: Also, in my next PR, I encapsulated it in a method of Worker, Worker::setMsgCollection(msg, collectionIndex). This method will decide whether to add the "collection" property based on _options

Replicator/ReplicatorOptions.hh
jianminzhao: The reason I adopted weird name if because push and pull were already in Options at the time I attempted progress gradually and kept both them both in Options and Options::CollectionOptions. (Also, the collectionIndex had default 0 and I later made it required) Later, I removed Options::push/pull. Do you prefer Options::push(CollectionIndex), I can change them. 
As for the index mapping, I think mapping to index is more useful because the index has special meaning in the program. We may always get the actual CollectionOptions from the index.

Replicator/ReplicatorOptions.hh
jianminzhao: @pasin this is because Options can be used as a data structure outside Replicator. I saw in code that we create an Options with both push/pull disabled which caused my initial attempt to fail. I later decided to enforce the sanity of if it is used in Replicator. Hence, verify() becomes a public method as opposed to doing it in the Options' constructor. 
I prefer not to use optional. Using optional sends a message to the called that every time using it, he/she needs to test it. I want to hint a semantics that isActive, collectionAware, etc. are attributes that are required for Replicators, and therefore, before reading them, code must verify them. Currently, failure of verify() throws error right away.

Replicator/Replicator.cc
borrrden: What I meant is to eventually avoid the use of the "collection" magic string and instead have it available as a constexpr.

Replicator/ReplicatorOptions.hh
borrrden: Well it seems like you are just lying if you put it that way.  You say you want to avoid "needing to test every time" but the "fix" here is to just pretend that it is always ok when it is not.  There is an actual tri-state value here unless you want to rework the way this happens.  I can see a method that returns a COPY of the options from verify that is actually completely immutable instead of just sort of frozen halfway through.

Replicator/Replicator.cc
jianminzhao: Oh, yes. Can I take this advice in the next PR, which is also due this sprint.

Replicator/ReplicatorOptions.hh
jianminzhao: Current architecture is that, for a particular replication work, there are a group of workers, headed by Replicator. All the workers in that group share the same Options object. In child workers, calling replicator()->isActive() is less safe than _options->isActive() because replicator() returns a bare pointer whereas _options is shared ponter.

jianminzhao: I submitted my motivation of having them in _options in another response.

jianminzhao: Changed them to methods

Replicator/ReplicatorOptions.hh
jianminzhao: I think this representation is clear: when used by Replicator, and all related workers, some constraints apply. As the head of group of workers, Replicator should call verify() to ensure the constraints. As by-products, with the constraints being satisfied, some derived properties can be defined, such as isActive. With the constraints being satisfied, isActive is not well defined, because some collection is active, some can be disabled, and yet some can be passive. 
Returning a copy indeed can maintain the mutability, but there is a practicality concern. The current architecture ensures that all workers are created with the shared Options object. For example, once inside Replicator's constructor, Connection, BLIPIO, etc., have already received shared Options. It's awkward to swap the Options for all of them. If isActive is an c++ optional, then, we either have to structurally guarantee that it contains a value after certain point, or every time we have to say
if (!isActive.has_value()) throw runtime logical error. else if (*isActive) { } else { }

Replicator/Replicator.cc
jianminzhao: Done. adopted kCollectionProperty from Jim's PR

Replicator/ReplicatorOptions.hh
jianminzhao: Changed accordingly,

borrrden (approved):
I think it has become satisfactory at this point, but since this is so disruptive let's discuss it finally at a realtime meeting.

PR #1514, opened 2022-07-14
CBL-3409 : Add basic collection tests for replicator

pasin:
* Added ReplicatorCollectionTest which inherits from ReplicatorLoopbackTest.

* Included basic replication tests with default collection, single collection, and multiple collection.

* Added gC4ExpectExceptions to the tests that could have exception thrown internally by LiteCore to avoid exception breakpoint in XCode.

* NOTE: The tests are disabled by #ifdef as the implementation in LiteCore is not done yet.

### DIFF
--- a/Replicator/tests/ReplicatorCollectionTest.cc
+++ b/Replicator/tests/ReplicatorCollectionTest.cc
@@ -106,10 +106,20 @@ public:
         return coll;
     }
     
-    int addDocs(C4Database* db, CollectionSpec spec, int total) {
+    // Add new docs to a collection
+    int addCollDocs(C4Database* db, CollectionSpec spec, int total) {
         C4Collection* coll = getCollection(db, spec);
-        string prefix = db == this->db ? "db1" : (db == this->db2 ? "db2" : "newdoc");
-        return ReplicatorLoopbackTest::addDocs(coll, 0ms, total, prefix);
+        string prefix = (db == this->db ? "newdoc-db1-" : (db == this->db2 ? "newdoc-db2-" : "newdoc-"));
+        int docNo = 1;
+        for (int i = 1; docNo <= total; i++) {
+            Log("-------- Creating %d docs --------", i);
+            TransactionHelper t(db);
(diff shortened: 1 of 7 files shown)


pasin:
Mac for PR Validation needs restart.

jianminzhao (approved):

Replicator/tests/ReplicatorCollectionTest.cc
```diff
@@ -0,0 +1,251 @@
...
+    _expectedDocumentCount += addDocs(db2, Default, 5);
+    runPushPullReplication({Default}, {Default});
+    validateCollectionCheckpoints(db, db2, 0, "{\"local\":15,\"remote\":15}");
+}
+
+TEST_CASE_METHOD(ReplicatorCollectionTest, "Push Single Collection", "[Push]") {
+    _expectedDocumentCount = addDocs(db, Guitars, 10);
+    runPushReplication({Guitars}, {Guitars});
+    validateCollectionCheckpoints(db, db2, 0, "{\"local\":10}");
+}
```
borrrden, line 170: These are getting a bit out of hand.  I suggest writing one test with sections rather than separate tests for every variation.  There is a handy `SECTION` keyword.  Alternatively, I know there are other tests that run in different variations for every test in the class, which seems like it would be applicable here.

borrrden (changes requested):

Replicator/tests/ReplicatorCollectionTest.cc
pasin: OK

borrrden (approved):
Not quite ideal the way it is, but that can be addressed another time.  You can factor out the common areas of the test outside of the SECTIONS, and they will still run every time.  Basically the entire test is executed each time, just conditionally running each SECTION sequentially (i.e. the whole test with only the first section, the whole test with only the second section, etc)

pasin:
I could restructure the common code and move it out of the SECTION.

borrrden:
Even better

PR #1516, opened 2022-07-19
CBL-3406 : refactor for the separation of logics of active and passiv…

jianminzhao:
…e Replicators.

For the active replicators, workers, such as pushers and pullers, are created in Replicator's constructors. We know what to create from Replicators' configs. For the passive Replicators, we don't know what collections are going to participate until we receive the first message, getCollections. So, we will create all necessary workers after afterwards. Also, all messages come with collection index, and we will dispatch the message to the target worker based on the collection index.

### DIFF
--- a/Replicator/Replicator.cc
+++ b/Replicator/Replicator.cc
@@ -74,7 +74,6 @@ namespace litecore { namespace repl {
     ,_delegate(&delegate)
     ,_connectionState(connection().state())
     ,_docsEnded(this, "docsEnded", &Replicator::notifyEndedDocuments, tuning::kMinDocEndedInterval, 100)
-    ,_subRepls(_options->collectionCount())
     {
         // Post-conditions:
         //   collectionOpts.size() > 0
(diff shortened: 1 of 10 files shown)


Replicator/Replicator.cc
```diff
@@ -88,66 +87,19 @@ namespace litecore { namespace repl {
...
-        })) {
-            for (auto profile : {"changes", "proposeChanges", "rev", "norev"})
-                registerHandler(profile,  &Replicator::returnForbidden);
+        // Following messages are handled by appropriate workers.
+        // Replicator receives all the messages. Based on collectionIndex,
+        // it dispatches the message to appropriate workers.
+        for (auto profile : {"subChanges", "getAttachment", "proveAttachment", // passive pushers
+                "changes", "proposeChanges", "rev", "norev"                    // passive pullers
+        }) {
+            registerHandler(profile, &Replicator::handleConnectionMessage);
```
pasin, line 102: I think this can be done in registerWorkerHandler() when registering the profile first time.

Replicator/Replicator.cc
```diff
@@ -172,6 +124,10 @@ namespace litecore { namespace repl {
             connection().start();
             // Now wait for _onConnect or _onClose...
 
+            if (!_options->isActive()) {
+                return;
```
pasin, line 128: Probably should be assert?

Replicator/Replicator.cc
```diff
@@ -182,12 +138,15 @@ namespace litecore { namespace repl {
...
-                if (_options->push(i) > kC4Passive || _options->pull(i) > kC4Passive) {
-                    // Get the checkpoints:
-                    if (getLocalCheckpoint(reset, i)) {
-                        getRemoteCheckpoint(false, i);
-                    }
+            bool goOn = true;
+            for (CollectionIndex i = 0; goOn && i < _subRepls.size(); ++i) {
+                goOn = goOn && getLocalCheckpoint(reset, i);
+            }
+            if (goOn) {
```
pasin, line 145: If not goOn, should be an error and stop as well? Not sure what will be the part after not goOn.

Replicator/Replicator.cc
```diff
@@ -777,21 +744,27 @@ namespace litecore { namespace repl {
...
-
+#ifdef TEMP_WORK_AROUND
+            }
+#endif
                 // Now we have the checkpoints! Time to start replicating:
                 startReplicating(coll);
-            }
+#ifndef TEMP_WORK_AROUND
+        }
+#endif
```
pasin, line 754: Probably I got confused with the ifdef stuff. But the `startReplicating(coll)` should be done only when `refresh` is `false`.

Replicator/Replicator.cc
jianminzhao: Well, passive replicator will also be started, but it will end here.

Replicator/Replicator.cc
```diff
@@ -1148,35 +1123,48 @@ namespace litecore { namespace repl {
...
+            if (!b) {
+                request->respondWithError({"BLIP"_sl, 400, "Invalid getCollections message: duplicate collection path"_sl});
+                return;
+            }
         }
-        
+
+        // Rearrange _options->collectionOpts according to collSpecs. If i-th CollectionSpec
+        // is not found, put an empty collectionOptions at i-th position.
+        _options->rearrangeCollections(collSpecs);
```
pasin, line 1150: I'm not a big fan of mutating the input configuration. This will work if _options is a copied config for the replicator. It's a pointer so I'm not sure.

Replicator/Replicator.cc
jianminzhao: c.f. getLocalCheckpoint() -- it returns false after "getError(), stop()" We cannot go on if any one of getLocalCheckpoint() fails.

Replicator/Replicator.cc
pasin: I understand that but I understand that registering the BLIP message handler can be done in registerWorkerHandler() as well so there is no need to do it separately upfront.

Replicator/Replicator.cc
jianminzhao: It is to prevent the active replicator to send message before it receiving the response of getCheckpoint. Currently, if the local checkpoint is available, we startReplicating() without waiting for the response of getCheckpoint. I gather that it is an optimization and 3.0 protocol does not provide that the first message being a handshake, getCollections. Without the handshake, the server won't be able to interpret the collection index. I temporarily disable this optimization, that is, the active replicator start replicating after the first message is received. Of course, I mean to remove it later to be a "genuine" 3.0 client.

Replicator/Replicator.cc
jianminzhao: (meant to insert.a link here but ...)
        // RearrangeCollections() is called only by the passive replicator. For the passive replicator, we presume
        // that the order of the collection properties is not important. So, we take it as legit to permutate
        // it in const method. (The Replicator holds a const Options object.) It is supposed to be called as it
        // starts to interact with the active replicator.
        // "collections" is a list of CollectionSpecs that the active replicatore proposes to replicate, and the
        // order will be used as index to refer to respective collections.
        // Post-conditions: collectionOpts[i] and collections[i] share the same CollectionSpec if
        //                    collections[i] is found in collectionOpts;
        //                  Otherwise, an empty collectionOptions is inserted in collectionOpts[i].
        // By empty collectionOptions we mean the collection path is a null slice.

jianminzhao: Let's discuss it in the meeting

Replicator/Replicator.cc
jianminzhao: We may have 5 pushers. Each pusher will register as a worker handler for, say, "change", message. But for Replicator, we only register once with BLIPIO

jianminzhao:
Per code review, filed the following tickets,
CBL-3448: manage mutable members of repl::Options
CBL-3449: Consider to replace registerWorkerHandler with a dispatch function

pasin (approved):

PR #1517, opened 2022-07-20
CBL-3419: Add collectionContext to C4DocumentEnded

borrrden:
:warning: API Change -> Added collectionContext to C4DocumentEnded

This change requires a lot of changes to other classes that interact with this struct.  The majority of the changes are to pass in the collection context where needed to ReplicatedRev (which is dynamically cast to C4DocumentEnded).  However, ChangesFeed needed to get a hold of the collection index to use, in order to know which context to add.

### DIFF
--- a/Replicator/ChangesFeed.cc
+++ b/Replicator/ChangesFeed.cc
@@ -38,19 +38,19 @@ namespace litecore { namespace repl {
 
 
     ChangesFeed::ChangesFeed(Delegate &delegate, const Options *options,
-                             DBAccess &db, Checkpointer *checkpointer,
-                             CollectionIndex collectionIndex)
+                             DBAccess &db, Checkpointer *checkpointer)
     :Logging(SyncLog)
     ,_delegate(delegate)
     ,_options(options)
     ,_db(db)
     ,_checkpointer(checkpointer)
     ,_skipDeleted(_options->skipDeleted())
-    ,_collectionIndex(collectionIndex)
     {
         DebugAssert(_checkpointer);
-        auto i = _options->collectionSpecToIndex().at(_checkpointer->collection()->getSpec());
-        _continuous = _options->push((CollectionIndex)i) == kC4Continuous;
+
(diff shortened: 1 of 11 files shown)


Replicator/ChangesFeed.cc
```diff
@@ -38,13 +38,15 @@ namespace litecore { namespace repl {
 
 
     ChangesFeed::ChangesFeed(Delegate &delegate, const Options *options,
-                             DBAccess &db, Checkpointer *checkpointer)
+                             DBAccess &db, Checkpointer *checkpointer,
+                             CollectionIndex collectionIndex)
```
jianminzhao, line 42: collectionIndex can be obtained by, _options->collectionSpecToIndex().find(checkpointer->getSpec())

borrrden: Hmmm not wild about it but it will work.  It's a lot of steps through several different types.

jianminzhao: Actually, thinking again, the map would work, because if the index is out of the range (workingCollectionCount), it would be internal logical error; checkpointer->getSpec() must have index in the range.

Replicator/ChangesFeed.cc
```diff
@@ -47,8 +47,10 @@ namespace litecore { namespace repl {
     ,_skipDeleted(_options->skipDeleted())
     {
         DebugAssert(_checkpointer);
-        auto i = _options->collectionSpecToIndex().at(_checkpointer->collection()->getSpec());
-        _continuous = _options->push((CollectionIndex)i) == kC4Continuous;
+
+        // JIM: This breaks tons of encapsulation, and should be reworked
```
jianminzhao, line 51: In a working session, collectionIndex and collectionSpec are synonyms: they should refer to the same thing. Most places we use the index, for it's faster. Very few places we use the collectionSpec.

jianminzhao (approved):

PR #1521, opened 2022-07-21
CBL-3422: Update protocol docs for collection-aware connections

borrrden:


modules/docs/pages/replication-protocol.adoc
```diff
@@ -105,22 +105,36 @@ below.
...
 `rev`: The MVCC revision ID of the checkpoint +
 Body: JSON data of the checkpoint
 
+[[getCollections]]
+getCollections [3.1+]
+^^^^^^^^^^^^^^^^^^^^^
+
+Body: JSON dictionary
+
+A collection aware parallelized version of `getCheckpoint`.  If this message is sent before any `getCheckpoint` calls then replication will operate in 3.1+ mode, which can fully handle collections.  The body of this message will contain two arrays (`collections` and `checkpoint_ids`) each of the same length.  The combination of the two at a given index provide information about a checkpoint to retrieve (i.e. request checkpoint X for collection Y).
```
pasin, line 127: Provide info about the collection name format which is `<scope-name>.<collection-name>` and the `<scope name>.` can be omitted if the scope name is the default scope.

modules/docs/pages/replication-protocol.adoc
```diff
@@ -105,22 +105,36 @@ below.
...
+getCollections [3.1+]
+^^^^^^^^^^^^^^^^^^^^^
+
+Body: JSON dictionary
+
+A collection aware parallelized version of `getCheckpoint`.  If this message is sent before any `getCheckpoint` calls then replication will operate in 3.1+ mode, which can fully handle collections.  The body of this message will contain two arrays (`collections` and `checkpoint_ids`) each of the same length.  The combination of the two at a given index provide information about a checkpoint to retrieve (i.e. request checkpoint X for collection Y).  
+
+Response:
+
+An array of items each identical in structure to the result of `getCheckpoint`.  The entries will be ordered according to the order in the request message.
```
pasin, line 131: Provide info about "null" value for the collection that is not found and empty dict for the collection that doesn't have the remote checkpoint.

pasin (approved):

PR #1522, opened 2022-07-22
Replication of named collections

jianminzhao:
CBL-3441 : integrate getCollections
CBL-3285, CBL-3280, CBL-3281, CBL-3286, CBL-3287, CBL-3284 : miscellaneous BLIP messages

Notes: with this PR, Replicators start to work with named collections in the simplest fashion: single shot Push or Pull (not mixed), without filters or callbacks.

### DIFF
--- a/Replicator/Inserter.cc
+++ b/Replicator/Inserter.cc
@@ -163,7 +163,10 @@ namespace litecore { namespace repl {
                 =_db->insertionDB().useLocked<Retained<C4Document>>([spec, outError, &put](C4Database* db) {
                     C4Collection* collection = db->getCollection(spec);
                     if (!collection || !collection->isValid()) {
-                        return Retained<C4Document>();
+                        C4Error err{LiteCoreDomain, kC4ErrorNotOpen};
+                        if (outError)
+                            *outError = err;
+                        C4Error::raise(err);
                     } else {
                         return collection->putDocument(put, nullptr, outError);
                     }
--- a/Replicator/Inserter.cc
+++ b/Replicator/Inserter.cc
@@ -158,7 +158,17 @@ namespace litecore { namespace repl {
                 put.allocedBody = {(void*)bodyForDB.buf, bodyForDB.size};
 
                 // The save!!
-                auto doc = _db->insertionDB().useLocked()->putDocument(put, nullptr, outError);
+                C4CollectionSpec spec = replicator()->collection(collectionIndex())->getSpec();
+                Retained<C4Document> doc
+                =_db->insertionDB().useLocked<Retained<C4Document>>([spec, outError, &put](C4Database* db) {
+                    C4Collection* collection = db->getCollection(spec);
+                    if (!collection || !collection->isValid()) {
+                        return Retained<C4Document>();
+                    } else {
+                        return collection->putDocument(put, nullptr, outError);
+                    }
+                });
+
                 if (!doc)
                     return false;
                 logVerbose("    {'%.*s' #%.*s <- %.*s} seq %" PRIu64,
--- a/Replicator/Replicator.cc
+++ b/Replicator/Replicator.cc
@@ -278,6 +278,8 @@ namespace litecore { namespace repl {
             ss << "/";
             ss << sub.pushStatus.progress.unitsTotal
                 + sub.pullStatus.progress.unitsTotal;
+            ss << ", docCount=" << sub.pushStatus.progress.documentCount
+                                + sub.pullStatus.progress.documentCount;
         }
         return ss.str();
     }
@@ -368,10 +370,11 @@ namespace litecore { namespace repl {
         setProgress(_pushStatus.progress + _pullStatus.progress);
 
         if (SyncBusyLog.willLog(LogLevel::Info)) {
-            logInfo("pushStatus=%-s, pullStatus=%-s, progress=%" PRIu64 "/%" PRIu64 "",
+            logInfo("pushStatus=%-s, pullStatus=%-s, progress=%" PRIu64 "/%" PRIu64 "/%" PRIu64 "",
                     kC4ReplicatorActivityLevelNames[_pushStatus.level],
                     kC4ReplicatorActivityLevelNames[_pullStatus.level],
-                    status().progress.unitsCompleted, status().progress.unitsTotal);
+                    status().progress.unitsCompleted, status().progress.unitsTotal,
+                    status().progress.documentCount);
         }
         if (SyncBusyLog.willLog(LogLevel::Verbose)) {
             logVerbose("Replicator status collection-wise: %s", statusVString().c_str());
@@ -702,7 +705,7 @@ namespace litecore { namespace repl {
         }
         MessageBuilder msg("getCheckpoint"_sl);
         msg["client"_sl] = sub.remoteCheckpointDocID;
-        setMsgCollection(msg, coll);
+        assignCollectionToMsg(msg, coll);
         Signpost::begin(Signpost::blipSent);
         sendRequest(msg, [this, refresh, coll, &sub](MessageProgress progress) {
             // ...after the checkpoint is received:
@@ -711,9 +714,12 @@ namespace litecore { namespace repl {
             Signpost::end(Signpost::blipSent);
             MessageIn *response = progress.reply;
 
-            auto collectionIn = response->intProperty(kCollectionProperty, kNotCollectionIndex);
-            (void)collectionIn; // The following statement only compiles in Debug build.
-            DebugAssert(collectionIn == kNotCollectionIndex || collectionIn == coll);
+            slice error;
+            std::tie(std::ignore, error) = checkCollectionOfMsg(*response, coll);
+            if (error) {
+                gotError(C4Error::make(LiteCoreDomain, kC4ErrorRemoteError, error));
+                return;
+            }
 
             Checkpoint remoteCheckpoint;
 
@@ -778,10 +784,10 @@ namespace litecore { namespace repl {
     void Replicator::getCollections() {
         if (_getCollectionsRequested)
             return;     // already in progress
-        
+
         if (_connectionState != Connection::kConnected)
             return;         // Not ready yet; Will be called again from _onConnect.
-        
+
         for (int i = 0; i < _subRepls.size(); i++) {
             if (!_subRepls[i].remoteCheckpointDocID)
                 _subRepls[i].remoteCheckpointDocID = _subRepls[i].checkpointer->initialCheckpointID();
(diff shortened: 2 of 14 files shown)


Replicator/Worker.cc
```diff
@@ -313,10 +313,37 @@ namespace litecore { namespace repl {
             _parent = nullptr;
     }
 
-    void Worker::setMsgCollection(blip::MessageBuilder& msg, CollectionIndex i) {
+    // Either there is error, or return a valid collection index
+    std::pair<CollectionIndex, slice>
+    Worker::checkCollectionOfMsg(const blip::MessageIn& msg, CollectionIndex i) const {
```
pasin, line 318: Personally I like a conservative function that does or returns one thing such as :

```
bool validateMessageCollectionIndex(CollectionIndex index, int refRequestIndex, C4Error* NULLABLE outError);
```

Then 
```
C4Error error;
CollectionIndex collIndex = getCollectionIndex(msg);
if (!validateMessageCollectionIndex(collIndex, -1, &error)) {
    ...
}
```

Replicator/Worker.cc
```diff
@@ -313,10 +313,37 @@ namespace litecore { namespace repl {
...
+            }
+        }
+
+        if (!err && collIn >= _options->workingCollectionCount()) {
+            err = error3;
+        }
+
+        if (!err && i != kNotCollectionIndex) {
+            // check against i if it is a valid collection index
+            if (collIn != i) {
```
pasin, line 342: This validation is confusing me a bit as I first thought this function is validating the collection index in blip message handler function. It seems like it's also for the REPLY message. Maybe a comment in the function description would be good.

Replicator/tests/ReplicatorCollectionTest.cc
```diff
@@ -178,87 +178,167 @@ TEST_CASE_METHOD(ReplicatorCollectionTest, "Use Zero Collections", "[Push][Pull]
...
+    vector<set<string>> _db2Before;
+};
+
 
 TEST_CASE_METHOD(ReplicatorCollectionTest, "Sync with Default Collection", "[Push][Pull]") {
     addCollDocs(db, Default, 10);
     addCollDocs(db2, Default, 10);
     
     SECTION("PUSH") {
+        CheckDBEntries check(db, db2, {Default});
```
pasin, line 248: It seems unnecessary to check as each test starts with all empty collections. If adding docs fails, test will fail. 

If you really think the test should check, move it outside SECTION as they are all the same.

Replicator/tests/ReplicatorLoopbackTest.hh
```diff
@@ -144,7 +144,7 @@ public:
         Log(">>> Replication complete (%.3f sec) <<<", st.elapsed());
         
         _checkpointIDs.clear();
-        for (int i = 0; i < opts1.collectionOpts.size(); ++i) {
+        for (int i = 0; i < optsRef1->collectionOpts.size(); ++i) {
```
pasin, line 147: Good catch!

Replicator/Inserter.cc
```diff
@@ -158,7 +158,17 @@ namespace litecore { namespace repl {
                 put.allocedBody = {(void*)bodyForDB.buf, bodyForDB.size};
 
                 // The save!!
-                auto doc = _db->insertionDB().useLocked()->putDocument(put, nullptr, outError);
+                C4CollectionSpec spec = replicator()->collection(collectionIndex())->getSpec();
+                Retained<C4Document> doc
+                =_db->insertionDB().useLocked<Retained<C4Document>>([spec, outError, &put](C4Database* db) {
+                    C4Collection* collection = db->getCollection(spec);
+                    if (!collection || !collection->isValid()) {
+                        return Retained<C4Document>();
```
pasin, line 166: This case also needs an out error as NotOpen. Would returning `nullptr` also work the same?

pasin (changes requested):

Replicator/tests/ReplicatorCollectionTest.cc
```diff
@@ -178,87 +178,167 @@ TEST_CASE_METHOD(ReplicatorCollectionTest, "Use Zero Collections", "[Push][Pull]
...
+        }
+    }
+    
+    ~CheckDBEntries() {
+        vector<set<string>> dbAfter;
+        vector<set<string>> db2After;
+        for (auto i = 0; i < _collSpecs.size(); ++i) {
+            dbAfter.push_back(getDocInfos(_db, _collSpecs[i]));
+            db2After.push_back(getDocInfos(_db2, _collSpecs[i]));
+            CHECK(dbAfter[i].size() == _dbBefore[i].size());
```
pasin, line 224: Not sure about the purpose of checking the same size here. For the case of push-pull at least, the size will not be the same.

Replicator/tests/ReplicatorCollectionTest.cc
pasin: I just realized your logic in CheckDBEntries.

Replicator/tests/ReplicatorCollectionTest.cc
pasin: A simple alternative to this class is a function to compare the end result at the end of the replication, whether same collection from the two dbs have the same docs or not, and check the number of docs.

Such as `compareCollections(roses, 20)` or `compareCollections({roses, tulips}, {10, 20})`

Replicator/Worker.cc
borrrden: I agree with this because a new smell is that the great majority of these function calls are ignoring the first returned argument from this function.  If it is being ignored more than used, something is wrong.

Replicator/Worker.cc
```diff
@@ -313,10 +313,37 @@ namespace litecore { namespace repl {
             _parent = nullptr;
     }
 
-    void Worker::setMsgCollection(blip::MessageBuilder& msg, CollectionIndex i) {
+    // Either there is error, or return a valid collection index
+    std::pair<CollectionIndex, slice>
+    Worker::checkCollectionOfMsg(const blip::MessageIn& msg, CollectionIndex i) const {
+        CollectionIndex collIn = getCollectionIndex(msg);
+        static slice error1 = "the collection property of the response does not match that of the request."_sl;
```
borrrden, line 320: Questionable variable naming tactics...

Replicator/Worker.hh
```diff
@@ -228,7 +228,18 @@ namespace litecore { namespace repl {
 
 #pragma mark - INSTANCE DATA:
     protected:
-        void setMsgCollection(blip::MessageBuilder& msg, CollectionIndex);
+        static CollectionIndex getCollectionIndex(const blip::MessageIn&  msgIn) {
+            long l = msgIn.intProperty(kCollectionProperty, kNotCollectionIndex);
+            Assert(l <= kNotCollectionIndex);
```
borrrden, line 233: If you are going to do this, it would be worth checking that l is also >= 0

Replicator/Worker.hh
```diff
@@ -228,7 +228,18 @@ namespace litecore { namespace repl {
...
+            return (CollectionIndex)l;
+        }
+        void assignCollectionToMsg(blip::MessageBuilder& msg, CollectionIndex i) const {
+            if (_options->collectionAware()) {
+                msg[kCollectionProperty] = i;
+            }
+        }
+        std::pair<CollectionIndex, slice>
```
borrrden, line 241: Whacky indentation problems

Replicator/Replicator.cc
```diff
@@ -1252,19 +1283,23 @@ namespace litecore { namespace repl {
...
 
-    void Replicator::handleConnectionMessage(Retained<blip::MessageIn> msg) {
-        slice profile = msg->property("Profile"_sl);
-        if (!profile)
-            return;
+    void Replicator::handleConnectionMessage(Retained<blip::MessageIn> request) {
+        slice profile = request->property("Profile"_sl);
+        Assert(profile);
+
+        auto [i, errMsg] = checkCollectionOfMsg(*request, kNotCollectionIndex);
```
borrrden, line 1290: Is `i` used?

borrrden (changes requested):

Replicator/Worker.cc
jianminzhao: getCollectionIndex and checkCollectionOfMsg are different. getCollectionIndex is at low level. It returns the index verbatim, except the handling of the nuance that the property is stored in message by long, but we are using it as unsigned. (typing here, I realized that I forgot to check the sign. I will change the assertion to
    Assert(0 <= l && l <= kNotCollectionIndex);
)
The value of kNotCollectionIndex will be interpreted at a higher level. For example, checkCollectionOfMsg() will never return kNotCollectionIndex unless there is an error. Another place it (kNotCollectionIndex) is interpreted is when the getCheckpoint message is received, as I try to tell whether it comes from 3.0 or 3.1 client.

jianminzhao: The reason I return error as slice rather than C4Error is because it will used in the context of request and response. For the former, we generate a local error as C4Error, but for that latter, we put the error in the response that we send back to the remote.

Replicator/Worker.cc
jianminzhao: yes, will add a comment to explain.

Replicator/Worker.hh
jianminzhao: Oh, great! We thought the same thing. See my earlier reply.

Replicator/Worker.hh
jianminzhao: Yes, I myself was changing it numerous times, I was trying to avoid too long a line. If you have suggestion, I am all ears.

Replicator/tests/ReplicatorCollectionTest.cc
jianminzhao: I will come to that when working on Push and Pull. Right now, the counts do not match yet.

Replicator/Worker.cc
borrrden: The issue for me is the lack of use of the first argument.  More often it is simply "ignore" which should be the exception rather than the norm.  The code should not just work, it should also be easy to reason about.  If we, as reviewers, have this many questions then what will an outside person trying to understand the code think?

Replicator/Worker.hh
borrrden: Nevermind, my screen is smaller than usual and it was adding its own line breaks which made it look different than I thought.  No issue here.

Replicator/Replicator.cc
borrrden: I don't see what line that is on.

borrrden: Ok I see it now!

borrrden (approved):
I still believe there is some tech debt being added here with regards to the suspicious return value of the function getting the collection index / error.  If it is functional though, I think it will be good enough to keep it in mind for optimization post merge.

Replicator/Worker.cc
jianminzhao: Oh, just noticed it. I think you mean argument 'i' of CollectionIndex. I take it and will make the following changes in the next PR: changing it to two overloaded functions: one having 1 argument, and the other having 2 arguments. This better reflects the meaning of the functions.

PR #1524, opened 2022-07-26
CBL-3195: Mangle table names for case sensitivity

borrrden:
SQLite is case insensitive, so the algorithm is a simple one, add a backslash before any capital letter in a collection name.  The reverse is also simple, since backslash is not an allowed character for a collection, the backslashes can simply be removed.


borrrden:
Please see if there are any areas that still need the name transformation that I missed.  I tried to cover everything I could think of in the test I wrote.

jianminzhao (approved):

PR #1525, opened 2022-07-27
CBL-3484: Don't require collection index in reply messages

borrrden:
Every reply message is paired with its original message, so there is no way for it to get mixed up.  It is safe to simply use the same collection index as the original.


pasin:
I think this also means that `checkCollectionOfMsg()` doesn't need reference collection index to be passed in.

borrrden:
Oh, you might be right!

borrrden:
Superceded by #1526

PR #1526, opened 2022-07-27
CBL-3467 : Replicator handles errors related to collections.

jianminzhao:
CBL-3484 : Do not check collection property in the reply messages.

### DIFF
--- a/Replicator/Replicator.cc
+++ b/Replicator/Replicator.cc
@@ -75,36 +75,42 @@ namespace litecore { namespace repl {
     ,_connectionState(connection().state())
     ,_docsEnded(this, "docsEnded", &Replicator::notifyEndedDocuments, tuning::kMinDocEndedInterval, 100)
     {
-        // Post-conditions:
-        //   collectionOpts.size() > 0
-        //   collectionAware == false if and only if collectionOpts.size() == 1 &&
-        //                                           collectionOpts[0].collectionPath == defaultCollectionPath
-        //   isActive == true ? all collections are active
-        //                    : all collections are passive.
-        _options->verify();
-        
-        _loggingID = string(db->getPath()) + " " + _loggingID;
-        _importance = 2;
+        try {
+            // Post-conditions:
+            //   collectionOpts.size() > 0
+            //   collectionAware == false if and only if collectionOpts.size() == 1 &&
+            //                                           collectionOpts[0].collectionPath == defaultCollectionPath
+            //   isActive == true ? all collections are active
+            //                    : all collections are passive.
+            _options->verify();
 
-        logInfo("%s", string(*options).c_str());
+            _loggingID = string(db->getPath()) + " " + _loggingID;
+            _importance = 2;
 
-        _remoteURL = webSocket->url();
-        if (_options->isActive()) {
-            prepareWorkers();
+            logInfo("%s", string(*options).c_str());
+
+            _remoteURL = webSocket->url();
+            if (_options->isActive()) {
+                prepareWorkers();
+            }
+
+            // Following messages are handled by appropriate workers.
+            // Replicator receives all the messages. Based on collectionIndex,
+            // it dispatches the message to appropriate workers.
+            for (auto profile : {"subChanges", "getAttachment", "proveAttachment", // passive pushers
+                    "changes", "proposeChanges", "rev", "norev"                    // passive pullers
+            }) {
+                registerHandler(profile, &Replicator::handleConnectionMessage);
+            }
+
+            registerHandler("getCheckpoint",    &Replicator::handleGetCheckpoint);
+            registerHandler("setCheckpoint",    &Replicator::handleSetCheckpoint);
+            registerHandler("getCollections",   &Replicator::handleGetCollections);
+        } catch (...) {
+            // terminate to break the circular references: connection -> BLIPIO -> connection.
+            terminate();
+            throw;
         }
-
-        // Following messages are handled by appropriate workers.
-        // Replicator receives all the messages. Based on collectionIndex,
-        // it dispatches the message to appropriate workers.
-        for (auto profile : {"subChanges", "getAttachment", "proveAttachment", // passive pushers
-                "changes", "proposeChanges", "rev", "norev"                    // passive pullers
-        }) {
-            registerHandler(profile, &Replicator::handleConnectionMessage);
-        }
-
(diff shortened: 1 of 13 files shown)


borrrden (changes requested):

Replicator/Replicator.cc
```diff
@@ -851,7 +849,8 @@ namespace litecore { namespace repl {
                     SubReplicator& sub = _subRepls[i];
                     Dict dict = checkpointArray[i].asDict();
                     if (!dict) {
-                        auto error = C4Error::printf(LiteCoreDomain, kC4ErrorNotFound,
+                        // Make it fatal error, UnexpectedError.
+                        auto error = C4Error::printf(LiteCoreDomain, kC4ErrorUnexpectedError,
```
pasin, line 853: Basically the collection is not found on the remote side. I think NotFound is suitable unless NotFound is not a permanent error?

Replicator/Replicator.cc
```diff
@@ -1101,7 +1092,7 @@ namespace litecore { namespace repl {
         if (!checkpointID)
             return;
 
-        auto [collectionIn, errMsg] = checkCollectionOfMsg(*request, kNotCollectionIndex);
+        auto [collectionIn, errMsg] = checkCollectionOfMsg(*request);
```
pasin, line 1095: For some reason, I don't see collectionIn is used.

borrrden:
I tested this with SG and it is functional, so mostly just the logical points above.

Replicator/Replicator.cc
jianminzhao: This is not the reason I changed the error code. It is because I have to stop the replicator and NotFound won't stop the  replicator.

Replicator/Replicator.cc
jianminzhao: Good catch. It will cause gcc to think it's unused error. I will make sure of it.

Replicator/Replicator.cc
```diff
@@ -1244,7 +1241,7 @@ namespace litecore { namespace repl {
                 C4Collection* c = db->getCollection(Options::collectionPathToSpec(
                                                     _options->collectionOpts[i].collectionPath));
                 if (c == nullptr) {
-                    error::_throw(error::UnexpectedError);
+                    error::_throw(error::NotFound);
```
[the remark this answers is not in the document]
jianminzhao: will do it

Replicator/Replicator.cc
pasin: Do you mean the replicator keeps retrying?

Replicator/Replicator.cc
```diff
@@ -1063,14 +1057,15 @@ namespace litecore { namespace repl {
...
-            auto [checked, err] = checkCollectionOfMsg(*request, kNotCollectionIndex);
+            auto [checked, err] = checkCollectionOfMsg(*request);
             if (err) {
                 request->respondWithError({"HTTP"_sl, 400, err});
                 return;
             }
             collectionIn = checked;
         }
+        // We don't save checkpointID by collection.
+        (void)collectionIn;
```
borrrden, line 1068: Then why bother returning it?

borrrden (approved):
The first argument of std::pair returned from checkCollectionMessage now seems to again be ignored more than used, which is a smell.

Replicator/Replicator.cc
jianminzhao: I don't know whether you mean why the function returns it or why in this particular call, why I use a variable to receive the returned value.
1. The reason the function returns it is that it is used in other places. Although textually it showed only once, most requests are routed by Replicator::handleConnectionMessages, where we this method is used to guarantee integrity and uniformity, whereby legacy absence of collection property is turned to index 0, etc. 
2. Second, I may choose to ignore the returned value at this place. The reason I used it here is to emphasize the protocol. The protocol requires a valid collection index, so as to allow the server save the checkpointID in place indexable by the collection index. I have the impression that SG uses the collection index. (I may be wrong.) But my current understanding is that the client is obliged to provide the collection index.

Replicator/Replicator.cc
jianminzhao: yes

Replicator/Worker.hh
```diff
@@ -247,13 +247,7 @@ namespace litecore { namespace repl {
...
         //   errorSlice != nullslice || (0 <= collectionIndex < _options->workingCollectionCount()
         // 'errorSlice' describes the nature of the violation.
-        // It may be used in two contexts. When handling a reqeust, the collection index must
-        // be in the range of 0, inclusive, and workingCollectionCount(), exclusive.
-        // When handling a response, we further check it against the collection index of
-        // the original request. This context is indicated by the second argument of
-        // CollectionIndex being not kNotCollectionIndex.
-        std::pair<CollectionIndex, slice> checkCollectionOfMsg(const blip::MessageIn& msg,
-                                                               CollectionIndex) const;
+        std::pair<CollectionIndex, slice> checkCollectionOfMsg(const blip::MessageIn& msg) const;
```
pasin, line 250: There is nothing wrong with the functionality, but again, I want to point out that I think we should avoid a function that returns two thing like this. There are standard ways of returning an error such as using exception or using an out error parameter. Also you could break the function into two as well if sometime the CollectionIndex is not used such as :

`bool validateCollectionIndex(CollectionIndex index, C4Error* error)`
`int getCollectionIndex(const blip::MessageIn& msg,  C4Error* error)` // Will call validateCollectionIndex() 

For error, you can use anything that we are using in the other places in LiteCore for consistency. I don't know if there is a case before that we return only error message. If so, you can follow that pattern. With C4Error, you can set the domain and error message.

pasin (approved):

Replicator/Worker.hh
jianminzhao: Got you. I chose this way for the reason of expediency because it's a private utility method. In fact, we have many methods returning pairs in Core, if you search "auto ["
The reason I return an error msg is that there are two difference uses, one will use the message in Error, and the other will use it as part of response back to remote.

pasin: > Got you. I chose this way for the reason of expediency because it's a private utility method. In fact, we have many methods returning pairs in Core, if you search "auto ["

I checked that but those are differences as they don't returns both result and error in pair.

PR #1527, opened 2022-07-29
CBL-3492 : Add C API test for syncing a named collection

pasin:
* Add a test for  CBL-3492, which has a fix in https://github.com/couchbase/couchbase-lite-core-EE/pull/24.

* Moved getCollection() and addCollDocs() from ReplicatorCollectionTest to c4Test, and renamed addCollDocs() to addDocs().

### DIFF
--- a/C/tests/c4Test.cc
+++ b/C/tests/c4Test.cc
@@ -328,6 +328,36 @@ void C4Test::deleteAndRecreateDB(C4Database* &db) {
 }
 
 
+/*static*/ C4Collection* C4Test::createCollection(C4Database* db, C4CollectionSpec spec) {
+    auto coll = c4db_createCollection(db, spec, ERROR_INFO());
+    Assert(coll);
+    return coll;
+}
+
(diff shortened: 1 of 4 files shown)


pasin:
PR Validation on Mac Mini is hanging.

C/tests/c4Test.cc
```diff
@@ -328,6 +328,36 @@ void C4Test::deleteAndRecreateDB(C4Database* &db) {
 }
 
 
+/*static*/ C4Collection* C4Test::createCollection(C4Database* db, C4CollectionSpec spec) {
+    auto coll = c4db_createCollection(db, spec, ERROR_INFO());
+    Assert(coll);
```
borrrden, line 333: Use REQUIRE instead of Assert?

borrrden (changes requested):
I think there is no reason to use Assert instead of REQUIRE where I pointed out (and one more location near it)

borrrden (approved):

PR #1528, opened 2022-07-29
CBL-501: Incorporate directional information in checkpoint ID

borrrden:
Since these are all new checkpoints they can change arbitrarily, so make all new checkpoints SHA-256 as well

### DIFF
--- a/Crypto/SecureDigest.cc
+++ b/Crypto/SecureDigest.cc
@@ -17,6 +17,7 @@
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wdocumentation-deprecated-sync"
 #include "mbedtls/sha1.h"
+#include "mbedtls/sha256.h"
 #pragma clang diagnostic pop
 
 #ifdef __APPLE__
@@ -33,19 +34,23 @@
 namespace litecore {
 
     void SHA1::computeFrom(fleece::slice s) {
-        (SHA1Builder() << s).finish(&bytes, sizeof(bytes));
+        (SHA1Builder() << s).finish(&_bytes, size());
     }
 
+    void SHA256::computeFrom(fleece::slice s) {
+        (SHA256Builder() << s).finish(&_bytes, size());
+    }
 
-    bool SHA1::setDigest(fleece::slice s) {
-        if (s.size != sizeof(bytes))
+    template<unsigned int N>
+    bool Hash<N>::setDigest(fleece::slice s) {
+        if (s.size != size())
             return false;
-        memcpy(bytes, s.buf, sizeof(bytes));
+        memcpy(_bytes, s.buf, size());
         return true;
     }
 
-
-    std::string SHA1::asBase64() const {
+    template<unsigned int N>
+    std::string Hash<N>::asBase64() const {
         return fleece::base64::encode(asSlice());
     }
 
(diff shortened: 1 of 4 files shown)


Replicator/Checkpointer.cc
```diff
@@ -234,9 +235,18 @@ namespace litecore { namespace repl {
...
-            enc.writeString(_collection->getSpec().name);
-            enc.writeString(_collection->getSpec().scope);
+            auto spec = _collection->getSpec();
+            auto index = _options->collectionSpecToIndex().at(spec);
+            enc.writeString(spec.name);
+            enc.writeString(spec.scope);
+
+            // CBL-501: Push only and pull only checkpoints create conflict
+            // So include them in the derivation
+            enc.writeBool(_options->pull(narrow_cast<unsigned int>(index)) != kC4Disabled);
```
jianminzhao, line 247: Why not to use CollectionIndex, in lieu of unsigned int?

borrrden: Reasonable point

jianminzhao (approved):

Crypto/SecureDigest.cc
```diff
@@ -82,4 +87,49 @@ namespace litecore {
...
+        DebugAssert(resultSize == sizeof(SHA256::_bytes));
+#ifdef USE_COMMON_CRYPTO
+        CC_SHA256_Final((uint8_t*)result, _CONTEXT);
+#else
+        mbedtls_sha256_finish(_CONTEXT, (uint8_t*)result);
+        mbedtls_sha256_free(_CONTEXT);
+#endif
+    }
+
+    // It seems odd to need these since they are base classes
```
snej, line 130: I think this happens because you've defined Hash::asBase64() outside the template declaration. Maybe? Templates are mysterious. But it's not a problem...

Crypto/SecureDigest.cc
```diff
@@ -82,4 +87,49 @@ namespace litecore {
 #endif
     }
 
+#undef _CONTEXT
+#ifdef USE_COMMON_CRYPTO
+#include <CommonCrypto/CommonDigest.h>
+#define _CONTEXT ((CC_SHA256_CTX*)_context)
```
snej, line 93: Is this macro necessary? All its uses below are already inside `#ifdef USE_COMMON_CRYPTO`, so the abstraction isn't needed.

Crypto/SecureDigest.hh
```diff
@@ -63,15 +72,52 @@ namespace litecore {
...
+    private:
+        friend class SHA256Builder;
+    };
+
+    /// Builder for creating SHA-1 digests from piece-by-piece data.
+    class SHA256Builder {
+    public:
+        SHA256Builder();
+
+        /// Add a single byte
```
snej, line 102: This comment and the next should be swapped

Replicator/Checkpointer.cc
```diff
@@ -233,9 +234,18 @@ namespace litecore { namespace repl {
         // Existing documents of are considered in the default collection
```
snej, line 234: existing documents of what?

Replicator/Checkpointer.cc
```diff
@@ -233,9 +234,18 @@ namespace litecore { namespace repl {
...
         // and they should keep the same docID or current checkpointers would be
         // inaccessible.
+        bool useSha1 = true;
         if (_collection != nullptr && !(_collection->getSpec() == kC4DefaultCollectionSpec)) {
```
snej, line 238: `_collection->getSpec() != kC4DefaultCollectionSpec` would be clearer

snej (approved):
Looks good, just a few little nits (some of which predate this commit, I realize)

Crypto/SecureDigest.cc
borrrden: `_CONTEXT` is defined differently for CommonCrypto vs mbed.  I just reused what was happening with SHA1

Replicator/Checkpointer.cc
borrrden: Documents that existed prior to a 3.1 upgrade

PR #1535, opened 2022-08-03
CBL-3526 : Continuous Pull Replicator went to stopped instead of idle

jianminzhao:
The bug is caused by a mistake when migrating to collection-aware.


pasin (approved):
Also confirmed that this PR fixes the test fails.

PR #1538, opened 2022-08-04
CBL-3514: Fix bug in collection case sensitivity

borrrden:
Queries didn't mangle the collection name which led to upper case giving a collection not found error

### DIFF
--- a/LiteCore/Storage/SQLiteDataFile.cc
+++ b/LiteCore/Storage/SQLiteDataFile.cc
@@ -685,8 +685,9 @@ namespace litecore {
             deletedTableName += name.substr(4);
             checkName = &deletedTableName;
         }
+        string finalName = SQLiteKeyStore::transformCollectionName(*checkName, true);
         string sql;
-        return getSchema(*checkName, "table", *checkName, sql);
+        return getSchema(finalName, "table", finalName, sql);
     }
 
 
@@ -813,16 +814,16 @@ namespace litecore {
     }
 
     string SQLiteDataFile::FTSTableName(const string &onTable, const string &property) const {
-        return onTable + "::" + property;
+        return onTable + "::" + SQLiteKeyStore::transformCollectionName(property, true);
     }
 
     string SQLiteDataFile::unnestedTableName(const string &onTable, const string &property) const {
-        return onTable + ":unnest:" + property;
+        return onTable + ":unnest:" + SQLiteKeyStore::transformCollectionName(property, true);
     }
 
 #ifdef COUCHBASE_ENTERPRISE
     string SQLiteDataFile::predictiveTableName(const string &onTable, const std::string &property) const {
-        return onTable + ":predict:" + property;
+        return onTable + ":predict:" + SQLiteKeyStore::transformCollectionName(property, true);
     }
 #endif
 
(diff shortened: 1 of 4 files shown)


LiteCore/Storage/SQLiteDataFile.cc
```diff
@@ -796,7 +796,7 @@ namespace litecore {
         if (collection == "_" || slice(collection) == KeyStore::kDefaultCollectionName || slice(collection) == KeyStore::kDefaultFullCollectionName)
             name += kDefaultKeyStoreName;
         else {
-            string candidate = name + string(KeyStore::kCollectionPrefix) + collection;
+            string candidate = name + string(KeyStore::kCollectionPrefix) + SQLiteKeyStore::transformCollectionName(collection, true);
```
jianminzhao, line 799: Why not hide the case mangling inside tableExists, before it hits SQLite table query? The case-mangling is a quirky thing owing to SQLite.

pasin: I think the current fix seems to appropriate to me as the function itself needs to return the table name as well.

pasin (approved):

LiteCore/Storage/SQLiteDataFile.cc
jianminzhao: " tableExists() will need to return ..." -- I assume you meant "collectionTableName() will ..."
It is opposite: with this change, collectionTableName() will return mangled table name.

LiteCore/Storage/SQLiteKeyStore.cc
```diff
@@ -259,13 +259,16 @@ namespace litecore {
 
     string SQLiteKeyStore::transformCollectionName(const string& name, bool mangle) {
         ostringstream ss;
-        for (uint8_t c : name) {
+        const char* name_cstr = name.c_str();
+        while(char c = *name_cstr++) {
+            if (c == '\\') {
+                continue;
```
jianminzhao, line 265: Is it not idempotent? Applying it to "Table" twice would generate R"(\\Table)" ?  (I noticed this change does not make it non-idempotent.

jianminzhao (changes requested):

LiteCore/Storage/SQLiteKeyStore.cc
borrrden: I did this on purpose because

A. We have no need to doubly encode things since backslash is not a valid character for the user.
B. There are so many places that implicitly pass through this area now I felt it was safer to leave the name as-is if it was already processed.

jianminzhao: So, your assumption here is, if we hit the above "continue", the argument mangle must be false. Assertion here?

borrrden: Not quite.  If mangle is true, the backslash is skipped, but then a capital letter is encountered so it is re-added

jianminzhao: Oh, right. My bad. De-mangle; if mangle then mangle. Perfect.

jianminzhao (approved):

PR #1543, opened 2022-08-10
CBL-3512: Fix some bugs regarding mismatched collections

borrrden:
For example, if the active side has more collections in the list than the passive one there needs to be a check that we are not running off the end of the passive side collection list.  Also clean up the code a bit to try to make it approachable for anyone seeing it for the first time.

Add in some mismatched collection test scenarios.

### DIFF
--- a/Replicator/Replicator.cc
+++ b/Replicator/Replicator.cc
@@ -598,9 +598,7 @@ namespace litecore { namespace repl {
 
         std::for_each(_subRepls.begin(), _subRepls.end(),
                       [](SubReplicator& sub) {
-            if (sub.checkpointer) {
-                sub.checkpointer->stopAutosave();
-            }
+            sub.checkpointer->stopAutosave();
         });
 
         // Clear connection() and notify the other agents to do the same:
@@ -1248,6 +1246,7 @@ namespace litecore { namespace repl {
                 C4Collection* c = db->getCollection(Options::collectionPathToSpec(
                                                     _options->collectionPath(i)));
                 if (c == nullptr) {
+                    _subRepls.clear();
                     error::_throw(error::NotFound, "collection %s is not found in the database.",
                                   _options->collectionPath(i).asString().c_str());
                 }
--- a/Replicator/Replicator.cc
+++ b/Replicator/Replicator.cc
@@ -183,7 +183,7 @@ namespace litecore { namespace repl {
                                                         info.flags & kDocDeleted,
                                                         false,
                                                         sub.collection->getSpec(),
-                                                        _options->collectionOpts[i].callbackContext));
+                                                        _options->callbackContextAt(collectionIndex())));
                     rev->error = C4Error::make(LiteCoreDomain, kC4ErrorConflict);
                     _docsEnded.push(rev);
                     ++nConflicts;
@@ -813,7 +813,7 @@ namespace litecore { namespace repl {
         enc.writeKey("collections"_sl);
         enc.beginArray();
         for (int i = 0; i < _subRepls.size(); i++) {
-            enc.writeString(_options->collectionOpts[i].collectionPath);
+            enc.writeString(_options->collectionPath(i));
         }
         enc.endArray();
         enc.endDict();
@@ -847,7 +847,7 @@ namespace litecore { namespace repl {
                 // Validate and read each checkpoints:
                 vector<Checkpoint> remoteCheckpoints(_subRepls.size());
                 for (int i = 0; i < _subRepls.size(); i++) {
-                    auto collPath = _options->collectionOpts[i].collectionPath;
+                    auto collPath = _options->collectionPath(i);
                     SubReplicator& sub = _subRepls[i];
                     Dict dict = checkpointArray[i].asDict();
                     if (!dict) {
@@ -1176,7 +1176,7 @@ namespace litecore { namespace repl {
             }
         }
 
-        // Rearrange _options->collectionOpts according to collSpecs. If i-th CollectionSpec
+        // Create options->workingCollections according to collSpecs. If i-th CollectionSpec
         // is not found, put an empty collectionOptions at i-th position.
         _options->rearrangeCollections(collSpecs);
         MessageBuilder response(request);
@@ -1190,7 +1190,7 @@ namespace litecore { namespace repl {
             logInfo("Request to get peer checkpoint '%.*s' for collection '%.*s'",
                     SPLAT(checkpointID), SPLAT(collections[i].asString()));
 
-            if (!_options->collectionOpts[i].collectionPath) {
+            if (!_options->collectionPath(i)) {
                 logVerbose("Get peer checkpoint '%.*s' for collection '%.*s' : Collection Not Found in the Replicator's config",
                         SPLAT(checkpointID), SPLAT(collections[i].asString()));
                 enc.writeNull();
@@ -1246,10 +1246,10 @@ namespace litecore { namespace repl {
         _db->useLocked([this](Retained<C4Database>& db) {
             for (CollectionIndex i = 0; i < _options->workingCollectionCount(); ++i) {
                 C4Collection* c = db->getCollection(Options::collectionPathToSpec(
-                                                    _options->collectionOpts[i].collectionPath));
+                                                    _options->collectionPath(i)));
                 if (c == nullptr) {
                     error::_throw(error::NotFound, "collection %s is not found in the database.",
-                                  _options->collectionOpts[i].collectionPath.asString().c_str());
+                                  _options->collectionPath(i).asString().c_str());
                 }
                 _subRepls[i].collection = c;
             }
--- a/Replicator/Replicator.cc
+++ b/Replicator/Replicator.cc
@@ -598,7 +598,9 @@ namespace litecore { namespace repl {
 
         std::for_each(_subRepls.begin(), _subRepls.end(),
                       [](SubReplicator& sub) {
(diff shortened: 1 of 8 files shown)


Replicator/ReplicatorOptions.hh
```diff
@@ -303,24 +303,37 @@ namespace litecore { namespace repl {
...
+                    if (activeIndex < origCount) {
+                        // Assertion: collectionOpts[j].collectionPath == nullslice
+                        // We haven't run out of passive entries yet, so make sure
+                        // the nullslice one is in the correct place alongside
+                        // its active counterpart.  The spec to index dictionary will
+                        // point to the out of bounds, and unusable, original entry
+                        // for the spec.
+                        _collectionSpecToIndex[specs[activeIndex]] = passiveCount;
+                        std::swap(collectionOpts[activeIndex], collectionOpts[passiveCount]);
+                    }
```
jianminzhao, line 322: we need the "else",
else {
    auto path = collectionOpts[activeIndex].collectionPath;
    if (path) {
        auto spec = Options::collectionPathToSpec(path);
       _collectionSpecToIndex[spec] = passiveCount;
    }
}
 
Looking at this example,
activeCollections:  A B C
this->collections:  C
We cannot lose track of C

Replicator/ReplicatorOptions.hh
```diff
@@ -303,24 +303,37 @@ namespace litecore { namespace repl {
...
+                    if (foundEntry->second > activeIndex) {
+                        // The entry is out of place, so swap it with the one
+                        // at the current position
+                        C4CollectionSpec foundSpec = foundEntry->first;
+                        size_t           foundIndex = foundEntry->second;
+
+                        // Since we are swapping the collectionOpts position,
+                        // the spec to index map needs to be updated for the new
+                        // positions inside of collectionOpts
+                        _collectionSpecToIndex[specs[activeIndex]] = foundIndex;
```
jianminzhao, line 334: Same problem as before.
if (activeIndex < origCount) {
    _collectionSpecToIndex[specs[activeIndex]] = foundIndex;
else {
       auto path = collectionOpts[activeIndex].collectionPath;
      if (path) {
          auto spec = Options::collectionPathToSpec(path);
         _collectionSpecToIndex[spec] = foundIndex;
      }
}

jianminzhao (changes requested):

Replicator/Replicator.cc
```diff
@@ -598,7 +598,9 @@ namespace litecore { namespace repl {
 
         std::for_each(_subRepls.begin(), _subRepls.end(),
                       [](SubReplicator& sub) {
-            sub.checkpointer->stopAutosave();
+            if (sub.checkpointer) {
```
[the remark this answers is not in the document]
borrrden: Well I found a situation where it is empty, that's why I changed it.  If prepareWorkers throws an error (like if a collection doesn't exist for the passive side) then the checkpointer never gets created.

Replicator/ReplicatorOptions.hh
borrrden: This is so overly complicated it drives me crazy....can't we make this simpler?

Replicator/Replicator.cc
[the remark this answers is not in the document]
jianminzhao: already throw, 
                if (c == nullptr) {
                    error::_throw(error::NotFound, "collection %s is not found in the database.",
                                  _options->collectionOpts[i].collectionPath.asString().c_str());
                }
I guess because it threw, it did not proceed to the ensuing for loop, 
for (CollectionIndex i = 0; i < _options->workingCollectionCount(); ++i) {

Replicator/ReplicatorOptions.hh
jianminzhao: I think I can make it simpler by getting rid of "specs" altogether. I used it to avoid converting collection path to spec every time a swap is needed. At that time, I think, my mental model is that collections in passive is greater than active collections.

Replicator/Replicator.cc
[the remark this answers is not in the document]
jianminzhao: Perhaps a fix closer to the root-cause could be, before error::throw above, in Replicator::prepareWorkers(),
if (c == nullptr) {
  _subRepls.resize(0);
  error::_throw(...)
}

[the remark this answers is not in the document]
borrrden: Won't this miss the call to stopAutoSave on any already processed entries?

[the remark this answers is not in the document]
jianminzhao: No, we haven't done anything yet. throw will take us out before we do anything

Replicator/ReplicatorOptions.hh
```diff
@@ -298,32 +298,21 @@ namespace litecore { namespace repl {
...
-                        size_t           indx;
-                        std::tie(spec, indx) = *findIt;
-                        _collectionSpecToIndex[specs[i]] = indx;
-                        _collectionSpecToIndex[spec] = i;
-                        std::swap(collectionOpts[i], collectionOpts[indx]);
-                    }
+                    reordered.push_back(collectionOpts[foundEntry->second]);
                 }
+
+                _collectionSpecToIndex[activeCollections[activeIndex]] = activeIndex;
```
jianminzhao, line 312: When not matched, activeCollections[activeIndex] won't be in reordered and ultimately the updated collectionOpts, and, so, should be in the map.

Replicator/ReplicatorOptions.hh
```diff
@@ -298,32 +298,21 @@ namespace litecore { namespace repl {
...
             }
+
+            collectionOpts = reordered;
```
jianminzhao, line 315: vector can be moved.

Replicator/ReplicatorOptions.hh
borrrden: I didn't follow what you want to change.

jianminzhao: _collectionSpecToIndex maps spec to index in vector collectionOpts. When not matched, activeCollections[activeIndex] won't exist in collectionOpts. Besides, the map has the specs that were in original collectionOpts, but are dropped out of "reordered". So they should be removed from the map.
With brutal-force, simplest is to recreate the index map.

jianminzhao:
I am little concerned about dropping the collection configs if they are matched by the active replicator, because code architecture allows to restart C4Replicator after Stopped. In that case, the original collection configs are gone. Basically, with this approach, we should not allow C4Replicator to start() second time.

Replicator/ReplicatorOptions.hh
borrrden: So....don't include them in that case?  It's sort of splitting hairs because the map will never even be used since the entire replication is invalid.

borrrden:
If they are to be restartable, then that makes me even more wary of rearrange to begin with and we might as well make a separate runtime list for them.

Replicator/ReplicatorOptions.hh
jianminzhao: correct.

Replicator/ReplicatorOptions.hh
```diff
@@ -248,39 +249,47 @@ namespace litecore { namespace repl {
...
+            return _workingCollections[i].pushFilter;
         }
 
-        Validator pullFilter(CollectionIndex i) const  {
-            return collectionOpts[i].pullFilter;
+        Validator pullFilter(CollectionIndex i) const {
+            return _workingCollections[i].pullFilter;
         }
 
         void* collectionCallbackContext(CollectionIndex i) const {
```
jianminzhao, line 271: This appears to be the same as as callbackContextAt().

Replicator/ReplicatorOptions.hh
```diff
@@ -298,22 +307,18 @@ namespace litecore { namespace repl {
             DebugAssert(!_isActive);
 
             size_t origCount = collectionOpts.size();
-            decltype(collectionOpts) reordered;
-            reordered.reserve(origCount);
+            _workingCollections.clear();
+            _workingCollections.reserve(origCount);
```
jianminzhao, line 311: reserve to activeCollections.size()

Replicator/ReplicatorOptions.hh
```diff
@@ -298,22 +307,18 @@ namespace litecore { namespace repl {
             DebugAssert(!_isActive);
 
             size_t origCount = collectionOpts.size();
-            decltype(collectionOpts) reordered;
-            reordered.reserve(origCount);
+            _workingCollections.clear();
```
jianminzhao, line 310: also clear _collectionSpecToIndex

Replicator/ReplicatorOptions.hh
borrrden: Oh I didn't notice it because so many places were going through collectionOpts directly!

jianminzhao: Perhaps it's time to make collectionOpts private, _collectionOpts. It also helps find out mistaken use of it in this file.
The exiting style is, all public properties are direct members.

Replicator/Replicator.cc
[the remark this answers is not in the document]
jianminzhao: Are you going to do this?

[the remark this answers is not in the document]
borrrden: I swear I did this and now it's gone, but yes....

jianminzhao (approved):

PR #1544, opened 2022-08-11
CBL-3489 : excess BLIPIO warning messages of "couldn't find any messa…

jianminzhao:
…ges to process"

Removed the warning for it's not that useful.

### DIFF
--- a/Networking/BLIP/BLIPConnection.cc
+++ b/Networking/BLIP/BLIPConnection.cc
@@ -414,6 +414,9 @@ namespace litecore { namespace blip {
         void _onWebSocketMessages(int gen =actor::AnyGen) {
             auto messages = _incomingFrames.pop(gen);
             if (!messages) {
+                if (gen != actor::AnyGen) {
+                    warn("onWebSocketMessages couldn't find any messages to process");
+                }
                 return;
             }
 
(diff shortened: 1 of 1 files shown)


borrrden:
This is an early return, so I assert that SOMETHING should be logged.  If it's noisy it can be moved to a more verbose level. If it is still noisy, the correct course of action is to find out why.

borrrden (commented):
See my above comment

jianminzhao:
I studied why. There are two cases,
  void _onWebSocketMessages(int gen =actor::AnyGen) {
              auto messages = _incomingFrames.pop(gen);
              if (!messages) {
                  warn("onWebSocketMessages couldn't find any messages to process");
                  return;
              }
(1) if gen == AnyGen. which only occurs with _close(). Most of time, there is no messages in the queue. It is quite normal.
(2) if gen != AnyGen. This is the normal case when an item is pushed to ActorBatcher. In this case, it should never happen. It's almost like assertion failure.
When running the cpptests, all occurrences are due to (1). It's very normal the queue is empty.

jianminzhao:
Perhaps we should keep the warning when it is not called from BLIPIO::_close(), or, when gen != actor::AnyGen ?

borrrden (commented):
I'd definitely prefer to have something there.  Almost universally I want to follow the rule that if we return early from a method, we log something.  If it is too noisy then the conditions of the log can be altered, or alternatively if the return early is the more frequent path then log something if the long path is taken.  Just bread crumbs to let us easily follow the path of the code via logs.

borrrden (approved):

PR #1545, opened 2022-08-15
CBL-3501 : Add incremental continuous sync tests with collections

jianminzhao:


Replicator/tests/ReplicatorCollectionTest.cc
```diff
@@ -511,6 +511,93 @@ TEST_CASE_METHOD(ReplicatorCollectionTest, "Multiple Collections Incremental Pus
...
+    addDocs(db, Roses, 2, "db-Roses-");
+    addDocs(db, Tulips, 2, "db-Tulips-");
+    C4Collection* roses = c4db_getCollection(db, Roses, nullptr);
+    C4Collection* tulips = c4db_getCollection(db, Tulips, nullptr);
+    C4Collection* roses2 = c4db_getCollection(db2, Roses, nullptr);
+    C4Collection* tulips2 = c4db_getCollection(db2, Tulips, nullptr);
+
+    SECTION("PUSH") {
+        _expectedDocumentCount = 4;
+        runPushReplication({Roses, Tulips}, {Tulips, Lavenders, Roses});
```
pasin, line 534: The goal of CBL-3501 is to test incremental sync with continuous replicators. So I think this should start a continuous push replicator? The logic could be:

1. Add docs / revs. 
2. Start the thread that adds new revs or docs (Maybe also wait until the replicator is IDLE before start adding new revs to ensue to test collection change observer inside the replicator).
3. Start a continuous push replicator.

Same comment for the other 2 sections.

Replicator/tests/ReplicatorCollectionTest.cc
```diff
@@ -511,6 +511,93 @@ TEST_CASE_METHOD(ReplicatorCollectionTest, "Multiple Collections Incremental Pus
...
+    {}
+    ~Jthread() {
+        thread.join();
+    }
+};
+
+TEST_CASE_METHOD(ReplicatorCollectionTest, "Multiple Collections Incremental Revisions", "[Push][Pull]") {
+    addDocs(db, Roses, 2, "db-Roses-");
+    addDocs(db, Tulips, 2, "db-Tulips-");
+    C4Collection* roses = c4db_getCollection(db, Roses, nullptr);
```
pasin, line 527: There is `getCollection(db, Roses)` small util that also checks if the returned collection exists.

pasin (changes requested):

Replicator/tests/ReplicatorCollectionTest.cc
jianminzhao: You want the above sync to be Continuous, and wait it until it goes to idle. Then, start the following background thread. Right?

Replicator/tests/ReplicatorCollectionTest.cc
jianminzhao: I saw it. Thanks

Replicator/tests/ReplicatorCollectionTest.cc
pasin: Yes or you can do the wait in the background thread as well, which I think it will be easier according how runPushReplication() is written now.

borrrden (commented):
I don't think I am in the loop about what this test does, so I defer to Pasin.

Replicator/tests/ReplicatorLoopbackTest.hh
```diff
@@ -261,6 +263,9 @@ public:
             {
                 _statusReceived = status;
                 _checkStopWhenIdle();
+                if (_statusReceived.level == kC4Idle && _callbackWhenIdle) {
```
pasin, line 266: I think it should be called before _checkStopWhenIdle(), otherwise, there is a case that it will be called after the replicator is called to stop.

pasin (changes requested):

Replicator/tests/ReplicatorCollectionTest.cc
```diff
@@ -511,6 +511,96 @@ TEST_CASE_METHOD(ReplicatorCollectionTest, "Multiple Collections Incremental Pus
     validateCollectionCheckpoints(db, db2, 1, "{\"local\":22,\"remote\":24}");
 }
 
+struct Jthread {
```
pasin, line 514: I have just noticed that there is `static std::thread* runInParallel(std::function<void()> callback)` in ReplicatorLoopbackTest.hh. Can `runInParallel` is used instead of Jthread? BTW, what is `J` stand for?

pasin (changes requested):

Replicator/tests/ReplicatorCollectionTest.cc
jianminzhao: c++20 introduced a new class, std::jthread. It is automatically joined when deleted. I just mimic it before we move to c++20.

Replicator/tests/ReplicatorLoopbackTest.hh
jianminzhao: _checkStopWhenIdle() does not stop unless the caller (test case) requests with stopWhenIdle(). The caller is responsible for synchronizing between stopWhenIdle and _callbackWhenIdle. If both are requested, It's kind of UD (undefined behavior). If do callback first, the state is still Idle, and we will stop. That's usually not we wanted. In my tests, I request StopWhenIdle() inside the callback and then nullify _callbackWhenIdle.

Replicator/tests/ReplicatorCollectionTest.cc
pasin: I'm curious why don't use runInParallel() instead of writing a new JThread? For this test, the test will not finish until the code running in the separate thread is done anyway. I don't see the benefit of using JThread.

pasin (approved):

PR #1546, opened 2022-08-23
CBL-3602: Freeing observers after deleting collection was crashing

borrrden:
This is a 2 part fix in which

1. The observers now retain their collections, since there is no communication about a collection deletion to its observer.  This meant that deleting the collection left a dangling pointer there.
2. The sequence tracker inside of the observer notifiers is a pointer now so that it can be set to null in the case of collection invalidation.  Otherwise, the notifier destructor attempts to use the now invalid tracker inside the invalid collection and causes an exception and/or crash.

### DIFF
--- a/C/c4Observer.cc
+++ b/C/c4Observer.cc
@@ -32,7 +32,7 @@ namespace litecore {
         ,_callback(move(callback))
         {
             _collection->sequenceTracker().useLocked<>([&](SequenceTracker &st) {
-                _notifier.emplace(st,
+                _notifier.emplace(&st,
                                   [this](CollectionChangeNotifier&) {_callback(this);},
                                   since);
             });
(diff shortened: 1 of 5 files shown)


C/c4Observer.cc
```diff
@@ -32,14 +32,21 @@ namespace litecore {
...
+                _notifier.emplace(&st,
                                   [this](CollectionChangeNotifier&) {_callback(this);},
                                   since);
             });
         }
 
 
         ~C4CollectionObserverImpl() {
+            if (!_collection->isValid()) {
+                // HACK: If the collection is not valid anymore, the notifier tracker is probably
```
jianminzhao, line 44: Why is this? tracker is owned by C4CollectionImpl object. If the collection object is good (in c++ sense, namely it's deleted yet), tracker should also be good, or not? _collection->isValid() == false means it's relation with the underlying database is broken. Did you see its crashing after you retained the collection object?

borrrden: Yes, it will crash when destroying the notifier unless I do this.

borrrden: This is what I explained in the 2nd point in the PR comment.

jianminzhao: Oh, that's because you retained only in C4CollectionObserverImpl. The notifier doesn't hold the ref count. Notifier needs to retain it as well.

borrrden: It has nothing to do with the retain count of the collection.  When the collection is invalidated, the tracker is destroyed.

jianminzhao: Ah, I see, close() is called if it's not valid. We destruct all members before the collection object itself is deleted. In other words, only thing from retained collection object is query if it's valid. Hopefully, this, the tracker, is the only sub-object that is leaked out from the collection obj.

jianminzhao (approved):

PR #1548, opened 2022-08-24
Fix build errors in Jenkins build 3.1.0-219

jianminzhao:


borrrden:
Oh duh of course, I blame jet lag still!  Ideally those would have a REQUIRE afterwords to clear the ERROR_INFO

PR #1551, opened 2022-08-25
CBL-3538 : allow "_default.collection_name" as valid collection path.

jianminzhao:
Note that user-created scope or collection cannot be led by underscore.

Also fixed a bug when collection names include upper case letters.

### DIFF
--- a/LiteCore/Storage/SQLiteDataFile.cc
+++ b/LiteCore/Storage/SQLiteDataFile.cc
@@ -547,7 +547,8 @@ namespace litecore {
             
             // Create a SQLite view of a union of both stores, for use in queries:
 #define COLUMNS "key,sequence,flags,version,body,extra,expiration"
-            const char *cname = name.c_str();
+            string tableName = keyStore->tableName().substr(3); // remove prefix "kv_"
+            const char* cname = tableName.c_str();
             _exec(format("CREATE TEMP VIEW IF NOT EXISTS \"all_%s\" (" COLUMNS ") AS "
                          "SELECT " COLUMNS " from \"kv_%s\" UNION ALL "
                          "SELECT " COLUMNS " from \"kv_del_%s\"",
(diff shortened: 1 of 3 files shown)


LiteCore/tests/QueryTest.hh
```diff
@@ -40,20 +40,26 @@ protected:
...
             case 0:
-                collectionName = KeyStore::kDefaultCollectionName;
+                jump = alter3++ % 3;
+                collectionName = (jump == 0) ? KeyStore::kDefaultCollectionName
+                    : (jump == 1) ? "_" : "cbl_core_temp";
                 break;
             case 1:
-                collectionName = "secondary";
-                store = &db->getKeyStore(".secondary");
+                collectionName = (alter2++ % 2 == 0) ? "Secondary" : "_default.Secondary";
```
borrrden, line 61: Usually prefer & 1 rather than % 2 but for a test either is fine.

LiteCore/Storage/SQLiteDataFile.cc
```diff
@@ -547,7 +547,8 @@ namespace litecore {
             
             // Create a SQLite view of a union of both stores, for use in queries:
 #define COLUMNS "key,sequence,flags,version,body,extra,expiration"
-            const char *cname = name.c_str();
+            string tableName = keyStore->tableName().substr(3); // remove prefix "kv_"
```
borrrden, line 550: Fragile.  Would probably be better to use a static integer here based on the length of the prefix, rather than the magic number 3.

LiteCore/Storage/SQLiteDataFile.cc
```diff
@@ -773,6 +774,27 @@ namespace litecore {
         return ksName == kDefaultKeyStoreName || ksName.hasPrefix(KeyStore::kCollectionPrefix);
     }
 
+    namespace {
+        std::pair<slice, slice> splitCollectionPath(const string& collectionPath) {
+            const char* sdata = collectionPath.data();
+            slice collection {sdata, collectionPath.length()};
```
borrrden, line 780: No need for both of these lines, `slice collection(collectionPath)` is fine.

borrrden (approved):

LiteCore/Storage/SQLiteDataFile.cc
jianminzhao: Add comments. Currently, "kv_" is used literally everywhere it's used. Agreed, we'd better avoid the use of magic "kv_v".

PR #1556, opened 2022-08-30
CBL-3626: Eliminate hang when calling setExpiration in a transaction

borrrden:
Part of this involved making sure that the input collection was valid, and so extended that protection to (almost) the entire C API involving collections

:note: There is a weird issue on Linux with calling enqueue from the Housekeeper constructor.  If there is an enqueue call there then `_housekeeper = new Housekeeper(this)` in CollectionImpl will end up somehow with a corrupted object (invalid ref count even though there is literally no place in there which supposedly touches the ref count).  This causes a crash in `_careful_retain`.  The only choice was to move the logic later, and retain the collection as a Housekeeper member until that time.

### DIFF
--- a/C/c4CAPI.cc
+++ b/C/c4CAPI.cc
@@ -33,6 +33,12 @@ using namespace fleece;
 using namespace litecore;
 
 
+#define checkCollectionValid(c, err, ret) if(_usuallyFalse(!c4coll_isValid(c))) { \
+    *err = c4error_make(LiteCoreDomain, kC4ErrorNotOpen, "Invalid collection: either deleted, or db closed"_sl); \
+    return ret; \
+}
+
+
 #pragma mark - BLOBS:
 
 
@@ -278,19 +284,21 @@ bool c4coll_isValid(C4Collection* coll) noexcept {
 }
 
 C4CollectionSpec c4coll_getSpec(C4Collection *coll) noexcept {
+    // Unlike the others, this continues to return valid data even
+    // after invalidation, so skip the validity check
     return coll->getSpec();
 }
 
 C4Database* c4coll_getDatabase(C4Collection *coll) noexcept {
-    return coll->isValid() ? coll->getDatabase() : nullptr;
+    return _usuallyTrue(c4coll_isValid(coll)) ? coll->getDatabase() : nullptr;
 }
 
 uint64_t c4coll_getDocumentCount(C4Collection *coll) noexcept {
-    return tryCatch<uint64_t>(nullptr, [=]{return coll->getDocumentCount();});
+    return tryCatch<uint64_t>(nullptr, [=]{return _usuallyTrue(c4coll_isValid(coll)) ? coll->getDocumentCount() : 0;});
 }
 
 C4SequenceNumber c4coll_getLastSequence(C4Collection *coll) noexcept {
-    return tryCatch<sequence_t>(nullptr, [=]{return coll->getLastSequence();});
+    return tryCatch<sequence_t>(nullptr, [=]{return _usuallyTrue(c4coll_isValid(coll)) ? coll->getLastSequence() : 0_seq;});
 }
 
 C4Document* c4coll_getDoc(C4Collection *coll,
@@ -299,6 +307,7 @@ C4Document* c4coll_getDoc(C4Collection *coll,
                           C4DocContentLevel content,
                           C4Error* C4NULLABLE outError) noexcept
 {
+    checkCollectionValid(coll, outError, nullptr);
     return tryCatch<C4Document*>(outError, [&]{
         Retained<C4Document> doc = coll->getDocument(docID, mustExist, content);
         if (!doc)
@@ -311,6 +320,7 @@ C4Document* c4coll_getDocBySequence(C4Collection *coll,
                                     C4SequenceNumber sequence,
                                     C4Error* C4NULLABLE outError) noexcept
 {
+    checkCollectionValid(coll, outError, nullptr);
     return tryCatch<C4Document*>(outError, [&]{
         auto doc = coll->getDocumentBySequence(sequence);
         if (!doc)
@@ -324,6 +334,7 @@ C4Document* c4coll_putDoc(C4Collection *coll,
                           size_t * C4NULLABLE outCommonAncestorIndex,
                           C4Error* C4NULLABLE outError) noexcept
 {
+    checkCollectionValid(coll, outError, nullptr);
     return tryCatch<C4Document*>(outError, [&]{
         return coll->putDocument(*rq, outCommonAncestorIndex, outError).detach();
     });
@@ -335,6 +346,7 @@ C4Document* c4coll_createDoc(C4Collection *coll,
                              C4RevisionFlags revFlags,
                              C4Error* C4NULLABLE outError) noexcept
 {
+    checkCollectionValid(coll, outError, nullptr);
     return tryCatch<C4Document*>(outError, [&]{
         return coll->createDocument(docID, revBody, revFlags, outError).detach();
     });
@@ -346,6 +358,7 @@ bool c4coll_moveDoc(C4Collection *coll,
                     C4String newDocID,
                     C4Error* C4NULLABLE outError) noexcept
 {
+    checkCollectionValid(coll, outError, false);
     return tryCatch(outError, [&]{
         coll->moveDocument(docID, toCollection, newDocID);
     });
@@ -355,6 +368,7 @@ bool c4coll_purgeDoc(C4Collection *coll,
                      C4String docID,
                      C4Error* C4NULLABLE outError) noexcept
 {
+    checkCollectionValid(coll, outError, false);
     try {
         if (coll->purgeDocument(docID))
             return true;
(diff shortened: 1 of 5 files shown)


borrrden:
@snej Do you have any ideas about what I am seeing with the enqueue call in the constructor.  I have some kind of paranoid theory that double inheritance is causing it to behave badly on Linux thereby destroying the object when enqueue is called.

C/c4CAPI.cc
```diff
@@ -33,6 +33,12 @@ using namespace fleece;
 using namespace litecore;
 
 
+#define checkCollectionValid(c, err, ret) if(_usuallyFalse(!c4coll_isValid(c))) { \
```
jianminzhao, line 36: Any reason to use #define instead of static inline?

C/c4CAPI.cc
```diff
@@ -278,19 +284,21 @@ bool c4coll_isValid(C4Collection* coll) noexcept {
...
 
 C4CollectionSpec c4coll_getSpec(C4Collection *coll) noexcept {
+    // Unlike the others, this continues to return valid data even
+    // after invalidation, so skip the validity check
     return coll->getSpec();
 }
 
 C4Database* c4coll_getDatabase(C4Collection *coll) noexcept {
-    return coll->isValid() ? coll->getDatabase() : nullptr;
+    return _usuallyTrue(c4coll_isValid(coll)) ? coll->getDatabase() : nullptr;
```
jianminzhao, line 293: Won't coll->getDatabase() throw if !coll->isValue(). ?

C/c4CAPI.cc
```diff
@@ -299,6 +307,7 @@ C4Document* c4coll_getDoc(C4Collection *coll,
                           C4DocContentLevel content,
                           C4Error* C4NULLABLE outError) noexcept
 {
+    checkCollectionValid(coll, outError, nullptr);
```
jianminzhao, line 310: Not clear why is this line necessary? Won't coll->getDocument throw if not valid?

C/c4CAPI.cc
borrrden: Want it to return from the method it is placed in.  You cannot do that with a static inline.

C/c4CAPI.cc
borrrden: That would be a bad move.  This method is marked noexcept.  Actually in this case it will just return null anyway, but for consistency sake I used this.

C/c4CAPI.cc
borrrden: Again, we cannot throw anything here, this method is marked noexcept!

borrrden: Nevermind, I see that you mean that the following will throw and be caught.  I also want to check for null here though.

C/c4CAPI.cc
jianminzhao: yes, we want noexcept, but if you catch it as all other c4 functions do, would it not work? It seems to be double protections and defeats the original design of C4Collection

C/c4CAPI.cc
jianminzhao: The cpp counterpart, coll->getDocument(), does not throw, why do we make it error here? 
There are two cases we get nullptr return. A. coll itself is invalid. In this case the underlying method throws, and we report to the C API with error. B. the doc to be looked for does not exist in the collection. In this case why make it error as well? I understand in one case the error is NotOpen and the other case NotFound.

borrrden: I don't know what you are talking about.  I want to check if `coll` is null, to void a SIGSEGV.  I figured if I was doing that much I might as well check early for an invalid collection as well.

C/c4CAPI.cc
borrrden: There is nothing to actually catch in this case, but I also want to protect against a null `coll` to avoid a SIGSEGV

C/c4CAPI.cc
jianminzhao: I see your point: you want to check if coll is null. My impression is that the style of c4 api, such as,
   c4coll_xyz(C4Collection* coll ...) 
is similar to an alias, coll->xyz(...). or use of C pointer for C++ implementation. The assumption is that the caller is responsible for using a non-null pointer. 
I checked c4doc_, c4db_, etc. and almost in all occurrences there is no checking pointer. In fact, in header file, we enclose all declarations with C4_ASSUME_NONNULL_BEGIN/C4_ASSUME_NONNULL_END, meaning all pointers are assumed to be non-null unless otherwise specified.

C/c4CAPI.cc
snej: Maybe rename it to make that clearer, since it's unusual behavior? Like, `returnIfCollectionInvalid`?

snej:
> @snej Do you have any ideas about what I am seeing with the enqueue call in the constructor. 

It's not safe for a RefCounted subclass's constructor to do anything that will cause itself to be retained/released. That's because it starts out with a 0 refcount -- it doesn't have any references until the constructor returns and the instance is assigned to a Retained object that holds it. If the constructor calls something that has a side effect of first retaining then releasing itself, that won't be a no-op -- the release will actually free the new instance.

jianminzhao:
@snej this becomes a problem only if the time span of Retain is completely inside the constructor, before another object has a chance to retain it. Right?

snej:
> @snej this becomes a problem only if the time span of Retain is completely inside the constructor, before another object has a chance to retain it. Right?

It happens only if there's a retain followed by a release while still in the constructor. It's OK if something retains the new object and doesn't release it.

C/c4CAPI.cc
borrrden: Those assumptions are useless in a language that doesn't understand them, like .NET.  C# has no visibility and so I think it would be better to return back an error that it can use as a .NET exception rather than simply hard crashing.

borrrden: In fact in *our own* deprecated API we make no attempt to check if the default collection is null or not before universally sending it to these APIs.

jianminzhao: Understood that the compiler directives, such as C4_ASSUME_NONNULL_BEGIN. does not help in runtime, but it is part of APIs, telling the reader (user) how to use the functions (passing the arguments). The vast majority of the functions expect that and therefore don't check for NULL. If there is an unchecked NULL, we should fix the code that passed NULL to the function, rather than reporting error that Platform code would have to mind it with every invocation.

borrrden: Yes, which is exactly why it's useless for bound languages.  What is going to happen is that the program will simply stop and say "exited with code <some meaningless code>" with no information on what went wrong.  This sinks a lot of wasted time trying to figure out what happened in a CI build or a user filed bug report.  You can argue that it is "just a null reference error" but that doesn't convince me because if you do something like `c4coll_XXX(c4db_getDefaultCollection(db, err))` the fact that the collection is null means that it is an invalid collection because it doesn't exist.  I am trying to make the library more easy to debug so we don't just get the equivalent of "my code crashed, here is a non-symbolicated stack trace" as a bug report.  Developers using a managed language are not likely to know the ins and outs of analyzing a native crash.  I do want to add more null checking to other parts of the library as well but I didn't want to weigh down this particular PR with it.  The fact is that without this check, or at least a check in the deprecated API, the test I wrote regarding deleting collections and then setting expiration will outright CRASH.   If you really insist then I can make this a debug build only check, but in my opinion the less we outright crash with signals the less time we will spend going back and forth trying to get context of a stack trace.

jianminzhao: This is an interesting example. In this case, indeed, a NULL is caused by "default collection not found." Make sense. But let's look at this example,
  c4coll_getDoc(collection, "docID", &error) ==>. error.code == NotFound
Is the document not found in the collection or the collection NULL? 
Now, the platform code check the error,
if (error.code != 0) { Log("document %s is not found in %s", docID, c4coll_getSpec(collection)); // (pseudo code)
oh, crashing, because c4coll_getSpec() does not check NULL.

borrrden: The code in question returns "NotOpen" instead of "NotFound" for the case that the collection is null or invalid.  It's kind of shaky but a null collection is definitely invalid.  So your argument of the overlapping error code is moot.  I'm not a fan of the way our API just says "oh don't worry, an error is thrown down the stack" or "oh don't worry the calling code is always going to honor our annotations" when literally the only way to know the former is to be familiar with the code base, and being wrong about the latter causes hours if not days of debugging.  There is no documentation or comments indicating which collection methods throw and in what situations.  This is how we get ourselves into situations where no one knows what is going on and we get caught by surprises time and time again.  

As for your example, ok let's add a custom null check (can't be the same one because we have tests that make sure that the name and scope are still accessible on an invalid collection) to that API as well.

C/c4CAPI.cc
```diff
@@ -278,19 +284,21 @@ bool c4coll_isValid(C4Collection* coll) noexcept {
 }
 
 C4CollectionSpec c4coll_getSpec(C4Collection *coll) noexcept {
+    // Unlike the others, this continues to return valid data even
+    // after invalidation, so skip the validity check
     return coll->getSpec();
```
jianminzhao, line 289: To be consistent, would you consider to check null?

jianminzhao (approved):

C/c4CAPI.cc
borrrden: Yes certainly I will.  I also have another style that I think might be useful for the future in which we put some contract comments on the collection class, and reference them here when we don't check collection validity.  Then it can be easily apparent to the reader why the check is "missing"

borrrden: Done!

PR #1557, opened 2022-08-30
CBL-932: add function to check if certificate exists

callumbirks:
Added functions to check if a certificate exists in the persistent key-store.
May still need testing in Windows as I cannot currently do that myself, and the solution required two different functions be implemented, one for Mac and one for Windows.
I've modified some tests in `Crypto/tests/CertificateTest.cc` to use `!Cert::exists()` rather than `Cert::load() == nullptr`, and all tests are passing in MacOS.

### DIFF
--- a/C/c4CAPI.cc
+++ b/C/c4CAPI.cc
@@ -1654,7 +1654,9 @@ C4Cert* c4cert_load(C4String name,
 bool c4cert_exists(C4String name,
                    C4Error *outError)
 {
-    return C4Cert::exists(name);
+    return tryCatch<bool>(outError, [&]() {
+        return C4Cert::exists(name);
+    });
 }
 
 
--- a/C/c4CAPI.cc
+++ b/C/c4CAPI.cc
@@ -1651,6 +1651,12 @@ C4Cert* c4cert_load(C4String name,
     });
 }
 
+bool c4cert_exists(C4String name,
+                   C4Error *outError)
+{
+    return C4Cert::exists(name);
+}
+
 
 #pragma mark - KEY PAIR API: (EE)
 
--- a/Crypto/PublicKey+Windows.cc
+++ b/Crypto/PublicKey+Windows.cc
@@ -201,9 +201,7 @@ namespace litecore::crypto {
                 nullptr
         );
 
-        DEFER {
-            CertCloseStore(store, 0);
-        };
+        CertCloseStore(store, 0);
 
         return winCert;
     }
@@ -647,9 +645,7 @@ namespace litecore::crypto {
 
         const auto* const winCert = getWinCert(persistentID);
 
-        DEFER {
-            CertFreeCertificateContext(winCert);
-        };
+        CertFreeCertificateContext(winCert);
 
         return !winCert ? false : true;
     }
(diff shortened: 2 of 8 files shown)


C/c4CAPI.cc
```diff
@@ -1651,6 +1651,12 @@ C4Cert* c4cert_load(C4String name,
     });
 }
 
+bool c4cert_exists(C4String name,
+                   C4Error *outError)
+{
+    return C4Cert::exists(name);
```
jianminzhao, line 1657: Please use tryCatch like the above.

Crypto/PublicKey+Windows.cc
```diff
@@ -685,6 +693,47 @@ namespace litecore::crypto {
         return new Cert(slice(winCert->pbCertEncoded, winCert->cbCertEncoded));
     }
 
+    bool Cert::exists(PublicKey *subjectKey) {
```
jianminzhao, line 696: Where is this used?

callumbirks: Nowhere where it’s used currently, but as there is a Cert::load function with the same parameter (as there is also a Cert::load function that takes a name parameter), I figured it may be useful to implement for future use. Of course if you think it isn’t necessary I can remove it.

jianminzhao: Cannot you refactor the code to avoid duplicated code with load() then?

callumbirks: Yes possibly, that’s a good idea. I’ll take a look at it tomorrow.

borrrden:
Please flip off from draft if you are going to request a review :)

Crypto/PublicKey+Windows.cc
```diff
@@ -168,6 +168,46 @@ namespace litecore::crypto {
...
+        auto winCert = CertFindCertificateInStore(
+                store,
+                X509_ASN_ENCODING,
+                0,
+                CERT_FIND_PUBKEY_MD5_HASH,
+                &hashBlob,
+                nullptr
+        );
+
+        DEFER {
```
borrrden, line 204: There is no need to use DEFER here since there is nothing left in the method.  You can simply call `CertCloseStore` without the DEFER.  DEFER is there to act sort of like a C++ destructor, so that no matter what happens after this call is guaranteed to happen.

Crypto/PublicKey+Windows.cc
```diff
@@ -601,6 +641,19 @@ namespace litecore::crypto {
         return cert;
     }
 
+    bool Cert::exists(const std::string &persistentID) {
+        LogTo(TLSLogDomain, "Checking if the certificate chain with id '%s' exists in the Keychain.",
+              persistentID.c_str());
+
+        const auto* const winCert = getWinCert(persistentID);
+
+        DEFER {
```
borrrden, line 650: Same for this DEFER

Crypto/PublicKey+Windows.cc
```diff
@@ -685,6 +707,16 @@ namespace litecore::crypto {
         return new Cert(slice(winCert->pbCertEncoded, winCert->cbCertEncoded));
     }
 
+    bool Cert::exists(PublicKey *subjectKey) {
+        auto winCert = getWinCert(subjectKey);
+
+        DEFER {
```
borrrden, line 713: And this one! 😄

borrrden (changes requested):

jianminzhao (approved):

borrrden (approved):

pasin (approved):

PR #1562, opened 2022-09-06
Add tests for pending document IDs with collections

callumbirks:
Add tests for pending document IDs for existing and non-existing collections.

Modify helper struct in c4ReplicatorHelpers.hh to be more generic, now can create it for any collection - not just default.
Had to modify some usages of the above helper struct as a variable name changed.

### DIFF
--- a/Replicator/c4ReplicatorHelpers.hh
+++ b/Replicator/c4ReplicatorHelpers.hh
@@ -15,24 +15,31 @@
 
 namespace litecore { namespace repl {
 
-    struct C4ReplParamsDefaultCollection : C4ReplicatorParameters {
-        C4ReplicationCollection defaultCollection{kC4DefaultCollectionSpec};
-
-        C4ReplParamsDefaultCollection()
-        : C4ReplicatorParameters()
-        , push(defaultCollection.push)
-        , pull(defaultCollection.pull)
-        , pushFilter(defaultCollection.pushFilter)
-        , validationFunc(defaultCollection.pullFilter)
+    // Helper struct to make testing with collections easier
+    struct C4ReplParamsCollection : C4ReplicatorParameters {
+        explicit C4ReplParamsCollection(
+                C4CollectionSpec collectionSpec
+        ):  C4ReplicatorParameters{ }
+            , replCollection { collectionSpec }
+            , push(replCollection.push)
+            , pull(replCollection.pull)
+            , pushFilter(replCollection.pushFilter)
+            , validationFunc(replCollection.pullFilter)
         {
-            collections = &defaultCollection;
+            collections = &replCollection;
             collectionCount = 1;
         }
-
+        C4ReplicationCollection replCollection;
         C4ReplicatorMode& push;
         C4ReplicatorMode& pull;
         C4ReplicatorValidationFunction C4NONNULL & pushFilter;
         C4ReplicatorValidationFunction C4NONNULL & validationFunc;
     };
+
+struct C4ReplParamsDefaultCollection : C4ReplParamsCollection {
+    C4ReplParamsDefaultCollection()
+            : C4ReplParamsCollection{ kC4DefaultCollectionSpec }
+    {}
+};
 
 }}
--- a/Replicator/c4ReplicatorHelpers.hh
+++ b/Replicator/c4ReplicatorHelpers.hh
@@ -16,10 +16,10 @@
 namespace litecore { namespace repl {
 
     // Helper struct to make testing with collections easier
-    struct C4ReplParamsCollection : C4ReplicatorParameters {
+    struct C4ReplParamsOneCollection : C4ReplicatorParameters {
         C4ReplicationCollection replCollection;
 
-        explicit C4ReplParamsCollection(
+        explicit C4ReplParamsOneCollection(
                 C4CollectionSpec collectionSpec
         ):  C4ReplicatorParameters{ }
             , replCollection { collectionSpec }
@@ -38,9 +38,9 @@ namespace litecore { namespace repl {
         C4ReplicatorValidationFunction C4NONNULL & validationFunc;
     };
 
-struct C4ReplParamsDefaultCollection : C4ReplParamsCollection {
+struct C4ReplParamsDefaultCollection : C4ReplParamsOneCollection {
     C4ReplParamsDefaultCollection()
-            : C4ReplParamsCollection{ kC4DefaultCollectionSpec }
+            : C4ReplParamsOneCollection{kC4DefaultCollectionSpec }
     {}
 };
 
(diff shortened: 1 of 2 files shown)


callumbirks:
Added test case for db with multiple collections. However not sure if my testing completely covers "Make sure only documents in the replicated collection are present in pending document ID call". Will bring up in tomorrow's meeting.

Replicator/tests/ReplicatorAPITest.cc
```diff
@@ -613,6 +613,215 @@ TEST_CASE_METHOD(ReplicatorAPITest, "Is Document Pending", "[C][Push]") {
 }
 #endif
 
+#ifdef COUCHBASE_ENTERPRISE
```
borrrden, line 616: No need to add ifdef for every test, they can be grouped inside (or added to an existing group)

Replicator/tests/ReplicatorAPITest.cc
```diff
@@ -613,6 +613,215 @@ TEST_CASE_METHOD(ReplicatorAPITest, "Is Document Pending", "[C][Push]") {
...
+    repl::C4ReplParamsCollection params{Kyber};
+    params.push = kC4OneShot;
+    params.pull = kC4Disabled;
+    params.callbackContext = this;
+    params.socketFactory = _socketFactory;
+
+    _repl = c4repl_newLocal(db, (C4Database*)db2, params, ERROR_INFO(err));
+    REQUIRE(_repl);
+
+    C4SliceResult encodedDocIDs = c4repl_getPendingDocIDs(_repl, Republic, ERROR_INFO(err));
```
borrrden, line 713: ERROR_INFO not really needed here.  Its purpose is to log the error that occurred if the next check fails, but the next check is checking the error code itself anyway.

Replicator/tests/ReplicatorAPITest.cc
```diff
@@ -613,6 +613,215 @@ TEST_CASE_METHOD(ReplicatorAPITest, "Is Document Pending", "[C][Push]") {
...
+    REQUIRE(encodedDocIDs == nullslice);
+
+    // Check that collection 2 still has all the documents pending
+    encodedDocIDs = c4repl_getPendingDocIDs(replFed, Federation, ERROR_INFO(err));
+    CHECK(err.code == 0);
+    REQUIRE(encodedDocIDs != nullslice);
+    docIDs = FLValue_AsArray(FLValue_FromData(C4Slice(encodedDocIDs), kFLTrusted));
+    CHECK(FLArray_Count(docIDs) == 100);
+
+    c4repl_free(replFed);
```
borrrden, line 821: Can probably use c4::ref to be safer here (it will wrap the return of c4repl_newLocal)

Replicator/tests/ReplicatorAPITest.cc
```diff
@@ -613,6 +613,215 @@ TEST_CASE_METHOD(ReplicatorAPITest, "Is Document Pending", "[C][Push]") {
...
+    encodedDocIDs = c4repl_getPendingDocIDs(_repl, Council, ERROR_INFO(err));
+    CHECK(err.code == 0);
+    REQUIRE(encodedDocIDs == nullslice);
+
+    // Check that collection 2 still has all the documents pending
+    encodedDocIDs = c4repl_getPendingDocIDs(replFed, Federation, ERROR_INFO(err));
+    CHECK(err.code == 0);
+    REQUIRE(encodedDocIDs != nullslice);
+    docIDs = FLValue_AsArray(FLValue_FromData(C4Slice(encodedDocIDs), kFLTrusted));
+    CHECK(FLArray_Count(docIDs) == 100);
```
borrrden, line 819: I'm a bit suspicious here.  Are there only 100 entries in the wikipedia json?

borrrden (changes requested):
Most of the inline comments are tweaks, but overall I think that these tests are repeating too much of the same work.  There is no need to have all the original variations in every test.  The goal of these tests is to make sure that the pending doucument IDs only return IDs from their respective collection.

Replicator/tests/ReplicatorAPITest.cc
```diff
@@ -613,6 +613,115 @@ TEST_CASE_METHOD(ReplicatorAPITest, "Is Document Pending", "[C][Push]") {
...
+    importJSONLines(sFixturesDir + "wikipedia_100.json", collFederation);
+
+    // Create the collections in the database to be replicated to
+    createDB2();
+    createCollection(db2, Council);
+    createCollection(db2, Federation);
+
+    // Set options for the replication
+    // Replicating only the Council collection
+    FLSliceResult options {};
```
jianminzhao, line 669: Where did you use options?

Replicator/tests/ReplicatorAPITest.cc
```diff
@@ -613,6 +613,115 @@ TEST_CASE_METHOD(ReplicatorAPITest, "Is Document Pending", "[C][Push]") {
...
+    // Create replicator config for Federation collection
+    // This won't actually be replicated, but is needed to call c4repl_getPendingDocIDs for the collection
+    repl::C4ReplParamsCollection paramsFederation { Federation };
+    paramsFederation.push = kC4OneShot;
+    paramsFederation.pull = kC4Disabled;
+    paramsCouncil.callbackContext = this;
+    paramsCouncil.socketFactory = _socketFactory;
+
+    int expectedPending = 0;
+    SECTION("Normal") {
```
jianminzhao, line 686: Why is this SECTION as there is one? SECTIONs are for running two test scenarios in one test case.
This is a good case when only one collection is involved in each different replicators. Maybe to consider a case where two collections are included in one replicator

Replicator/c4ReplicatorHelpers.hh
```diff
@@ -15,17 +15,20 @@
 
 namespace litecore { namespace repl {
 
-    struct C4ReplParamsDefaultCollection : C4ReplicatorParameters {
-        C4ReplicationCollection defaultCollection{kC4DefaultCollectionSpec};
+    // Helper struct to make testing with collections easier
```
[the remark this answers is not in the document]
callumbirks: Thanks, that's a better sounding name 🙂

Replicator/tests/ReplicatorAPITest.cc
callumbirks: Ah sorry, that was left over from when the test had multiple sections

Replicator/tests/ReplicatorAPITest.cc
callumbirks: Oh I see, again as above that was left over from when the test had multiple sections. I wasn't sure if it could be removed or not. Thank you 🙂

jianminzhao (approved):

borrrden (approved):

PR #1563, opened 2022-09-06
CBL-3009 : quoting collection names.

jianminzhao:
By N1QL grammar, collection names and scope names are IDENTIFIERs, which can be quoted, just like the property names in a property path. That is, we may say, "SELECT ssn FROM `human.resource`.payroll", where "human.resouce" represents the scope name in the grammar, and "payroll" represents the collection name. Currently, however, the parser collapses them into one collection path, "human.resource.payroll". The original scope/collection names are not recoverable. Usually, this is not a problem, because no matter where you put the divide, the part with dot in it won't qualifiy as a scope or collection name in the later phase of the parser. However, the loss of the original intent of the N1QL expression can cause a problem. If unqualified -- not led by a scope name -- a collection name may represent the default collection. For example, if "human.resource.db" is the name of the database, "SELECT name FROM `human.resource.db` should mean to select the names from the default collection of the database.

Therefore,
1. we parse the collection path in N1QL expression into separate properties, "SCOPE" and "COLLECTION" in JSON.
2. We allow each properties, of SCOPE & COLLECTION, to be quoted, and save them in JSON string with escaped dot.
3. In the JSON form, we continue to allow omitting SCOPE and the value of COLLECTION to be dot-separated path, but being a separator, it must not be escaped.

If dot is not used in scope or collection names, there should not be any change from the current behavior. If they include dots, the parser will determine whether it corresponds to the database name and represent the default collection, or invalid collection names.

### DIFF
--- a/LiteCore/Query/N1QL_Parser/n1ql.cc
+++ b/LiteCore/Query/N1QL_Parser/n1ql.cc
@@ -2530,6 +2530,24 @@ YY_ACTION(void) yy_1_join(yycontext *yy, char *yytext, int yyleng)
 #undef s
 #undef o
 }
+YY_ACTION(void) yy_3_collectionName(yycontext *yy, char *yytext, int yyleng)
+{
+#define c2 yy->__val[-1]
+#define c yy->__val[-2]
+#define __ yy->__
+#define yypos yy->__pos
+#define yythunkpos yy->__thunkpos
+  yyprintf((stderr, "do yy_3_collectionName\n"));
+  {
+#line 88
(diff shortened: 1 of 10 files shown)


snej (commented):
I don't see anything wrong, but I've been away from this code too long to feel comfortable reviewing it.

borrrden (approved):
My first reaction was incorrect in that I thought "we don't allow dots in database names" but unfortunately we do.  This is the reason this has become this complex, unfortunately.

PR #1568, opened 2022-09-13
CBL-3442: handle legacy 3.0 client

jianminzhao:
With 3.1, the first message that the client sends to the server is "getCollections". This allows the server to align the list of collections with order of the client. For 3.0, there is not a clear first message.

### DIFF
--- a/Replicator/Replicator.cc
+++ b/Replicator/Replicator.cc
@@ -1026,7 +1026,6 @@ namespace litecore { namespace repl {
 
     // Handles a "getCheckpoint" request by looking up a peer checkpoint.
     void Replicator::handleGetCheckpoint(Retained<MessageIn> request) {
-        // 3.0 client may send this message as the first one.
         setMsgHandlerFor3_0_Client(request);
 
         slice checkpointID = getPeerCheckpointDocID(request, "get");
@@ -1067,7 +1066,6 @@ namespace litecore { namespace repl {
 
     // Handles a "setCheckpoint" request by storing a peer checkpoint.
     void Replicator::handleSetCheckpoint(Retained<MessageIn> request) {
-        // 3.0 client may send this message as the first one.
         setMsgHandlerFor3_0_Client(request);
 
         slice checkpointID = getPeerCheckpointDocID(request, "set");
@@ -1108,6 +1106,8 @@ namespace litecore { namespace repl {
 
     // Handles a "getCollections" request by looking up a peer checkpoint of each collection.
     void Replicator::handleGetCollections(Retained<blip::MessageIn> request) {
+        // This message only comes from 3.1+ client.
+
         if (_subRepls.size() != 0) {
             // Request of 3.0 style has been received prior to this one.
             logError("Some message has preceded 'getCollections'");
@@ -1317,6 +1317,13 @@ namespace litecore { namespace repl {
             != kNotCollectionIndex) {
             return; // 3.0 message should not include the collection property
         }
+
+        // At this point, we are dealing with a 3.0 style replicator which can only have exactly
+        // one collection, which is the default one.  If the default collection is not specified in
+        // the passive config, rearrangeCollectionsFor3_0_Client() will put a null collection path
+        // at the place of index 0, and then we return an error here.
+        // (If the collection does not exist in the database, prepareWorkers() will fail and an
+        //  error will be returned from there.)
 
         _options->rearrangeCollectionsFor3_0_Client();
         DebugAssert(!_options->collectionAware());
--- a/Replicator/Replicator.cc
+++ b/Replicator/Replicator.cc
@@ -1295,6 +1295,12 @@ namespace litecore { namespace repl {
         }
     }
 
+    // This method is to properly initialize the passive replicator to get ready to
+    // serve 3.0 replicator, which is unaware of the collection.
+    // It is a no-op if
+    // 1. it is working in the active mode,                             or
+    // 2. the incoming meesage includes explicit "collection" property, or
+    // 3. the second time and after that this method is called.
     void Replicator::setMsgHandlerFor3_0_Client(Retained<blip::MessageIn> request)
     {
         if (_setMsgHandlerFor3_0_ClientDone) {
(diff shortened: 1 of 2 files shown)


Replicator/Replicator.cc
```diff
@@ -1091,6 +1067,9 @@ namespace litecore { namespace repl {
 
     // Handles a "setCheckpoint" request by storing a peer checkpoint.
     void Replicator::handleSetCheckpoint(Retained<MessageIn> request) {
+        // 3.0 client may send this message as the first one.
```
borrrden, line 1070: Incorrect comment

Replicator/Replicator.cc
```diff
@@ -1307,4 +1295,36 @@ namespace litecore { namespace repl {
         }
     }
 
+    void Replicator::setMsgHandlerFor3_0_Client(Retained<blip::MessageIn> request)
```
borrrden, line 1298: This method takes a while to understand and so deserves some comments.  It is a no-op for active replicators, messages that include an explicit collection property, and the second time and after that this method is called.  The middle part is particularly important to note because that is what stops every 3.1 based message from causing trouble here.

Replicator/Replicator.cc
```diff
@@ -1307,4 +1295,36 @@ namespace litecore { namespace repl {
...
+        if (_options->isActive()) {
+            return; // only deal with passive replicator.
+        }
+
+        if (request->intProperty(kCollectionProperty, kNotCollectionIndex)
+            != kNotCollectionIndex) {
+            return; // 3.0 message should not include the collection property
+        }
+
+        std::vector<C4CollectionSpec> v;
```
borrrden, line 1315: Where is this actually used?

Replicator/Replicator.cc
```diff
@@ -1307,4 +1295,36 @@ namespace litecore { namespace repl {
...
+        v.emplace_back(kC4DefaultCollectionSpec);
+        _options->rearrangeCollectionsFor3_0_Client();
+        DebugAssert(!_options->collectionAware());
+        DebugAssert(_options->workingCollectionCount() == 1);
+        if (!_options->collectionPath(0)) {
```
borrrden, line 1320: What is this testing?  The log message doesn't seem to line up with the test.  Shouldn't this be compared to the default collection path rather than just `!`?

Replicator/Replicator.cc
```diff
@@ -1307,4 +1295,36 @@ namespace litecore { namespace repl {
...
+            logVerbose("Client is legacy 3.0, but the default collection is not in the config of this 3.1 replicator.");
+            // what should the message be? 3.0 client is not aware of default collection.
+            request->respondWithError({"BLIP"_sl, 400, "This server does not support 3.0 client"_sl});
```
borrrden, line 1323: Suggestion:  This server is not configured for 3.0 client support

borrrden (changes requested):

Replicator/Replicator.cc
jianminzhao: Oh, copy/paste mistake.

Replicator/Replicator.cc
jianminzhao: when the default collection is not listed in the array of ReplicationCollection of ReplicatorParameters. rearrangeCollectionFor3_0_Client() will pull the default collection to the beginning of the array. If it's missing, nullslice will take that place.

Replicator/Replicator.cc
jianminzhao: Agreed, it's better.

Replicator/Replicator.cc
jianminzhao: Which comment are you referring to? The one above the function header or below it?

Replicator/Replicator.cc
borrrden: This doesn't prove that.  What if there is collection A, instead of the default collection, in position 0?

Replicator/Replicator.cc
borrrden: `// 3.0 client may send this message as the first one.`

jianminzhao: Actually, I meant it. Because 3.0 client may send this message as the first message, I have to apply the following line. In fact, I mean to add the following line to any handlers that may handle the first message, except for getCollections, which 3.0 client won't send.

Replicator/Replicator.cc
jianminzhao: We are following the general rule: if at index i, the collection is not specified by the passive replicator, we substitute null collection path. In this case, there is only one collection in the config, with index 0.

Replicator/Replicator.cc
borrrden: If you are just adding this comment to everything, the comment value is less.  Have you actually seen a situation in which setCheckpoint comes first?  I can't imagine that it would.

Replicator/Replicator.cc
borrrden: I see, this is more context that will not make sense to a new reader without comments.  The key here seems to be that we have the assertion `!_options->collectionAware()`.  This means that we are forced into only having at most 1 collection, which can only be the default one or nothing.  See how just saying "the general rule that if the collection is not specified then we substitute null" doesn't tell the whole story?

Replicator/Replicator.cc
```diff
@@ -1307,4 +1295,40 @@ namespace litecore { namespace repl {
...
+
+        if (_options->isActive()) {
+            return; // only deal with passive replicator.
+        }
+
+        if (request->intProperty(kCollectionProperty, kNotCollectionIndex)
+            != kNotCollectionIndex) {
+            return; // 3.0 message should not include the collection property
+        }
+
```
borrrden, line 1320:
```suggestion
// At this point, we are dealing with a 3.0 style replicator which can only have exactly one collection, which is the default one.  If the default collection is not specified in the passive config (or the default collection does not exist), then we return an error.
```

Replicator/Replicator.cc
jianminzhao: I can be 100% certain. Without the certainly, having this line here won't hurt because it would be a no-op after the first call through any other handler.

Replicator/Replicator.cc
jianminzhao: "forced into only having at most 1 collection" -- not "at most". 0 collection would be ruled out before arriving here. At this place, there is exactly one collection.

borrrden:
Have you noticed the suggestion block above? https://github.com/couchbase/couchbase-lite-core/pull/1568#discussion_r972489781

Replicator/Replicator.cc
jianminzhao: good point. removing them.

Replicator/Replicator.cc
jianminzhao: Sure sure how this "commit suggestion" work, and, since I need to do some other changes, plus little modification of the above comment, I copy-pasted the comment to my local machine and pushed a new revision.

borrrden (approved):

Replicator/Replicator.cc
```diff
@@ -1307,4 +1295,47 @@ namespace litecore { namespace repl {
...
+        // at the place of index 0, and then we return an error here.
+        // (If the collection does not exist in the database, prepareWorkers() will fail and an
+        //  error will be returned from there.)
+
+        _options->rearrangeCollectionsFor3_0_Client();
+        DebugAssert(!_options->collectionAware());
+        DebugAssert(_options->workingCollectionCount() == 1);
+        if (!_options->collectionPath(0)) {
+            logVerbose("Client is legacy 3.0, but the default collection is not in the config of this 3.1 replicator.");
+            // what should the message be? 3.0 client is not aware of default collection.
```
pasin, line 1333: Minor : Seem like a question was left in the comment.

pasin (approved):

callumbirks (approved):

PR #1569, opened 2022-09-14
Bump fleece to its master branch.

jianminzhao:


borrrden (approved):
If the fix is in Fleece then I suggest not including the ticket number in the PR title.  Otherwise the ticket will get cluttered with different PRs.  The build bot is already going to show which build of core got the fix by virtue of having the ticket number in a Fleece commit.

PR #1570, opened 2022-09-16
CBL-3531: Change ReplicatorLoopbackTest to use named collections

callumbirks:
Changed ReplicatorLoopbackTest to use a named collection rather than the default collection.
Had to change usages of functions that use default collection to their named collection variants.
Some functions also needed new overloads that allow them to be used with named collections.
Because of [this change](https://github.com/couchbase/couchbase-lite-core/blob/master/Replicator/Checkpointer.cc#L239-L250) to how checkpoint IDs are generated for named collections, the expected ID in checkpoint validation tests had to be changed.

### DIFF
--- a/C/tests/c4Test.cc
+++ b/C/tests/c4Test.cc
@@ -522,7 +522,11 @@ string C4Test::listSharedKeys(string delimiter) {
 
 
 string C4Test::getDocJSON(C4Database* inDB, C4Slice docID) {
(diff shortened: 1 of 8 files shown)


jianminzhao (approved):

borrrden:
In the future, rebase instead of merging to avoid long commit histories including commits like "merge master into CBL-3531"

borrrden (approved):

PR #1571, opened 2022-09-16
CBL-3612: lastSequence may lag behind the actual maximum sequence in …

jianminzhao:
…the kv table.

This problem is reported by CBSE ticket 12382. It is manifested by SQLite error, "SQLite error (code 2067): ... UNIQUE constraint failed: kv_default.sequence". Analysis shows it is caused by the fact that the last sequence as stored in the meta data, which is supposed to be the maximum among all revisions in the database table, is outdated (lags behind) due to exceptional situaions, out-of-memory error in particular.

It is hard to devise a test case for it but I strive to remedy, or ameliorate, the aftermath, without addressing root-cause why the last sequence gets outdated.

I intend it to be 100% progressive: it would not change the program's behavior in the absence of that particular error, "UNIQUE constraint failed." When the event does happen, an exception will bubble up all the way to the library boundary without the fix, and this fix will make a second attempt to verify the cause and bump the last sequence, namely A. Compute "SELECT MAX(sequence) FROM kv_table > [current last sequence]" B. If the above is true, bump the last sequence by the result of SELECT. C. In all other cases, rethrow the exception.

Caveat: the solution will get us overcome the imminent obstacle of UNIQUE constraint and put the record to the database, but we are not concerned of other possible impact of the root-cuase and don't even know whether this new sequence number would collide with a revision that is not currently selected for the records of the table.

### DIFF
--- a/LiteCore/Storage/SQLiteKeyStore.cc
+++ b/LiteCore/Storage/SQLiteKeyStore.cc
@@ -331,52 +331,102 @@ namespace litecore {
 
         enum { VersionParam = 1, BodyParam, ExtraParam, FlagsParam, SequenceParam, KeyParam,
                OldSequenceParam, OldSubsequenceParam };
-        const char *opName;
-        SQLite::Statement *stmt;
-        if (rec.sequence == 0_seq) {
-            // Insert only:
-            stmt = &compileCached(
-                    "INSERT OR IGNORE INTO kv_@ (version, body, extra, flags, sequence, key)"
-                    " VALUES (?, ?, ?, ?, ?, ?)");
-            opName = "insert";
-        } else {
-            // Replace only:
-            stmt = &compileCached(
-                    "UPDATE kv_@ SET version=?, body=?, extra=?, flags=?, sequence=?"
-                    " WHERE key=? AND sequence=? AND (flags >> 16) = ?");
-            stmt->bind(OldSequenceParam,    (long long)rec.sequence);
-            stmt->bind(OldSubsequenceParam, (long long)rec.subsequence);
-            opName = "update";
-        }
 
-        sequence_t seq;
-        int64_t rawFlags = int(rec.flags);
-        if (updateSequence) {
-            seq = lastSequence() + 1;
-        } else {
-            Assert(rec.sequence > 0_seq);
-            seq = rec.sequence;
-            // If we don't update the sequence, update the subsequence so MVCC can work:
-            rawFlags |= (rec.subsequence + 1) << 16;
-        }
+        bool tryAgain = false;
+        sequence_t ret;
+        std::tuple<std::string, int, int> lastExcArgs;
 
-        stmt->bindNoCopy(VersionParam, rec.version.buf, (int)rec.version.size);
-        stmt->bindNoCopy(BodyParam,    rec.body.buf, (int)rec.body.size);
-        stmt->bindNoCopy(ExtraParam,   rec.extra.buf, (int)rec.extra.size);
-        stmt->bind      (FlagsParam,   (long long)rawFlags);
-        stmt->bindNoCopy(KeyParam,     (const char*)rec.key.buf, (int)rec.key.size);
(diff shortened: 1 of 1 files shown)


LiteCore/Storage/SQLiteKeyStore.cc
```diff
@@ -331,52 +331,102 @@ namespace litecore {
...
+            try {
+                status = stmt->exec();
+                // We are good, don't try again.
+                tryAgain = false;
+            } catch (const SQLite::Exception& exc) {
+                // We won't try again twice.
+                tryAgain = !tryAgain;
+                int sqliteErrorCode = exc.getErrorCode();
+                int sqliteExtErrorCode = exc.getExtendedErrorCode();
+                if (tryAgain && sqliteErrorCode == 19 && sqliteExtErrorCode == 2067) {
```
borrrden, line 406: Use SQLITE defines `SQLITE_CONSTRAINT` and `SQLITE_CONSTRAINT_UNIQUE`

LiteCore/Storage/SQLiteKeyStore.cc
```diff
@@ -331,52 +331,102 @@ namespace litecore {
...
+                    throw SQLite::Exception(std::get<0>(lastExcArgs), std::get<1>(lastExcArgs), std::get<2>(lastExcArgs));
+                }
 
-        stmt->bindNoCopy(VersionParam, rec.version.buf, (int)rec.version.size);
-        stmt->bindNoCopy(BodyParam,    rec.body.buf, (int)rec.body.size);
-        stmt->bindNoCopy(ExtraParam,   rec.extra.buf, (int)rec.extra.size);
-        stmt->bind      (FlagsParam,   (long long)rawFlags);
-        stmt->bindNoCopy(KeyParam,     (const char*)rec.key.buf, (int)rec.key.size);
-        stmt->bind      (SequenceParam,(long long)seq);
+                setLastSequence((sequence_t)maxSeq);
```
borrrden, line 353: Just to verify, does this update `rec.sequence` to the new value?

borrrden (commented):
I need to hear what Jens has to say about this, but I am slightly worried that as this point the disk value is not the latest.  Since at this point there is a write transaction, I'm guessing it should be but there is a part of my mind that says instead of reacting like this, the latest sequence should just be loaded once when the db is open (comparing the highest sequence in kv to the one in the meta table).  Once it is in memory, it won't get corrupted like this, it's only after a crash that this happens.

LiteCore/Storage/SQLiteKeyStore.cc
jianminzhao: No, it does not. It merely resets the last sequence to maxSeq. We then restart the process.

LiteCore/Storage/SQLiteKeyStore.cc
```diff
@@ -331,52 +331,103 @@ namespace litecore {
...
+                // We won't try again twice.
+                tryAgain = !tryAgain;
+                int sqliteErrorCode = exc.getErrorCode();
+                int sqliteExtErrorCode = exc.getExtendedErrorCode();
+                if (tryAgain && sqliteErrorCode == SQLITE_CONSTRAINT &&
+                    sqliteExtErrorCode == SQLITE_CONSTRAINT_UNIQUE) {
+                    // We only handle this error,
+                    // Log: Database | UNIQUE constraint failed: kv_default.sequence (19/2067)
+                    // Jot down the exception that makes us to try again.
+                    lastExcArgs = std::make_tuple(exc.what(), sqliteErrorCode, sqliteExtErrorCode);
```
snej, line 411: Could you just store the SQLite::Exception object itself, instead of having to break its parts into a tuple?

snej (approved):
The logic looks OK. It would be good to add a comment at the top explaining that this is a band-aid for an undiagnosed bug, so we don't forget what this code is for, and can remove it once the real bug is fixed.

PR #1572, opened 2022-09-18
Prevent unnecessary growth of `RevTree._insertedData`

blaugold:
Currently, `RevTree._insertedData` is copied *as is* when copying a `RevTree`.

Why is this a problem?

When a document is repeatedly updated, `Rev`s start to be purged eventually. Also, `Rev` bodies that aren't needed anymore are removed.
`RevTree._insertedData` is only ever added to and copied, keeping more and more data unnecessarily alive with each update. This data occupies memory, and the work required to copy `RevTree._insertedData` starts to dominate save operations.

This is currently making the [fast path](https://github.com/blaugold/couchbase-lite-core/blob/b5cb97c26a8d392b95afc33e161f666e30e554fe/C/c4Document.cc#L209-L224) of `C4Document::update` worse than the slow-path after a few updates.

This change keeps track of which data in a `Rev` is owned by `RevTree._insertedData` and only adds that data to `_insertedData` of a copied `RevTree`.

### DIFF
--- a/LiteCore/RevTrees/RevTree.hh
+++ b/LiteCore/RevTrees/RevTree.hh
@@ -67,11 +67,15 @@ namespace litecore {
     private:
         slice       _body;          /**< Revision body (JSON), or empty if not stored in this tree*/
 
+        bool _hasInsertedRevID;     /**< Whether revID is owned by RevTree._insertedData */
+        bool _hasInsertedBody;      /**< Whether body is owned by RevTree._insertedData */
+
         void addFlag(Flags f)           {flags = (Flags)(flags | f);}
         void clearFlag(Flags f)         {flags = (Flags)(flags & ~f);}
         void removeBody()               {clearFlag((Flags)(kKeepBody | kHasAttachments));
+                                         _hasInsertedBody = false;
                                          _body = nullslice;}
-        bool isMarkedForPurge() const FLPURE   {return (flags & kPurge) != 0;}
+        bool isMarkedForPurge() const FLPURE    {return (flags & kPurge) != 0;}
 #if DEBUG
         void dump(std::ostream&);
 #endif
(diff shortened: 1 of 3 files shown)


LiteCore/RevTrees/RevTree.hh
```diff
@@ -57,7 +57,7 @@ namespace litecore {
             kNew            = 0x04, /**< Has this rev been inserted since decoding? */
             kHasAttachments = 0x08, /**< Does this rev's body contain attachments? */
             kKeepBody       = 0x10, /**< Body will not be discarded after I'm a non-leaf */
-            kIsConflict     = 0x20, /**< Unresolved conflicting revision; should never be current */
+            kIsConflict      = 0x20, /**< Unresolved conflicting revision; should never be current */
```
snej, line 60: Looks like a mistake, a stray space character?

LiteCore/RevTrees/RevTree.hh
```diff
@@ -67,11 +67,29 @@ namespace litecore {
     private:
         slice       _body;          /**< Revision body (JSON), or empty if not stored in this tree*/
 
+        enum DataFlags : uint8_t {
```
snej, line 70: It would be simpler just to add two bool fields, `_hasInsertedRevID` and `_hasInsertedBody`. Then you can get and set them directly instead of creating a bunch of accessors. If you're really concerned about the extra byte you could make them bitfields, but I think alignment is going to cause the struct to grow by more than 2 bytes no matter what you add.

snej (changes requested):

LiteCore/RevTrees/RevTree.hh
blaugold: I was trying out font ligatures and forgot about it. Changed it back.

LiteCore/RevTrees/RevTree.hh
blaugold: Done

snej (approved):

snej:
Looks good, thanks!

PR #1573, opened 2022-09-20
CBL-3700: Upgrade to SQLite 3.39.3

borrrden:


borrrden:
False detection in blackduck check, since two products live in the same repo and only one was updated (SQLiteCpp remains at 2.0.0)

PR #1574, opened 2022-09-20
CBL-3669: Change occurrences of sprintf to snprintf

callumbirks:


borrrden:
Unassigning until the validation failures are fixed

callumbirks:
What's the reason the build and tests would be passing fine on my machine but failing here, are there things that aren't built by default locally that I can enable to make sure everything is building fine? @borrrden

callumbirks:
Nevermind, I think I figured it out. I was only building the C++ tests before. Selecting the C Tests scheme showed me the other errors that weren't showing up before.

pasin (approved):

borrrden:
Don't merge this just yet, I want to include the Fleece changes as well.

PR #1575, opened 2022-09-20
CBL-3624: double free of BLIPIO

jianminzhao:
The cause of this bug is in WeakHolder -- it currently works only if the callback it invokes is synchronous. We fix it by holding an extra strong reference inside WeakHolder.

### DIFF
--- a/LiteCore/Support/WeakHolder.hh
+++ b/LiteCore/Support/WeakHolder.hh
@@ -12,13 +12,17 @@
 
 #pragma once
 #include "fleece/RefCounted.hh"
-#include <shared_mutex>
 
 namespace litecore {
 
 /** WeakHolder<T>: holds a pointer to T weakly. Unlike general weak reference, one cannot get a strong holder from it.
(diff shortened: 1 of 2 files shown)


LiteCore/Support/WeakHolder.hh
```diff
@@ -12,29 +12,26 @@
...
-
-    // Only the original owner may rescind the pointer.
-    // After rescind(), the held pointer becomes nullptr and invoke() returns false.
-    void rescind(T* owner) {
-        if (owner == _pointer) {
-            std::lock_guard<std::shared_mutex> lock(_mutex);
-            _pointer = nullptr;
-        }
+        _holder = dynamic_cast<RefCounted*>(_pointer);
+        Assert(_holder);
```
borrrden, line 34: I feel like this is something that can be asserted at compile time somehow

`static_assert(std::is_assignable<T*, RefCounted*>::value)`

borrrden: Nevermind...the dual inheritance makes this painful....

borrrden (approved):

PR #1579, opened 2022-09-23
CBL-3699: Update fleece to include snprintf changes

callumbirks:
Fleece also needed sprintf changing to snprintf, forgot to bump it to that commit in the last PR for this issue. Thanks @borrrden for help with that.


borrrden (commented):
There is a weird first commit in here that changes only whitespace.  What is that about?

jianminzhao (approved):

PR #1580, opened 2022-09-23
CBD-5140: Correctly install icu4c libs

ceejatec:


borrrden:
Don't worry about blackduck check, there is something else wrong with it now 😩 .  I'll have a look at it but the output shows the manifest to be correct, it's just returning an error status for whatever reason.

borrrden (approved):

PR #1582, opened 2022-09-27
CBL-3706: Add calls to returnIfCollectionInvalid to CAPI functions.

callumbirks:
CAPI functions using default collection can pass nullptr to other functions if default collection has been deleted, adding call to returnIfCollectionInvalid in these functions stops this from causing any crashes.
Also add tests which call all these functions, after deleting default collection, to reproduce the issue and ensure no crashes occur.

### DIFF
--- a/C/c4CAPI.cc
+++ b/C/c4CAPI.cc
@@ -527,15 +527,17 @@ bool c4db_maintenance(C4Database* database, C4MaintenanceType type, C4Error *out
 // semi-deprecated
 C4Timestamp c4db_nextDocExpiration(C4Database *db) noexcept {
     C4Error err;
-    returnIfCollectionInvalid(db->getDefaultCollection(), &err, C4Timestamp::Error);
-    return c4coll_nextDocExpiration(db->getDefaultCollection());
+    auto coll = db->getDefaultCollection();
+    returnIfCollectionInvalid(coll, &err, C4Timestamp::Error);
+    return c4coll_nextDocExpiration(coll);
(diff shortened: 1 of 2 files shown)


C/c4CAPI.cc
```diff
@@ -526,12 +526,15 @@ bool c4db_maintenance(C4Database* database, C4MaintenanceType type, C4Error *out
 
 // semi-deprecated
 C4Timestamp c4db_nextDocExpiration(C4Database *db) noexcept {
+    C4Error err;
+    returnIfCollectionInvalid(db->getDefaultCollection(), &err, C4Timestamp::Error);
```
pasin, line 530: This doesn't solve the problem as calling db->getDefaultCollection() on the next line could still fail. You could get the collection, test it with `returnIfCollectionInvalid()` then use it on the next line.

```
auto collection = db->getDefaultCollection();
returnIfCollectionInvalid(collection, ...);
...
```

pasin (changes requested):

borrrden (commented):
Agree with Pasin

pasin:
@callumbirks if it's ready to re-review, please re-assign.

callumbirks:
> @callumbirks if it's ready to re-review, please re-assign.

Thanks Pasin, I didn’t even realise I could do that.

pasin (approved):

borrrden (approved):

jianminzhao (approved):

PR #1585, opened 2022-09-29
CBL-3698: Repeat the tests in ReplicatorCollectionTest.cc with Sync-g…

jianminzhao:
…ateway server.

Tests in ReplicatorCollectionTest.cc are performed with loopback connections between two replicators in the same executable. We port those tests with Sync-gateway server on the local machine, as appropriate to test the interaction between the replicator in CBL as client and Sync-gateway as the server. The new tests are put in ReplicatorCollectionSGTest.cc. Still to be done is "Resolve Conflict", "Reset Checkpoint with Pull", and "Filters & docIDs with Multiple Collections".

### DIFF
--- a/Replicator/tests/ReplicatorCollectionSGTest.cc
+++ b/Replicator/tests/ReplicatorCollectionSGTest.cc
@@ -1,0 +1,1087 @@
+//
+//  ReplicatorCollectionSGTest.cc
+//
+//  Copyright 2022-Present Couchbase, Inc.
+//
+//  Use of this software is governed by the Business Source License included
+//  in the file licenses/BSL-Couchbase.txt.  As of the Change Date specified
+//  in that file, in accordance with the Business Source License, use of this
+//  software will be governed by the Apache License, Version 2.0, included in
+//  the file licenses/APL2.txt.
+//
+
+#include "c4Base.h"
+#include "ReplicatorAPITest.hh"
+#include "ReplicatorOptions.hh"
+#include "ReplicatorLoopbackTest.hh"
+#include "Base64.hh"
+#include "Defer.hh"
+#include "c4Collection.h"
+#include "c4DocEnumerator.h"
+#include "c4Document.h"
+#include "c4Database.h"
+#include "c4Replicator.hh"
+#include "fleece/Mutable.hh"
+#include <array>
+
+// Tests in this file, tagged by [.SyncServerCollection], are not done automatically in the
+// Jenkins/GitHub CI. They can be run in locally with the following environment.
+// Couchbase DB server, with docker, for example,
+//   docker run -d --name cbserver -p 8091-8096:8091-8096 -p 11210-11211:11210-11211 couchbase:7.1.1
+//   bucket configuration:
+//     user    : Administrator
+//     password: password
+//     name    : any
+//     scope   : flowers
+//     collection: roses
+// Sync-gateway:
+//   config.json:
+/*
+ {
+   "bootstrap": {
+     "server": "couchbase://localhost",
+     "username": "Administrator",
+     "password": "password",
+     "use_tls_server": false
+   },
+   "logging": {
+     "console": {
+       "log_level": "info",
+       "log_keys": ["*"]
+     }
+   }
+ }
+ */
+//  config db:
+/*
+ curl --location --request PUT "localhost:4985/scratch/" \
+ --header "Content-Type: application/json" \
+ --header "Authorization: Basic QWRtaW5pc3RyYXRvcjpwYXNzd29yZA==" \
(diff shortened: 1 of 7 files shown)


borrrden (commented):
This PR is too huge to reasonably review, as it does far more than simply add tests for SG it seems.  It changes a whole bunch of functions that every other test also uses, and for what?

jianminzhao:
Perhaps it's candidate of in-person live review.

Replicator/tests/ReplicatorCollectionSGTest.cc
```diff
@@ -0,0 +1,1094 @@
...
+ --header "Authorization: Basic QWRtaW5pc3RyYXRvcjpwYXNzd29yZA==" \
+ --data-raw '{"num_index_replicas": 0, "bucket": "your_bucket_name", "scopes": {"flowers": {"collections":{"flowers":{}}}}}'
+ */
+//  config SG user:
+/*
+ curl --location --request POST "localhost:4985/scratch/_user/" \
+ --header "Content-Type: application/json" \
+ --header "Authorization: Basic QWRtaW5pc3RyYXRvcjpwYXNzd29yZA==" \
+ --data-raw '{"name": "sguser", "password": "password", "admin_channels": ["*"]}'
+ */
```
pasin, line 68: * Would it be possible to create a static SG config file to configure SG? Previous test setup has two config files : TLS and Non-TLS.

* Would it be possible to have a script that helps running the docker and also configure the buckets?

jianminzhao: 1. I got the impression that static SG config is obsolete, contraindicated
2. yes. we may provide the script for Unix shell.

pasin: Here is a bit more info from Adam:

```
It's correct to say that we won't support file-based database configuration forever, and the recommended
approach for 3.0 and later is to use file-based config only for cluster bootstrapping, and to use the REST API
for database and user creation/management
```

pasin: As the config for tests are pretty much static, we could move the CURL commands to the same script that starts SG. If there are some changes in the test code, the test code will need to do the REST API call.

Replicator/tests/ReplicatorCollectionSGTest.cc
```diff
@@ -0,0 +1,1094 @@
...
+        c4Params.collections = replCollections.data();
+    };
+    replicate(paramsSetter);
+    verifyDocs(collectionSpecs, docIDs);
+    for (size_t i = 0; i < collectionCount; ++i) {
+        checkAttachments(verifyDb, blobKeys[i], attachments1);
+    }
+}
+
+#if 0
```
pasin, line 683: Is this test failed? Can you give info and ticket number if having one about the reason to disable this test?

Replicator/tests/ReplicatorCollectionSGTest.cc
```diff
@@ -0,0 +1,1094 @@
...
+static constexpr C4CollectionSpec Guitars = { GuitarsName, kC4DefaultScopeID };
+
+static constexpr slice RosesName = "roses"_sl;
+static constexpr slice TulipsName = "tulips"_sl;
+//static constexpr slice LavenderName = "lavenders"_sl;
+static constexpr slice FlowersScopeName = "flowers"_sl;
+
+static constexpr C4CollectionSpec Roses = { RosesName, FlowersScopeName };
+static constexpr C4CollectionSpec Tulips = { TulipsName, FlowersScopeName };
+//static constexpr C4CollectionSpec Lavenders = { LavenderName, FlowersScopeName };
```
pasin, line 84: Remove if not needed.

pasin (changes requested):

Replicator/tests/ReplicatorCollectionSGTest.cc
jianminzhao: I haven't checked in any script yet. I hope we will script to do the SG and CB server. Jim has script in his docker compose. I just did not get to it and manually set up the cb server. I hope not to include it in this PR if okay.

Replicator/tests/ReplicatorCollectionSGTest.cc
```diff
@@ -0,0 +1,1094 @@
...
+        thread.join();
+    }
+};
+
+TEST_CASE_METHOD(ReplicatorCollectionSGTest, "Multiple Collections Incremental Revisions SG", "[.SyncServerCollection]") {
+    string idPrefix = timePrefix();
+    constexpr size_t collectionCount = 1;
+    std::array<C4CollectionSpec, collectionCount> collectionSpecs = {
+        Roses
+        //, Tulips
```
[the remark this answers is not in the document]
borrrden: The estimate for that is the end of October

pasin (approved):

PR #1586, opened 2022-09-30
CBL-3712: Backport snprintf changes to Lithium.

callumbirks:
Cherry-picked snprintf changes from CBL-3669 onto lithium branch. 
Also did the same for fleece, and updated fleece commit in this branch.

### DIFF
--- a/C/tests/c4AllDocsPerformanceTest.cc
+++ b/C/tests/c4AllDocsPerformanceTest.cc
@@ -16,8 +16,8 @@
 #include "SecureRandomize.hh"
 #include <chrono>
 
-static const size_t kSizeOfDocument = 1000;
-static const unsigned kNumDocuments = 100000;
+static constexpr size_t kSizeOfDocument = 1000;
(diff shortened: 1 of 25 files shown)


borrrden (changes requested):
The master branch would have caught this but you have a chicken and egg problem.  You can't point to the PR branch commit of fleece, the PR in fleece needs to be merged first, and then this PR updated to the latest release/lithium fleece

callumbirks:
> The master branch would have caught this but you have a chicken and egg problem. You can't point to the PR branch commit of fleece, the PR in fleece needs to be merged first, and then this PR updated to the latest release/lithium fleece

Ah I see, thanks Jim 🙂

jianminzhao (approved):

borrrden (changes requested):
The latest commit just reset fleece back to where it was, which is also wrong.  This PR needs to point to the current tip of fleece lithium which is `d9ad8314b0c1dc424a9b7d1ec9e88f154ac5923d`

borrrden (approved):

PR #1587, opened 2022-10-05
CBL-3682: LiteCore fails due to unclosed transaction

jianminzhao:


LiteCore/Database/DatabaseImpl.cc
```diff
@@ -1040,10 +1047,8 @@ namespace litecore {
                 }
             }
 
-            if (remoteID > 0) {
-                // Found the remote ID!
-                return remoteID;
-            } else if (!canCreate) {
+            if (remoteID > 0 // Found the remote ID!
+                || !canCreate) {
```
pasin, line 1051: For the case of `!canCreate`, just curious why we don't break or return without even begin transaction. Is it because we want to try to check the existing remoteID for the second time?

jianminzhao: I may not understand your question. We do break without beginning transaction.

pasin:
I totally agreed that the current code could cause the transaction left open. But I couldn't find anything that is calling getRemoteDBID() with canCreate = false. If that the case, does it mean that the Java or the test code is somehow calling c4db_getRemoteDBID() with canCreate = false? I don't expect Java or the platform to call this function by itself.

jianminzhao:
The fix does deal with canCreate == true.
This is what happened, as far as I can see from the logs.
We do the loop twice. First time, there is no transaction. We simply read off the kv-table. Since we are reading outside transaction, some other transaction may write to the table.
Now, we run the loop second time. This time, we open a transaction. We read the table again. This time, we found the remoteID. So, we simply return that remoteID, but forgot close the transaction.

LiteCore/Database/DatabaseImpl.cc
pasin: I mean for the case that creating = true and canCreate = false.

pasin: But as nobody is actually use canCreate option = false so I think preserve the current logic is totally fine.

pasin (approved):

borrrden:
There is no description whatsoever of what the problem was....

PR #1589, opened 2022-10-10
CBL-3720: porting test, "Resolve Conflict", to use SG.

jianminzhao:


borrrden (commented):
What is this test checking?  I don't think this test is particularly needed in this form since it is testing a behavior that is not present in LiteCore.  LiteCore's job is to deliver the notification to the platform and that's it.  The more we do this, the more we are just testing implementation that exists only in the test suite.  Furthermore, are you sure that REQUIRE is ok to use inside of a callback?  It might cause the replication to simply stop and you would not find out the error in  Catch because it would be swallowed.

jianminzhao:
It tests the same logic as its counterpart does with regard to those using the loopback connections, the test in ReplicatorCollectionTest.cc.
We have three suites of tests now, A. those use local passive replicator as the server, B. those use Sync-gateway server with walrus database, and C. those use sync-gateway plus a real CB server. B is to be deprecated. We really have two classes, A (no SG( and C (with SG). This particular test, I think, tests that Core properly generates docEnded calls as it syncs with SG, and c4doc_resolveConflict works as designed.
I will check with how CHECK works inside the callback.

jianminzhao:
It seems CHECK inside callback works just fine. 
This is the message, 
-----------------------------------------------------------------------------------------------------------------------
Resolve Conflict SG
-----------------------------------------------------------------------------------------------------------------------
/Users/jzhao/work/Git/couchbase-lite-core/Replicator/tests/ReplicatorCollectionSGTest.cc:556

/Users/jzhao/work/Git/couchbase-lite-core/Replicator/tests/ReplicatorCollectionSGTest.cc:628: FAILED:
  CHECK( error.code != 0 )
with expansion:
  0 != 0

and the line number is also correct.

borrrden:
CHECK will work because it doesn't try to halt execution like REQUIRE does.

Replicator/tests/ReplicatorCollectionSGTest.cc
```diff
@@ -553,6 +553,130 @@ TEST_CASE_METHOD(ReplicatorCollectionSGTest, "Push and Pull Attachments SG", "[.
...
 }
 
+TEST_CASE_METHOD(ReplicatorCollectionSGTest, "Resolve Conflict SG", "[.SyncServerCollection]") {
+    string idPrefix = timePrefix();
+    constexpr size_t collectionCount = 1;
+    std::array<C4CollectionSpec, collectionCount> collectionSpecs = {
+        Roses
+    };
+    std::array<C4Collection*, collectionCount> collections =
+        collectionPreamble(collectionSpecs, "sguser", "password", kC4LogNone);
```
pasin, line 563: I don't remember the detail of collectionPreamble() function, but what is the kC4LogNone for?

Replicator/tests/ReplicatorCollectionSGTest.cc
```diff
@@ -553,6 +553,130 @@ TEST_CASE_METHOD(ReplicatorCollectionSGTest, "Push and Pull Attachments SG", "[.
...
+            enc.endArray();
+            Doc doc {enc.finish()};
+            docIDsDicts[i] =
+                repl::Options::updateProperties(
+                    AllocedDict(c4Params.collections[i].optionsDictFleece),
+                    kC4ReplicatorOptionDocIDs,
+                    doc.root());
+            c4Params.collections[i].optionsDictFleece = docIDsDicts[i].data();
+        }
+    };
```
pasin, line 612: If this needs to be done on every test, I think we should have a utility function for this.

Replicator/tests/ReplicatorCollectionSGTest.cc
jianminzhao: This is for debugging purpose. If not None, this function will set the kSyncLog to that level. I probably should remove it.

Replicator/tests/ReplicatorCollectionSGTest.cc
jianminzhao: I can certainly do it, but I feel every test case should have its parameter setting. The first two lines are absolutely common, but the rest are per-test.

Replicator/tests/ReplicatorCollectionSGTest.cc
pasin: If all the tests are going to set it to None, it's not much useful. When debugging an issue, you can just enable the debug log for the test case-by-case.

Replicator/tests/ReplicatorCollectionSGTest.cc
pasin: The test can setup the param but have a function that takes some params to setup docIDs? It is really adding some complexity to the test. Jim is working on some idea in this area, hope that can be applied here as well.

Replicator/tests/ReplicatorCollectionSGTest.cc
jianminzhao: yes, I am going to remove it. It's mostly useful for my testing by per-case basis.

Replicator/tests/ReplicatorCollectionSGTest.cc
jianminzhao: yes, docIDs are used to work around the issue with flushing before every test. Let's see if we want to abandon this approach altogether.

borrrden (approved):
I still think this test goes too heavy into doing actual conflict resolution, which is something that can be tested separately without replication at all, but I won't block on that.

pasin (approved):

PR #1591, opened 2022-10-12
CBL-3448: Manage mutable members of repl::Options

velicuvlad:
- Moved all mutables under `Mutables` struct
- Updated use/calls to reflect the struct change

### DIFF
--- a/Replicator/ReplicatorOptions.hh
+++ b/Replicator/ReplicatorOptions.hh
@@ -40,10 +40,10 @@ namespace litecore { namespace repl {
         void*                   callbackContext         {nullptr};
         std::atomic<C4ReplicatorProgressLevel> progressLevel {kC4ReplProgressOverall};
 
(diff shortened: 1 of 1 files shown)


callumbirks (commented):
LGTM, seems like it does what the ticket asks. I'm not 100% on understanding exactly what the ticket needs though, so I can't say for sure.

borrrden:
I remember asking for this ticket, and what I imagined was that the mutable members were moved to the replicator itself, so that was the options could be purely const and never need to be modified.  Is that something that would be possible?

borrrden:
Also please avoid making merge commits inside of PRs.  Rebase if needed.

borrrden (approved):
This is a step in the right direction, but as discussed the next step is likely time consuming and not worth focusing on at the moment so other work should be prioritized.

pasin (approved):

PR #1593, opened 2022-10-17
CBL-3794: Errors in CollectionSGTest config

callumbirks:
Fix error in listed curl request.
Rename config.json to walrus_config.json.
Create a permanent config for CollectionSGTest, config.json.


callumbirks:
I believe it was mentioned that we eventually plan to deprecate walrus from our testing, so I believed it suitable to rename it, as `config.json` seems more suitable for the future default way of testing (if it is) - the CollectionSGTest using cb server.

pasin (changes requested):
As discussed in the sync meeting today:

1. For consistency, we should rename https://github.com/couchbase/couchbase-lite-core/blob/master/Replicator/tests/data/ssl_config.json as well.

2. Need to update README on https://github.com/couchbase/couchbase-lite-core/tree/master/Replicator/tests/data.

In addition, for the config that test with SG, by default we should test with SSL (This could be in a separate ticket).

pasin (approved):

PR #1594, opened 2022-10-17
CBL-3793: CollectionSGTest "Resolve Conflict SG" C4Error uninitialized

callumbirks:
Modify an error parameter so that C4Error is initialized properly.


pasin (changes requested):
As discussed in the sync meeting today:

1. For consistency, we should rename https://github.com/couchbase/couchbase-lite-core/blob/master/Replicator/tests/data/ssl_config.json as well.

2. Need to update README on https://github.com/couchbase/couchbase-lite-core/tree/master/Replicator/tests/data.

In addition, for the config that test with SG, by default we should test with SSL (This could be in a separate ticket).

pasin (approved):

jianminzhao (approved):

PR #1597, opened 2022-10-18
CBL-3806: Easier setup of CollectionSGTest

callumbirks:
Create shell script for easy SG setup.


callumbirks:
This shouldn't be merged until CBL-3794 is merged, as it relies on the config rename I made there.

jianminzhao (approved):

pasin (approved):
This would work but for enhancement, we could use the docker to run pull and start both SG and CB Server?

borrrden (approved):

callumbirks:
> This would work but for enhancement, we could use the docker to run pull and start both SG and CB Server?

Yeah I think Jim is working on something for this.

PR #1598, opened 2022-10-18
CBL-3774: Collection SG test, "Auto Purge Enabled - Revoke Access"

callumbirks:
Port test. Several adjustments made to fit into this test suite.


callumbirks:
Looks like I accidentally included a change in the comments at the top of the file, this change is also present though in the ongoing PR CBL-3794.

pasin:
PR Validation Fail.

pasin:
@jianminzhao What is the current strategy (e.g. using channels or docIDs filters) for setting up the replicator in tests with SG?

jianminzhao:
This test is not included in PR validation. The validation failure must be due to something else, a.k.a flaky test.

Replicator/tests/ReplicatorCollectionSGTest.cc
```diff
@@ -722,6 +722,103 @@ TEST_CASE_METHOD(ReplicatorCollectionSGTest, "Resolve Conflict SG", "[.SyncServe
     }
 }
 
+TEST_CASE_METHOD(ReplicatorCollectionSGTest, "Auto Purge Enabled - Revoke Access - SGColl", "[.SyncServerCollection]") {
+    _authHeader = "Basic c2d1c2VyOnBhc3N3b3Jk"_sl; // "sguser:password" base64 encoded
```
pasin, line 726: It seems like the other tests are using collectionPreamble() for bootstrapping the test with some common configuration. The test might also needs an extra setting that isolate each test's docs push to SG. Please check with @jianminzhao.

Replicator/tests/ReplicatorCollectionSGTest.cc
```diff
@@ -722,6 +722,103 @@ TEST_CASE_METHOD(ReplicatorCollectionSGTest, "Resolve Conflict SG", "[.SyncServe
...
+            if ((doc->flags & kRevPurged) == kRevPurged) {
+                ((ReplicatorAPITest*)context)->_docsEnded++;
+            }
+        }
+    };
+
+    auto collRoses = collections[0];
+
+    // Put doc in remote DB, in channels a and b
+    sendRemoteRequest("PUT", docID.asString(), R"({"channels":["a", "b"]})");
```
jianminzhao, line 775: This will put the doc to the default collection. Have you checked that doc actually landed on the CB server?

Replicator/tests/ReplicatorCollectionSGTest.cc
```diff
@@ -722,6 +722,103 @@ TEST_CASE_METHOD(ReplicatorCollectionSGTest, "Resolve Conflict SG", "[.SyncServe
...
+
+    // Pull doc into CBL:
+    C4Log("-------- Pulling");
+    replicate(paramsSetter);
```
jianminzhao, line 779: Not sure how you managed server db. I guess you empty the db before this test. Otherwise, you will pull all the docs from the db. I used docIDs to manage it.

Replicator/tests/ReplicatorCollectionSGTest.cc
```diff
@@ -722,6 +722,103 @@ TEST_CASE_METHOD(ReplicatorCollectionSGTest, "Resolve Conflict SG", "[.SyncServe
...
+
+    // Setup onDocsEnded:
+    _enableDocProgressNotifications = true;
+    _onDocsEnded = [](C4Replicator* repl,
+                      bool pushing,
+                      size_t numDocs,
+                      const C4DocumentEnded* docs[],
+                      void* context) {
+        for (size_t i = 0; i < numDocs; ++i) {
+            auto doc = docs[i];
```
jianminzhao, line 765: Have you checked whether numDocs == 0 or the following test failed?

callumbirks:
> This test is not included in PR validation. The validation failure must be due to something else, a.k.a flaky test.

I think it failed during the jenkins Windows or MacOS build because of an unused variable or something.

Replicator/tests/ReplicatorCollectionSGTest.cc
callumbirks: Yeah, because currently SGW only supports single collection, PUT request will put the document in the named collection. (If you are using a named collection with SGW, default collection is not used)

callumbirks: BUT when SGW supports multiple collections, that could change, so it would probably be a good idea for me to specify the collection to future-proof the test.

Replicator/tests/ReplicatorCollectionSGTest.cc
callumbirks: Ah yeah, that's a good point... I'll change it to use docIDs and also think I will make the test randomly generate docID at the beginning of the test so that previous test runs can never interfere.

Replicator/tests/ReplicatorCollectionSGTest.cc
```diff
@@ -722,6 +722,103 @@ TEST_CASE_METHOD(ReplicatorCollectionSGTest, "Resolve Conflict SG", "[.SyncServe
...
+
+    // Verify the update:
+    doc1 = c4coll_getDoc(collRoses, docID, true, kDocGetAll, nullptr);
+    REQUIRE(doc1);
+    CHECK(slice(doc1->revID).hasPrefix("2-"_sl));
+    CHECK(_docsEnded == 0);
+    CHECK(_counter == 0);
+
+    // Revoke access to all channels:
+    sendRemoteRequest("PUT", "_user/sguser", &status, &error, R"({"admin_channels":[]})", true);
```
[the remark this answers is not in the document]
callumbirks: collectionPreamble() does restore the admin channels for 'sguser'. But that's probably not clear enough in the test and you're right, test should guarantee this. And yeah creating another user could be a good idea to make sure 0 interference with other tests.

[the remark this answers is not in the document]
callumbirks: Nevermind, I think I was remembering the wrong thing. collectionPreamble() doesn't seem to restore admin channels.

Replicator/tests/ReplicatorCollectionSGTest.cc
jianminzhao: you are right, sendRemoteRequest uses REST API directly. It uses _authHeader, whereas _options is used by Replicator

Replicator/tests/ReplicatorCollectionSGTest.cc
jianminzhao: You're right, I heard the same thing: they only more collection, default or named. I prefer to do test on the named-collection. I have extended sendRemoteRequest to include collection path in the arguments. I cannot send the PR because test is not passing. I have submitted a ticket against SGW.

Replicator/tests/ReplicatorCollectionSGTest.cc
jianminzhao: "idPrefix" in the existing tests just does it.

callumbirks:
Test is now passing multiple times in a row without purging bucket or restarting SGW.
Achieved this by purging doc at the end of test.
Also now have temp user for test, as Pasin suggested.
Test is now also future-proofed against the multiple-collection support in SGW as I have specified the document's collection in the PUT request.

pasin (approved):

Replicator/tests/ReplicatorCollectionSGTest.cc
```diff
@@ -722,6 +722,116 @@ TEST_CASE_METHOD(ReplicatorCollectionSGTest, "Resolve Conflict SG", "[.SyncServe
...
+    replicate(paramsSetter);
+
+    // Verify that doc1 is purged:
+    doc1 = c4coll_getDoc(collRoses, slice(docIDstr), true, kDocGetAll, nullptr);
+    REQUIRE(!doc1);
+    CHECK(_docsEnded == 1);
+    CHECK(_counter == 1);
+
+    // Purge remote doc so test will succeed multiple times without flushing bucket
+    sendRemoteRequest("POST", "_purge", &status, &error, "{\"" + docIDstr + "\":[\"*\"]}", true);
```
jianminzhao, line 831: In this case, you don't rely on the uniqueness of the docIDs, 
   const string docIDstr = timePrefix() + "apera-doc1"; 
right?

callumbirks: Yeah, const string docIDstr = timePrefix() + "apera-doc1", so doc has unique ID.

Replicator/tests/ReplicatorCollectionSGTest.cc
```diff
@@ -849,6 +849,148 @@ TEST_CASE_METHOD(ReplicatorCollectionSGTest, "Resolve Conflict SG", "[.SyncServe
...
+    replicate(paramsSetter);
+
+    // Verify that doc1 is purged:
+    doc1 = c4coll_getDoc(collRoses, slice(docIDstr), true, kDocGetAll, nullptr);
+    REQUIRE(!doc1);
+    CHECK(_docsEnded == 1);
+    CHECK(_counter == 1);
+
+    // Purge remote doc so test will succeed multiple times without flushing bucket
+//    sendRemoteRequest("POST", "_purge", &status, &error, "{\"" + docIDstr + "\":[\"*\"]}", true);
```
pasin, line 990: If this is not necessary for cleaning up, you can remove it.

pasin (approved):

Replicator/tests/ReplicatorCollectionSGTest.cc
callumbirks: Ah, I tried to look through the code for any parts I'd left commented, thank you for spotting!

Replicator/tests/ReplicatorCollectionSGTest.cc
```diff
@@ -849,6 +849,148 @@ TEST_CASE_METHOD(ReplicatorCollectionSGTest, "Resolve Conflict SG", "[.SyncServe
...
+        {{ // three sets of braces? because Xcode
+            collectionSpecs[0], kC4Disabled, kC4OneShot,
+            opts.data(), nullptr, _pullFilter, this
+        }}
+    };
+
+    C4ParamsSetter paramsSetter = [&replCollections, &opts](C4ReplicatorParameters& c4Params) {
+        c4Params.collectionCount = replCollections.size();
+        c4Params.collections = replCollections.data();
+        for(auto it = Dict(opts).begin(); it != Dict(opts).end(); ++it) {
```
jianminzhao, line 908: I don't think you need this. kC4ReplicatorOptionChannels is an option per-collection. You already added it to the collection.

Replicator/tests/ReplicatorCollectionSGTest.cc
```diff
@@ -849,6 +849,148 @@ TEST_CASE_METHOD(ReplicatorCollectionSGTest, "Resolve Conflict SG", "[.SyncServe
...
+    const string docIDstr = idPrefix + "apera-doc1";
+    const string channelIDa = idPrefix + "a";
+    const string channelIDb = idPrefix + "b";
+
+    _authHeader = SGUserAuthHeader;
+
+    // Create a temporary user for this test
+    HTTPStatus status;
+    C4Error error;
+    sendRemoteRequest("POST", "_user", &status, &error, R"({"name":"purgeRevoke","password":"password"})", true);
```
jianminzhao, line 863: There is helper for this purpose, createTestUser, which is meant for user per test. Have. you rebased to the latest master?

Replicator/tests/ReplicatorCollectionSGTest.cc
jianminzhao: It won't do anything, but will be confusing

callumbirks: I see, I think some other error before tripped me up and made me think this was necessary, but I just tested the same point which tripped me up before and it's fine now. Was also confusing to me, so I'm glad you pointed it out.

Replicator/tests/ReplicatorCollectionSGTest.cc
callumbirks: Ah yeah, I forgot about that function. Will amend now.

callumbirks:
Tests are passing, turns out it was due to some strange SGW bug. Possibly related to [CBG-2487](https://issues.couchbase.com/browse/CBG-2487).

callumbirks:
I've had to modify the assignUserChannel function to take a parameter of user because it was hardcoded "sguser"

jianminzhao (approved):

callumbirks:
I've temporarily set the test's tag to "SyncServerCollectionTEMP" until we decide what to do with the test while CBG-2487 is still causing flaky failures.

Replicator/tests/ReplicatorCollectionSGTest.cc
```diff
@@ -1008,6 +1008,9 @@ TEST_CASE_METHOD(ReplicatorCollectionSGTest, "Update Once-Conflicted Doc - SGCol
     SG::TestUser testUser { _sg, "uocd", chIDs, collectionSpecs };
     _sg.authHeader = testUser.authHeader();
 
+    _sg.assignUserChannel("sguser", { collectionSpecs.begin(), collectionSpecs.end() }, { chIDs });
+    _sg.authHeader = HTTPLogic::basicAuth("sguser", "password");
+
```
jianminzhao, line 1013: Why are the above two lines?
"sguser" is a permanent user. My current thoughts are, we don't want to modify its channel access, because, if we do, we need to restore its original channel access after the test. Your TestUser is convenient class to use since the user would be automatically deleted after the test,

Replicator/tests/ReplicatorCollectionSGTest.cc
```diff
@@ -1068,6 +1071,116 @@ TEST_CASE_METHOD(ReplicatorCollectionSGTest, "Update Once-Conflicted Doc - SGCol
...
+        c4::ref<C4Document> doc1 = c4coll_getDoc(collections[i], slice(docIDstr), true, kDocGetAll, nullptr);
+        REQUIRE(doc1);
+        CHECK(slice(doc1->revID).hasPrefix("1-"_sl));
+
+        // Update doc to only channel 'b'
+        auto oRevID = slice(doc1->revID).asString();
+        REQUIRE(_sg.upsertDoc(collectionSpecs[i], docIDstr, oRevID, "{}", { channelIDb }));
+    }
+    CHECK(_docsEnded == 0);
+    CHECK(_counter == 0);
```
jianminzhao, line 1155: Odd to check them here, as if having anything to do with the preceding for-loop. Suggesting to move up right after replicate.

Replicator/tests/ReplicatorCollectionSGTest.cc
callumbirks: You're right, this looks like what used to be in the test. I think this was added back during a merge conflict resolution.

PR #1599, opened 2022-10-19
CBL-3775: Collection SG test, "Auto Purge Enabled - Filter Revoked Revision"

callumbirks:


pasin (commented):
My comment is the same as in https://github.com/couchbase/couchbase-lite-core/pull/1598

Replicator/tests/ReplicatorCollectionSGTest.cc
```diff
@@ -722,6 +722,97 @@ TEST_CASE_METHOD(ReplicatorCollectionSGTest, "Resolve Conflict SG", "[.SyncServe
     }
 }
 
+TEST_CASE_METHOD(ReplicatorCollectionSGTest, "Auto Purge Enabled - Filter Revoked Revision - SGColl", "[.SyncServerCollection]") {
+    _authHeader = "Basic c2d1c2VyOnBhc3N3b3Jk"_sl; // Admin user for REST requests
```
pasin, line 726: As we seem to need this in multiple tests, it's better to have it as a function that tests can be called.

pasin: Or if this is required whenever we call SG REST API, it could be automatically set in the test's setup?

pasin (changes requested):

Replicator/tests/ReplicatorCollectionSGTest.cc
```diff
@@ -722,6 +722,97 @@ TEST_CASE_METHOD(ReplicatorCollectionSGTest, "Resolve Conflict SG", "[.SyncServe
...
+    const string docIDstr = timePrefix() + "apefrr-doc1";
+
+
+    HTTPStatus status;
+    C4Error error;
+    sendRemoteRequest("POST", "_user", &status, &error, R"({"name":"filterRevoked","password":"password"})", true);
+    sendRemoteRequest("PUT", "_user/filterRevoked", &status, &error, R"({"admin_channels":["*"]})", true);
```
pasin, line 733: (Not for this PR) Maybe we should create a SG class that does all of the REST API with the APIs such as createUser(), deleteUser(), addDoc(s), getDoc(s), getDocIDs(), getDocCount(), purgeDoc() and etc.

borrrden: We absolutely should, and I think this is a good collaboration for @velicuvlad and @callumbirks to work on.  If there is a certain action that starts to take multiple steps each time it can be brought into the macro server as well.

Replicator/tests/ReplicatorCollectionSGTest.cc
```diff
@@ -722,6 +730,96 @@ TEST_CASE_METHOD(ReplicatorCollectionSGTest, "Resolve Conflict SG", "[.SyncServe
...
+            auto doc = docs[i];
+            if ((doc->flags & kRevPurged) == kRevPurged) {
+                ((ReplicatorAPITest*)context)->_docsEnded++;
+            }
+        }
+    };
+
+    sendRemoteRequest("PUT", docIDstr, &status, &error,
+                      "{\"channels\":[\"a\"],\"scopes\":{\"flowers\":{\"collections\":{\"roses\":{}}}}}"_sl);
+    REQUIRE(status == HTTPStatus::Created);
```
jianminzhao, line 790: I don't know whether the "scope" property will make SG to put the doc to the desired collection. If not, please use the overload that takes collection path as a paramter.

callumbirks: In order to do that would I need to merge/rebase master in order to access that function?

jianminzhao (approved):

pasin (approved):

borrrden:
Something has happened here again to require a rebase

callumbirks:
> Something has happened here again to require a rebase

I think it’s because lots of PRs going on modifying the same file, probably another test has been added. That happened before and I had to rebase

borrrden (approved):
Not wild about an instance variable with the nondescript name `_counter` (what is it counting?) but that is not part of this PR.  Also it would be better to use test specific context objects rather than just passing the entire `ReplicatorAPITest` as the context since that adds in variables that are specific to certain subsets of tests but not others.

callumbirks:
Similar to CBL-3774, have set the test tag to "SyncServerCollectionTEMP" until we've had a discussion about what to do until CBG-2487 is solved.

Replicator/tests/ReplicatorCollectionSGTest.cc
```diff
@@ -1014,6 +1014,92 @@ TEST_CASE_METHOD(ReplicatorCollectionSGTest, "Update Once-Conflicted Doc - SGCol
...
+
+    C4Log("-------- Pull the revoked");
+    replicate(replParams);
+
+    // Verify if doc1 is not purged as the revoked rev is filtered:
+    for(auto& coll : collections) {
+        c4::ref<C4Document> doc1 = c4coll_getDoc(coll, slice(docIDstr), true, kDocGetAll, nullptr);
+        REQUIRE(doc1);
+    }
+    CHECK(_docsEnded == collectionCount);
```
jianminzhao, line 1099: Is this CHECK failing now? Could you make a note if so?

jianminzhao (changes requested):

Replicator/tests/ReplicatorCollectionSGTest.cc
callumbirks: Yeah, that's right. Just done so in latest commit

jianminzhao (approved):

PR #1603, opened 2022-10-21
CBL-3790: Collection SG test, "Push & Pull Deletion"

velicuvlad:
- Test ported from ReplicatorSGTest to ReplicatorCollectionSGTest

### DIFF
--- a/Replicator/tests/ReplicatorCollectionSGTest.cc
+++ b/Replicator/tests/ReplicatorCollectionSGTest.cc
@@ -623,14 +623,12 @@ TEST_CASE_METHOD(ReplicatorCollectionSGTest, "Push & Pull Deletion SG", "[.SyncS
 
     std::array<C4CollectionSpec, collectionCount> collectionSpecs;
     std::array<C4Collection *, collectionCount> collections;
-    std::array<unordered_map<alloc_slice, unsigned>, collectionCount> docIDs;
     std::array<C4ReplicationCollection, collectionCount> replCollections;
 
     collectionSpecs = {
         Roses
     };
     collections = collectionPreamble(collectionSpecs, "sguser", "password");
-    docIDs[0] = getDocIDs(collections[0]);
     replCollections = {
         C4ReplicationCollection{collectionSpecs[0], kC4OneShot, kC4Disabled},
     };
@@ -657,6 +655,7 @@ TEST_CASE_METHOD(ReplicatorCollectionSGTest, "Push & Pull Deletion SG", "[.SyncS
     };
 
     createRev(collections[0], slice(docID), kRevID, kFleeceBody);
+
     replicate(paramsSetter);
 
     c4::ref<C4Document> remoteDoc = c4coll_getDoc(collections[0], slice(docID), true, kDocGetAll, nullptr);
--- a/Replicator/tests/ReplicatorCollectionSGTest.cc
+++ b/Replicator/tests/ReplicatorCollectionSGTest.cc
@@ -615,6 +615,59 @@ TEST_CASE_METHOD(ReplicatorCollectionSGTest, "Push and Pull Attachments SG", "[.
     }
 }
 
+TEST_CASE_METHOD(ReplicatorCollectionSGTest, "Push & Pull Deletion SG", "[.SyncServerCollection]") {
+    string idPrefix = timePrefix();
+    const string docID = idPrefix + "ppd-doc1";
+    constexpr size_t collectionCount = 1;
+    string revID;
+
+    std::array<C4CollectionSpec, collectionCount> collectionSpecs;
+    std::array<C4Collection *, collectionCount> collections;
+    std::array<unordered_map<alloc_slice, unsigned>, collectionCount> docIDs;
+    std::array<C4ReplicationCollection, collectionCount> replCollections;
+
+    collectionSpecs = {
+        Roses
+    };
+    collections = collectionPreamble(collectionSpecs, "sguser", "password");
+    docIDs[0] = getDocIDs(collections[0]);
+    replCollections = {
+        C4ReplicationCollection{collectionSpecs[0], kC4OneShot, kC4Disabled},
+    };
+
+    C4ParamsSetter paramsSetter = [&replCollections](C4ReplicatorParameters& c4Params) {
+        c4Params.collectionCount = replCollections.size();
+        c4Params.collections = replCollections.data();
+    };
+    
+    createRev(collections[0], slice(docID), kRevID, kFleeceBody);
+    createRev(collections[0], slice(docID), kRev2ID, kEmptyFleeceBody, kRevDeleted);
+    replicate(paramsSetter);
+
+    C4Log("-------- Deleting and re-creating database --------");
+    deleteAndRecreateDB();
+
+    collections = collectionPreamble(collectionSpecs, "sguser", "password");
+    replCollections = {
+        C4ReplicationCollection{collectionSpecs[0], kC4Disabled, kC4OneShot},
+    };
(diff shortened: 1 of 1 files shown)


pasin (approved):

Replicator/tests/ReplicatorCollectionSGTest.cc
```diff
@@ -615,6 +615,59 @@ TEST_CASE_METHOD(ReplicatorCollectionSGTest, "Push and Pull Attachments SG", "[.
...
+    };
+    
+    createRev(collections[0], slice(docID), kRevID, kFleeceBody);
+    createRev(collections[0], slice(docID), kRev2ID, kEmptyFleeceBody, kRevDeleted);
+    replicate(paramsSetter);
+
+    C4Log("-------- Deleting and re-creating database --------");
+    deleteAndRecreateDB();
+
+    collections = collectionPreamble(collectionSpecs, "sguser", "password");
```
borrrden, line 650: redundant line?

Replicator/tests/ReplicatorCollectionSGTest.cc
```diff
@@ -615,6 +615,59 @@ TEST_CASE_METHOD(ReplicatorCollectionSGTest, "Push and Pull Attachments SG", "[.
...
+    replCollections = {
+        C4ReplicationCollection{collectionSpecs[0], kC4Disabled, kC4OneShot},
+    };
+    paramsSetter = [&replCollections](C4ReplicatorParameters& c4Params) {
+        c4Params.collectionCount = replCollections.size();
+        c4Params.collections = replCollections.data();
+    };
```
borrrden, line 657: Seems also redundant

borrrden (approved):

Replicator/tests/ReplicatorCollectionSGTest.cc
callumbirks: When deleteAndRecreateDB is called, collections are invalidated, need to be created in DB again.

Replicator/tests/ReplicatorCollectionSGTest.cc
callumbirks: Is necessary to set up pull replication (previous replCollections, paramsSetter was for push). Perhaps this could be moved to an external function though because this is necessary in every test to change replication mode.

velicuvlad: - size will always be constant within a test.
- .data() returns the pointer and still refer to the first (only) element in of (which is the same) 
`replCollections = {
        C4ReplicationCollection{collectionSpecs[0], kC4OneShot, kC4Disabled},
    };`
`replCollections = {
        C4ReplicationCollection{collectionSpecs[0], kC4Disabled, kC4OneShot},
    };`

So in this case, this might be redundant if the memory if not altered... but can we assure this?

Replicator/tests/ReplicatorCollectionSGTest.cc
```diff
@@ -615,6 +615,59 @@ TEST_CASE_METHOD(ReplicatorCollectionSGTest, "Push and Pull Attachments SG", "[.
...
+    std::array<C4CollectionSpec, collectionCount> collectionSpecs;
+    std::array<C4Collection *, collectionCount> collections;
+    std::array<unordered_map<alloc_slice, unsigned>, collectionCount> docIDs;
+    std::array<C4ReplicationCollection, collectionCount> replCollections;
+
+    collectionSpecs = {
+        Roses
+    };
+    collections = collectionPreamble(collectionSpecs, "sguser", "password");
+    docIDs[0] = getDocIDs(collections[0]);
```
jianminzhao, line 633: Why do this? This will contain all exiting docs before this test.

Replicator/tests/ReplicatorCollectionSGTest.cc
jianminzhao: You may take advantage of verifyDB

Replicator/tests/ReplicatorCollectionSGTest.cc
jianminzhao: verifyDocs can help here.

Replicator/tests/ReplicatorCollectionSGTest.cc
```diff
@@ -615,6 +615,59 @@ TEST_CASE_METHOD(ReplicatorCollectionSGTest, "Push and Pull Attachments SG", "[.
...
+    collections = collectionPreamble(collectionSpecs, "sguser", "password");
+    replCollections = {
+        C4ReplicationCollection{collectionSpecs[0], kC4Disabled, kC4OneShot},
+    };
+    paramsSetter = [&replCollections](C4ReplicatorParameters& c4Params) {
+        c4Params.collectionCount = replCollections.size();
+        c4Params.collections = replCollections.data();
+    };
+
+    createRev(collections[0], slice(docID), kRevID, kFleeceBody);
```
jianminzhao, line 659: if you use verifyDocs, you don't have to recreate the 1-rev doc here.

Replicator/tests/ReplicatorCollectionSGTest.cc
borrrden: Does this actually create the collections?  The function name sounds like it is just creating a struct or something, in which case it is identical to the last time it was created.

Replicator/tests/ReplicatorCollectionSGTest.cc
borrrden: I retract this redundant comment since the underlying data changed (replCollections)

Replicator/tests/ReplicatorCollectionSGTest.cc
velicuvlad:
```// This function should be called before replicating against the Couchbase server.
    // It does the following:
    //  - sets up _options for authenticaton
    //  - assigns _collection with input "collection"
    //  - creates the collection if it is not the default collection
    //  - sets up the log level with input "logLevel"
    //  - returns the C4Collection object.
    template<size_t N>
    std::array<C4Collection*, N>
    collectionPreamble(std::array<C4CollectionSpec, N> collections,
                            const char* user, const char* password)

Replicator/tests/ReplicatorCollectionSGTest.cc
velicuvlad: what is verifyDB ?

Replicator/tests/ReplicatorCollectionSGTest.cc
velicuvlad: what is verifyDB ?

Replicator/tests/ReplicatorCollectionSGTest.cc
velicuvlad: copied form another test... is redundant, my bad - removed

Replicator/tests/ReplicatorCollectionSGTest.cc
callumbirks: The idea of the test is to: create a doc with 2 revs, 1st normal and 2nd deleted. Push this doc to SGW. Clear local DB. Create doc again but only with 1st (not deleted) rev, then pull into local again and make sure that the local doc was updated with the deleted rev.

velicuvlad: is using verifyDocs preffered?

Replicator/tests/ReplicatorCollectionSGTest.cc
borrrden: Yikes...that's a smelly function with a very underdescriptive name then but you are right that it needs to be called again.

borrrden (approved):

PR #1609, opened 2022-10-25
CBL-3820: Restore log levels of BLIPLog and BLIPMessagesLog to the le…

jianminzhao:
…vels prior to CBL- 2670/2671

### DIFF
--- a/Networking/BLIP/BLIPConnection.cc
+++ b/Networking/BLIP/BLIPConnection.cc
@@ -45,8 +45,8 @@ namespace litecore { namespace blip {
     const char* const kMessageTypeNames[8] = {"REQ", "RES", "ERR", "?3?",
                                               "ACKREQ", "AKRES", "?6?", "?7?"};
 
-    LogDomain BLIPLog("BLIP", LogLevel::Debug);
-    static LogDomain BLIPMessagesLog("BLIPMessages", LogLevel::Debug);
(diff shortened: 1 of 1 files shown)


Networking/BLIP/BLIPConnection.cc
```diff
@@ -45,8 +45,8 @@ namespace litecore { namespace blip {
     const char* const kMessageTypeNames[8] = {"REQ", "RES", "ERR", "?3?",
                                               "ACKREQ", "AKRES", "?6?", "?7?"};
 
-    LogDomain BLIPLog("BLIP", LogLevel::Debug);
-    static LogDomain BLIPMessagesLog("BLIPMessages", LogLevel::Debug);
+    LogDomain BLIPLog("BLIP", LogLevel::Warning);
+    static LogDomain BLIPMessagesLog("BLIPMessages", LogLevel::None);
```
borrrden, line 49: Is this changeable via CBL API?  If not, then what good is this domain doing?

jianminzhao: I don't know the original intent, but I think it's useful for Core debugging with Xcode.

jianminzhao: All I am here doing is to restore their values to the values before I changed them to catch more infos for the flaky failures I could only hope to see in Jenkins build.

borrrden (approved):

PR #1610, opened 2022-10-25
CBL-3783: Collection SG test, "Pinned Certificate Success"

callumbirks:
Ported test.
Update config and script for TLS.
Add new certs in Replicator/tests/data/cert.
Test passing.


Replicator/tests/data/cert/cert.pem
```diff
@@ -0,0 +1,17 @@
+-----BEGIN CERTIFICATE-----
```
jianminzhao, line 1: Where is this cert used?

jianminzhao (approved):

Replicator/tests/data/cert/cert.pem
callumbirks: I believe it's the same as the leaf certificate (the first cert in sg_cert.pem). It was generated by the scripts/gen_cert.sh script

PR #1612, opened 2022-10-25
CBL-3778: Collection SG test, "Auto Purge Enabled - Filter Removed Re…

jianminzhao:
…vision"

### DIFF
--- a/Replicator/tests/ReplicatorCollectionSGTest.cc
+++ b/Replicator/tests/ReplicatorCollectionSGTest.cc
@@ -109,6 +109,17 @@ public:
             c4db_release(verifyDb);
             verifyDb = nullptr;
         }
+        
+        if (hasCreatedTestUser) {
+            HTTPStatus status;
+            alloc_slice saveAuthHeader = _authHeader;
+            _authHeader = AdminAuthHeader;
+            sendRemoteRequest("DELETE", "_user/"s+TestUser, &status, ERROR_INFO(), nullslice, true);
+            _authHeader = saveAuthHeader;
+            if (status != HTTPStatus::OK) {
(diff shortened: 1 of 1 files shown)


Replicator/tests/ReplicatorCollectionSGTest.cc
```diff
@@ -1070,3 +1099,121 @@ TEST_CASE_METHOD(ReplicatorCollectionSGTest, "Auto Purge Enabled - Remove Doc Fr
...
+         int docsEndedPurge = 0;
+         int pullFilterTotal = 0;
+         int pullFilterPurge = 0;
+         void reset() {
+             docsEndedTotal = 0;
+             docsEndedPurge = 0;
+             pullFilterTotal = 0;
+             pullFilterPurge = 0;
+         }
+     } cbContext;
```
borrrden, line 1139: This is tons of lines duplicated with other tests, which is a recipe for unreliable maintainability.

borrrden (approved):
Careful of this building tech debt in the form of copy pasted code

Replicator/tests/ReplicatorCollectionSGTest.cc
jianminzhao: Do you refer to the above or below?

borrrden: Always above....below does not show up in the snippet above

jianminzhao: It's just like 4 local variables. Thought to take them out but could not find a meaningful standalone description and there are only 3 or 4 tests may share.

PR #1616, opened 2022-10-27
CBL-3781: Collection SG test, "Auto Purge Enabled(default) - Delete t…

jianminzhao:
…hen Create Doc"

### DIFF
--- a/Replicator/tests/ReplicatorCollectionSGTest.cc
+++ b/Replicator/tests/ReplicatorCollectionSGTest.cc
@@ -983,9 +983,7 @@ lTIN5f2LxWf+8kJqfjlj
     const std::array<C4CollectionSpec, collectionCount> collectionSpecs = {
             Roses
     };
-    std::array<C4Collection*, collectionCount> collections =
-            collectionPreamble(collectionSpecs, "sguser", "password");
-    (void)collections;
+    collectionPreamble(collectionSpecs, "sguser", "password");
     std::array<C4ReplicationCollection, collectionCount> replCollections {
             {{ // three sets of braces? because Xcode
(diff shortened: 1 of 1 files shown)


Replicator/tests/ReplicatorCollectionSGTest.cc
```diff
@@ -1419,44 +1419,83 @@ TEST_CASE_METHOD(ReplicatorCollectionSGTest, "Auto Purge Enabled(default) - Dele
...
+                REQUIRE(error.code == 0);
+                REQUIRE(docs[i] != nullptr);
+            }
+        }
+        for (auto coll : collections) {
+            REQUIRE(c4coll_getDocumentCount(coll) == 1);
+        }
+    }
+
+    SECTION("Delete without Create") {
```
borrrden, line 1467: This test is called "delete then create" and it has a section called "Delete without Create".  This seems to just add noise without meaning.

jianminzhao: I combined two tests in ReplicatorSGTest, "Auto Purge Enabled(default) - Delete Doc" & "Auto Purge Enabled(default) - Delete then Create Doc", into one with two SECTIONs.

borrrden: Then the naming of the test doesn't really match anymore, since it describes only one of the scenarios being tested.

jianminzhao: updated the name of the test to "Delete Doc or Delete then Create Doc SG"

borrrden (approved):

PR #1617, opened 2022-10-27
CBL-3831 : Prevent unnecessary growth of `RevTree._insertedData` (Port)

pasin:
* Port fix from the master branch (https://github.com/couchbase/couchbase-lite-core/commit/7b47459416ae572fa5b884bd8c690195acd82986) for 3.0.6 MR.

* The following message is copied from the original PR (https://github.com/couchbase/couchbase-lite-core/pull/1572)

> Currently, `RevTree._insertedData` is copied *as is* when copying a `RevTree`.
> 
> Why is this a problem?
> 
> When a document is repeatedly updated, `Rev`s start to be purged eventually. Also, `Rev` bodies that aren't needed anymore are removed.
> `RevTree._insertedData` is only ever added to and copied, keeping more and more data unnecessarily alive with each update. This data occupies memory, and the work required to copy `RevTree._insertedData` starts to dominate save operations.
> 
> This is currently making the [fast path](https://github.com/blaugold/couchbase-lite-core/blob/b5cb97c26a8d392b95afc33e161f666e30e554fe/C/c4Document.cc#L209-L224) of `C4Document::update` worse than the slow-path after a few updates.
> 
> This change keeps track of which data in a `Rev` is owned by `RevTree._insertedData` and only adds that data to `_insertedData` of a copied `RevTree`.

### DIFF
--- a/LiteCore/RevTrees/RawRevTree.cc
+++ b/LiteCore/RevTrees/RawRevTree.cc
@@ -143,6 +143,8 @@ namespace litecore {
 
     void RawRevision::copyTo(Rev &dst, const deque<Rev> &revs) const {
         const void* end = this->next();
+        dst._hasInsertedRevID = false;
+        dst._hasInsertedBody = false;
         dst.revID = {this->revID, this->revIDLen};
         dst.flags = (Rev::Flags)(this->flags & ~kPersistentOnlyFlags);
         auto parentIndex = endian::dec16(this->parentIndex_BE);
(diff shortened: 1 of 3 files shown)


pasin:
Note: The fix was directly ported from the master branch without modification.

pasin:
Note : this cannot be merged until 3.0.4 and 3.0.5 are done.

jianminzhao (approved):

PR #1620, opened 2022-10-28
CBL-3785: Collection SG test, "API Push Conflict"

jianminzhao:


borrrden (approved):
This test file name is starting to get confusing because the majority of these tests do not relate directly to collections.  What they are actually doing is just using a non-walrus sync gateway.  Even in the future there is not likely to be value added to many of them by using multiple collections.

jianminzhao:
I think mostly two things, 1. moving away from walrus, 2. using named collection. Where applicable, there may be multiple collections, or many collections. So far, we found one case where walrus and CB-backed SG exhibit different behavior. (where walrus behaves correct.)

PR #1622, opened 2022-11-01
CBL-3704: porting 3624, double free of BLIPIO (#1575)

jianminzhao:
The cause of this bug is in WeakHolder -- it currently works only if the callback it invokes is synchronous. We fix it by holding an extra strong reference inside WeakHolder.


Replicator/Replicator.cc
```diff
@@ -492,6 +487,7 @@ namespace litecore { namespace repl {
             notifyEndedDocuments();
             _delegate->replicatorConnectionClosed(this, status);
         }
+        _weakConnectionDelegateThis = nullptr;
```
pasin, line 490: This is an edge case ...  What if a replicator is created but is released without start, will it cause a memory leak?

jianminzhao: It's an excellent point. Must be paired. Will fix it. I need to find a solid pair.

pasin (approved):

PR #1623, opened 2022-11-02
CBL-3865: Remove Win32 Unicode implementation in favor of ICU

borrrden:
It is guaranteed to be available on all Windows versions that we support.


borrrden:
This actually cannot merge without a change to our backend, so I'm removing the reviewers.

PR #1628, opened 2022-11-04
CBL-3823: Extract SG REST API calls into a utility class for testing with SG

callumbirks:
Extracted SGW connection parameters out of ReplicatorAPITest into a separate struct SGConnection.
Created SGRest class with high-level functions for REST calls to SGW.
Moved sendRemoteRequest to this class.
Created SGRest::TestUser class for easier creation of user per test. Due to the class's destructor, it will be automatically deleted from SGW when it goes out of scope (at the end of the test).
Tested all SyncServer and SyncServerCollection tests, all are passing. Although seemingly related to [CBG-2487](https://issues.couchbase.com/browse/CBG-2487), the bucket must be flushed first. This may be something I have to investigate further next week to confirm it is the same issue.

The reason for using SGConnection everywhere is so that the SGRest class is not tied to ReplicatorAPITest.


borrrden (approved):
This is fine as a first step, but there is much more work to be done here.  The idea is to have actual calls that describe what is happening to replace all of the "sendRemoteRequest" calls.  For example, createDocs, deleteUser, etc

callumbirks:
That’s my plan. I wish to phase out usage of sendRemoteRequest. The rest of the API I’ve created operates independently of those calls, just there’s a lot of usages of sendRemoteRequest still and I don’t have time or scope to modify all of them, for this ticket anyway.

pasin (commented):
The PR would work but I think we can simplify the API by having only SGRest? For example, instead of using static function, we can have something like :

SGRest rest = SGRest(url );                 // https://localhost:4984/db
// OR SGRest rest = SGRest(config ); // config is like SGConnection
rest.config().basicAuthHeader = "sssss";
rest.config().pinnedCert = pinnedCert;
rest.flushDatabase();
rest.getServerName();
...

pasin (approved):
Then you could also have only one which is SGConnection or just SG instead of SGConnection and SGRest? Anyway, I will let SG team decide about the API. As it works, I will approve the PR.

callumbirks:
I've refactored per comments, into a class called SG, so that a call looks like `_sg.upsertDoc(...)` rather than `SGRest::upsertDoc(_sgConnection, ...)`.
Added some descriptive comments to SG.hh and SGTestUser.hh to describe the classes and how they can be used.

borrrden:
Looks like the CE builds broke with the latest commits, so add me back as a reviewer when that is fixed.

jianminzhao (approved):

borrrden (approved):

PR #1629, opened 2022-11-08
CBL-3824: Re-opening a database with an Index causes failure

callumbirks:
Wrote test to reproduce issue with some help from Blake.
Fixed issue (there was an assertion where there shouldn't be, so a function was not doing what it said it should).
Ran all LiteCore C and Cpp tests, all passing.

### DIFF
--- a/LiteCore/Database/DatabaseImpl.cc
+++ b/LiteCore/Database/DatabaseImpl.cc
@@ -579,17 +579,12 @@ namespace litecore {
             // CBL-3824: This is necessary for determining if the keystore is an index, which may occur in
             // the case of a Full-Text Index. The keystore name for an FTI looks like: <scope>.<collection>::<index>
             slice idxName = nullslice;
-            // Find the first ':'
-            if (auto idxSep = name.findByte(KeyStore::kIndexSeparator)) {
-                idxName = slice(idxSep+1, name.end());
-                // Find the second colon and set index name accordingly
-                if (auto idxSep2 = idxName.findByte(KeyStore::kIndexSeparator); idxSep2 == idxSep+1) {
-                    idxName.setStart(idxSep2+1);
-                    // make sure name does not contain the index name, otherwise later assertion will fail
-                    name.setEnd(idxSep);
-                } else {
-                    idxName = nullslice;
-                }
+            // Find the index separator "::"
+            if (auto idxSep = name.find(KeyStore::kIndexSeparator)) {
+                idxName = slice(idxSep.offset(2), name.end());
+                // Assert that index name is > 0
+                DebugAssert(idxName.size > 0);
+                name.setEnd(idxSep.buf);
             }
 
             DebugAssert((name == kC4DefaultCollectionName && scope == kC4DefaultScopeID)
--- a/LiteCore/Database/DatabaseImpl.cc
+++ b/LiteCore/Database/DatabaseImpl.cc
@@ -576,8 +576,26 @@ namespace litecore {
                 DebugAssert(isValidScopeNameOrDefault(scope));
                 name.setStart(slash + 1);
             }
-            if((name == kC4DefaultCollectionName && scope == kC4DefaultScopeID)
-                        || KeyStore::isValidCollectionName(name))
+            // CBL-3824: This is necessary for determining if the keystore is an index, which may occur in
+            // the case of a Full-Text Index. The keystore name for an FTI looks like: <scope>.<collection>::<index>
+            slice idxName = nullslice;
+            // Find the first ':'
+            if (auto idxSep = name.findByte(KeyStore::kIndexSeparator)) {
+                idxName = slice(idxSep+1, name.end());
+                // Find the second colon and set index name accordingly
+                if (auto idxSep2 = idxName.findByte(KeyStore::kIndexSeparator); idxSep2 == idxSep+1) {
+                    idxName.setStart(idxSep2+1);
+                    // make sure name does not contain the index name, otherwise later assertion will fail
+                    name.setEnd(idxSep);
+                } else {
+                    idxName = nullslice;
+                }
+            }
+
+            DebugAssert((name == kC4DefaultCollectionName && scope == kC4DefaultScopeID)
+                        || KeyStore::isValidCollectionName(name));
+            // If keystore name was not an index name
+            if(!idxName)
                 return {name, scope};
         }
         return {nullslice, nullslice};
--- a/LiteCore/Database/DatabaseImpl.cc
+++ b/LiteCore/Database/DatabaseImpl.cc
@@ -576,12 +576,11 @@ namespace litecore {
                 DebugAssert(isValidScopeNameOrDefault(scope));
                 name.setStart(slash + 1);
             }
-            DebugAssert((name == kC4DefaultCollectionName && scope == kC4DefaultScopeID)
-                        || KeyStore::isValidCollectionName(name));
-            return {name, scope};
-        } else {
-            return {nullslice, nullslice};
+            if((name == kC4DefaultCollectionName && scope == kC4DefaultScopeID)
+                        || KeyStore::isValidCollectionName(name))
+                return {name, scope};
         }
+        return {nullslice, nullslice};
     }
 
 
--- a/LiteCore/Storage/KeyStore.hh
+++ b/LiteCore/Storage/KeyStore.hh
@@ -75,6 +75,8 @@ namespace litecore {
         /// Set of characters allowed in a collection or scope name:
         static constexpr slice kCollectionNameCharacterSet
                             = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ01234567890_-%";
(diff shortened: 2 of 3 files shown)


pasin (commented):
Can you explain a bit why DebugAssert is causing "virtual tables may not be altered" error? and why did the assertion fail?

LiteCore/Database/DatabaseImpl.cc
```diff
@@ -576,12 +576,11 @@ namespace litecore {
                 DebugAssert(isValidScopeNameOrDefault(scope));
                 name.setStart(slash + 1);
             }
-            DebugAssert((name == kC4DefaultCollectionName && scope == kC4DefaultScopeID)
```
jianminzhao, line 579: So, we have have an instance of violating this assertion. What's it?

callumbirks: If the database has a Full Text Index, the function is called with “<scope>.<collection>::<index name>. This violates one of the checks, and therefore should return { nullslice, nullslice } as it is not a collection. In the test I added the REQURE(c4coll_deleteIndex) to ensure the index is still available once the database has been re-opened, and it is. So the change I’ve made here produces what the expected behaviour should have been.

callumbirks: Github seems to have taken some of my formatting, function is called with [scope].[collection]::[index name]

jianminzhao: I see. We probably should handle this exception rather than taking away the assertion altogether. Also make sure whether we should return the collection part of the key store name of simply null based on the calling code thread.

callumbirks: Per the comments above it, that function’s purpose is just to determine if whatever’s passed to it is a collection. And it’s up to the caller whether { nullptr, nullptr } is expected return or not. So for a FTI, it is expected to get back { nullptr, nullptr } as it is not a collection. Additionally, as it is DebugAssert, it will not do anything when built for release (the assertion only runs when building in Debug mode).

jianminzhao: Key table names are completely derived by us (the program). We should know the format. Here we encountered a surprise. We can account it. I don't at the moment whether we should return the collection part or null, but at least we should note that this a valid case.

jianminzhao:
I also have the same question as @pasin did.

callumbirks:
> Can you explain a bit why DebugAssert is causing "virtual tables may not be altered" error? and why did the assertion fail?

@pasin  I’m not exactly sure why it’s causing that specific error on the Lite platform side, I just know that assertion was failing when trying to open a DB that has a Full Text Index. Probably showing that error as this function was producing unexpected behaviour, and therefore was returning an improper error to Lite. 
The assertion failed because the check “isValidCollectionName()” failed, which checks that the name passed to it only contains characters in a specific list. Colon is passed to it with the FullTextIndex “scope.collection::indexName”, which is not within the list of characters for a valid collection name. Again though, this is expected as the FTI is not a collection, so we would expect { nullptr, nullptr } to be returned from the function, as is specified in the comments above the function.

LiteCore/Database/DatabaseImpl.cc
callumbirks: I’m not sure, I could debug it more and see where it’s called from. But due to the description above the function: “ If the KeyStore is not a collection, the collection & scope names will be `nullslice`”, I assumed it is expected that not all KeyStores passed to the function will be collections. 
My assumption is that all keyStores are passed to this function upon loading the DB to determine which are collections.

callumbirks: Although thinking about it, this issue is only encountered with Full Text Indexes, so a question I have is: Are all indexes saved as a keystore? If so, it may be that for some reason, other indexes are not being passed to this function, but FTIs are.

jianminzhao:
Were you able to repro the bug in the ticket if you remote the DebugAssert?

callumbirks:
The bug in the ticket seemed to be caused by the debugAssert. With the debugAssert there, I repro’d the ticket per the instructions Blake gave me, and fixed the error by removing the debugAssert.

jianminzhao:
I meant by only removing the assert, without your changes.

borrrden:
Something doesn't smell right about this.  I think the original problem would have been run in release mode, which skips this assert all together, and therefore removing the assertion cannot possibly be the answer.  Have you seen the problem reported in the ticket, or just an assertion failure?

LiteCore/Database/DatabaseImpl.cc
borrrden: Usually indexes are simply created in SQLite, and then recorded in an indexes table to ensure that they are not created again.  FTS indexes are special though because the full text engine inside of SQLite requires the use of virtual tables in order to function.

callumbirks:
@borrrden  I’ll have to run it again in the morning to verify, will do some extra debugging to figure out where this is being called from etc.

callumbirks:
This is the function which calls the function. So the intention here is to call the keyStoreNameToCollectionSpec function for every keystore, in order to determine which keystores are collections. So the behaviour I have produced with my changes is the expected behaviour.
```
void DatabaseImpl::startBackgroundTasks() {
        for (const string &name : _dataFile->allKeyStoreNames()) {
            if (CollectionSpec collSpec = keyStoreNameToCollectionSpec(name); collSpec.name) {
                if (_dataFile->getKeyStore(name).nextExpiration() > C4Timestamp::None) {
                    asInternal(getCollection(collSpec))->startHousekeeping();
                }
            }
        }
    }

**Review by pasin — COMMENTED** [review]
Thanks @callumbirks, now it's clear what is going on. Changing the `DebugAssert` to `if` check would work, but I think it's better to be specific by detecting whether the keystore name is the index one then return a null spec.

**Comment by callumbirks** [comment]
@pasin Yeah that makes sense, thanks. Will implement that 🙂.

File: LiteCore/Database/DatabaseImpl.cc [hunk 2]
```diff
@@ -576,12 +576,29 @@ namespace litecore {
                 DebugAssert(isValidScopeNameOrDefault(scope));
                 name.setStart(slash + 1);
             }
+            // CBL-3824: This is necessary for determining if the keystore is an index, which may occur in
+            // the case of a Full-Text Index. The keystore name for an FTI looks like: <scope>.<collection>::<index>
+            slice idxName = nullslice;
+            // Find the first ':'
+            if (auto idxSep = name.findByte(KeyStore::kIndexSeparator)) {
+                idxName = slice(idxSep+1, name.end());
+                // Find the second colon and set index name accordingly
```
borrrden, line 585: Is it necessary to find the second colon?  In your example it just seems to be one position after the first.  If that's the case then the second find is overkill and you can just check the following byte, or better yet use find("::"_sl)

callumbirks: Thanks Jim 🙂 I wasn't sure on the best solution, but this didn't feel like it. Yours is a much better one.

LiteCore/Storage/KeyStore.hh
```diff
@@ -75,6 +75,8 @@ namespace litecore {
         /// Set of characters allowed in a collection or scope name:
         static constexpr slice kCollectionNameCharacterSet
                             = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ01234567890_-%";
+        // Separator for an FTI keystore (i.e. <scope>.<collection>::<index>
+        static constexpr char kIndexSeparator = ':';
```
jianminzhao, line 79: Is it "::" ?

LiteCore/Database/DatabaseImpl.cc
```diff
@@ -576,12 +576,29 @@ namespace litecore {
...
+                }
+            }
+
             DebugAssert((name == kC4DefaultCollectionName && scope == kC4DefaultScopeID)
                         || KeyStore::isValidCollectionName(name));
-            return {name, scope};
-        } else {
-            return {nullslice, nullslice};
+            // If keystore name was not an index name
+            if(!idxName)
```
jianminzhao, line 598: It is okay, but in general empty() is a better predicate, for size == 0 makes !!idName == true

LiteCore/Storage/KeyStore.hh
callumbirks: Yeah, but because I did search for first colon, then second, I had to use ‘:’ twice. Per Jim’s suggestion though, I will adjust it to simply search for “::” instead of ‘:’ then ‘:’ again

LiteCore/Database/DatabaseImpl.cc
callumbirks: I simply did that because if the first or second colon are not found, idxName is set to nullslice. Would .empty() be invalid on a nullslice?

jianminzhao: It does not matter in this case, because you are so close to where it is set.

jianminzhao (approved):

pasin (approved):

PR #1630, opened 2022-11-11
CBL-3835: Function for creating SGCollectionTest repl params

callumbirks:
Created a wrapper class around C4ReplicatorParameters "ReplParams" which reduces a lot of re-used code in CollectionSGTest suite and provides higher-level functions for modifying replication parameters.
Modified all tests in the suite to use ReplParams rather than C4ParamsSetter.
ReplParams has a function that mimics C4ParamsSetter, which I've used to set up the replicator in the same way it would be with ParamsSetter.
The class manages a lot of its own memory for things like the options AllocedDicts, and the memory for ParamsSetter functions created from it, to prevent accidental access of de-allocated memory.

I've merged the ongoing PR branch CBL-3823 into this one, as that branch has a lot of refactoring work and I wanted to make sure all the tests will be working. This also will make merge easier later. (Therefore, CBL-3823 should be merged before this branch, although I believe CBL-3823 is ready to merge now).
I've added brief comments above the functions to describe what they do.


Replicator/tests/ReplParams.cc
```diff
@@ -0,0 +1,129 @@
...
+// Created by Callum Birks on 07/11/2022.
+//
+
+#include "ReplParams.hh"
+
+ReplParams::ReplParams(std::vector<C4ReplicationCollection> collections_, const AllocedDict& options,
+                       StatusCallback statusCallback, DocsEndedCallback docsEndedCallback,
+                       BlobProgressCallback blobProgressCallback, EncryptionCallback encryptionCallback,
+                       DecryptionCallback decryptionCallback, void *callbackContext_, C4SocketFactory *socketFactory_)
+{
```
jianminzhao, line 11: There are so many arguments in the constructor. Would you consider cascadable pattern, like
replParams.setA()
                   .setB()
...

Replicator/tests/ReplParams.cc
```diff
@@ -0,0 +1,129 @@
...
+    collections = _collectionVector.data();
+    collectionCount = _collectionVector.size();
+}
+
+AllocedDict ReplParams::setOptions(const AllocedDict& params, const AllocedDict& options) {
+    AllocedDict result = params;
+    for(Dict::iterator i(options); i; ++i) {
+        result = repl::Options::updateProperties(result, i.keyString(), i.value());
+    }
+    return result;
```
jianminzhao, line 55: Who is going to hold it? Should this class do?
I see, it is to be used internally for other setOptions methods. More apt as a private static method.

Replicator/tests/ReplParams.cc
```diff
@@ -0,0 +1,129 @@
...
+    for(auto& c : _collectionVector) {
+        c.pushFilter = pushFilter;
+    }
+}
+
+void ReplParams::setPullFilter(ValidationFunction pullFilter) {
+    for(auto& c : _collectionVector) {
+        c.pullFilter = pullFilter;
+    }
+}
```
jianminzhao, line 129: Is there a method assigning collectionContext?

borrrden (approved):
Jianmin has some valid suggestions.  I also notice that despite the huge constructor, you rarely (never?) use more than the first argument, so why is it there?  If those get fixed up the rest looks good!

Replicator/tests/ReplParams.cc
callumbirks: I simply did that so that the user can construct it like a C4ReplicatorParameters if they wish. As it inherits from C4ReplicatorParameters, they can set the other options directly; i.e. replParams.statusCallback = x.
Although, a cascadable pattern may also be something nice to implement.

Replicator/tests/ReplParams.cc
callumbirks: Yeah it is currently used only internally, I left it public simply to give a user a simple method for updating a dict. Although you're probably right, should be a private method.

Replicator/tests/ReplParams.cc
callumbirks: Ah you're right, something I missed.

callumbirks:
> Jianmin has some valid suggestions. I also notice that despite the huge constructor, you rarely (never?) use more than the first argument, so why is it there? If those get fixed up the rest looks good!

Simply as an option for the user, so they can construct the object almost exactly like they would construct a C4ReplicatorParameters. Although it does look ugly, so Jianmin's cascadable suggestion may be a good one.

jianminzhao (approved):

PR #1633, opened 2022-11-12
CBL-2532: Assertion failure when stopping replicator while replicator…

jianminzhao:
… is connecting

### DIFF
--- a/Networking/BLIP/LoopbackProvider.hh
+++ b/Networking/BLIP/LoopbackProvider.hh
@@ -63,7 +63,7 @@ namespace litecore { namespace websocket {
         }
 
         virtual void connect() override {
-            Assert(_driver && _driver->_peer);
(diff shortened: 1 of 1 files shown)


borrrden:
Context?

jianminzhao:
Both start and stop are asynchronous. Start moves to connect, and stop moves to close. So, assertion, Assert(_state < State::connecting) may not hold. It should be either State::unconnected or closed. In the case "closed," _peer will NULL.

borrrden:
How is the peer related to this?  I noticed the _peer non-null assertion moved from connect to _connect

jianminzhao:
From connection() to _connect(), it jumps a thread. We don't need to assert here. We assert in the worker's thread queue where _state is changed.

borrrden (approved):

PR #1636, opened 2022-11-16
CBL-3494: Change Options to use collectionSpec instead of collectionPath

callumbirks:
Changed Options::CollectionOptions::collectionPath to collectionSpec.
This has resulted in fewer conversions between path and spec, as most usages are of collectionSpec rather than collectionPath.

There was another suggestion in the ticket to make the `collectionOpts` vector private, but after attempting this I think it is unfeasible, or at least out of scope, as there are many usages in tests that directly access and modify the vector in a complex way.

### DIFF
--- a/Replicator/Replicator.cc
+++ b/Replicator/Replicator.cc
@@ -1229,8 +1229,7 @@ namespace litecore { namespace repl {
         // and this is expected.
         _db->useLocked([this](Retained<C4Database>& db) {
             for (CollectionIndex i = 0; i < _options->workingCollectionCount(); ++i) {
-                C4Collection* c = db->getCollection(Options::collectionPathToSpec(
-                                                    _options->collectionPath(i)));
+                C4Collection* c = db->getCollection(_options->collectionSpec(i));
                 if (c == nullptr) {
                     _subRepls.clear();
                     error::_throw(error::NotFound, "collection %s is not found in the database.",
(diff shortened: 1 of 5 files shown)


callumbirks:
> CollectionSpec is not heap-indigenous. Someone has to be responsible to hold it. You need to hold it in alloc_slices somewhere.

I ran all the C and C++ tests and they all ran fine, didn’t cause any memory issues

jianminzhao:
That's because someone is holding them.

borrrden (changes requested):
Jianmin is correct, this object should guarantee the lifetime of the thing it is holding.  I think instead of changing the constructor, you could make collectionPath private, and only expose a collectionSpec that is created from the passed collectionPath.

Also all EE builds failed because of a forgotten change somewhere it looks like?  Or is there another EE PR?

callumbirks:
> Jianmin is correct, this object should guarantee the lifetime of the thing it is holding. I think instead of changing the constructor, you could make collectionPath private, and only expose a collectionSpec that is created from the passed collectionPath.
> 
> Also all EE builds failed because of a forgotten change somewhere it looks like? Or is there another EE PR?

I don't understand this entirely, would this be because C4CollectionSpec consists of two `slice`s, rather than an `alloc_slice` - therefore it doesn't own the memory of the collectionSpec?

I built in EE, so I don't think so. Tested build in both CLion and Xcode.

callumbirks:
Added collectionPath back in as a private member, per Jim's recommendation. collectionSpec is now dependent on collectionPath (guaranteed lifetime). Merged master for CollectionSGTest updates so I could test to make sure all is working as expected.

borrrden (changes requested):
EE PR validation still failing, please check it

callumbirks:
> EE PR validation still failing, please check it

@borrrden It seems like the file "C4LocalReplicator.hh" is not being built in any of my builds (For any of the build targets I selected), which'll be why I missed the problem. I do have EE enabled so I'm not sure why. Seems like the file is from the EE repo, so that'll require a separate PR.

jianminzhao (approved):

borrrden (approved):

pasin (approved):

PR #1641, opened 2022-11-25
CBL-3921 (workaround): Batch bulk docs

callumbirks:
Modify insertBulkDocs function to batch docs into groups of 200.
I've tested this with as many as 10,000 docs and it works successfully.
This is a workaround to the macOS-specific issue detailed in CBL-3921.

### DIFF
--- a/Replicator/tests/SG.cc
+++ b/Replicator/tests/SG.cc
@@ -175,14 +175,14 @@ bool SG::insertBulkDocs(C4CollectionSpec collectionSpec, slice docsDict_) const 
         return false;
     }
 
+    MutableArray remainingDocsArray = remainingDocs["docs"_sl].asArray().mutableCopy();
+
     C4Log("Bulk Docs: Sending %u docs to SGW", remainingDocs["docs"_sl].asArray().count());
 
     size_t batchNum = 1;
 
     // While there are docs remaining to send
-    for(fleece::Array remainingDocsArray = remainingDocs["docs"_sl].asArray();
-                      remainingDocsArray.count() > 0;
-                      remainingDocsArray = remainingDocs["docs"_sl].asArray())
+    while(remainingDocsArray.count() > 0)
     {
         const uint32_t remainingDocsCount = remainingDocsArray.count();
         // batchSize is the smallest of remainingDocsCount and bulkDocsBatchSize
@@ -191,7 +191,6 @@ bool SG::insertBulkDocs(C4CollectionSpec collectionSpec, slice docsDict_) const 
         C4Log("Bulk Docs: Batch #%zu - %u docs - %u remaining", batchNum++, batchSize, remainingDocsCount - batchSize);
 
         JSONEncoder encSend;
-        JSONEncoder encRetain;
         // Start current batch
         encSend.beginDict();
         encSend.writeKey("docs"_sl);
@@ -208,17 +207,8 @@ bool SG::insertBulkDocs(C4CollectionSpec collectionSpec, slice docsDict_) const 
             C4Log("Bulk Docs: ERROR completing REST request: %s", error.description().c_str());
             return false;
         }
-
-        // Put all remaining docs back into `remainingDocs`
-        encRetain.beginDict();
-        encRetain.writeKey("docs"_sl);
-        encRetain.beginArray();
-        for(uint32_t i = (uint32_t)batchSize; i < remainingDocsCount; ++i) {
-            encRetain.writeValue(remainingDocsArray.get(i));
-        }
-        encRetain.endArray();
-        encRetain.endDict();
-        remainingDocs = FLDoc_FromJSON(encRetain.finish(), nullptr);
+        // Remove the batch from remainingDocs
+        remainingDocsArray.remove(0, batchSize);
     }
     C4Log("Bulk Docs: Complete!");
     return status == HTTPStatus::Created;
--- a/Replicator/tests/SG.cc
+++ b/Replicator/tests/SG.cc
@@ -161,9 +161,66 @@ void SG::flushDatabase() const {
     runRequest("POST", { }, "_flush", nullslice, true);
 }
 
-bool SG::insertBulkDocs(C4CollectionSpec collectionSpec, const slice docsDict) const {
-    HTTPStatus status;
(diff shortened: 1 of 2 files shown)


velicuvlad (approved):

Replicator/tests/SG.cc
```diff
@@ -161,9 +161,66 @@ void SG::flushDatabase() const {
...
+                      remainingDocsArray = remainingDocs["docs"_sl].asArray())
+    {
+        const uint32_t remainingDocsCount = remainingDocsArray.count();
+        // batchSize is the smallest of remainingDocsCount and bulkDocsBatchSize
+        const uint32_t batchSize = remainingDocsCount < bulkDocsBatchSize ? remainingDocsCount : bulkDocsBatchSize;
+
+        C4Log("Bulk Docs: Batch #%zu - %u docs - %u remaining", batchNum++, batchSize, remainingDocsCount - batchSize);
+
+        JSONEncoder encSend;
+        JSONEncoder encRetain;
```
borrrden, line 194: Instead of making this, and recreating a slightly smaller array each time, I suggest making the original array mutable and removing the first `batchSize` elements each time.

Replicator/tests/SG.cc
```diff
@@ -161,9 +161,66 @@ void SG::flushDatabase() const {
...
+    fleece::Doc remainingDocs = FLDoc_FromJSON(docsDict, nullptr);
+
+    // If dict is not in correct format for bulk docs
+    if(!remainingDocs["docs"_sl] || remainingDocs.asDict().count() != 1) {
+        C4Log("ERROR: Incorrect format for bulk docs!");
+        return false;
+    }
+
+    C4Log("Bulk Docs: Sending %u docs to SGW", remainingDocs["docs"_sl].asArray().count());
+
```
borrrden, line 179: Create remainingDocsArray here as `remainingDocsArray = remainingDocs["docs"_sl].asArray().mutableCopy()`

Replicator/tests/SG.cc
```diff
@@ -161,9 +161,66 @@ void SG::flushDatabase() const {
...
+    size_t batchNum = 1;
+
+    // While there are docs remaining to send
+    for(fleece::Array remainingDocsArray = remainingDocs["docs"_sl].asArray();
```
borrrden, line 183: `while(remainingDocsArray.count() > 0)` after the other changes

Replicator/tests/SG.cc
```diff
@@ -161,9 +161,66 @@ void SG::flushDatabase() const {
...
+        encSend.endArray();
+        encSend.endDict();
+        runRequest("POST", collectionSpec, "_bulk_docs", encSend.finish(), false, WITH_ERROR(error), &status, false);
+        if (status != HTTPStatus::Created) {
+            C4Log("Bulk Docs: ERROR completing REST request: %s", error.description().c_str());
+            return false;
+        }
+
+        // Put all remaining docs back into `remainingDocs`
+        encRetain.beginDict();
```
borrrden, line 213: This entire block can be replaced, after the other changes, with `remainingDocsArray.remove(0, batchSize)`

borrrden (changes requested):

Replicator/tests/SG.cc
callumbirks: Thanks Jim, I knew there must be a more efficient way of doing this

callumbirks:
@borrrden Made the changes you suggested, tested with 10k docs.

jianminzhao:
We'd better understand why we have to do it, I think.

callumbirks:
@jianminzhao I’d agree. This is simply a workaround (for now) if we want it. Or if this is down to something to do with MacOS that we can’t fix.

jianminzhao:
Then, put whatever you know in the comment. You mentioned the problem does not appear in Windows, what about Linux?

callumbirks:
Jim has run the test himself with 1000 docs and the test passes fine, the issue does not seem to appear. I’m also going to test Windows myself now that I have a Windows machine. Not sure about Linux, would have to set up some sort of VM or a dual boot on my Windows machine.

callumbirks:
The “Pull deltas …” collectionSGTest that Vlad ported will trigger this issue if you set kNumDocs to 1000 (or anything above ~200). 1000 is what the test should be set to, but we changed it to 50 because of this issue.

jianminzhao:
I used WSL and it worked for me.

borrrden (approved):

jianminzhao:
Let's compare the behavior on individual machines w/o this workaround. I am having Mac/x64. For me, the whole bulk, 1000 docs and 1000 revs, all landed on SGW. The return status  of the REST call is not good, but all ensuing checks of "Pull deltas from Collection SG" are passed.

callumbirks:
In the case of running the test on my MacOS/arm64 machine. Setting the number of docs to 1000, I get the error output described in the test. Test continues but fails on `CHECK(slice(info.revID).hasPrefix("2-"_sl));`. I believe this is because it failed to write all docs during the `_bulk_docs` call, so some docs are still on 1st rev generation.

If I wrap line 707 `_sg.insertBulkDocs(...)` in a `REQUIRE()` statement, it fails on that line.

callumbirks:
Correction: I just got it to pass by doing `setTimeout(60)` during `SG::createRequest()`. This is what Jim said before got his test to pass on Windows. I'm sure I tried that and it didn't work for me, but that may have been for my other test which has over 16k docs (iTunesMusicLibrary). So perhaps if I add a timeout parameter to the `insertBulkDocs` function, and the test implementer will have to find a timeout value that doesn't fail for their test and use that.

callumbirks:
Yeah if I set kNumDocs to 10,000, the test fails with timeout of 60. So I think that test (CBL-3789) will need a much higher timeout value.

callumbirks:
I can close this PR and create a PR which just adds a timeout parameter to `insertBulkDocs()` and the other functions such as `runRequest()`.

callumbirks:
I think it could also be useful if we can get a more helpful error output when this does occur (such as "Request timed out" or similar)

callumbirks:
Closing because we can fix this problem by setting timeout on a per-test basis where required.

PR #1642, opened 2022-11-29
Allow DictKeys to cache shared keys from query results

blaugold:
This is a follow up to https://github.com/couchbase/fleece/pull/168 to ensure shared keys from query results will still be cachable in `DictKey`s once that PR makes it into LiteCore.

### DIFF
--- a/LiteCore/Query/SQLiteQuery.cc
+++ b/LiteCore/Query/SQLiteQuery.cc
@@ -401,6 +401,18 @@ namespace litecore {
     };
 
 
+    // In-memory SharedKeys implementation that is used exactly once to encode keys in a
+    // query result set.
+    class QueryResultSetKeys final : public fleece::impl::PersistentSharedKeys {
+    protected:
+        virtual bool read() override {
+            return true; // No-op
+        }
+        virtual void write(slice encodedData) override {
+            assert(false); // Should never be called
+        }
+    };
+
 
     // Reads from 'live' SQLite statement and records the results into a Fleece array,
     // which is then used as the data source of a SQLiteQueryEnum.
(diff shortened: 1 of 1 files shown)


snej (changes requested):
Hm. This workaround should not be needed. DictKeys should be able to cache keys from a regular non-persistent SharedKeys. Something must be wrong in that prior Fleece change.

... OK, it looks like SharedKeys::isInTransaction() is returning true for a base SharedKeys, when it shouldn't. Yes, `_inTransaction` is true, but that's just for simplicity in the implementation; it basically means `_isMutable`. Probably the best fix is to make the new `inTransaction` method virtual and have the SharedKeys implementation return false, while the PersistentSharedKeys implementation returns `_inTransaction`.

blaugold:
> ... OK, it looks like SharedKeys::isInTransaction() is returning true for a base SharedKeys, when it shouldn't. Yes, _inTransaction is true, but that's just for simplicity in the implementation; it basically means _isMutable. Probably the best fix is to make the new inTransaction method virtual and have the SharedKeys implementation return false, while the PersistentSharedKeys implementation returns _inTransaction.

Sounds good. But we'll need to address an issue in replication pull filters, that implicitly got fixed by basic `SharedKeys` always returning `true` for `inTransaction()`. The body of documents passed to pull filters are encoded with temporary `SharedKeys`. 

https://github.com/couchbase/couchbase-lite-core/blob/3e2d634633fc2b1334d3b969bde04ad211c46fa1/Replicator/IncomingRev.cc#L169

https://github.com/couchbase/couchbase-lite-core/blob/3e2d634633fc2b1334d3b969bde04ad211c46fa1/Replicator/DBAccess.cc#L288-L333

If `DictKey`s are used with these documents and allowed to cache shared keys, the `DictKey`s might pick up shared keys from the temporary `SharedKeys` which are not valid to use in the context of documents that got inserted into the database.

snej:
I think the solution may be to add a `cacheable` flag to SharedKeys. It defaults to true, but there's a `disableCaching` method that forces it to return false. PersistentSharedKeys sets it to false while it's in a transaction.

DictKey can then check `sk->cacheable()` instead of `!sk->inTransaction()`.

And DBAccess would call `disableCaching` on its tempSharedKeys.

jianminzhao:
This issue is moved to CBL-4547 where the underlying issue in Fleece is addressed.

PR #1643, opened 2022-11-29
CBL-3915: Add kC4NetErrUnknownInterface to C4NetworkErrorCode

velicuvlad:


C/include/c4Error.h
```diff
@@ -116,7 +116,8 @@ typedef C4_ENUM(int32_t, C4NetworkErrorCode) {
     kC4NetErrHostDown,                  // The other side reports it is down [EHOSTDOWN, retryable]
     kC4NetErrHostUnreachable,           // There is no network path to the host [EHOSTUNREACH, retryable]
     kC4NetErrAddressNotAvailable,       // The address in question is already being used [EADDRNOTAVAIL, retryable]
-    kC4NetErrBrokenPipe,                // Broken pipe [EPIPE, retryable]
+    kC4NetErrBrokenPipe,
+    kC4NetErrUnknownInterface,                // Broken pipe [EPIPE, retryable]
```
pasin, line 120: The comment is in the wrong line and please add the comment for the `kC4NetErrUnknownInterface`(e.g. An unknown network interface is specified) as well.

Networking/WebSockets/WebSocketInterface.hh
```diff
@@ -83,8 +83,9 @@ namespace litecore { namespace websocket {
         kNetErrHostUnreachable,
         kNetErrAddressNotAvailableAIL,
         kNetErrBrokenPipe,
+        kC4NetErrUnknownInterface,
         // Add new codes here. You MUST add messages to kLiteCoreMessages!
-        // You MUST add corresponding kC4NetErr codes to the enum in C4Base.h!
+        // You MUST add corresponding kC4NetErr codes to the enum in c4Error.h!
```
pasin, line 88: Not sure about adding message to kLiteCoreMessages. It seems like that should be done as well.

pasin (changes requested):

Networking/WebSockets/WebSocketInterface.hh
```diff
@@ -83,8 +83,9 @@ namespace litecore { namespace websocket {
         kNetErrHostUnreachable,
         kNetErrAddressNotAvailableAIL,
         kNetErrBrokenPipe,
+        kC4NetErrUnknownInterface,
```
jianminzhao, line 86: Here, by convention, we drop C4 in the name, but double check it gets the number as the C4 one.

Networking/WebSockets/WebSocketInterface.hh
velicuvlad: added inside kNetworkMessages as is a NetErr

pasin (changes requested):
There is also fleece submodule update in this PR. If it's not related to this PR, can you back that off?

velicuvlad:
removed the fleece update

pasin (approved):

PR #1644, opened 2022-11-29
CBL-3921: Error when sending _bulk_docs request with large number of …

jianminzhao:
…docs

There are two occasions of errors:
1.  [TLS] ERROR: mbedTLS(C): mbedtls_ssl_send_alert_message() returned -80 (-0x0050) This error is due to the fact that we send the request twice. First time we send it without proper auth-header, and SG may responds with 401/authentation without reading out the entire body of the request. At the socket level, it finds socket closed as it is finishing up the write. We can disregard this error in this case.
2. [WS] WARNING: ClientSocket got POSIX error 35 "Resource temporarily unavailable." This is actually due to socket timeout. I added warning in this case, so that we can decide whether to dial up the timeout or reduce the size of the bulk. To match the size (1000 docs) used in the corresponding test with Walrus, I needed to bump the timeout from default 5 seconds to 20 seconds.

### DIFF
--- a/REST/Response.hh
+++ b/REST/Response.hh
@@ -87,6 +87,7 @@ namespace litecore { namespace REST {
         Response& setBody(fleece::slice body);
         Response& setTLSContext(net::TLSContext*);
         Response& setProxy(const net::ProxySpec&);
+        double    getTimeout() const                    { return _timeout; }
         Response& setTimeout(double timeoutSecs)        {_timeout = timeoutSecs; return *this;}
 
         Response& allowOnlyCert(fleece::slice certData);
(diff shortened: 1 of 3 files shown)


pasin (approved):

borrrden (commented):
Where is the timeout increase?

callumbirks:
Agree with Jim, I was expecting adding a timeout parameter to `insertBulkDocs`, `createRequest` and `runRequest`. Then the line inside `createRequest` with `setTimeout(5)` which would need changing to `setTimeout(timeout)`. Timeout parameter could have a default value of 5, as this was the setting being used before.

callumbirks (commented):
^ Above comment, meant to do it as a review.

callumbirks:
I see, that makes a lot of sense. Will be helpful to have good error output when we are hitting a timeout, and easier diagnose the problem.

PR #1646, opened 2022-11-30
CBL-2870: Change SQL Log Level

callumbirks:
(I believe) I've changed the SQL log level to Verbose (per Jim's suggestion). Although I'm not entirely sure if this is the right thing to change, could anyone confirm this for me?


borrrden (approved):
LGTM.  Without that change SQL domain will never be able to log anything above the warning level, even if the overall log level is set otherwise.

jianminzhao:
Reading the CBL ticket, @pasin, it seems that the intention is to reduce the logging for SQL, but it also suggests to change the log level of the from Info to Debug, which will allow more logs. 
Current it is Warning, which means we won't see logs unless they are warnings or errors, and changing it to Verbose will allow more logs. 
Confused about the purpose of the PR.
By the way, SQL currently is not exposed to C4 and I had to play hack to show SQL logs when I debugged in CBL-ios the other day.

callumbirks:
This was solved in a previous commit https://github.com/couchbase/couchbase-lite-core/commit/a82a5bbc1e56051c8552727b8081516342b490c0

PR #1647, opened 2022-11-30
CBL-3829: Create query with SQL++ is slow on complex query

jianminzhao:


borrrden:
Does this test belong in the perf suite for showfast?

borrrden (approved):
How will we be alerted if this test "fails" ?

jianminzhao:
"Does this test belong in the perf suite for showfast?"  -- do you mean "N1QL benchmark

jianminzhao:
Did not mean to close the PR. I thought I was cancelling the comment editing.

borrrden:
Yes....the test you added in this PR "N1QL performance"

jianminzhao:
1. "Does this test belong in the perf suite for showfast?"
2."How will we be alerted if this test "fails" ?
Do you refer to the new tests in N1QL? I assume you do.
I don't know much about "showfast," but it's not a big performance thing; each test is only about half second thing. The purpose I put them, so that next time if we introduce new grammar rules, the tests will bark if the they degrade the performance significantly. Of course, if the grammar rules are very valuable, we may still allow them and increase the time allowance, but at least we won't get caught too late.

borrrden:
The tests may bark but will anyone hear it?  I am assuming it's just a log line in the middle of a gigantic log.  If on showfast there will be a graph that is easy to quickly notice a performance regression -> http://showfast.sc.couchbase.com/#/timeline/Linux/lite/core/macOS (VPN needed)

jianminzhao:
The PR validation will fail, forcing the author to find out why and raise the issue as whether to increase the time allowance. Right now, two tests each is limited to 0.5sec.

pasin (approved):

Replicator/tests/ReplicatorCollectionTest.cc
```diff
@@ -566,23 +566,41 @@ TEST_CASE_METHOD(ReplicatorCollectionTest, "Multiple Collections Incremental Rev
...
+            }
+            if (_statusReceived.progress.documentCount < _expectedDocumentCount) {
+                return;
+            }
+            // It seems that windows will delete the lambda object  after
+            // _callbackWhenIdle = nullptr. Therefore, capture it beforehand.
+            auto self = this;
+            self->_callbackWhenIdle = nullptr;
+            self->_stopOnIdle = true;
+            self->_checkStopWhenIdle();
```
pasin, line 594: The change here doesn't seem to be related to the PR.

pasin (approved):

borrrden:
Ok I see, that's somewhat fragile (depends on the machine's power) so let's see how often it gets hit.

Replicator/tests/ReplicatorCollectionTest.cc
jianminzhao: yes. It is to push the PR through validation pipeline more predictably. I have recently seen it fail many times, that forcing me to restart.

PR #1648, opened 2022-12-01
CBL-3961: Add timeout parameter to SG functions

callumbirks:
Added timeout parameter to `SG::insertBulkDocs()`, `SG::runRequest()` and `SG::createRequest()`.
`runRequest()` and `createRequest()` have default values of 5 for the timeout, as this was the value being used before.
`insertBulkDocs()` has a default value of 30 for the timeout, as this seems reasonable for most `_bulk_docs` requests to pass.
Have tested "Pull deltas from Collection SG" with a value of 5, confirmed that Jianmin's timeout error was output.
Updated "Pull deltas from Collection SG" to call `insertBulkDocs()` with a timeout of 30, tested and passing fine.

### DIFF
--- a/Replicator/tests/ReplicatorCollectionSGTest.cc
+++ b/Replicator/tests/ReplicatorCollectionSGTest.cc
@@ -617,7 +617,7 @@ TEST_CASE_METHOD(ReplicatorCollectionSGTest, "Pull deltas from Collection SG", "
     constexpr size_t kDocBufSize = 60;
 
     // connection closing from SGW for 1000 docs => reduced to 50
-    constexpr int kNumDocs = 50, kNumProps = 50;
+    constexpr int kNumDocs = 1000, kNumProps = 50;
     string revID;
 
     const string docIDPref = idPrefix + "doc";
@@ -704,7 +704,7 @@ TEST_CASE_METHOD(ReplicatorCollectionSGTest, "Pull deltas from Collection SG", "
         encUpdate.endArray();
         encUpdate.endDict();
         for (size_t i = 0; i < collectionCount; ++i) {
-            REQUIRE(_sg.insertBulkDocs(collectionSpecs[i], encUpdate.finish()));
+            REQUIRE(_sg.insertBulkDocs(collectionSpecs[i], encUpdate.finish(), 30.0));
         }
     }
 
--- a/Replicator/tests/SG.cc
+++ b/Replicator/tests/SG.cc
@@ -14,6 +14,7 @@ std::unique_ptr<REST::Response> SG::createRequest(
         std::string path,
         slice body,
         bool admin,
+        double timeout,
         bool logRequests
 ) const {
     auto port = uint16_t(address.port + !!admin);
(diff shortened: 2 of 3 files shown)


Replicator/tests/ReplicatorCollectionSGTest.cc
```diff
@@ -617,7 +617,7 @@ TEST_CASE_METHOD(ReplicatorCollectionSGTest, "Pull deltas from Collection SG", "
     constexpr size_t kDocBufSize = 60;
 
     // connection closing from SGW for 1000 docs => reduced to 50
-    constexpr int kNumDocs = 50, kNumProps = 50;
+    constexpr int kNumDocs = 1000, kNumProps = 50;
```
velicuvlad, line 620: The original test with Walrus had both **kNumDocs = 1000, kNumProps = 1000;**. Could you please change it back to that and test as well?

velicuvlad (changes requested):

callumbirks:
Tested with kNumProps at 1000 too, passing.

Replicator/tests/ReplicatorCollectionSGTest.cc
```diff
@@ -617,7 +617,7 @@ TEST_CASE_METHOD(ReplicatorCollectionSGTest, "Pull deltas from Collection SG", "
     constexpr size_t kDocBufSize = 60;
 
     // connection closing from SGW for 1000 docs => reduced to 50
```
jianminzhao, line 619: Could you remove this outdated comment?

Replicator/tests/SG.cc
```diff
@@ -166,9 +168,9 @@ void SG::flushDatabase() const {
     runRequest("POST", { }, "_flush", nullslice, true);
 }
 
-bool SG::insertBulkDocs(C4CollectionSpec collectionSpec, const slice docsDict) const {
+bool SG::insertBulkDocs(C4CollectionSpec collectionSpec, const slice docsDict, double timeout) const {
```
borrrden, line 171: Have a default value so that you don't need to type in one every time?

callumbirks: It's in the `.hh` file

Replicator/tests/ReplicatorCollectionSGTest.cc
```diff
@@ -704,7 +703,7 @@ TEST_CASE_METHOD(ReplicatorCollectionSGTest, "Pull deltas from Collection SG", "
         encUpdate.endArray();
         encUpdate.endDict();
         for (size_t i = 0; i < collectionCount; ++i) {
-            REQUIRE(_sg.insertBulkDocs(collectionSpecs[i], encUpdate.finish()));
+            REQUIRE(_sg.insertBulkDocs(collectionSpecs[i], encUpdate.finish(), 30.0));
```
borrrden, line 706: This 30.0 is not needed explicitly here anymore right?  No need to change just for this PR though.

borrrden (approved):

Replicator/tests/ReplicatorCollectionSGTest.cc
callumbirks: I simply did it so it is clear to the reader of the test what timeout value is being used, understand it’s not really necessary though

velicuvlad (approved):

PR #1649, opened 2022-12-02
CBL-3010: LiteCore delivers unrecoverable error in BUSY state

callumbirks:
I've changed `c4ReplicatorImpl::onStateChanged` so that errors are only logged when `status.level` <= kC4Offline.

### DIFF
--- a/Replicator/c4ReplicatorImpl.hh
+++ b/Replicator/c4ReplicatorImpl.hh
@@ -485,9 +485,11 @@ namespace litecore {
                 }
             }
 
-            auto onStatusChanged = _onStatusChanged.load();
-            if (onStatusChanged && status.level != kC4Stopping /* Don't notify about internal state */)
-                onStatusChanged(this, status, _options->callbackContext);
+            if( !(status.error.code && status.level > kC4Offline) ) {
+                auto onStatusChanged = _onStatusChanged.load();
+                if (onStatusChanged && status.level != kC4Stopping /* Don't notify about internal state */)
+                    onStatusChanged(this, status, _options->callbackContext);
+            }
         }
 
 
--- a/Replicator/tests/ReplicatorAPITest.cc
+++ b/Replicator/tests/ReplicatorAPITest.cc
@@ -851,11 +851,9 @@ TEST_CASE_METHOD(ReplicatorAPITest, "Stop after transient connect failure", "[C]
     C4Error err;
     importJSONLines(sFixturesDir + "names_100.json");
     REQUIRE(startReplicator(kC4Disabled, kC4Continuous, WITH_ERROR(&err)));
-    
+
+    _numCallbacksWithLevel[kC4Offline] = 0;
     waitForStatus(kC4Offline);
-    
-    _numCallbacksWithLevel[kC4Connecting] = 0;
-    waitForStatus(kC4Connecting);
     c4repl_stop(_repl);
     
     waitForStatus(kC4Stopped);
@@ -887,10 +885,8 @@ TEST_CASE_METHOD(ReplicatorAPITest, "Calling c4socket_ method after STOP", "[C][
     importJSONLines(sFixturesDir + "names_100.json");
     REQUIRE(startReplicator(kC4Disabled, kC4Continuous, WITH_ERROR(&err)));
 
+    _numCallbacksWithLevel[kC4Offline] = 0;
     waitForStatus(kC4Offline);
-
-    _numCallbacksWithLevel[kC4Connecting] = 0;
-    waitForStatus(kC4Connecting);
     c4repl_stop(_repl);
 
     waitForStatus(kC4Stopped);


Replicator/c4ReplicatorImpl.hh
```diff
@@ -475,7 +475,7 @@ namespace litecore {
                 if (status.progress.unitsTotal > 0)
                     progress = 100.0 * double(status.progress.unitsCompleted)
                                      / status.progress.unitsTotal;
-                if (status.error.code) {
+                if (status.error.code && status.level <= kC4Offline) {
```
jianminzhao, line 478: Not sure if this is requested by the ticket.

borrrden: It is not.  This merely prevents a log statement from coming.  What the ticket is after is stopping the entire callback from happening.

borrrden (changes requested):
The goal of the ticket is to stop the onStatusChanged callback on the bottom of this function.  These changes simply remove an error log.

Replicator/c4ReplicatorImpl.hh
callumbirks: To update this PR; Jianmin, Blake and I have discussed it. What is necessary to solve the ticket is to clear the error code from the status when level <= kC4Offline. Not something I understood before

callumbirks:
Merge master because Actions builds are failing on a bunch of unrelated stuff, hoping merge will fix.

callumbirks:
So based on the conversation I had with Jianmin and Blake, I've changed the behaviour so that the callback will not be called if there is an error in the status and `status.level > kC4Offline`.
This broke a couple of tests that had `waitForStatus(kC4Connecting)` as they were waiting for a status that contained an error, I adjusted those tests to match the new behaviour. Have tested all CppTests, all Walrus SG tests, and all SGColl tests, everything is passing.

borrrden (approved):
Hopefully this is not a big surprise for QE....

jianminzhao:
I am okay that we drop the callback when the replicator suffered from Unrecoverable Error when the activity level is anywhere but kC4Stopped, for the only outcome of Unrecoverable Error is Stopped.
I am yet to be convinced that we should drop the callback just because there is an (recoverable) error.

jianminzhao:
Actually, I don't know what exactly Unrecoverable Errors are in C4. I assumed they be StoppingErrors in Core, which are     static constexpr StoppingErrorEntry StoppingErrors[] = {
        {{ LiteCoreDomain, kC4ErrorUnexpectedError,0 }, true, "An exception was thrown"_sl},
        {{ WebSocketDomain, 403, 0}, true, "An attempt was made to perform an unauthorized action"_sl},
        {{ WebSocketDomain, 503, 0 }, false, "The server is over capacity"_sl}
    };
Only those errors will take the replicator to Stopped or Offline, as far as I understand it now.
@borrrden do you assume any error will take the replicator to either Stopped or Offline? This would be hard-invariant if we move to drop the status callback whenever there is error in the status.

borrrden:
You are looking at the protocol level errors in this case.  Most of the things that will stop or take a replicator offline are the connection level errors, which have no need to be enumerated in the same way.  All connection level errors will cut off the connection and then be analyzed to see whether or not to retry the connection.  I think it is safe to do this.

Replicator/tests/ReplicatorAPITest.cc
```diff
@@ -854,8 +854,8 @@ TEST_CASE_METHOD(ReplicatorAPITest, "Stop after transient connect failure", "[C]
     
     waitForStatus(kC4Offline);
     
-    _numCallbacksWithLevel[kC4Connecting] = 0;
-    waitForStatus(kC4Connecting);
+    _numCallbacksWithLevel[kC4Offline] = 0;
+    waitForStatus(kC4Offline);
```
pasin, line 858: There is also waitForStatus(kC4Offline) in line 855. Should one get removed? or Is the test also testing retry (if so, please add a comment there as well) ?

Replicator/tests/ReplicatorAPITest.cc
```diff
@@ -889,8 +889,8 @@ TEST_CASE_METHOD(ReplicatorAPITest, "Calling c4socket_ method after STOP", "[C][
 
     waitForStatus(kC4Offline);
 
-    _numCallbacksWithLevel[kC4Connecting] = 0;
-    waitForStatus(kC4Connecting);
+    _numCallbacksWithLevel[kC4Offline] = 0;
+    waitForStatus(kC4Offline);
```
pasin, line 893: Same as above, why there are two wait for offline? Please add a comment there if it's intended.

Replicator/c4ReplicatorImpl.hh
```diff
@@ -485,9 +485,11 @@ namespace litecore {
...
             }
 
-            auto onStatusChanged = _onStatusChanged.load();
-            if (onStatusChanged && status.level != kC4Stopping /* Don't notify about internal state */)
-                onStatusChanged(this, status, _options->callbackContext);
+            if( !(status.error.code && status.level > kC4Offline) ) {
+                auto onStatusChanged = _onStatusChanged.load();
+                if (onStatusChanged && status.level != kC4Stopping /* Don't notify about internal state */)
+                    onStatusChanged(this, status, _options->callbackContext);
+            }
```
pasin, line 492: status.level != kC4Stopping -> means that it will only report when the status == kC4Offline after this change.

Why not just do `if (onStatusChanged && status.level == kC4Offline` and remove the outer if?

snej: No, it will report kC4Stopped and kC4Offline. (@pasin, I think you mixed up kC4Stopped and kC4Stopping.)

snej (approved):

pasin (approved):

Replicator/c4ReplicatorImpl.hh
pasin: Right, that is kC4Stopping not kC4Stopped.

jianminzhao:
This might be an interesting case, 
TEST_CASE_METHOD(ReplicatorCollectionTest, "Use Unmatched Collections", "[Push][Pull]") {
    _expectedError = {WebSocketDomain, 404};
    runPushPullReplication({Roses, Lavenders}, {Tulips, Lavenders});
}

If I manually injected an error in Replicator::getCollections(),
 if (root) { <<<====== the original correct error test is "if ( ! root )"
                    auto error = C4Error::printf(LiteCoreDomain, kC4ErrorRemoteError,
                                                 "Unparseable checkpoints: %.*s", SPLAT(json));
                    return gotError(error);
                }
the replicator will hang. (This by itself is problematic; let's put it aside.)
Currently the client will see from the status callback,
8:21:36.993122| WARNING: >> Replicator is busy, progress 0/0, 0 docs, errorCode = 0. 
08:21:37.196484| WARNING: >> Replicator is busy, progress 0/0, 0 docs, errorCode = 26. <=== Busy/Error(26)
In theory, clients may attend to errors in the status and call stop(). (They ultimately hold the replicator via C4Replicator.) But with the PR, they will see Busy/Error(0) only.
There are many appearances of this pattern, "return gotError()," and they are hardly covered by tests.

Replicator/tests/ReplicatorAPITest.cc
callumbirks: Ah, you could be right, something I missed. I will investigate and update the test where necessary.

callumbirks:
@jianminzhao thanks, I'll take a look at this and see what I find.

callumbirks:
After discussion with Jianmin, as with current behaviour we are unsure that Replicator will always go to Stopped or Offline, I have now enforced that principle.
`Replicator::onError` will now call `_disconnect` for any error. This takes internal state to Stopped, then the existing `c4RemoteReplicator::handleStopped()` decides if we should retry the connection (and therefore go to Offline).

Replicator/Replicator.cc
```diff
@@ -467,21 +457,14 @@ namespace litecore { namespace repl {
...
-                }
-                return;
-            }
-        }
+
+        // Log error and stop replicator
+        // The logic in c4RemoteReplicator::handleStopped() will decide if we should retry connection
+        string message = error.description();
+        logError("Stopping due to error: %s", message.c_str());
+        _disconnect(websocket::kCloseAppTransient, message);
```
pasin, line 467: Does this mean all errors coming to there will be transient and will cause the replicator to get retired? If this is for fixing hang in busy / error, what is the error that is reported here for that case?

Replicator/Replicator.cc
```diff
@@ -846,8 +831,7 @@ namespace litecore { namespace repl {
                         auto error = C4Error::printf(WebSocketDomain, 404,
                                                      "Collection '%.*s' is not found on the remote server",
                                                      SPLAT(collPath));
-                        gotError(error);
-                        _stop();
+                        onError(error);
```
pasin, line 834: Do we want to make it a permanent or transient error? With the PR, the error will be transient?

pasin: OK, I think the replicator will stop() with 404 error.

callumbirks: Yeah, _disconnect() is essentially the same as _stop(). _stop() just calls _disconnect

Replicator/Replicator.cc
callumbirks: The original error Jianmin found that causes Replicator to hang in Busy is if the checkpoints JSON received from remote is unparseable. This is what fuelled the conversation on enforcing the principle of "error always leads to Stopped or Offline". Because we previously have no certainty about the behaviour when encountering an error.
The transient code seems the most suitable to use (to me) out of the available `websocket::CloseCode`s. The actual error that we encountered is still logged, and sent to remote. But also 4001 (kCloseAppTransient) is there with it.

borrrden (changes requested):
I think this has gotten out of scope and control from the original intention.  I don't see anything wrong with the first attempt, and recommend going back to it.  I don't think it will cause any issues.  If there is still doubt, then this should be dropped and pushed out to another release.

callumbirks:
> I think this has gotten out of scope and control from the original intention. I don't see anything wrong with the first attempt, and recommend going back to it. I don't think it will cause any issues. If there is still doubt, then this should be dropped and pushed out to another release.

I'd pretty much agree with this. There are errors which can be triggered which are not handled properly, and that's what the second solution solves, but that should maybe be a different ticket. In the case of those errors, they would normally send the Replicator to (Busy, error), but this will no longer be reported to the callback. Although I guess if the errors are causing the Replicator to hang anyway, it doesn't make much difference whether we make the callback or not. (Because from the customer's POV, the Replicator is hanging).

Either way, both issues need to be solved. Either in this ticket, or separate ones. So if I have the go ahead, I'll create a separate issue for the second solution, and separate it out from this branch into that one.

jianminzhao:
"Either in this ticket, or separate ones. "  -- I prefer the latter, with separate ticket, particular because we are very near code-freeze for a release.

callumbirks:
Have reverted the changes to error handling, those have been split into a separate branch. Now this branch is only concerned with the behaviour of reporting to callback.

jianminzhao (approved):

borrrden (approved):

PR #1657, opened 2022-12-14
CBL-4022 : Update BuiltInWebSocket to do preemptive auth by default

pasin:
* Updated BuiltInWebSocket and HTTPLogic to do preemptive auth by default with an option for doing challenge based auth.
* This change is required by CBL-C to do basic auth preemptively to avoid the issue that users can get authenticated as a GUEST user when they don't specify credentials.

### DIFF
--- a/Networking/HTTP/HTTPLogic.cc
+++ b/Networking/HTTP/HTTPLogic.cc
@@ -89,6 +89,13 @@ namespace litecore { namespace net {
 
 
     string HTTPLogic::requestToSend() {
+        // This logic prevents calling requestToSend() more than once to respond to the
+        // auth challenge after the the authentication from the first challenge failed
+        // (when failed the dispostion will still be kAuthenticate).
+        //
+        // In handleResponse(), after it handles the challenge first time, it will reset
+        // the _proxy's credentials for proxy authentication or _authHeader for server
+        // authentication.
         if (_lastDisposition == kAuthenticate) {
             if (_httpStatus == HTTPStatus::ProxyAuthRequired)
                 Assert(_proxy && _proxy->username);
@@ -194,6 +201,8 @@ namespace litecore { namespace net {
             case HTTPStatus::UseProxy:
                 return handleRedirect();
             case HTTPStatus::Unauthorized:
+                // When challenge auth is not enabled (preemptive auth), fail right away
+                // as the authentication has been done and the result was unauthorized.
                 if (_enableChallengeAuth) {
                     if (_authChallenged)
                         _authHeader = nullslice;
--- a/Networking/HTTP/HTTPLogic.cc
+++ b/Networking/HTTP/HTTPLogic.cc
@@ -93,7 +93,7 @@ namespace litecore { namespace net {
             if (_httpStatus == HTTPStatus::ProxyAuthRequired)
                 Assert(_proxy && _proxy->username);
             else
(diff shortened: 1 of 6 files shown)


Replicator/tests/ReplicatorSGTest.cc
```diff
@@ -85,6 +85,34 @@ class ReplicatorSGTest : public ReplicatorAPITest {
...
+                enc.writeKey(C4STR(kC4ReplicatorAuthType));
+                enc.writeString("Basic"_sl);
+                enc.writeKey(C4STR(kC4ReplicatorAuthUserName));
+                enc.writeString("brown");
+                enc.writeKey(C4STR(kC4ReplicatorAuthPassword));
+                enc.writeString("sugar");
+                enc.writeKey(C4STR(kC4ReplicatorAuthEnableChallengeAuth));
+        
+            SECTION("Preemptive Auth") {
+                enc.writeBool(false);
```
jianminzhao, line 105: Will this trigger Assert?

borrrden (commented):
This assertion makes me uncomfortable.  Can you explain in what situations it will be hit?

Networking/HTTP/HTTPLogic.cc
```diff
@@ -93,7 +93,7 @@ namespace litecore { namespace net {
             if (_httpStatus == HTTPStatus::ProxyAuthRequired)
                 Assert(_proxy && _proxy->username);
             else
-                Assert(_authHeader);
+                Assert(_enableChallengeAuth && _authHeader);
```
borrrden, line 96: Sorry it's this assertion...

Replicator/tests/ReplicatorSGTest.cc
pasin: What Assert?

Networking/HTTP/HTTPLogic.cc
pasin: It will be hit only when there is a challenge. So if _enableChallengeAuth == false, it should be hit at all.

Replicator/tests/ReplicatorSGTest.cc
pasin: If it's the same assert as @borrrden mentioned, the answer is YES.

jianminzhao (approved):

borrrden:
Update: I was right to doubt the assertion because the logic itself will cause it to be hit in seemingly normal cases.  The reason this doesn't actually happen, however, is because there is more logic in BuiltInWebSocket that stops it.  This creates a confusing implicit agreement to only do one challenge response.  Maybe that's a normal thing to expect, maybe not, but HTTPLogic is silent about it to the consumer, it expects them to know that trying to do another challenge response without setting the auth header again will cause a crash.

pasin:
Regardless the fact that the logic is controlled by the assertion, in the scope of replicator's authentication, doing only one challenge is correct as it will always be the same credentials.

borrrden (approved):
After spending probably longer than I should have thinking about what HTTPLogic's job is, I have concluded that this PR is now alright, but a future improvement could be made to have it not outright crash in silly mistakes like "oops I sent no auth after I received an auth challenge".  Having it repeat the same action and returning the same result (it could probably even short circuit here and just skip the action and return authenticate immediately) would be less jarring.

PR #1661, opened 2022-12-15
CBL-4023: Make SyncServerCollection tests compatible with SGW 3.1.0

callumbirks:
Tested with SGW 3.1.0-438, all tests are passing.

### DIFF
--- a/Replicator/tests/SG.cc
+++ b/Replicator/tests/SG.cc
@@ -138,9 +138,37 @@ bool SG::deleteUser(const string &username) const {
 }
 
 bool SG::assignUserChannel(const std::string& username, const std::vector<std::string> &channelIDs) const {
-    alloc_slice bodyWithChannel = addChannelToJSON("{}"_sl, "admin_channels"_sl, channelIDs);
+    Encoder enc;
+    enc.beginDict();
+    enc.writeKey("collection_access"_sl);
+    {
+        enc.beginDict();
+        enc.writeKey("flowers"_sl);
+        {
+            enc.beginDict();
+            enc.writeKey("roses"_sl);
+            {
+                enc.beginDict();
+                enc.writeKey("admin_channels"_sl);
+                {
+                    enc.beginArray();
+                    for (const auto& chID : channelIDs) {
+                        FLEncoder_WriteString(enc, slice(chID));
+                    }
+                    enc.endArray();
+                }
+                enc.endDict();
+            }
+            enc.endDict();
+        }
+        enc.endDict();
+    }
+    enc.endDict();
+    Doc doc {enc.finish()};
+    Value v = doc.root();
+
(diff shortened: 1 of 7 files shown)


callumbirks:
Realized the channel_access fixes are hardcoded for the "Roses" collection, working on fixing this now

Replicator/tests/SG.cc
```diff
@@ -138,9 +138,37 @@ bool SG::deleteUser(const string &username) const {
 }
 
 bool SG::assignUserChannel(const std::string& username, const std::vector<std::string> &channelIDs) const {
```
jianminzhao, line 140: We probably need an argument telling what collections will be involved. Or we may want to test different channelIDs for different collections,

callumbirks: I believe the commit I just made should address those two concerns

callumbirks:
The reason for using a multimap is to make parsing the multiple collections easier. With this method, the user should be able to provide collectionSpecs with different scopes and collections, and it will all work as expected.

pasin (approved):

Replicator/tests/SG.cc
```diff
@@ -123,11 +124,10 @@ alloc_slice SG::addChannelToJSON(slice json, slice ckey, const std::vector<std::
 bool SG::createUser(const std::string& username, const std::string& password,
                     const std::vector<std::string> &channelIDs) const {
```
jianminzhao, line 125: Where is "channelIds" used in this function?

Replicator/tests/SG.hh
```diff
@@ -50,7 +50,13 @@ public:
...
                     const std::vector<std::string> &channelIDs) const;
     bool deleteUser(const std::string& username) const;
-    bool assignUserChannel(const std::string& username, const std::vector<std::string>& channelIDs) const;
+    // Assign given channels to the user with given username, in the given collections
+    bool assignUserChannel(const std::string& username, const std::vector<C4CollectionSpec>& collectionSpecs, const std::vector<std::string>& channelIDs) const;
+    // Same as above, but accepts an array parameter
+    template <size_t N>
+    bool assignUserChannel(const std::string& username, const std::array<C4CollectionSpec, N>& collectionSpecs, const std::vector<std::string>& channelIDs) const {
+        return assignUserChannel(username, std::vector(collectionSpecs.begin(), collectionSpecs.end()), channelIDs);
+    }
```
jianminzhao, line 59: Why not the other way around? From vector to array, I guess, is simple.

Replicator/tests/SG.cc
callumbirks: Ah you’re right, I’ve removed the use from this function now

velicuvlad (approved):

jianminzhao (approved):

PR #1663, opened 2022-12-19
CBL-3994: Query with FTS with multiple collections.

jianminzhao:
* Extended SQL++ to be able specify the collection when referring to the FTS index in MATCH() function.
* Extended QueryParser to handle multiple collections.
* Fixed prediction with non-default collections.

### DIFF
--- a/LiteCore/Query/QueryParser.cc
+++ b/LiteCore/Query/QueryParser.cc
@@ -598,8 +598,30 @@ namespace litecore {
         for (auto &ftsTable : _indexJoinTables) {
             auto &table = ftsTable.first;
             auto &alias = ftsTable.second;
+            auto idxAt = table.find(KeyStore::kIndexSeparator);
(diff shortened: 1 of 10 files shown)


snej (approved):

LiteCore/Query/QueryParser.cc
```diff
@@ -598,8 +598,30 @@ namespace litecore {
         for (auto &ftsTable : _indexJoinTables) {
             auto &table = ftsTable.first;
             auto &alias = ftsTable.second;
+            auto idxAt = table.find(KeyStore::kIndexSeparator);
```
borrrden, line 601: Below this area could stand some comments about the logic

jianminzhao: Agreed and will do.

borrrden (approved):
Is there a plan for handling ambiguous collection names?

PR #1665, opened 2022-12-21
CBL-4004: Extend test of "[.SyncServerCollection]" to multiple collec…

jianminzhao:
…tions - part I

### DIFF
--- a/Replicator/tests/ReplicatorCollectionSGTest.cc
+++ b/Replicator/tests/ReplicatorCollectionSGTest.cc
@@ -229,20 +229,18 @@ public:
 
         for (size_t i = 0; i < N; ++i) {
             if (checkRev) {
+                c4::ref<C4DocEnumerator> e = c4coll_enumerateAllDocs
+                    (collections[i], nullptr, ERROR_INFO());
                 unsigned count = 0;
-                c4::ref<C4DocEnumerator> e = c4coll_enumerateAllDocs(collections[i],
-                                                                     nullptr, ERROR_INFO());
-                {
+                while (c4enum_next(e, ERROR_INFO())) {
                     ++count;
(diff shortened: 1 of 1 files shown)


velicuvlad:
This looks good. 

It is worth mentioning that me and Callum started using SG::TestUser to set the user and channels.
```
SG::TestUser testUser { _sg, "pdfcsg", chIDs, collectionSpecs }; // if no channels, use {"*"}
_sg.authHeader = testUser.authHeader();
std::array<C4Collection *, collectionCount> collections 
            = collectionPreamble(collectionSpecs, testUser);
```


Also, for `replParams`. 
```
std::vector<C4ReplicationCollection> replCollections {collectionCount};
for(size_t i = 0; i < collectionCount; ++i) {
        replCollections[i] = { collectionSpecs[i] };
    }
ReplParams replParams { replCollections };
replParams.setPushPull(<pushType>, <pullType>)
```

If time permits, it would be a good change in order to maintain structure across tests. For reference, these can be seen in our open PRs for our tests.

velicuvlad (approved):

callumbirks (approved):
I'd agree with Vlad's comment, `SG::TestUser` allows for consistent isolation between tests. I also think it makes it somewhat easier to understand what's going on. 
Of course, it is down to your decision whether to use it.

jianminzhao:
Thanks for mentioning it. I noticed that too. In some tests I did change it to use TestUser. It's succinct and convenient, but I did not change all of them so as to have tests testing the DocID filter where appropriate, because DocID is a product feature that ought to work reliably. Moving forward, we should try to use TestUser as much as possible. By in TestUser, we may want to add a method to populate a local db with all the docs that the test-user has access based on channel

PR #1666, opened 2022-12-23
CBL-4059: compiler complaint on Centos build.

jianminzhao:


jianminzhao:
reopen after verifying on the build image.

PR #1668, opened 2022-12-23
CBL-2694: Prevent rev list gaps

callumbirks:
With this change, we will now keep ~100~ 50 revs in the history of a doc, rather than 20. This shouldn't make a huge impact on storage requirements, because we only keep bodies for the current rev, and the last rev we synced with SG.

~With this current implementation, if the client has somehow made > 100 revisions since they last synced, SG will receive a rev list with gap. I can change this so we give SG the random revIDs we generate (I've tested the random revIDs, and they're fine). But I'm not sure if we want to do that, or just allow the warning to happen for > 100 revisions because if a customer is making > 100 revisions between syncs frequently, that may be a sign of another problem.~

With this implementation, if the client has made > 50 revisions since they last synced, the gap in the history to be sent to SG will be filled with randomly generated revIDs from `c4doc_getRevisionHistory()`. The client will also be warned of the large amount of mutations between sync, if we had to use random revIDs. We will send SG a history of up to 100 revs (50 real, 50 randomly generated). If there is more than 100 revisions since last sync, SG will receive a rev list with a gap, and therefore SG will warn the user about a gap.

### DIFF
--- a/LiteCore/Database/DatabaseImpl.cc
+++ b/LiteCore/Database/DatabaseImpl.cc
@@ -51,7 +51,7 @@ namespace litecore {
 
 
     static constexpr slice kMaxRevTreeDepthKey = "maxRevTreeDepth";
-    static constexpr uint32_t kDefaultMaxRevTreeDepth = 20;
+    static constexpr uint32_t kDefaultMaxRevTreeDepth = 100;
 
 
     static string collectionNameToKeyStoreName(C4Database::CollectionSpec);
--- a/LiteCore/Database/DatabaseImpl.cc
+++ b/LiteCore/Database/DatabaseImpl.cc
@@ -51,7 +51,7 @@ namespace litecore {
 
 
     static constexpr slice kMaxRevTreeDepthKey = "maxRevTreeDepth";
-    static constexpr uint32_t kDefaultMaxRevTreeDepth = 100;
+    static constexpr uint32_t kDefaultMaxRevTreeDepth = 50;
 
 
     static string collectionNameToKeyStoreName(C4Database::CollectionSpec);
--- a/Replicator/RevFinder.cc
+++ b/Replicator/RevFinder.cc
@@ -121,6 +121,7 @@ namespace litecore::repl {
                 MessageBuilder response(req);
                 response.compressed = true;
                 if (!_db->usingVersionVectors()) {
+                    // Depth of rev history SG should send to us
                     response["maxHistory"_sl] = tuning::kDefaultMaxHistory;
                 }
                 if (!_db->disableBlobSupport())
--- a/Replicator/RevFinder.cc
+++ b/Replicator/RevFinder.cc
@@ -122,7 +122,7 @@ namespace litecore::repl {
                 response.compressed = true;
                 if (!_db->usingVersionVectors()) {
 #if 1
-                    response["maxHistory"_sl] = 20;
+                    response["maxHistory"_sl] = 100;
 #else
                     response["maxHistory"_sl] = _db->useLocked()->getMaxRevTreeDepth();
 #endif
--- a/Replicator/RevFinder.cc
+++ b/Replicator/RevFinder.cc
@@ -121,11 +121,7 @@ namespace litecore::repl {
                 MessageBuilder response(req);
                 response.compressed = true;
                 if (!_db->usingVersionVectors()) {
-#if 1
-                    response["maxHistory"_sl] = 100;
-#else
-                    response["maxHistory"_sl] = _db->useLocked()->getMaxRevTreeDepth();
-#endif
+                    response["maxHistory"_sl] = tuning::kDefaultMaxHistory;
                 }
                 if (!_db->disableBlobSupport())
                     response["blobs"_sl] = "true"_sl;
--- a/Replicator/tests/ReplicatorCollectionSGTest.cc
+++ b/Replicator/tests/ReplicatorCollectionSGTest.cc
@@ -1737,12 +1737,23 @@ TEST_CASE_METHOD(ReplicatorCollectionSGTest, "Pull iTunes deltas from Collection
 // in the gap of the other client's rev history.
 TEST_CASE_METHOD(ReplicatorCollectionSGTest, "Give SG a rev history with a gap", "[.SyncServerCollection]") {
     constexpr size_t maxHistory = tuning::kDefaultMaxHistory;
-    constexpr size_t numRevs1 = maxHistory + 50;
-    // num revs of the second 'client' land within the gap of the first 'client' history
-    constexpr size_t numRevs2 = numRevs1 - (maxHistory + 10);
     constexpr size_t numInitialRevs = 2;
     constexpr const char * saveDBName = "revsgap";
     const string docID = timePrefix() + "doc1";
+
+    size_t numRevs1;
+    size_t numRevs2;
+
+    SECTION("No gap in history") {
+        numRevs1 = maxHistory - 5;
+        numRevs2 = numRevs1 - 10;
+    }
+
+    SECTION("Gap in history") {
+        numRevs1 = maxHistory + 50;
+        // num revs of the second 'client' land within the gap of the first 'client' history
+        numRevs2 = numRevs1 - (maxHistory + 10);
+    }
 
     initTest({ Roses, Tulips, Lavenders });
 
@@ -1784,11 +1795,11 @@ TEST_CASE_METHOD(ReplicatorCollectionSGTest, "Give SG a rev history with a gap",
     // 'client 2' mutations
     for(auto& coll : _collections) {
         for(int i = numInitialRevs; i < numRevs2 + numInitialRevs; ++i) {
-            createFleeceRev(coll, slice(docID), nullslice, slice(R"({"a":)" + to_string(i) + "}"));
+            createFleeceRev(coll, slice(docID), nullslice, slice(R"({"b":)" + to_string(i) + "}"));
         }
     }
 
-    // There will be a conflict in the push because different sets of mutations from each 'client'
+    // There will be a conflict in the push because different revIDs from separate 'clients'
     // We need to make sure the test suite knows we expect this, otherwise an assertion fails
     _expectedDocPullErrors = {docID};
 
--- a/Replicator/tests/ReplicatorCollectionSGTest.cc
+++ b/Replicator/tests/ReplicatorCollectionSGTest.cc
@@ -1890,6 +1890,9 @@ TEST_CASE_METHOD(ReplicatorCollectionSGTest, "Pull iTunes deltas from Collection
           timeWithDelta, timeWithoutDelta, timeWithoutDelta/timeWithDelta);
 }
 
+// This test may be used in future, if kDefaultMaxHistory > kDefaultMaxRevTreeDepth, and we wish to test
+// the conflict resolution behaviour for randomly generated rev history.
+#if 0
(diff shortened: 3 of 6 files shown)


borrrden:
It doesn't have to be either or, we should do both the random IDs AND a warning that we are doing something funky.

Replicator/RevFinder.cc
```diff
@@ -122,7 +122,7 @@ namespace litecore::repl {
                 response.compressed = true;
                 if (!_db->usingVersionVectors()) {
 #if 1
-                    response["maxHistory"_sl] = 20;
+                    response["maxHistory"_sl] = 100;
```
borrrden, line 125: Is it possible to get this value from ReplicatorTuning.hh ?

callumbirks: You're right, it is.

callumbirks:
I've changed the revTree depth to 50, and left the replication maxHistory at 100. These numbers can be discussed if we think something else is suitable.
I've added a log warning for a gap in the revision history, with some (hopefully) helpful info for the client.

callumbirks:
With these numbers (revTree depth 50, max history 100): above 50 revisions since last sync, we will send SG random revIDs and there will be a warning logged to the client. Above 100 revisions, there will also be a gap in the rev list sent to SG, so a warning will appear there as well.

pasin (commented):
I'm not sure about changing the max depth = 20 if it doesn't cause the issue.

callumbirks:
> I'm not sure about changing the max depth = 20 if it doesn't cause any issues.

So the original issue is SG receiving a rev list with gaps. Increasing the `maxHistory` for the Replicator means we will generate random revIDs to fill in the gap in the rev history. This stops SG from receiving a rev list with gaps (up to the `maxHistory`). However, the concern is that because the gap is filled with randomly generated revIDs, this causes conflicts with other clients' rev history.

The revTree depth is not as storage-expensive as it was in the past, because now we only store bodies of the current rev, and the last rev we synced with remote. And increasing the revTree depth will reduce the number of cases where we have conflicts with another client.

I have written a full test case to test the conflicts, and they seem to resolve fine, but the less the better.

snej (approved):
Not fully comfortable with storing more revIDs locally, but the performance impact probably is negligible...

Replicator/tests/ReplicatorCollectionSGTest.cc
```diff
@@ -1889,3 +1889,133 @@ TEST_CASE_METHOD(ReplicatorCollectionSGTest, "Pull iTunes deltas from Collection
     C4Log("-------- %.3f sec with deltas, %.3f sec without; %.2fx speed",
           timeWithDelta, timeWithoutDelta, timeWithoutDelta/timeWithDelta);
 }
+
+TEST_CASE_METHOD(ReplicatorCollectionSGTest, "Give SGW random rev history and conflicts", "[.CBL-2694]") {
+    // The idea of this test is to exceed the RevTree history limit, so we can see what happens when we use the
```
jianminzhao, line 1894: I am not sure what's the motivation of the changes to the two constants, 20 -> 50 and 20 -> 100. I restored that changes, and this test also passes.

jianminzhao (changes requested):

LiteCore/Database/DatabaseImpl.cc
```diff
@@ -51,7 +51,7 @@ namespace litecore {
 
 
     static constexpr slice kMaxRevTreeDepthKey = "maxRevTreeDepth";
-    static constexpr uint32_t kDefaultMaxRevTreeDepth = 20;
+    static constexpr uint32_t kDefaultMaxRevTreeDepth = 100;
```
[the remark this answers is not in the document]
callumbirks: The max revTree depth is how many revs to store in the history of the revTree document, the Replicator's maxHistory is the max depth of history to send to SG during sync. We have a facility in `c4doc_getRevisionHistory` to generate random revIDs to fill in the gap between `maxRevTreeDepth` and Replicator's `maxHistory`, which seems like it was intended to be used based on the comments in the function, but never was because previously the values of `maxRevTreeDepth` and `maxHistory` were the same.

Replicator/RevFinder.cc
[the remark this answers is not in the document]
callumbirks: It does in the most recent commit

Replicator/tests/ReplicatorCollectionSGTest.cc
callumbirks: The test is intended to see the behaviour of conflict resolution when we are sending SG a history that contains randomly generated rev history. So the test just uses whatever the values are of `maxHistory` and `maxRevTreeDepth`, creates some conflicts in the rev history, and ensures that conflict resolution behaves as expected. 

Looking at the test now though, I do need to update it a bit since I changed the values from (100/20) to (100/50) since I first made the test.

callumbirks:
I've updated the test and changed it to use `kDefaultMaxHistory` rather than a hardcoded number. I did a lot of refactoring to try and make the test neater.

LiteCore/Database/DatabaseImpl.cc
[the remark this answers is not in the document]
callumbirks: In the latest commit, `maxRevTreeDepth`/`maxHistory` are different (`50`/`100`)

Replicator/tests/ReplicatorCollectionSGTest.cc
jianminzhao: My wonder is actually the following. There are three parts of the in the PR.
1. Substance, it changed two constant, kDefaultMaxRevTreeDepth, kDefaultMaxHistory
2. Log the gap when the random revIDs are substituted in the history.
3. This test.
I tried to remove (1) and the test also passed. So, is the main object of PR is to add a test? Is there a thing that did not work without this PR?

callumbirks: The test doesn't test something that's internally broken, the issue stems from SG receiving a rev history from us that has a gap. There were complaints from certain customers about this warning in SG. The test is simply to test that with the increased values, and thereby the usage of randomly generated revIDs, the conflict resolution will still behave as it does normally.

I discussed it with SG team, Jim and Jens and SG team considered whether they could just drop the warning from SG when they receive a gap. This is what Ben said to me:

> Adam replied saying this and then I pointed him to the litecore code generating non-deterministic revision IDs:
> > We wouldn't create fake conflicts if we had a deterministic approach for generating the intermediate digests (either clone the digest of the latest revision, or hash from parent rev ID).  Have we already reviewed whether the rev tree pruning algorithm (even for non-branching rev trees) is happy with gaps in generation numbers?

> It seems odd that the comment around that area of litecore code suggests its generating them in order to send them, but then never does

> It still feels like litecore should be filling that gap for us - couchdb (and therefore SG) formats revisions in a way that doesn't allow you to have gaps

So I decided the best course of action was to increase the length of history that we send to SG, so it is much harder to perform enough local mutations between syncs to end up sending a rev history with a gap to SG. That is why I propose to increase `ReplicatorTuning::kDefaultMaxHistory`.

The drawback with that, is any history in-between the revTree depth and maxHistory is randomly generated (within the function `TreeDocument::_getRevisionHistory`). As it is randomly generated, this could easily cause conflicts with other clients, and we want to cause as little conflicts as we can.

The current max revTree depth of 20 was set when we used to store the bodies of all revs in the tree, so each rev we store in the history was quite storage-expensive. As we now only store the bodies of the current rev and the last rev we synced with remote, additional revs in the revTree aren't too expensive (I'm assuming just however many bytes the revIDs are). 

Therefore, to reduce the usage of randomly generated revIDs, I want to increase `kDefaultMaxRevTreeDepth`, as it shouldn't be too storage-expensive and will significantly reduce the possibility of conflicts between clients' rev history.

callumbirks: Additionally, the reason for logging a warning to the client when we had to use randomly generated revIDs is because if the client is making more than 50 mutations between syncs with SG, there is probably something going wrong.

jianminzhao: So, the fundamental problem is that the client voluntarily limit the max-history. The protocol allows the reply to include "maxHistory." If it is absent, the client would voluntarily limit it to a number, which I assume is not part of the protocol. Now, we are in a bad situation. On one hand, we limit the history to, say, 20 or 100, but the last revision that the remote knows is deeper than the max. On the other hand, we want to include the last synced revision in the history. 
Either we don't self-impose the maxHistory, or the protocol allows gaps and SG should handle it. Of course, if SG puts maxHistory in the reply, it knows the history may be truncated.

jianminzhao: `                if (hasRemoteAncestor(revID)) {
                    // Always write the common ancestor, making room if necessary:
                    if (revsWritten == maxRevs)
                        removeLast();
                    append(revID);
                    break;
                } else {`
^^^ from _getRevisionHistory

Replicator/tests/ReplicatorCollectionSGTest.cc
```diff
@@ -1730,3 +1730,111 @@ TEST_CASE_METHOD(ReplicatorCollectionSGTest, "Pull iTunes deltas from Collection
     C4Log("-------- %.3f sec with deltas, %.3f sec without; %.2fx speed",
           timeWithDelta, timeWithoutDelta, timeWithoutDelta/timeWithDelta);
 }
+
+// This test may be used in future, if kDefaultMaxHistory > kDefaultMaxRevTreeDepth, and we wish to test
+// the conflict resolution behaviour for randomly generated rev history.
+#if 0
```
jianminzhao, line 1736: Why disable this test? It is still valuable to test sync in the presence of gaps, IMHO.

borrrden (approved):
I am also curious what happens when gaps are present, so tend to agree with Jianmin.

Replicator/tests/ReplicatorCollectionSGTest.cc
```diff
@@ -1730,3 +1730,94 @@ TEST_CASE_METHOD(ReplicatorCollectionSGTest, "Pull iTunes deltas from Collection
     C4Log("-------- %.3f sec with deltas, %.3f sec without; %.2fx speed",
           timeWithDelta, timeWithoutDelta, timeWithoutDelta/timeWithDelta);
 }
+
+// Test replication and conflict resolution behaviour when we send SG a rev history with a gap (number of new
+// revs between syncs > repl::tuning::kDefaultMaxHistory).
+// This test attempts to emulate two separate clients with the same doc, where one client's rev history lands
+// in the gap of the other client's rev history.
+TEST_CASE_METHOD(ReplicatorCollectionSGTest, "Give SG a rev history with a gap", "[.SyncServerCollection]") {
```
jianminzhao, line 1738: I think. it's more interesting to demonstrate the difference between the current set of constants, 50/50, v.s. ideal ones, infinity/infinity, or 200/200 in your example. I know we cannot do it without modifying the source code, but perhaps you can shrinking the size of numRev1/numRev2 proportionally to fit to maxTreeDepth.

callumbirks: I don't quite understand your meaning. Would you mean potentially having two Catch Sections for the test, one with numRev values that won't cause a gap, one with numRev values that will cause a gap? Then we can show that the behaviour is the same with or without a gap?

jianminzhao: exactly

jianminzhao (approved):

PR #1672, opened 2023-01-05
CBL-4051: Fix a deadlock in ReplicatorAPITest

jianminzhao:
This problem is revealed by a DebugAssert.


callumbirks (approved):
Changes look good, pulled and ran some tests locally, behaves fine too.

PR #1673, opened 2023-01-06
CBL-4083: Polish CollectionSGTest Suite

callumbirks:
Moved the test class definition and functions into a header file to clean up test file.
Created new functions and variables in the test class to reduce repeated code and make tests cleaner and easier to write.
Modify some existing test functions to facilitate the new functions.
Modify all CollectionSG tests to use the new functions.
Have cleaned up tests where necessary, and adjusted tests to fit a similar pattern.
Also cleaned up Xcode project because of the CollectionSG helper files that Pasin noted were in the root of the project directory.

Ran all tests, all passing. Making all tests fit this same pattern, I also noticed a test that was not behaving as expected, and commented out that section of the test because it will not work until SG has multi-scope support.

### DIFF
--- a/Replicator/tests/ReplParams.cc
+++ b/Replicator/tests/ReplParams.cc
@@ -20,6 +20,24 @@ ReplParams::ReplParams(const std::vector<C4ReplicationCollection>& collections_)
     socketFactory = nullptr;
 }
 
+ReplParams::ReplParams(const std::vector<C4CollectionSpec> &collSpecs, C4ReplicatorMode push, C4ReplicatorMode pull) {
+    _collectionVector = std::vector<C4ReplicationCollection>{ collSpecs.size() };
+    for(int i = 0; i < collSpecs.size(); ++i) {
+        _collectionVector[i] = { collSpecs[i], push, pull };
+    }
+    collections = _collectionVector.data();
+    collectionCount = _collectionVector.size();
+    _optionsDict = {};
+    optionsDictFleece = _optionsDict.data();
+    onStatusChanged = nullptr;
+    onDocumentsEnded = nullptr;
+    onBlobProgress = nullptr;
+    propertyDecryptor = nullptr;
+    propertyEncryptor = nullptr;
+    callbackContext = nullptr;
+    socketFactory = nullptr;
+}
+
 ReplParams::ReplParams(const ReplParams &other) {
     _collectionVector = { other._collectionVector };
     _collectionsOptionsDict = { other._collectionsOptionsDict };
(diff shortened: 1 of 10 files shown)


Replicator/tests/SGTestUser.cc
```diff
@@ -3,21 +3,39 @@
 //
 
 #include "SGTestUser.hh"
-#include "Defer.hh"
 
 bool SG::TestUser::addChannels(const std::vector<std::string>& channels) {
+    REQUIRE(_sg);
```
jianminzhao, line 8: This is more related to the test framework itself. Perhaps Assert is more apt.

jianminzhao (approved):

Replicator/tests/ReplParams.cc
```diff
@@ -20,6 +20,24 @@ ReplParams::ReplParams(const std::vector<C4ReplicationCollection>& collections_)
...
+    collectionCount = _collectionVector.size();
+    _optionsDict = {};
+    optionsDictFleece = _optionsDict.data();
+    onStatusChanged = nullptr;
+    onDocumentsEnded = nullptr;
+    onBlobProgress = nullptr;
+    propertyDecryptor = nullptr;
+    propertyEncryptor = nullptr;
+    callbackContext = nullptr;
+    socketFactory = nullptr;
```
borrrden, line 38: This is too noisy.  You can define default values for these null items in the header and there is no need to explicitly set them all to null here.

Replicator/tests/ReplicatorCollectionSGTest.hh
```diff
@@ -0,0 +1,269 @@
+//
+// Created by Callum Birks on 05/01/2023.
+//
+
+#ifndef LITECORE_REPLICATORCOLLECTIONSGTEST_HH
+#define LITECORE_REPLICATORCOLLECTIONSGTEST_HH
```
borrrden, line 6: `#pragma once`

borrrden (approved):
Would like these changes but since it is test code it is not essential.

Replicator/tests/ReplParams.cc
callumbirks: It extends from a C-style struct, which is where the values come from. Is there a way to still have default values?

Replicator/tests/SGTestUser.cc
callumbirks: Thank you, I was unsure about this

Replicator/tests/ReplParams.cc
borrrden: Hmmm, well if the header is also used form C then it will be more trouble than it is worth.  If the header is C++ only then it doesn't matter and you can do it.  Or you can use the modern C++ initialization

```
Constructor()
    :field1(value1)
    ,field2(value2)
{ }
```

callumbirks: Seems I can't use that type of initialisation for the C-struct values, but now I've fixed my IDE being slow, I found out I can just do `: C4ReplicatorParameters()` and all the struct values get initialised.

PR #1677, opened 2023-01-12
CBL-4075: Add more multi-collection tests.

jianminzhao:
- testing nonexistent collections.
- testing bad configurations. (Also made C4Replicator's error checking more consistent)
- test that does Push, Pull, and Push/Pull, with respective collections.
- test Auto Purge w/ and w/o Filter with regard to different collections.
- test pushing to Default collection when it is not in the SG config.

### DIFF
--- a/Replicator/tests/ReplicatorCollectionSGTest.cc
+++ b/Replicator/tests/ReplicatorCollectionSGTest.cc
@@ -988,7 +988,8 @@ TEST_CASE_METHOD(ReplicatorCollectionSGTest, "Auto Purge Enabled - w/ and w/o Fi
             pullFilterPurge = 0;
         }
         unsigned collIndex;
-    } cbContexts[_collectionCount];
+    } cbContexts[2];
+    Assert( 2 == _collectionCount );
     for (unsigned i = 0; i < _collectionCount; ++i) {
         cbContexts[i].collIndex = i;
     }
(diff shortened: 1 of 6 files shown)


Replicator/tests/ReplicatorCollectionSGTest.cc
```diff
@@ -281,6 +365,125 @@ TEST_CASE_METHOD(ReplicatorCollectionSGTest, "Multiple Collections Push & Pull S
...
+
+    {
+        // Send the docs to remote
+        ReplParams replParams { _collectionSpecs };
+        replParams.setPushPull(kC4OneShot, kC4Disabled);
+        replicate(replParams);
+    }
+
+    deleteAndRecreateDB();
+    initTest({Lavenders, Roses, Tulips}, {channelID});
```
callumbirks, line 411: I suggest these two lines can be replaced with `deleteAndRecreateDBAndCollections()`. This will achieve the same outcome you're looking for, without the overhead of all the additional stuff in initTest.

Replicator/tests/ReplicatorCollectionSGTest.cc
```diff
@@ -281,6 +365,125 @@ TEST_CASE_METHOD(ReplicatorCollectionSGTest, "Multiple Collections Push & Pull S
...
+                default:
+                    break;
+            }
+        }
+    };
+    check();
+
+    // Clear the local database and check again.
+    deleteAndRecreateDB();
+    initTest({Lavenders, Roses, Tulips}, {channelID});
```
callumbirks, line 477: Same here as previous comment. Just a suggestion though, isn't necessary to change.

callumbirks (approved):
Pulled branch, all tests passing.
Test code looks good, using the new functions and pattern.
Left a couple comments about potential improvement using another one of the new functions.

Replicator/tests/ReplicatorCollectionSGTest.cc
jianminzhao: I will try that, thanks

PR #1680, opened 2023-01-23
CBL-4091: make c4cert_getValidTimespan available in CE

callumbirks:


C/include/c4Certificate.h
```diff
@@ -67,6 +67,8 @@ CBL_CORE_API bool c4cert_subjectNameAtIndex(C4Cert* cert,
                                unsigned index,
                                C4CertNameInfo *outInfo) C4API;
 
+#endif // COUCHBASE_ENTERPRISE
+
 /** Returns the time range during which a (signed) certificate is valid.
```
jianminzhao, line 72: Why not moving up this function so that "#ifdef COUCHBASE_ENTERPRISE" won't have to be broken?

jianminzhao (approved):

PR #1682, opened 2023-01-25
CBL-4118: Replicating with non-default coll against SG3.0 hangs Replicator

callumbirks:


callumbirks:
I tested applying these changes against my CBL-4100 branch, the test where this issue shows up now passes, and Replicator stops as it should.

Replicator/Replicator.cc
```diff
@@ -817,6 +820,11 @@ namespace litecore { namespace repl {
             MessageIn *response = progress.reply;
 
             if (response->isError()) {
+                if (response->getError().message == "No handler for BLIP request") {
+                    auto error = C4Error::printf(LiteCoreDomain, kC4ErrorRemoteError,
+                                                 "No handler for BLIP request 'getCollections'");
+                    return onError(error);
+                }
                 return gotError(response);
```
jianminzhao, line 828: Have you tried to find out why gotError not working generally? It appears that gotError is kind-of used pretty standardly.

jianminzhao (approved):

Replicator/Replicator.cc
borrrden: Also how is the error turning into 26 instead of 404?

Replicator/Replicator.cc
```diff
@@ -48,7 +50,8 @@ namespace litecore { namespace repl {
     static constexpr StoppingErrorEntry StoppingErrors[] = {
         {{ LiteCoreDomain, kC4ErrorUnexpectedError,0 }, true, "An exception was thrown"_sl},
         {{ WebSocketDomain, 403, 0}, true, "An attempt was made to perform an unauthorized action"_sl},
-        {{ WebSocketDomain, 503, 0 }, false, "The server is over capacity"_sl}
+        {{ WebSocketDomain, 503, 0 }, false, "The server is over capacity"_sl},
+        {{ LiteCoreDomain, 26, 1 }, true, "No handler for BLIP request 'getCollections'"_sl }
```
pasin, line 54: Better to use kC4ErrorRemoteError instead of 26.

Replicator/Replicator.cc
```diff
@@ -817,6 +820,11 @@ namespace litecore { namespace repl {
             MessageIn *response = progress.reply;
 
             if (response->isError()) {
+                if (response->getError().message == "No handler for BLIP request") {
```
pasin, line 823: Is it possible to check error domain and code instead of message?

Replicator/Replicator.cc
```diff
@@ -817,6 +820,11 @@ namespace litecore { namespace repl {
...
+                    auto error = C4Error::printf(LiteCoreDomain, kC4ErrorRemoteError,
+                                                 "No handler for BLIP request 'getCollections'");
+                    return onError(error);
```
pasin, line 826: It would be better if gotError(response) works but if not, I think there is also gotError(error) as well.

pasin (changes requested):

Replicator/Replicator.cc
callumbirks: All `gotError(MessageIn response)` does is extract the error from a response, and pass that to `onError()`. 1. When I let that error get passed to gotError, it goes to blipToC4Error() and causes "Received unknown error" to be logged a bunch of times.

You may wonder why I didn't add handling for this error into blipToC4Error(). The reason is because "No handler for BLIP request" could be for any request that we send to SG, and I don't want to silence future potential errors. Additionally, handling it here in getCollections, I can add in the context ('getCollections') to the error message to be logged.

I changed the error into 26 (kC4ErrorRemoteError) myself. 404 is the BLIP error, in Core context I believe it should be 26 (remote error).

Replicator/Replicator.cc
callumbirks: I am worried that the error code (404) is quite general and not specific to this error. All I could find internally about it was that 404 corresponds to "Not Found" (for anything). And I spoke to Ben about it, he said there could be other errors which return a 404.

Replicator/Replicator.cc
callumbirks: `gotError(error)` just does an extra logging call, then calls `onError()`. The logs were very bloated when I used `gotError`, because it's already logged about 4 times, without the extra log.

Replicator/Replicator.cc
borrrden: The same logic applies for your addition of "remote error" into the stopping error list.  This applies to any error that happened on the Sync Gateway side, basically, and not just in getCollections.

Replicator/Replicator.cc
borrrden: I worry that remote error is too vague because now any remote error is going to stop the replicator, with a context specific message about "getCollections" (as added in the stopping entry list)

Replicator/Replicator.cc
borrrden: I also suggest changing the message here, if you think keeping 26 is the right call.  You can log a context specific message before you get to this point, but if this is the message here in the stopping error list then it's going to log this message here regardless of the cause or context.

Replicator/Replicator.cc
callumbirks: Ah you're right, I didn't realise the message string wasn't compared as well.

callumbirks: So the actual domain of the error is "BLIP". We don't have an error::Domain or C4ErrorDomain for that (so `blipToC4Error()` will cause "Received unknown error from sever" to be logged). So I either need to add a new error domain to Core, or convert the error (i.e. LiteCoreDomain, kC4RemoteError, original err msg). I imagine it will be the latter (convert the error). 
Then, we need to decide whether to: stop the Replicator on any kC4RemoteError, or add in a new conditional to onError - for that specific error - and stop the Replicator in there. I don't love the latter, but I guess it would do for now.

callumbirks:
I've taken in the advice you all left, and from today's discussion with Jianmin and Pasin, and changed the behaviour to this:
- In `getCollections()`, if an error was received from SG, we log a contextual message "Error response from remote for request 'getCollections'".
- We pass the response to gotError(response), which calls `onError(blipToC4Error())`
- There is a new conditional in `blipToC4Error()` which checks for (BLIP, 404) and translates the error to C4Error (LiteCoreDomain, RemoteError)
- This new error is passed to `Replicator::onError`, which will stop the Replicator as (LiteCoreDomain, RemoteError) has been added to `StoppingErrors`
- The message in the `StoppingErrors` entry is "Unexpected error from remote", because this entry will catch any `kC4ErrorRemoteError`

The log output now looks like this:
```
{Repl#8} Error response from remote for request 'getCollections'
{Repl#8} Got error response: BLIP 404 'No handler for BLIP request'
{Repl#8} Stopping due to fatal error: LiteCore RemoteError, "No handler for BLIP request"
{Repl#8} Connection closed with WebSocket/HTTP status 4002: "Unexpected error from remote" (state=3)
BLIP sent 1 msgs (269 bytes), rcvd 0 msgs (67 bytes) in 0.017 sec. Max outbox depth was 1, avg 1.00
{Repl#8} Got LiteCore error: WebSocket error 4002, "Unexpected error from remote"
{C4Replicator#12} State: busy, progress=0.00%, error=LiteCore RemoteError, "No handler for BLIP request"
```

Replicator/Worker.cc
```diff
@@ -176,6 +176,9 @@ namespace litecore { namespace repl {
         if (err.domain == "HTTP"_sl) {
             domain = WebSocketDomain;
             code = err.code;
+        } else if (err.domain == "BLIP"_sl && err.code == 404) {
+            domain = LiteCoreDomain;
```
jianminzhao, line 180: Did we way that just let the code flow to the end and (having the same effect) ?

callumbirks: It has almost the same effect, apart from it also logs "Received unknown error from remote", which is A. One extra unnecessary logging call and B. The error is known, so slightly inaccurate.

jianminzhao: I meant you may remove this "else if" and won't change anything, right? Without this "else if", it will progress to the following "else', and, if in the future, we decide to add a custom error for it, it will be picked up automatically.
By the way, I don't see "log" in this "else if".

callumbirks: It will hit the log in the final `else` if this elseif does not exist, which I wanted to avoid because of the extra log. But no, it won't change any functionality.

jianminzhao: I see you wanted avoid that extra log, but you may change the log there. 
if (code == 0) { if (err.domain == "BLIP") ...
This makes the code clear that there is no domain in Core named "BLIP"

Replicator/Replicator.cc
jianminzhao: I strongly agree.

callumbirks: ^^ I've done that now, on latest commit

Replicator/Worker.cc
jianminzhao: The impression of this "else if" is that you wanted to avoid the following "else", but I think if we have general treatment of BLIP error, we should let the general error logic to work.

callumbirks: Yeah, I understand your meaning now. I do agree

Replicator/Replicator.cc
```diff
@@ -48,7 +50,8 @@ namespace litecore { namespace repl {
     static constexpr StoppingErrorEntry StoppingErrors[] = {
         {{ LiteCoreDomain, kC4ErrorUnexpectedError,0 }, true, "An exception was thrown"_sl},
         {{ WebSocketDomain, 403, 0}, true, "An attempt was made to perform an unauthorized action"_sl},
-        {{ WebSocketDomain, 503, 0 }, false, "The server is over capacity"_sl}
+        {{ WebSocketDomain, 503, 0 }, false, "The server is over capacity"_sl},
+        {{ LiteCoreDomain, kC4ErrorRemoteError, 1 }, true, "Unexpected error from remote"_sl }
```
borrrden, line 54: Why is this 1 here instead of 0 for the third part of of C4Error?

Replicator/Replicator.cc
```diff
@@ -30,6 +30,8 @@
 #include "Instrumentation.hh"
 #include "fleece/Mutable.hh"
 
+#include <iostream>
```
pasin, line 33: Is this include required?

Replicator/Replicator.cc
```diff
@@ -817,6 +820,7 @@ namespace litecore { namespace repl {
             MessageIn *response = progress.reply;
 
             if (response->isError()) {
+                logError("Error response from remote for request 'getCollections'");
```
pasin, line 823: Maybe add error code and domain as well? It seems missing the information here.

Replicator/Replicator.cc
callumbirks: Now that you point it out, I see that it's probably unnecessary here, as it's not compared. But that's what the error from the SG response had. (`internal_info == 1`)

Replicator/Replicator.cc
callumbirks: You're right, it isn't. I think it got auto-included by the IDE when I was trying to output something.

Replicator/Replicator.cc
callumbirks: The code and domain of the response's error is already logged in `gotError()`, so I wanted to let that function handle error reporting, rather than bloating up this function. I just wanted to add some context as to where the (next line in the log) error came from.

pasin (approved):

PR #1685, opened 2023-01-26
CBL-4129: Inconsistent serialization/deserialization of Checkpoint::_…

jianminzhao:
…completed


pasin (approved):
The fix looks right to me that the logic b/w write and read is not consistent.

PR #1695, opened 2023-02-02
CBL-4131: SQL++ to allow UNICODE with LOCALE parameter.

jianminzhao:
The JSON format already has LOCALE property. This PR extends the SQL++ grammar to include it.

### DIFF
--- a/LiteCore/Query/N1QL_Parser/n1ql.leg
+++ b/LiteCore/Query/N1QL_Parser/n1ql.leg
@@ -282,8 +282,16 @@ collateSuffix =
         )+ ')' _
     )                                   { $$ = co; }
 
+collationLang =
+    <"NOUNICODE"i>                      { $$ = string(yytext); }
+  | (
+        "UNICODE"i                      { $$ = string("UNICODE"); }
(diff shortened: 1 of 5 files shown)


borrrden (approved):

LiteCore/Query/N1QL_Parser/n1ql.leg
```diff
@@ -282,8 +282,16 @@ collateSuffix =
         )+ ')' _
     )                                   { $$ = co; }
 
+collationLang =
+    <"NOUNICODE"i>                      { $$ = string(yytext); }
+  | (
+        "UNICODE"i                      { $$ = string("UNICODE"); }
+        <(':' [a-zA-Z] [a-zA-Z0-9_]*)?> { $$ = $$.as<string>() + string(yytext); }
```
pasin, line 289: I'm getting confused about '_' as it could also mean whitespace. Do I understand correctly that something like this is allowed, "en_US".

jianminzhao: yes, it is allowed as I understand. It's not?

PR #1696, opened 2023-02-03
CBL-4020: Handle missing trailing slashes in db copy 

velicuvlad:


velicuvlad:
reported for C initially, but as is just functionality exposed from Core, it was decided that this is handled in Core

C/c4Database.cc
```diff
@@ -153,7 +153,12 @@ static C4DatabaseConfig newToOldConfig(const C4DatabaseConfig2 &config2) {
 
 /*static*/ void C4Database::copyNamed(slice sourcePath, slice destinationName, const Config &config) {
     ensureConfigDirExists(config);
-    FilePath from(sourcePath);
+    std::string sourcePathString = sourcePath.asString();
```
bmeike, line 156: What about `config.parentDirectory`?  Does it need a trailing '/'?  Does it get one?

bmeike (approved):
Looks good as far as it goes.  If config.parentDirectory needs a trailing slash, though, it would be good if LiteCore fixed that too.
If it doesn't need one, let me know.

C/c4Database.cc
```diff
@@ -153,7 +153,12 @@ static C4DatabaseConfig newToOldConfig(const C4DatabaseConfig2 &config2) {
...
+        if (sourcePathString.back() != '/')
+        {
+            sourcePathString.append("/");
+        }
+    FilePath from((slice(sourcePathString)));
```
jianminzhao, line 161: I think the real problem is in FilePath. Currently, FilePath::isDir() is based on syntax, whether there is a trailing slash. This can be a problem other place. /path/to_db, whether it ends with slash better not to be significant. 
For Apple, for example, we pass it to Apple's API that has RECURSIVE option. In Linux, we currently implement our own RECURSIVE based on the syntax of the source path. In my opinion, we'd better make our recursive copy more robust.

jianminzhao (changes requested):

C/c4Database.cc
velicuvlad: From what I found, FilePath `_dir` is enforced to end with separator `/`.

C/c4Database.cc
velicuvlad: from the tests Core has seems that the utils that call on `parentDir` (exists and delete) are not bothered if there is/not a trailing "/".

C/c4Database.cc
jianminzhao: yes, that's at the syntactic level. But at the time of execution, we may adjust it based on the actual facts. For example, we pass fromPathSrc to Apple API, which has COPYFILE_RECURSIVE flag.

C/c4Database.cc
velicuvlad: But `config.parentDirectory` does end with `/` when set at creation.

pasin, line 156: `FilePath from(sourcePath, "");`

When doing above, the FilePath will not try to split sourcePath into dir and file and the trailing '/' will be handled correctly as it is telling that sourcePath is a directory. We are doing this in multiple places in LiteCore.

pasin (changes requested):

C/c4Database.cc
pasin: I think @bmeike has a point her to check. What if the parentDirectory doesn't have a trailing '/', has LiteCore handled this already?

jianminzhao: yes, this is a nice way telling the library that its a directory. The issue is whether should we rely on the syntactic input to do the recursive copy. Here is the snippet #if __APPLE__,
   copyfile_flags_t flags = COPYFILE_ALL | COPYFILE_RECURSIVE;
   result = copyfile(fromPathStr, toPathStr, nullptr, flags);
They don't reply on the syntax of the path. There is a disparity here between Linux and Apple.

jianminzhao: parentDir is find w/ or w/o trailing slash, ::mkdir() does not care of the trailing slash. mkdir A == mkdir A/

pasin: I think the scope to enhance isDir() and recursive copying is much bigger than this ticket itself. I suggested that it should be done in a separated ticket (post helium).

velicuvlad: as Jianmin confirmed, indeed it doesn't make a difference. By default config parent is taken with "/" and in the case it is manually set, I have checked for all cases where it is used and tests pass even without "/"

C/c4Database.cc
pasin: > I think the real problem is in FilePath. Currently, FilePath::isDir() is based on syntax, whether there is a trailing slash.

With the fact that a FilePath object can be created from a non-existing path, probably the problem is how to use the current FilePath the right way?

pasin: From another angle, may this problem can be solved by adding `FilePath(std::string_view path, bool isDir)` so that it's clear that the path is dir or file.

C/c4Database.cc
jianminzhao: okay

jianminzhao (approved):

C/c4Database.cc
borrrden: What happened to the initial change that Pasin suggested here?  That would do what these lines of code are attempting to replicate.

pasin (approved):

PR #1699, opened 2023-02-14
CBL-4175: Errors from the encryptor callback should cause the checkpo…

jianminzhao:
…int to move past it

### DIFF
--- a/Replicator/Pusher+Revs.cc
+++ b/Replicator/Pusher+Revs.cc
@@ -70,6 +70,13 @@ namespace litecore::repl {
             c4err = C4Error::make(LiteCoreDomain, kC4ErrorNotFound);
         }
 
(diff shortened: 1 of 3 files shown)


bmeike:
Shouldn't this target the release/lithium branch?

pasin (approved):

pasin (commented):
~~It would be good if we have a test that tests this logic without going through the replicator. All tests are also passed without the fix so I think at least we need a test that tests the issue?~~

jianminzhao:
Right now, after this PR, in terms of the code, lithium and helium are matched. Perhaps we can add tests in postHelium

jianminzhao:
I answered to wrong context. For this PR, yes, we have test cases.

PR #1700, opened 2023-02-15
CBL-4191: Fix a bug in Checkpoint.cc

jianminzhao:


borrrden (approved):

bmeike (approved):

pasin (commented):
It would be good if we have a test that tests this logic without going through the replicator. All tests are also passed without the fix so I think at least we need a test that tests the issue?

jianminzhao:
Yes, it's logically a problem of consistency between serialization & deserialization. It's so easy to change on and forget the other part.

bmeike (approved):

pasin (approved):

PR #1701, opened 2023-02-15
CBL-4125: Port Replicator test helpers, SG, SGTestUser, ReplParams to Lithium

callumbirks:


jianminzhao:
I am paying particular attention to the files shared by the product, as opposed to the test. I see 3 files, REST/Response.hh, Replicator/ReplicatorOptions.hh, and Crypto/PublicKey+Apple.mm. The first two seem okay. Do you need changes to PublicKey. Looks to me all formatting. If so, I would suggest not to touch it.

callumbirks:
@jianminzhao There are some differences in PublicKey. I believe this is the one from master. I copied it to here because there were some additions to the one in master, and I was having issues with some function in this file.

jianminzhao:
In general, we don't want to touch the release branch. Changes, again in general, are due to CBSEs. Since tests do not go to the product, we can be little lax on them. Better to hold back the large amount of formatting changes. Then, for the material changes, we need to assess whether it's critical to the customers.

bmeike:
Not just "in general".  Release branches are off limits unless:
- There is a critical or higher priority ticket requiring the change
- The change is the minimum change necessary to effect the fix
- in the absence of consensus about how to handle vulnerability releases, the previous release must be clearly tagged in the git root and all submodules

bmeike (changes requested):
I think this brings up an issue that I will discuss with the Principles: since we maintain several versions at the same time, we may, at some point, want to make more pervasive changes than are necessary for a simple bug fix.  I am not certain that this is something we want to do, but it has not been discussed and it might be.
For now, though, this change is just nothing that can go into the 3.0 line.  It is a huge (though, I can see, relatively benign) change, we don't have QE resources to test it, and it not a response to a user issue.

callumbirks:
Apologies for the delay. Changes are now constrained to `test/` folders, re-ran tests and all passing still.

jianminzhao:
Excellent. Thanks!

bmeike:
Ok.  I still have a question about this commit: Why is it going to the 3.1 branch?  Will it cause us to make any change in the 3.1 branch?  Are these tests failing in Beryllium?

Can someone cite, specifically, the advantage of having these tests in a branch that will not change in a way that these tests will test?

bmeike (changes requested):
I appreciate all the work that has gone into this.  It still seems to me, though, like a somewhat arbitrary change.  I'd like a clear reason for putting them in the 3.1, instead of the 3.2 branch.

jianminzhao:
"As we tried to test 3.0 Lite against 3.1 SGW, Callum and myself wrote bunch of test code on the Lithium line. (no panic, not merged in yet). In the spirit of the recent meeting on proper control of all the release branches, I want to have your opinions as how to manage those code. The code definitely not meant for the release line." -- excerpted from an earlier email.
Basically, we want to test against 3.1 SGW of non-walrus mode. (All 3.0 LiteCore' test cases are against walrus SG)

bmeike:
I'm not clear on what you are trying to say with that excerpt.  I am, here, giving you my opinion on how to manage this code.  I do not think it should be checked in to the Lithium branch.

There is nothing stopping you from running these tests against the Lithium branch.  Please do that.  If any of the tests fail, please create a ticket for the failure and check in a fix and a test for that failure.
If these tests are not in the Helium branch, please do the same thing there: run the tests.  If any of them fail create a ticket describing the failure and check in a fix and a test for the failure.

The idea is to make minimal, necessary changes to release branches.  I have not heard any reason that these changes, to a release branch, are necessary.

bmeike:
I discussed this with @callumbirks .  He explained that these tests are not unit tests (which would be useless, because this code base is frozen) but system tests that verify that changes in **the SGW** are still compatible with the 3.0 line of LiteCore.  We all agree that they are low risk.  Protecting against changes in SGW also seems necessary.  Approved.

bmeike (approved):
Low risk; validates changes in SGW.
Go for it.

PR #1704, opened 2023-02-17
CBL-4205: Replicator hangs if it uses only the default collection that is not in Sync-Gateway.

jianminzhao:


pasin (commented):
From the discussion in https://docs.google.com/document/d/1kbPUuWgtnAgV9Oj55_soWmIPA8f9v3RBrjMXVvbn4p4/edit#

Is 503 expected to be returned instead of 400?

jianminzhao:
This is the current response from SG. We are dealing the error when Lite starts replicator with collection that is not recognized by SG. I think 400/Bad Request makes sense: the request violates the protocol.

pasin (approved):
I think the PR title and commit message should say that this is the case for using only the default collection and the default collection is not configured on SG. This is very specific to the default collection because GetCheckpoint is used instead of GetCollections.

pasin (commented):
I did approved but I have another question. For the case of non-default collection, what is the current behavior? Is the replicator stopped or offline / retry in that case?

jianminzhao:
test case for collection not existing in SG:   
 SECTION("Collection absent at the remote") {
        expectedErrorMsg = "Collection 'flowers.dummy' is not found on the remote server"_sl;
        expectedError = { WebSocketDomain, 404 };
    }

borrrden (approved):

PR #1706, opened 2023-02-23
CBL-4174: Consistent formatting across Core

callumbirks:
There's a PDF document attached to the [Jira ticket](https://issues.couchbase.com/browse/CBL-4174) which outlines the formatting style I think we should be following. Do note that one of the code blocks' formatting was messed up by the conversion to PDF, and the correct formatting for that code block can be found in the comments of the Jira ticket.

This PR contains the formatting changes adhering to the style outlined in the PDF.


C/c4Document.cc
```diff
@@ -110,28 +108,24 @@ C4Document::~C4Document() {
...
 C4Document::C4Document(const C4Document &doc)
-:_flags(doc._flags)
-,_docID(doc._docID)
-,_revID(doc._revID)
-,_sequence(doc._sequence)
-,_selected(doc._selected)
-,_selectedRevID(doc._selectedRevID)
-,_collection(doc._collection)
-{
+        : _flags(doc._flags), _docID(doc._docID), _revID(doc._revID), _sequence(doc._sequence), _selected(doc._selected),
```
snej, line 111: Shoving all the initializers together makes them much harder to read. If these were assignment statements, they'd be on separate lines, right? Initializers are very important, including the order they happen in, so they should be as clear as possible.

C/c4Document.cc
```diff
@@ -39,17 +39,15 @@ using namespace litecore;
...
     // Quick sanity test of the docID, but no need to scan for valid UTF-8 since we're not inserting.
     if (_docID.size < 1 || _docID.size > kMaxDocIDLength)
         error::_throw(error::BadDocID, "Invalid docID \"%.*s\"", SPLAT(_docID));
 
 #if DEBUG
     // Make sure that C4Document and C4Document_C line up so that the same object can serve as both:
-    auto asStruct = (C4Document_C*)this;
-    if (&_flags != &(asStruct)->flags || (void*)&_docID != (void*)&(asStruct)->docID) {
+    auto asStruct = (C4Document_C *)this;
+    if (&_flags != &(asStruct)->flags || (void *)&_docID != (void *)&(asStruct)->docID) {
```
snej, line 50: Putting spaces in a typecast like this looks awkward IMO, and makes it less clear that it's a single operation.

C/c4Document.cc
```diff
@@ -39,17 +39,15 @@ using namespace litecore;
 
 
 C4Document::C4Document(C4Collection *collection, alloc_slice docID_)
-:_collection(asInternal(collection))
-,_docID(move(docID_))
-{
+        : _collection(asInternal(collection)), _docID(move(docID_)) {
```
snej, line 42: Why are the initializers indented way over like this?

C/c4DocEnumerator.cc
```diff
@@ -40,27 +40,22 @@ class C4DocEnumerator::Impl : public RecordEnumerator, public fleece::InstanceCo
...
-    {
+    static RecordEnumerator::Options recordOptions(const C4EnumeratorOptions &c4options) {
         RecordEnumerator::Options options;
         if (c4options.flags & kC4Descending)
             options.sortOption = kDescending;
         else if (c4options.flags & kC4Unsorted)
             options.sortOption = kUnsorted;
         options.includeDeleted = (c4options.flags & kC4IncludeDeleted) != 0;
-        options.onlyConflicts  = (c4options.flags & kC4IncludeNonConflicted) == 0;
+        options.onlyConflicts = (c4options.flags & kC4IncludeNonConflicted) == 0;
```
snej, line 58: The extra space was to line up this statement with the one above it since they are very similar; it exposes the similarity.

C/c4DocEnumerator.cc
```diff
@@ -31,7 +31,7 @@ using namespace litecore;
 #pragma mark - DOC ENUMERATION:
 
 CBL_CORE_API const C4EnumeratorOptions kC4DefaultEnumeratorOptions = {
-    kC4IncludeNonConflicted | kC4IncludeBodies
+        kC4IncludeNonConflicted | kC4IncludeBodies
```
snej, line 34: Why is this indented two levels at once?

C/c4Database.cc
```diff
@@ -324,8 +316,8 @@ Retained<C4Document> C4Database::getDocumentBySequence(C4SequenceNumber sequence
 }
 
 Retained<C4Document> C4Database::putDocument(const C4DocPutRequest &rq,
-                                 size_t* C4NULLABLE outCommonAncestorIndex,
-                                 C4Error *outError) {
+                                             size_t *C4NULLABLE outCommonAncestorIndex,
```
snej, line 319: This spacing is weird because the `*` has nothing to do with the C4NULLABLE keyword, it's part of the `size_t`.

C/c4Collection.cc
```diff
@@ -24,42 +24,39 @@ using namespace litecore;
...
     return doc;
 }
 
 
-Retained<C4Query> C4Collection::newQuery(C4QueryLanguage language, slice expr,int *errPos) const {
+Retained<C4Query> C4Collection::newQuery(C4QueryLanguage language, slice expr, int *errPos) const {
     if (_usuallyFalse(!_database))
         failClosed();
-    return C4Query::newQuery(const_cast<C4Collection*>(this), language, expr, errPos);
+    return C4Query::newQuery(const_cast<C4Collection *>(this), language, expr, errPos);
```
snej, line 61: Putting a space before the `*` doesn't make sense here; `C4Collection*` is a type.

The reason for the spacing comes from pointer variable declarations like `char *foo` where it's supposed to highlight that `*foo` is a `char`. It also promotes writing multiple declarations correctly, like `char *foo, *bar` instead of `char* foo, bar` (where bar gets the wrong type.)

There's no reason to put a space before a `*` (or `&`) in another context; it's not clarifying anything then.

C/c4BlobStore.cc
```diff
@@ -282,29 +279,25 @@ void C4BlobStore::replaceWith(C4BlobStore &other) {
...
-:_impl(move(other._impl))
-{ }
+C4ReadStream::C4ReadStream(C4ReadStream &&other)
+        : _impl(move(other._impl)) {}
 
 C4ReadStream::~C4ReadStream() = default;
-size_t C4ReadStream::read(void *dst, size_t mx)     {return _impl->read(dst, mx);}
-int64_t C4ReadStream::getLength() const             {return _impl->getLength();}
-void C4ReadStream::seek(int64_t pos)                {_impl->seek(pos);}
+
```
snej, line 288: In these single-line inline methods, if the method bodies aren't lined up they get a lot harder to read. I was really meticulous about making these line up nicely.

Also, adding blank lines between them doesn't seem to add anything (they're one per line so it's already clear where the separations are), and it takes up more vertical space so you can't see as much onscreen.

C/c4BlobStore.cc
```diff
@@ -34,21 +34,21 @@ namespace C4Blob {
...
 
 
-
 #pragma mark - C4BLOBKEY:
 
 
 static constexpr slice
-    kBlobDigestStringPrefix = "sha1-",  // prefix of ASCII form of blob key ("digest" property)
-    kBlobFilenameSuffix = ".blob";      // suffix of blob files in the store
+        kBlobDigestStringPrefix = "sha1-",  // prefix of ASCII form of blob key ("digest" property)
```
snej, line 41: The alignment of the comments got lost here.

C/tests/c4DatabaseEncryptionTest.cc
```diff
@@ -46,12 +47,11 @@ TEST_CASE("Database Key Derivation", "[Database][Encryption][C]") {
...
                                   C4EncryptionAlgorithm alg) = nullptr;
     string expectedKey;
-    C4EncryptionKey key {};
+    C4EncryptionKey key{};
     SECTION("SHA256") {
         c4key_setPasswordFunc = c4key_setPassword;
         expectedKey = "ad3470ce03363552b20a4a70a4aec02cb7439f6202e75b231ab57f2d5e716909";
-    }
-    SECTION("SHA1") {
+    }SECTION("SHA1") {
```
snej, line 54: Why on earth did it collapse the next SECTION onto the same line as the closing brace?

I'm guessing the formatter spazzed because Catch's macros confuse it.

LiteCore/Database/DatabaseImpl.cc
```diff
@@ -1025,17 +1022,17 @@ namespace litecore {
...
 
     C4RemoteID DatabaseImpl::getRemoteDBID(slice remoteAddress, bool canCreate) {
         bool inTransaction = false;
         bool commit = false;
         DEFER {
-            if (inTransaction) {
-                endTransaction(commit);
-            }
-        };
+                  if (inTransaction) {
```
snej, line 1032: This spacing is weird! Especially that the closing brace is indented farther than the opening line. I suspect the DEFER macro confused the formatter.

snej (commented):
I was just skipping through the diffs, looking for types of things I dislike, and also found a few specific places where the formatter seemed to be confused by macros that hide control flow (Catch has a bunch of these like SECTION, also we have one called DEFER.)

C/c4Document.cc
callumbirks: I think CLion's formatter was doing something weird with indents. Luckily it comes with built-in Clang-Format support, so I've switched to that

C/tests/c4DatabaseEncryptionTest.cc
callumbirks: Luckily not a problem with Clang-Format, just seems CLion's formatter has some limitations

C/c4Document.cc
callumbirks: Fixed with move to Clang-Format

C/c4Document.cc
callumbirks: Fixed with move to Clang-Format

C/c4DocEnumerator.cc
callumbirks: Fixed with move to Clang-Format

C/c4DocEnumerator.cc
callumbirks: Fixed with move to Clang-Format

C/c4Database.cc
callumbirks: This one is difficult. We'll have to discuss more. I don't think Clang-Format has an option to cover this case.

C/c4Collection.cc
callumbirks: Fixed with move to Clang-Format

C/c4BlobStore.cc
callumbirks: Unfortunately, I couldn't find an option to do this automagically with Clang-Format. Luckily, you can just disable Clang-Format for sections of code, so I've returned your code to it's previous beautiful alignment, and added `// clang-format off` for that section of code.

C/c4BlobStore.cc
callumbirks: Move to Clang-Format has fixed the alignment of the comments, although it has aligned the variables weirdly. A small thing, that I'm not sure how to fix, but I can investigate if needed.

LiteCore/Database/DatabaseImpl.cc
callumbirks: Fixed with move to Clang-Format

callumbirks:
It turns out the order of our includes is extremely sensitive. It'd be nice to have a consistent order to the headers, and Clang-Format is pretty flexible with categorising headers to ensure specific orders, but I couldn't narrow down which file's include order was breaking the builds, so I've turned include re-ordering off.

C/Cpp_include/access_lock.hh
```diff
@@ -35,128 +35,135 @@ namespace litecore {
...
-            access(MUTEX &mut, const Retained<R>& ref, SENTRY& sentry):access(mut, ref.get()) {
-                if (sentry) sentry(ref);
+
+            access(MUTEX &mut, const Retained<R> &ref, SENTRY &sentry) : access(mut, ref.get()) {
+                if ( sentry ) sentry(ref);
             }
+
             access_lock::LOCK_GUARD _lock;
-            const R*                _ref;
+            const R                *_ref;
```
borrrden, line 150: Not a big fan of this format, as it's easy to glance down the types and miss that this is `R*` instead of just `R`

C/Cpp_include/access_lock.hh
```diff
@@ -167,56 +174,50 @@ namespace litecore {
...
+            if ( _sentry ) _sentry(_contents);
             return callback(_contents);
         }
 
         // const versions:
 
-        auto useLocked() const  {
-            return access<const T&>(getMutex(), _contents, const_cast<SENTRY&>(_sentry));
-        }
+        auto useLocked() const { return access<const T &>(getMutex(), _contents, const_cast<SENTRY &>(_sentry)); }
```
borrrden, line 183: How long are lines allowed to be in this format?

C/Cpp_include/c4BlobStore.hh
```diff
@@ -86,43 +85,48 @@ namespace C4Blob {
...
 
-struct C4ReadStream : public fleece::InstanceCounted, C4Base {
+struct C4ReadStream
+    : public fleece::InstanceCounted
+    , C4Base {
     C4ReadStream(const C4BlobStore&, C4BlobKey);
     C4ReadStream(C4ReadStream&&);
     ~C4ReadStream();
-    size_t read(void *buffer, size_t maxBytes);
+    size_t  read(void* buffer, size_t maxBytes);
```
borrrden, line 97: What is the reasoning for this one being different than the others with the `*` placement?

borrrden (commented):
I've gotten through c4Collection.hh and I think I see the same pattern of changes happening.  I commented on some of the ones that stood out but I noticed a slight inconsistency as well.

C/Cpp_include/c4BlobStore.hh
callumbirks: That's very strange. I have no idea. Possibly some of the settings I chose

C/Cpp_include/access_lock.hh
callumbirks: 120. Most standards have it at 80, but this feels quite strict (especially with higher resolution displays becoming increasingly common)

C/Cpp_include/access_lock.hh
callumbirks: There will probably be numerous examples of this throughout the code - where we have broken the rules, and Clang-Tidy would not be aware we have broken the rules. Each one either needs formatting manually, and disabling Clang-Tidy for that section of code, or needs accepting the regular rule.

C/Cpp_include/c4BlobStore.hh
callumbirks: Turns out `DerivePointerAlignment` was on. This means Clang-Format will analyze each file's most common pointer placement, and use that for that file. In this file, most pointers and references were attached to the type, so it followed that.

callumbirks:
Changed PointerAlignment to Left (next to typename). This will suit most of our cases, and Clang-Tidy will put pointers next to the variables for multiple variable assignment, so that case is also covered.

jianminzhao:
I read through the pdf description of the new style and I am all for it (pretty much my usual style). My only concern is blame. You mentioned that there is something you may do to have blame skip this commit. Is it feasible to include that as part of this PR?

callumbirks:
@jianminzhao 
The instructions for ignoring specific commits for BLAME are as follows:
Create a file (i.e. `.git-blame-ignore-revs`) and put the hash of each commit you want to ignore in the file (one per line).
Then do `git config blame.ignoreRevsFile .git-blame-ignore-revs`.

So as it depends on the commit hash, I think this will be a new commit hash when the branch is squashed + merged? So I think the file would have to be created in a follow-up PR. Then configuring git blame to ignore that commit is a local thing, so each person will have to call that `git config` command manually for their machine. For that reason, I think it may be sensible for me to also update the README with information about this - in the PR where I add the `.git-blame-ignore` file.

jianminzhao (approved):

C/Cpp_include/access_lock.hh
```diff
@@ -35,128 +35,135 @@ namespace litecore {
...
 #else
-            access(access&&) = delete;       // I cannot be moved
+            access(access&&) = delete;  // I cannot be moved
 #endif
 
-            auto get()                      {return _ref;}
-            operator const R* ()            {return _ref;}
-            auto operator * ()              {return _ref;}
-            auto operator -> ()             {return _ref;}
+            auto get() { return _ref; }
```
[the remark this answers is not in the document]
callumbirks: `SeparateDefinitionBlocks`. Doesn't seem like there's an option to change this behaviour for single-line functions. I can put `// clang-format off` and `// clang-format on` around the block of functions, and manually remove the newlines.

C/Cpp_include/access_lock.hh
```diff
@@ -35,128 +35,135 @@ namespace litecore {
...
+
+            operator const R*() { return _ref; }
```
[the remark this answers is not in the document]
callumbirks: I can change `AlignConsecutiveDeclarations` from `Consecutive` to `AcrossEmptyLines`

C/c4Collection.cc
```diff
@@ -22,44 +22,30 @@
...
 
-
-C4Collection::C4Collection(C4Database *db, C4CollectionSpec spec)
-:_database(db)
-,_scope(spec.scope)
-,_name(spec.name)
-{ }
-
+C4Collection::C4Collection(C4Database* db, C4CollectionSpec spec)
+    : _database(db), _scope(spec.scope), _name(spec.name) {}
```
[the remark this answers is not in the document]
callumbirks: I can change that behaviour, I think the current behaviour is it will put them on a single line if they fit, otherwise will put them on newlines

C/c4Collection.cc
```diff
@@ -22,44 +22,30 @@
...
+    if ( !doc ) doc = TreeDocumentFactory::documentContaining(value);
     return doc;
 }
 
-
-Retained<C4Query> C4Collection::newQuery(C4QueryLanguage language, slice expr,int *errPos) const {
-    if (_usuallyFalse(!_database))
-        failClosed();
+Retained<C4Query> C4Collection::newQuery(C4QueryLanguage language, slice expr, int* errPos) const {
+    if ( _usuallyFalse(!_database) ) failClosed();
```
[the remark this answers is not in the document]
callumbirks: `AllowShortIfStatementsOnASingleLine: Always` ('Always' meaning whether it is 'if', 'if-else' etc, each statement can still always be on a single line if it is short)

snej (approved):

C/Cpp_include/access_lock.hh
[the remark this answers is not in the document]
callumbirks: It turned out that `AcrossEmptyLines` caused a lot of other weird formatting behaviour, so I'm going to leave this as is for now.

callumbirks:
Discussed some final points about the formatting in sync-up with Jim, hopefully this is the last change needed for formatting. I've moved the `run-clang-format.sh` script into `build_cmake/scripts` and added an Xcode scheme which runs the script when you build that scheme (Tested and it works as expected).

borrrden (approved):

PR #1713, opened 2023-03-03
CBL-4220: Sync via RESTListener /_replicate does not include User-Agent

jianminzhao:
We add user agent header for BuiltInWebSocket if it is not already in the header. This header will be defaulted to, like "User-Agent: couchbase-lite-core/3.1.0-EE".

### DIFF
--- a/Networking/WebSockets/BuiltInWebSocket.cc
+++ b/Networking/WebSockets/BuiltInWebSocket.cc
@@ -184,6 +184,25 @@ namespace litecore { namespace websocket {
         // Create the HTTPLogic object:
         Dict headers = options()[kC4ReplicatorOptionExtraHeaders].asDict();
         HTTPLogic logic {Address(url()), Headers(headers)};
+        bool foundUserAgent = false;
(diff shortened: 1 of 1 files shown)


Networking/WebSockets/BuiltInWebSocket.cc
```diff
@@ -184,6 +184,25 @@ namespace litecore { namespace websocket {
...
+        bool foundUserAgent = false;
+        for (auto iter = headers.begin(); iter != headers.end(); ++iter) {
+            if (iter.keyString().caseEquivalent("User-Agent")) {
+                foundUserAgent = true;
+                break;
+            }
+        }
+        if (!foundUserAgent) {
+            string ua = "couchbase-lite-core/";
+            alloc_slice ver = c4_getVersion();
```
pasin, line 196: I think we can just put whatever returned here without discarding the info -The CBL platforms include this info to the User-Agent as well.

pasin (approved):

PR #1714, opened 2023-03-08
debugging cbl-4208 which repros only in the build machine.

jianminzhao:
Won't merge it in.


jianminzhao:
It was for debugging purpose.

PR #1715, opened 2023-03-08
CBL-4254: P2P tests giving invalid collection error

jianminzhao:
The error occurs When the database name contains dots and URL parser takes the suffix after the dot as the collection, which is not valid in this context. Since the P2P protocol only uses the database, we simply take the whole string in the path as the name of the database which may contain dots.


pasin (approved):

bmeike:
This change affects only the REST code, which, in turn, is used only in the cblite tool?

borrrden:
No, this is the P2P listener code that will go into production via URLEndpointListener.  It is mixed in with the REST code that is only used in the cblite tool, which makes it confusing though.

PR #1719, opened 2023-03-10
CBL-4221: Analysis Warnings (1/13)

callumbirks:
### Notable changes
- c4BlobStore MIME type checks changed to use `std::array` and `std::any_of`
- `c4doc_selectNextLeafRevision` was previously not passing parameter "withBody" to `C4Document::selectNextLeafRevision`, now it does. Although this is apparently hardcoded to 'true' in CBL, so probably why nothing was broken.
- `C4Document::resolveConflict` was not passing parameter "pruneLosingBranch" to the next `resolveConflict` function, now it does. However, we don't expose that param to API, so it probably doesn't make a difference.
- The `C4QueryOptions` parameter has been removed from `C4Query::_createEnumerator` as it was not being used. Although, I have left the API (i.e. `c4query_run`) alone, so it still has that (unused) parameter.
- The `error` struct now has a new constructor including a parameter for `backtrace`. This is to enable using a temporary anonymous value for the error to throw, in order to avoid unnecessary copies (which is what was previously happening).
- Unqualified usages of `std::move` (just `move()`) have been fixed. There's a good conversation about the reasons behind this [here](https://reviews.llvm.org/D119670)
- `virtual` has been removed from functions marked `virtual [...] override` as it has no effect
- Deleted constructors moved out of `private` because deleted private functions will give you the wrong warning (function is private rather than function is deleted), and because there is no reason to make the function private as it has no implementation.
- Some keywords replaced on destructors (`override` instead of `virtual`), see [modernize-use-override](https://clang.llvm.org/extra/clang-tidy/checks/modernize/use-override.html) 

Many instances of `// NOLINT` to disable certain Clang-Tidy warnings. Either because the warning doesn't apply, or it would be too difficult to fix (requiring API or major internal changes).


jianminzhao (approved):

C/Cpp_include/c4BlobStore.hh
```diff
@@ -106,15 +107,15 @@ struct C4WriteStream
     : public fleece::InstanceCounted
     , C4Base {
     explicit C4WriteStream(C4BlobStore&);
-    C4WriteStream(C4WriteStream&&);
-    ~C4WriteStream();
+    C4WriteStream(C4WriteStream&&) noexcept;
+    ~C4WriteStream() override;
 
-    C4BlobStore& blobStore() const { return _store; }
+    [[nodiscard]] C4BlobStore& blobStore() const { return _store; }
```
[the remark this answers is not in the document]
callumbirks: I believe Clang-Tidy recommends marking method as `nodiscard` if a method is returning a reference or pointer, because not using the return value from such a method is usually a programming error.

C/Cpp_include/c4Collection.hh
```diff
@@ -64,6 +65,13 @@ struct C4Collection
 
     static C4Document* documentContainingValue(FLValue) noexcept;
 
+    /**
+ * Default arguments on virtual methods are strongly argued against by the C++ standards, Google goes so far as to
```
pasin, line 69: The indent seems off.

C/Cpp_include/c4Query.hh
```diff
@@ -89,28 +89,28 @@ struct C4Query final
     /// auto e = query.run();
     /// while (e.next()) { ... }
     /// ```
-    Enumerator run(const C4QueryOptions* C4NULLABLE opt = nullptr, slice params = fleece::nullslice);
+    Enumerator run(slice params = fleece::nullslice);
```
pasin, line 92: If nobody is using this C4QueryOption, we may remove this from the API as well.

```
/** Options for running queries. */
typedef struct {
    bool rankFullText_DEPRECATED;      ///< Ignored; use the `rank()` query function instead.
} C4QueryOptions;
```

C/Cpp_include/c4Query.hh
```diff
@@ -134,22 +134,22 @@ struct C4QueryObserver
...
 
     /// If the latest run of the query failed, the error will be stored here, with nonzero `code`.
     /// Always check the error before getting the enumerator.
-    C4Error getError() const { return _currentError; }
+    [[nodiscard]] C4Error getError() const { return _currentError; }
 
     /// Returns a new enumerator on the query results.
     /// If the query failed, throws that error as an exception.
-    virtual C4Query::Enumerator getEnumerator(bool forget = true) = 0;
+    virtual C4Query::Enumerator getEnumerator(bool forget = true) = 0;  // NOLINT(google-default-arguments)
```
pasin, line 149: As we have done this in many places, I'm wondering if Clang-tidy has a config file that we can ignore this rule by default?  What is your opinion on this?

C/c4Certificate.cc
```diff
@@ -173,6 +170,7 @@ Retained<C4Cert> C4Cert::requestFromData(slice certRequestData) {
 
 bool C4Cert::isSigned() { return _impl->isSigned(); }
 
+// NOLINTBEGIN(readability-convert-member-functions-to-static)
```
pasin, line 173: Format: I have seen that there is always an extra empty line after the comment somewhere else.

pasin (commented):
LGTM, I just have a few questions.

C/Cpp_include/c4BlobStore.hh
```diff
@@ -161,15 +162,16 @@ struct C4BlobStore : C4Base {
...
 
-    C4BlobStore(const C4BlobStore&) = delete;
-    litecore::FilePath                            dir() const;
-    litecore::FilePath                            pathForKey(C4BlobKey) const;
-    std::unique_ptr<litecore::SeekableReadStream> getReadStream(C4BlobKey) const;
-    std::unique_ptr<litecore::BlobWriteStream>    getWriteStream();
+    [[nodiscard]] litecore::FilePath                            dir() const;
+    [[nodiscard]] litecore::FilePath                            pathForKey(C4BlobKey) const;
+    [[nodiscard]] std::unique_ptr<litecore::SeekableReadStream> getReadStream(C4BlobKey) const;
+    std::unique_ptr<litecore::BlobWriteStream>                  getWriteStream();
```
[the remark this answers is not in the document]
borrrden: This also interests me.  Is it perhaps because the method is not const?  That's the only thing I find different here.

C/c4CAPI.cc
```diff
@@ -147,7 +151,9 @@ size_t c4stream_read(C4ReadStream* stream, void* buffer, size_t maxBytes, C4Erro
 
 int64_t c4stream_getLength(C4ReadStream* stream, C4Error* outError) noexcept {
     try {
-        return stream->getLength();
+        uint64_t streamLength = stream->getLength();
+        DebugAssert(streamLength < std::numeric_limits<int64_t>::max());
+        return static_cast<int64_t>(streamLength);
```
borrrden, line 156: I actually already have a facility for this, it is called `narrow_cast` in NumConversion.hh

borrrden (approved):

C/Cpp_include/c4BlobStore.hh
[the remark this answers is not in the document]
jianminzhao: I think only const method requires [[nodiscard]]

C/c4CAPI.cc
jianminzhao: Probably better use narrow_cast, which checks the bounds.

C/Cpp_include/c4BlobStore.hh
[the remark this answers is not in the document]
callumbirks: Yeah I guess if the method is not const, it assumes you may only want to trigger the modifying-effects of the function, without being interested in the return value

C/Cpp_include/c4Query.hh
callumbirks: My opinion is there's probably a better way to do this than the way we do it, but it would be a tremendous change to fix it. So in the meantime, yes we could create a `.clang-tidy` file I think in order to disable this warning.

C/c4CAPI.cc
callumbirks: Oh awesome, thanks.

C/Cpp_include/c4Query.hh
callumbirks: Should I create a separate ticket for that?

callumbirks:
I've added a `.clang-tidy` file, copied in the current configuration from my IDE, and remove `google-default-arguments`, which is the warning we are explicitly ignoring everywhere as we can't adhere to it.

C/Cpp_include/c4Collection.hh
callumbirks: I think Clang-Format wanted to align the comment with the code, this comment is removed in latest commit now anyway.

C/Cpp_include/c4Query.hh
pasin: Yes

pasin (approved):

C/Cpp_include/c4Query.hh
callumbirks: https://issues.couchbase.com/browse/CBL-4331

C/Cpp_include/c4BlobStore.hh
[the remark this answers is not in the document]
pasin: I think it the result is intended to be used, we could add [[nodiscard]] as well for the information even though it's OK by the tidy.

PR #1725, opened 2023-03-16
CBL-4328: CMake error with MSVC using Ninja

callumbirks:


callumbirks:
@borrrden I was hoping this would still work on MSVC, I guess it doesn't 😭

PR #1726, opened 2023-03-16
CBL-4333: Fix binary log encoding (Helium port)

callumbirks:


callumbirks:
Tested that logs are decoding properly, and manually inspect `.cbllog` files to ensure there is no duplication.

borrrden (approved):
Might be worth putting a note on insert to clarify its behavior since I'm guessing most people would think its behavior is to overwrite.

pasin (approved):

PR #1727, opened 2023-03-17
CBL-4288: Data getting corrupted during PushPull collection replication

borrrden:
This only happens on Windows, seemingly, and the only theory I have is that there is an optimization issue that causes some stale reads of a variable.  The details of the issue is that when reading the Fleece shared keys out of the "remote" db in a db to db pushpull replication, the number of keys in the copy of the shared keys object is 0 instead of what it should be.  This causes a faulty check for "unchanged" shared keys (memory address equal, and count > initial count) which in turn encodes incorrect keys onto the document.

This change seems to fix it, though it is unclear why.  The only theory I have is that reading the "count" variable first is forcing a re-evaluation which then uses the correct value when copying.

### DIFF
--- a/Replicator/DBAccess.cc
+++ b/Replicator/DBAccess.cc
@@ -42,8 +42,6 @@ namespace litecore { namespace repl {
     ,_timer(bind(&DBAccess::markRevsSyncedNow, this))
     ,_usingVersionVectors((db->getConfiguration().flags & kC4DB_VersionVectors) != 0)
     {
-        // Copy database's sharedKeys:
-        SharedKeys dbsk = db->getFleeceSharedKeys();
     }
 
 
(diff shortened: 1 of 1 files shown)


borrrden:
mac failure is addressed in #1726

Replicator/DBAccess.cc
```diff
@@ -305,8 +305,8 @@ namespace litecore { namespace repl {
             lock_guard<mutex> lock(_tempSharedKeysMutex);
             if (!_tempSharedKeys || _tempSharedKeysInitialCount < dbsk.count()) {
                 // Copy database's sharedKeys:
-                _tempSharedKeys = SharedKeys::create(dbsk.stateData());
                 _tempSharedKeysInitialCount = dbsk.count();
```
bmeike, line 308: Are you sure you don't want to put the loop here?  I was much persuaded by your argument that you really didn't have a clear root cause, and that it was unclear why this tweaky change fixed it.

bmeike (approved):
I'd feel better if we were sure the change always fixed the issue... but this is way out of my ken.  I'll totally take your word for ti.

Replicator/DBAccess.cc
jianminzhao: That is, the original first statement relies on the side-effect of second statement that reads the count. Reading the if statement, dbsk.count() may not be called if !_tempSharedKey == true. I wonder if the problem could also be fixed by swapping the two expressions in the if statement.

jianminzhao (approved):

Replicator/DBAccess.cc
pasin: Will getting the count from the `_tempSharedKeys` also solve the issue? I think doing that will eliminate the atomic issue b/w creating a temp shared key and getting its initial count?

pasin: From the ShredKeys.cc in fleece, there is a mutex lock to guard `_count`. It seems like there is one place doesn't do that and it's in the `writeState()` used by `stateData()` method. I don't know if this contributes to the issue or not but it's a suspicous.

https://github.com/couchbase/fleece/blob/dcaa54943e6fcf05a610476912299873b5328bf7/Fleece/Core/SharedKeys.cc#L91

borrrden: @bmeike I suppose I could add it anyway, but this change made just as many passes as the loop did.

@pasin No, in fact that is the core of the problem.  It is getting created incorrectly, so from the temp shared keys point of view the count legitimately is zero. I am going to try using the mutex in stateData as well.

pasin (approved):

borrrden:
I would also feel better if we had proof this was actually the fix, but if we did then this probably would have already been fixed because "that's easy" 😅

pasin (approved):

jianminzhao (approved):

PR #1730, opened 2023-03-20
CBL-4222: Analysis Warnings (2/13) C/include + C/tests

callumbirks:
Only notable change is the macros `__C4_ENUM_ATTRIBUTES` and `__C4_OPTIONS_ATTRIBUTES` have been changed to remove the starting underscores. These names are reserved by C/C++ standard. See https://en.cppreference.com/w/cpp/language/identifiers.

Other changes are mostly things like `[[nodiscard]]` and `const`

### DIFF
--- a/C/include/c4.h
+++ b/C/include/c4.h
@@ -12,16 +12,18 @@
 
 #pragma once
 
-#include "c4BlobStore.h"
-#include "c4Certificate.h"
-#include "c4Collection.h"
-#include "c4Database.h"
-#include "c4Document.h"
-#include "c4Document+Fleece.h"
-#include "c4DocEnumerator.h"
-#include "c4Index.h"
-#include "c4Listener.h"
-#include "c4Observer.h"
-#include "c4Query.h"
-#include "c4Replicator.h"
-#include "c4Socket.h"
+// Mark these includes as 'keep' so Clangd doesn't warn about them not being used in this file
+
+#include "c4BlobStore.h"        // IWYU pragma: keep
+#include "c4Certificate.h"      // IWYU pragma: keep
+#include "c4Collection.h"       // IWYU pragma: keep
+#include "c4Database.h"         // IWYU pragma: keep
+#include "c4Document.h"         // IWYU pragma: keep
+#include "c4Document+Fleece.h"  // IWYU pragma: keep
+#include "c4DocEnumerator.h"    // IWYU pragma: keep
+#include "c4Index.h"            // IWYU pragma: keep
+#include "c4Listener.h"         // IWYU pragma: keep
+#include "c4Observer.h"         // IWYU pragma: keep
+#include "c4Query.h"            // IWYU pragma: keep
+#include "c4Replicator.h"       // IWYU pragma: keep
+#include "c4Socket.h"           // IWYU pragma: keep
--- a/C/include/c4.h
+++ b/C/include/c4.h
@@ -14,16 +14,16 @@
 
 // Mark these includes as 'keep' so Clangd doesn't warn about them not being used in this file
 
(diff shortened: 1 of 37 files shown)


pasin (approved):

C/include/c4.h
```diff
@@ -12,16 +12,18 @@
...
+#include "c4Database.h"         // IWYU pragma: keep
+#include "c4Document.h"         // IWYU pragma: keep
+#include "c4Document+Fleece.h"  // IWYU pragma: keep
+#include "c4DocEnumerator.h"    // IWYU pragma: keep
+#include "c4Index.h"            // IWYU pragma: keep
+#include "c4Listener.h"         // IWYU pragma: keep
+#include "c4Observer.h"         // IWYU pragma: keep
+#include "c4Query.h"            // IWYU pragma: keep
+#include "c4Replicator.h"       // IWYU pragma: keep
+#include "c4Socket.h"           // IWYU pragma: keep
```
pasin, line 29: c4.h is an umbrella header. Wish there is a way to tell as feeling a little weird when seeing `IWYU pragma: keep` in the public API headers.

pasin (approved):

C/include/c4BlobStore.h
```diff
@@ -138,7 +142,7 @@ CBL_CORE_API int64_t c4stream_getLength(C4ReadStream*, C4Error* C4NULLABLE) C4AP
 CBL_CORE_API bool c4stream_seek(C4ReadStream*, uint64_t position, C4Error* C4NULLABLE) C4API;
 
 /** Closes a read-stream. (A NULL parameter is allowed.) */
-CBL_CORE_API void c4stream_close(C4ReadStream* C4NULLABLE) C4API;
+CBL_CORE_API void c4stream_close(C4ReadStream* C4NULLABLE) C4API;  // NOLINT(readability-redundant-declaration)
```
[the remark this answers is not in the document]
callumbirks: It's a redundant declaration because `c4stream_close` is declared in both `c4Base.h` and `c4BlobStore.h`. One option is to remove it from `c4BlobStore.h`.

C/include/c4.h
callumbirks: From some quick research it seems the general consensus is: this is a bad, but common, practice. And the problem of increased compile times etc gets worse the larger your library (I'd say this one is rather large).

C/include/c4BlobStore.h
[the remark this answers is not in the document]
borrrden: That should be done if it hasn't already

C/include/c4BlobStore.h
```diff
@@ -171,7 +175,7 @@ CBL_CORE_API bool c4stream_install(C4WriteStream*, const C4BlobKey* C4NULLABLE e
 /** Closes a blob write-stream. If c4stream_install was not already called (or was called but
         failed), the temporary file will be deleted without adding the blob to the store. 
         (A NULL parameter is allowed, and is a no-op.) */
-CBL_CORE_API void c4stream_closeWriter(C4WriteStream* C4NULLABLE) C4API;
+CBL_CORE_API void c4stream_closeWriter(C4WriteStream* C4NULLABLE) C4API;  // NOLINT(readability-redundant-declaration)
```
borrrden, line 178: Ditto for this.  I see no reason to ever have multiple declarations.

borrrden (changes requested):

C/include/c4.h
callumbirks: I've removed these comments now

jianminzhao: As a side note, I used to work with coding guidelines that recommend to include only necessary headers.

jianminzhao (approved):

borrrden (approved):

PR #1733, opened 2023-03-28
CBL-4368: ReplicatorLoopbackTest flaky failure

callumbirks:
~~Attempted fix ReplicatorLoopbackTest flaky failure. Made DB names in tests random (based on time prefix), not sure if this will actually fix anything.~~

~~Fix collection log format cache flaky failure:~~
~~- There was a flaky failure in `formatWithCollection()` that appeared very occasionally when running certain tests, seemed related to string allocation. After this change, I haven't seen the same failure again.~~

~~I have run all CppTest with these changes, all passing.~~
See recent commits/comment

### DIFF
--- a/Replicator/Worker.hh
+++ b/Replicator/Worker.hh
@@ -134,10 +134,13 @@ namespace litecore { namespace repl {
 
         // Add the token for collection info to the format string
         static inline const char* formatWithCollection(const char* fmt) {
-            const std::string            fmtStr = format("%s %s", kCollectionLogFormat, fmt);
-            std::unique_lock<std::mutex> lock(_formatMutex);
-            const auto                   found = _formatCache.insert(fmtStr);
-            return found.first->data();
+            const std::string fmtStr = format("%s %s", kCollectionLogFormat, fmt);
+            const auto        found  = _formatCache.find(fmtStr);
+            if ( found == _formatCache.end() ) {  // Not found, attempt to insert and return
+                std::unique_lock<std::mutex> lock(_formatMutex);
+                return _formatCache.insert(fmtStr).first->data();
+            }  // Was found, return existing
+            return found->data();
         }
 
         // overrides for Logging functions which insert collection index to the format string
--- a/Replicator/Worker.hh
+++ b/Replicator/Worker.hh
@@ -134,9 +134,10 @@ namespace litecore { namespace repl {
 
         // Add the token for collection info to the format string
         static inline const char* formatWithCollection(const char* fmt) {
-            const std::string fmtStr = format("%s %s", kCollectionLogFormat, fmt);
-            const auto        found  = _formatCache.insert({fmtStr, 0});
-            return found.first->first.data();
+            std::unique_lock<std::mutex> lock(_formatMutex);
+            std::string                  fmtStr = format("%s %s", kCollectionLogFormat, fmt);
+            const auto                   found  = _formatCache.insert(fmtStr);
+            return found.first->data();
         }
 
         // overrides for Logging functions which insert collection index to the format string
@@ -288,12 +289,12 @@ namespace litecore { namespace repl {
         std::string               _loggingID;      // My name in the log
         uint8_t                   _importance{1};  // Higher values log more
       private:
-        Retained<blip::Connection> _connection;               // BLIP connection
-        int                        _pendingResponseCount{0};  // # of responses I'm awaiting
-        Status                     _status{kC4Idle};          // My status
-        bool                       _statusChanged{false};     // Status changed during this event
-        const CollectionIndex      _collectionIndex;
-        static std::unordered_map<std::string, unsigned short> _formatCache;  // Using map as vector causes some weird
-                                                                              // memory issue on Windows
+        Retained<blip::Connection>             _connection;               // BLIP connection
+        int                                    _pendingResponseCount{0};  // # of responses I'm awaiting
+        Status                                 _status{kC4Idle};          // My status
+        bool                                   _statusChanged{false};     // Status changed during this event
+        const CollectionIndex                  _collectionIndex;
+        static std::unordered_set<std::string> _formatCache;  // Store collection format strings for LogEncoders benefit
+        static std::mutex                      _formatMutex;  // Ensure we don't hit a race condition in cache insert
     };
 }}  // namespace litecore::repl
--- a/Replicator/Worker.hh
+++ b/Replicator/Worker.hh
@@ -135,7 +135,7 @@ namespace litecore { namespace repl {
         // Add the token for collection info to the format string
         static inline const char* formatWithCollection(const char* fmt) {
             std::unique_lock<std::mutex> lock(_formatMutex);
-            std::string                  fmtStr = format("%s %s", kCollectionLogFormat, fmt);
+            const std::string            fmtStr = format("%s %s", kCollectionLogFormat, fmt);
             const auto                   found  = _formatCache.insert(fmtStr);
             return found.first->data();
         }
--- a/Replicator/Worker.hh
+++ b/Replicator/Worker.hh
@@ -134,9 +134,9 @@ namespace litecore { namespace repl {
 
         // Add the token for collection info to the format string
         static inline const char* formatWithCollection(const char* fmt) {
-            std::unique_lock<std::mutex> lock(_formatMutex);
             const std::string            fmtStr = format("%s %s", kCollectionLogFormat, fmt);
-            const auto                   found  = _formatCache.insert(fmtStr);
+            std::unique_lock<std::mutex> lock(_formatMutex);
+            const auto                   found = _formatCache.insert(fmtStr);
             return found.first->data();
         }
 
--- a/Replicator/Worker.hh
+++ b/Replicator/Worker.hh
@@ -24,6 +24,7 @@
 #include <atomic>
 #include <functional>
 #include <memory>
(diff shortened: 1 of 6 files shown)


callumbirks:
Undid changing test db name to random. It seemed to have no effect on test failures (Still failing consistently in Jenkins). Only changes in this PR now are; fixing flaky collection logging failure, and adding timeout to ReplicatorLoopbackTest.

callumbirks:
I don't think I've actually done anything to fix the failure - it just seems to have passed on this run. But at least now, it won't run for 2 hours if Replicator is hanging.

callumbirks:
It seems the problem was a race condition in `_formatCache.insert` inside `formatWithCollection()`. If two threads try to insert a value at the same time to a STL container, it's fine. But if two threads try to insert _the same_ value at the same time, it is not. I've changed the data type of the `_formatCache` to a set, as it's more suitable and should be faster, and I've added a mutex to that function to ensure thread safety for insertion.

I've also changed the `formatWithCollection()` back to using `format()` rather than `stringstream` as I did some performance measuring, and at worst format is the same performance as stringstream, at best it is ~5 microseconds faster.

callumbirks:
I think the reason this race condition was hit consistently in ReplicatorLoopbackTest is that it runs two Replicators which often log the same thing at the same time.

pasin (approved):

Replicator/Worker.hh
```diff
@@ -134,9 +134,10 @@ namespace litecore { namespace repl {
 
         // Add the token for collection info to the format string
         static inline const char* formatWithCollection(const char* fmt) {
-            const std::string fmtStr = format("%s %s", kCollectionLogFormat, fmt);
-            const auto        found  = _formatCache.insert({fmtStr, 0});
-            return found.first->first.data();
+            std::unique_lock<std::mutex> lock(_formatMutex);
```
jianminzhao, line 137: It gets more and more expensive now. Consider use read/write lock. I assume most read would be successful.

callumbirks: What do you mean exactly? I thought of doing:
- check for existing string
- if it exists already, return pointer to it
- if it doesn't, lock mutex and insert the string, then return a pointer to it

The issue with this, is if we are hitting the case which is not thread-safe, those two threads are trying to insert the same value at the exact same time. This means, it's pretty likely that the log string might not exist yet, both threads see that it doesn't exist as they're checking at the same time, and both threads will again try to insert at the same time.

I also would not worry about performance. I just did some performance testing, and over an average of 5,000 calls, that function still only takes ~1.5 microseconds - that is including the mutex lock.

jianminzhao: Oh, your preceding 3 bullets smell something fishing: it's not thread safe to read outside lock. The first bullet ought to be, check for existing string inside readLock.
By the way, from 0.9 to 1.5 is a 30% increase.

jianminzhao: Are you saying that STL container may handle one-write and multi-read concurrently ?

callumbirks: Yes. Additionally, a scenario which you may think is dangerous:

thread A wants to log "Finished!"
thread B also wants to log "Finished!"

both threads read concurrently, "Finished!" does not exist. `mutex` is always thread-safe, so only one of them (say A) can grab it. A calls `_formatCache.insert()` and inserts "Finished!" to the cache. Now A releases the mutex, and B grabs it, then calls `_formatCache.insert()`. As `_formatCache.insert()` does a lookup first, it finds that "Finished!" already exists, so it does no insertion and just returns the iterator to the existing element.

callumbirks: From cppreference:

> All const member functions can be called concurrently by different threads on the same container. In addition, the member functions begin(), end(), rbegin(), rend(), front(), back(), data(), find(), lower_bound(), upper_bound(), equal_range(), at(), and, except in associative containers, operator[], behave as const for the purposes of thread safety (that is, they can also be called concurrently by different threads on the same container)

And:

> Different elements in the same container can be modified concurrently by different threads, except for the elements of std::vector<bool> (for example, a vector of [std::future](https://en.cppreference.com/w/cpp/thread/future) objects can be receiving values from multiple threads).

jianminzhao: My understanding is that STL won't touch the elements themselves. So, different elements, provided you already have the references to them, can be modified concurrently. But STL does not protect the wiring from concurrent read/write.
"find" will go through the wiring of the container, "insert", may modify the wiring. My understanding is they cannot be performed concurrently.
But you may do "find" inside a readLock, and only it fails do a "insert" under normal lock.

callumbirks: unordered_set is based on a hash-map. Elements are placed in "buckets" based on the hash of the element. The container is not consecutive or ordered, so insertion of any element does not invalidate held iterators. There is one scenario in which iterators can be invalidated, described here:
> If rehashing occurs due to the insertion, all iterators are invalidated. Otherwise iterators are not affected. References are not invalidated. Rehashing occurs only if the new number of elements is greater than max_load_factor() * bucket_count()

Additionally:
>References and pointers to data stored in the container are only invalidated by erasing that element, even when the corresponding iterator is invalidated.

So the pointers returned from `formatWithCollection()` are never invalidated.

callumbirks: If we are worried about unknown behaviour of the following scenario:
- One thread `find`s the string it's looking for, has the `found` iterator which points to it.
- Before it can return `found->data()`, another thread performs an `insert()` operation which exceeds the available capacity of the `unordered_set`, and a rehash occurs - thus invalidating the reading thread's iterator before it got chance to return.
We can use an `std::shared_mutex` to allow multi-read and single-write locking.

jianminzhao: I am not worried about the pointer returned by "find". What I am concerned is, during the operation of "find", it may read the container's wiring, which "insert" may modify. In this case, it's the hash table itself.

callumbirks: That's what I mean. One thread may have an iterator returned from `find`, which may be invalidated if another thread performs an `insert` that causes a rehash of the container. I am currently modifying the function to use a shared_mutex which allows multiple reads, but blocks reads when a write needs to occur.

jianminzhao: (:- The last 2nd and 3rd msgs are example of concurrent posts, orders got screwed.

Replicator/Worker.hh
```diff
@@ -135,8 +135,12 @@ namespace litecore { namespace repl {
         // Add the token for collection info to the format string
         static inline const char* formatWithCollection(const char* fmt) {
             const std::string fmtStr = format("%s %s", kCollectionLogFormat, fmt);
-            const auto        found  = _formatCache.insert({fmtStr, 0});
-            return found.first->first.data();
+            const auto        found  = _formatCache.find(fmtStr);
```
pasin, line 138: I didn't follow the discussion about the performance but the update is not correct.

This is not thread-safe. There could be one thread inserting and another thread finding. 

This is a double work? I understand that insert() will also check if the item exists before insert.

pasin (changes requested):

Replicator/Worker.hh
callumbirks: It is _technically_ thread-safe, as an insert will not immediately invalidate iterators (ie the one returned from `find()`). `insert` can invalidate existing iterators if it causes a rehash (capacity of the `set` is exceeded by the `insert` operation). I am improving this thread-safety in the commit I am about to push so it will be 100% thread-safe.

And yes, there is not really any way to do it (and be thread-safe) without doing a `find` and an `insert`, but the extra work done is minimal, as `find` is extremely fast in an `unordered_set` (average O(1) complexity).

callumbirks:
Now using a `std::shared_mutex` to allow multiple threads to read at the same time, but block all threads from reading if an insert needs to occur.

Replicator/Worker.hh
```diff
@@ -132,11 +133,19 @@ namespace litecore { namespace repl {
...
             const std::string fmtStr = format("%s %s", kCollectionLogFormat, fmt);
-            const auto        found  = _formatCache.insert({fmtStr, 0});
-            return found.first->first.data();
+            {
+                std::shared_lock sharedLock(_formatMutex);  // Multiple threads can read concurrently
+                const auto       found = _formatCache.find(fmtStr);
+                if ( found != _formatCache.end() ) { return found->data(); }
+            }
+            std::unique_lock lock(_formatMutex);  // Block all other threads if we need to perform an insert
+            return _formatCache.insert(fmtStr).first->data();
```
pasin, line 148: I think we are over-engineering this. Why can't we just lock `_formatCache.insert(fmtStr).first->data();`. 

Note that based on the doc, insert() also check if the item exists or not.

pasin (changes requested):

Replicator/Worker.hh
callumbirks: I assume you mean just basically
```c++
static inline const char* formatWithCollection(const char* fmt) {
        const std::string fmtStr = format("%s %s", kCollectionLogFormat, fmt);
        std::unique_lock lock(_formatMutex);
        const auto found = _formatCache.insert(fmtStr);
        return found.first->data();
    }
```
This method has the largest performance hit, as only a single thread can read or write at once. The implementation of the latest commit allows concurrent reads from any number of threads, until a write needs to occur.

jianminzhao (approved):

pasin (approved):

PR #1734, opened 2023-03-28
CBL-4369: Adjust SG test cases after SG build 576.

jianminzhao:
Prior to SG build 557, we received status 201 when assigning user channels on collections that were not registered with SG. After having updated to build 576, we received status 404. We make necessary changes for tests (testing negative situations) that use unregistered collections.

### DIFF
--- a/Replicator/tests/SG.hh
+++ b/Replicator/tests/SG.hh
@@ -22,6 +22,7 @@
 #include <fleece/Expert.hh>
 #include <fleece/Mutable.hh>
 #include <fleece/slice.hh>
+#include <optional>
 #include <utility>
 #include <vector>
 #include <memory>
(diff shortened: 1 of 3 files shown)


Replicator/tests/SG.hh
```diff
@@ -124,6 +125,7 @@ public:
     c4::ref<C4Cert> identityCert{nullptr};
     c4::ref<C4KeyPair> identityKey{nullptr};
 #endif
+    std::optional<std::vector<C4CollectionSpec>> collectionsAtSG;
```
pasin, line 128: It will be good to have comment here about what is this for.

pasin (approved):

borrrden (approved):
I have no idea what this change is for (I only know that an error code changed in SG 576) but it's only in testing so I won't make a big deal out of it since the changes seem small and limited.

PR #1737, opened 2023-03-31
CBL-4376: (Helium port) formatWithCollection thread safety

callumbirks:
Helium port of CBL-4368.

Make formatWithCollection thread-safe.
Change _formatCache to unordered_set.
Add timeout to ReplicatorLoopbackTest.

### DIFF
--- a/Replicator/Worker.cc
+++ b/Replicator/Worker.cc
@@ -21,6 +21,7 @@
 #include "BLIP.hh"
 #include "HTTPTypes.hh"
 #include <sstream>
+#include <thread>
 
 #if defined(__clang__) && !defined(__ANDROID__)
 #include <cxxabi.h>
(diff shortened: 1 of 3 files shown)


bmeike (changes requested):
This cannot go into 3.1 as is.  There is no Crit/Block bug that requires it.
It can totally go into 3.2 but unless there is a clear reason to put it in 3.1, nope....

jianminzhao:
We need this to make Core thread-safe. We observed some occurrences of racing.

callumbirks:
@bmeike The issue that's solved with this change is: There's potential of undefined behaviour if two threads try to log the same thing at roughly the same time. This was the bug causing ReplicatorLoopbackTest to fail consistently. It would probably be less likely to be hit in production, but it's still an issue.

jianminzhao (approved):

borrrden:
It's more of a procedural thing.  The port needs its own ticket first that can be approved to go into 3.1.

borrrden (changes requested):
The code is fine but the title, commit, etc needs to be changed to a Helium specific ticket.

jianminzhao:
This is a Helium specific ticket. The Beryllium ticket is 4368

borrrden (approved):
Be sure to not put the beryllium ticket number as-is into the commit message.  Also add the helium ticket number to the commit title when squashing.

pasin (approved):

bmeike (approved):
This patch fixes a bug introduced after RC1 in PR https://github.com/couchbase/couchbase-lite-core/pull/1726

PR #1739, opened 2023-04-04
CBL-4377 : Not moving deleted docs when open database prior 3.1

pasin:
**Problem:**

When opening the database prior 3.1, there will be two kv-tables (live and deleted table) created for each collection. The deleted docs will be moved from the live table to the deleted table. However, moving the deleted docs are expensive operation and could take several seconds or mins depending on the database and the platform.

**Changes:**

* Not moving the deleted docs at all for the default collection (Note: we do not support collections prior 3.1 so this change only affects the default collection).

* As BothKeyStore is used in 3.1 by default and it has already handled the case when the deleted doc are in both live and deleted table, so no code change needed at the KeyStore level.

* Updated Query parser to use all_default table to query the default collection when the deleted meta property is accessed.

* Updated Query parser to add (flags & 1 = 0) check when querying kv_default for the default collection (with no accessing deleted meta property).

* Updated Query parser tests that queries against the default collection accordingly. Note: for improvement, we may add more tests to cover the same scenarios but using the non-default collections.

**Pending Failed Tests:**

* The "Database Upgrade From 2.7 to Version Vectors" test failed after the change. This is failed because the upgrader cannot move the deleted rev using `updateSequence` == false from live table to the deleted table with the assertion failure [here](https://github.com/couchbase/couchbase-lite-core/blob/master/LiteCore/Storage/SQLiteKeyStore.cc#L338). The assumption seems to be that the deleted has already moved to the deleted table fore the upgrade. Maybe somehow don't use `BothKeyStore` to do the Version Vector upgrade for the default collection.

**Fixed Flaky Tests:**

* Add more delay to "Multiple Collections Incremental Revisions" test to reduce the flakiness of the test as the replicator may become IDLE while some docs are not replicated yet due to the delay of the changed added to the collections. Noted the flaky is a test issue, not related to the change.

### DIFF
--- a/C/tests/c4DatabaseTest.cc
+++ b/C/tests/c4DatabaseTest.cc
@@ -1075,7 +1075,9 @@ TEST_CASE("Database Upgrade From 2.7", "[Database][Upgrade][C]") {
 }
 
 
-TEST_CASE("Database Upgrade From 2.7 to Version Vectors", "[Database][Upgrade][C]") {
+TEST_CASE("Database Upgrade From 2.7 to Version Vectors", "[Database][Upgrade][C][.failed]") {
+    auto dbDomain = c4log_getDomain("DB", false);
+    c4log_setLevel(dbDomain, kC4LogDebug);
     testOpeningOlderDBFixture("upgrade_2.7.cblite2", kC4DB_VersionVectors);
 }
 
--- a/LiteCore/Query/QueryParser.cc
+++ b/LiteCore/Query/QueryParser.cc
@@ -354,9 +354,18 @@ namespace litecore {
 
 
     void QueryParser::writeWhereClause(const Value *where) {
+        _checkedDeleted = false;
         if (where) {
             _sql << " WHERE ";
             parseNode(where);
+        }
+        
+        // CBL-4377:
+        // If the query doesn't access the deleted meta property, add (flags & 1 = 0) test to
+        // the where clause when querying against the default collection.
+        if (!_checkedDeleted) {
+            bool needAnd = (where != NULL);
+            writeNotDeletionTestWhenQueryDefaultCollection(_dbAlias, needAnd, !needAnd, false);
         }
     }
 
@@ -578,10 +587,20 @@ namespace litecore {
                         _sql << " " << kJoinTypeNames[ joinType ] << " JOIN "
                              << sqlIdentifier(entry.tableName)
                              << " AS " << sqlIdentifier(entry.alias);
+                        
+                        _checkedDeleted = false;
                         if (entry.on) {
                             _sql << " ON (";
                             parseNode(entry.on);
                             _sql << ")";
+                        }
+                        
+                        // CBL-4377:
+                        // If the join condition doesn't access the deleted meta property, add (flags & 1 = 0) test to
+                        // to the join condition when the joined collection is the default collection.
+                        if (!_checkedDeleted) {
+                            bool needAnd = (entry.on != NULL);
+                            writeNotDeletionTestWhenQueryDefaultCollection(entry.alias, needAnd, false, !needAnd);
                         }
                         break;
                     }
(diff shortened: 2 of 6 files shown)


snej:
> Not moving the deleted docs at all for the default collection (Note: we do not support collections prior 3.1 so this change only affects the default collection).

How about checking the size of the db first, and moving the deleted docs if the doc count is less than some threshold? Since it's only for very large databases that the move is problematic.

borrrden:
Are there any gains to be had from converting small databases?  The logic to handle both tables will continue to be there and we want to develop a system later to asynchronously move documents into the new table (and probably bump the db user_version when finished) so I am interested in the reasoning to do it immediately for small DBs.

jianminzhao:
Here are my understanding of the issue.
- For each collection, there are two tables, kv_collection, kv_del_collection,
- After the current upgrading step, the db is completely separated into live and deleted docs.
- With this PR, we only separate the named collections.
- We may conceptually think the default collection is partially separated: the live store contains all live documents plus some deleted docs, and the del store contains only deleted  docs. With this model, we may gradually move deleted docs into del store without violating the invariant. When all deleted docs are in the deleted store, then we can operate under invariant that all collections have completely separated live and deleted stores.
- With this model, we can make changes closer to that state before the commit of the feature. I guess you also referred to state before that commit for with _checkedDeleted.
- One caveat is after the current upgrade, the db will be marked as new version. We cannot say those collections are upgraded. If were able to mark upgraded version per collection, then we could gradually update the default collections. But this can be dealt with later.

snej:
> Are there any gains to be had from converting small databases? 

I believe there are gains for converting large-but-not-huge databases. AFAIK the only situation where the conversion was problematic was with an extremely large multi-gigabyte db, right?

borrrden:
Yes, that's the current information we have.  It's also the only scenario aside from small testing data sets that we have available currently.

jianminzhao:
If we can have a intermediate version between 3.0 and 3.1, say 3.0.5, then we can move the deleted docs gradually.
For v3.0.5, kv_default has all the live docs plus part of deleted docs, and kv_del_default contains the rest of the deleted docs. Of course, our code must work correctly in 3.0.5 state. With this, every time we open the db, we can run that query with LIMIT 1000, for example. This is what I aspire to get to.
With the current code, the parser parses the query and able to determine the 3 cases, for each collection,
1. Query only involves live docs. Hence the data source will be kv_default
2. Query only involves the deleted docs. Hence, the data source will be kv_del_default
3. Cannot determine at the parse time, (1) or (2) above. So, the data source will be all_default
all_default is a view, may not be optimal for queries, but hopefully it should represent small portion in practice.
For v3.0.5, if the code determines case 1 or 3, we will use the corresponding table, but, because kv_default may contain deleted docs, we have to use flag&1 to filter out the selection. (The beauty is, the more deleted docs are moved out of kv_default, the smaller this table would be.) Same holds for case (3). It it is determined all deleted docs, we have to turn it to case (3) because some deleted docs are in kv_default, and some are in kv_del_default. 
I am going to send out a PR based on this model.

C/tests/c4DatabaseTest.cc
```diff
@@ -1075,7 +1075,9 @@ TEST_CASE("Database Upgrade From 2.7", "[Database][Upgrade][C]") {
 }
 
 
-TEST_CASE("Database Upgrade From 2.7 to Version Vectors", "[Database][Upgrade][C]") {
+TEST_CASE("Database Upgrade From 2.7 to Version Vectors", "[Database][Upgrade][C][.failed]") {
+    auto dbDomain = c4log_getDomain("DB", false);
+    c4log_setLevel(dbDomain, kC4LogDebug);
```
borrrden, line 1080: Can we remove this?  It will cause log spam in normal cases.

borrrden (changes requested):
I'd like to remove the debug log level change to avoid large log spam in CI

C/tests/c4DatabaseTest.cc
pasin: Oh, I forgot to remove that.

callumbirks:
Do we send deleted docs at all to SG? If so, would this be a concern at all?

PR #1740, opened 2023-04-04
CBL-4379: CertHelper leaves persistent certs in the keychain

callumbirks:
Potentially a fix for CBL-2985, but can't be certain.


borrrden (commented):
What about client persistent identity?

callumbirks:
> What about client persistent identity?

@borrrden Oh yes, I think I forgot because it isn't actually used. Will make that change now.

borrrden (approved):

jianminzhao (approved):

PR #1742, opened 2023-04-07
CBL-4398 : Fix URL scheme in HTTP message when using proxy

pasin:
When using proxy, the request target in the start line of the HTTP message is the full URL instead of just the URL path and the URL scheme should be http / https instead of ws / wss.

### DIFF
--- a/Networking/HTTP/HTTPLogic.cc
+++ b/Networking/HTTP/HTTPLogic.cc
@@ -109,9 +109,15 @@ namespace litecore { namespace net {
             rq << "CONNECT " << string(slice(_address.hostname)) << ":" << _address.port;
         } else {
             rq << MethodName(_method) << " ";
-            if (_proxy && _proxy->type == ProxyType::HTTP)
-                rq << string(_address.url());
-            else
+            if (_proxy && _proxy->type == ProxyType::HTTP) {
+                if (_isWebSocket) {
+                    Address address = _address;
+                    address.scheme = address.scheme == "wss"_sl ? "https"_sl : "http"_sl;
+                    rq << string(Address::toURL(*(C4Address*)&address));
+                } else {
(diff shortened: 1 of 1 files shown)


bmeike (approved):

jianminzhao (approved):

Networking/HTTP/HTTPLogic.cc
```diff
@@ -109,9 +109,15 @@ namespace litecore { namespace net {
             rq << "CONNECT " << string(slice(_address.hostname)) << ":" << _address.port;
         } else {
             rq << MethodName(_method) << " ";
-            if (_proxy && _proxy->type == ProxyType::HTTP)
-                rq << string(_address.url());
-            else
+            if (_proxy && _proxy->type == ProxyType::HTTP) {
```
borrrden, line 112: What happens in the case of `ProxyType::HTTPS`?

pasin: I don't know that it has HTTPS in the enum as well. The original code doesn't handle that so that seems to be a bug here. It should handle the same way.

pasin: Will fix this.

borrrden: My guess is that it is meant to connect to a proxy that is serving SSL?  So the handshake needs to happen?  In which case it is not relevant here and we probably don't support it?

pasin: Yes, the handshake should happen with the proxy the same way when no-proxy is used, but regarding the HTTP message, it should be the same when using HTTP and HTTPS. 

Note: I have never verified with HTTPS proxy so whether there could be a handshake issue or not, it's a separate issue.

borrrden (approved):

PR #1743, opened 2023-04-07
Add more scenario checks in testOpeningOlderDBFixture()

pasin:
Add more scenarios checks in testOpeningOlderDBFixture() including :

* Check document counts
* Enumerate only non-deleted doc
* Enumerate in descending order (This case is currently failed so the test code is commented out).
* Update and delete docs


pasin:
The PR validation failed on Windows due to timeout. I think the process was terminated.

borrrden (commented):
Not approving this because I think it should be a part of the PR for CBL-4377

PR #1747, opened 2023-04-12
CBL-4407: "Persistent cert chain" test flaky failure

callumbirks:
Avoid the flaky failure of "Cert already exists" by adding randomness to the cert parameters.

### DIFF
--- a/Crypto/CertificateTest.cc
+++ b/Crypto/CertificateTest.cc
@@ -289,10 +289,20 @@ TEST_CASE("Persistent save duplicate cert or id", "[Certs]") {
     key2->remove();
 }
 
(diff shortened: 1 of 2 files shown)


pasin (approved):

pasin:
@callumbirks it's likely that there are other tests that may create the cert with the same serial number and issuer. Maybe we have a static function that could return a serial number with a running number that all cert tests can use?

jianminzhao (approved):

callumbirks:
@pasin You're right... I will hold off on merging for now. Tomorrow I will look into how I can best do that, so we have the same system for all cert tests.

pasin (approved):

PR #1750, opened 2023-04-14
CBL-4351: Generated file n1ql.cc violates C++ reserved identifiers

callumbirks:
PR is merged on `snej/peg` which makes sure tool doesn't generate code with '__'
Updated n1ql.leg to remove '__' and make compatible with tool changes.
Re-generated n1ql.cc using updated tool and n1ql.leg.

I have run the N1QL Parser tests to verify, and all is good.
There is one notable change in the way to use n1ql.leg, which is the scenarios where you would use `_`, you should instead use `y_`.

### DIFF
--- a/LiteCore/Query/N1QL_Parser/n1ql.leg
+++ b/LiteCore/Query/N1QL_Parser/n1ql.leg
@@ -22,8 +22,8 @@ namespace litecore { namespace n1ql {
 %}
 
 
-n1ql = y_
-    s:selectStatement y_ !.            { $$ = s;}
+n1ql = _
+    s:selectStatement _ !.            { $$ = s;}
     |
     sr:selectResults !.               { MutableDict dict = MutableDict::newDict();
                                         setAny(dict, "WHAT", sr);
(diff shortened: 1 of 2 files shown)


jianminzhao:
Defer to @snej since he pulled in the changes in peg. I guess I need to rebuild peg.

LiteCore/Query/N1QL_Parser/n1ql.leg
```diff
@@ -22,8 +22,8 @@ namespace litecore { namespace n1ql {
 %}
 
 
-n1ql = _
-    s:selectStatement _ !.            { $$ = s;}
+n1ql = y_
```
snej, line 25: Why did the grammar syntax change? I thought the changes to the tool only affected the generated C code. (I guess I must have missed this when reviewing the `peg` PR, sorry)

I don't want to change the grammar syntax if we can help it, since this is a breaking change for anything using `peg`, not just us.

snej (changes requested):

LiteCore/Query/N1QL_Parser/n1ql.leg
callumbirks: Because `peg` generates function names dependent on the name of the function in the grammar. I believe it's `yy_` + [function name]. So `_` generates a function called `yy__`. I guess if we don't want the grammar of n1ql.leg to change, we could adjust the generation in `peg` to not be `yy_` + [function name].

callumbirks:
Merged new change on `snej/peg`, now we can avoid '__' without having to change grammar.

snej (approved):

PR #1751, opened 2023-04-14
CBL-4430: Update clang-format-check to clang-format-16

callumbirks:
And some formatting changes between clang-format v15 and v16.

### DIFF
--- a/.github/workflows/clang-format-check.yml
+++ b/.github/workflows/clang-format-check.yml
@@ -38,6 +38,6 @@ jobs:
       - name: Lint Code Base
         uses: jidicula/clang-format-action@v4.11.0
         with:
-          clang-format-version: '15'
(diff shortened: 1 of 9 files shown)


callumbirks:
After chatting to someone in the MCR office, he made the good point that if we have a huge, auto-generated file (`n1ql.cc`), it doesn't really need to be formatted. I'd tend to agree with that, so I've ignored `n1ql.cc` from both the `run-clang-format.sh` script and the Actions Linter. This will avoid 10k line changes, which would be unnecessary anyway as the file is still unreadable.

pasin (approved):
I totally agreed that the generated n1ql.cc should be ignored especially if we are using an external tool to generate as we couldn't control the format of the generated file.

borrrden (approved):

PR #1753, opened 2023-04-18
CBL-4223: Analysis Warnings (3/13) LiteCore/Query

callumbirks:
No notable changes. Mostly just optimizing imports, adhering to language standards, marking unused parameters, etc.


LiteCore/Query/SQLiteFleeceEach.cc
```diff
@@ -176,7 +177,7 @@ namespace litecore {
 
 #pragma mark - INSTANCE METHODS:
 
-        FleeceCursor(FleeceVTab* vtab) : _vtab(vtab) {}
+        explicit FleeceCursor(FleeceVTab* vtab) : sqlite3_vtab_cursor{vtab}, _vtab(vtab) {}
```
borrrden, line 180: What is the righthand side of this change doing?

LiteCore/Query/SQLiteFleeceUtil.cc
```diff
@@ -150,7 +156,9 @@ namespace litecore {
                 case kNumber:
                     if ( val->isInteger() )
                         if ( val->isUnsigned() ) {
-                            sqlite3_result_int64(ctx, val->asUnsigned());
+                            // val->asUnsigned() does cause some overflow error with narrow_cast, so we can't use it
```
borrrden, line 159: Actually in release builds it does not, but this is fine as well

borrrden (approved):

jianminzhao (approved):

LiteCore/Query/SQLiteFleeceEach.cc
callumbirks: Initializes the base class `sqlite3_vtab_cursor`. 
There is another constructor for the object:
`explicit FleeceCursor(FleeceVTab* vtab) : sqlite3_vtab_cursor{vtab}, _vtab(vtab) {}`
Which initializes both the base class, and `_vtab`.
Interestingly, `_vtab` has the same data type as the base class, so it's possible we only need one of them. Going to investigate that now.

callumbirks: Hmm yeah, removing the inheritance from `sqlite3_vtab_cursor`, although it causes no compile errors, causes fatal errors during QueryTest. So I guess some sqlite3 functions that takes pointers etc, rely on the passed object inheriting from sqlite3_vtab_cursor.

LiteCore/Query/SQLiteFleeceUtil.cc
callumbirks: 😮

PR #1754, opened 2023-04-19
CBL-4418: Replicator is stuck in busy state when there is an error th…

jianminzhao:
…rown while applying delta to create full fleece doc

This PR does not fix the delta error. (The reason of the delta errors as shown in the ticket is still unknown.) Instead, I try to fix the problem of "stuck in busy state." In fact, our code already has paths to deal with the possible delta error. After injecting delta errors for push and pull, bugs emerged as the ticket describes.

After the fix, the delta error will trigger the second round of "rev" message which uses full rev as opposed to the delta rev.

### DIFF
--- a/Replicator/Pusher+Revs.cc
+++ b/Replicator/Pusher+Revs.cc
@@ -240,7 +240,12 @@ namespace litecore::repl {
                 }
                 doneWithRev(rev, completed, synced);
                 switch (retry) {
-                    case kRetryNow:   retryRevs({rev}, true); break;
+                    case kRetryNow:
+                        rev->error.code = 0;
+                        rev->isWarning = 0;
+                        rev->errorIsTransient = false;
+                        retryRevs({rev}, true);
+                        break;
                     case kRetryLater: _revsToRetry.push_back(rev); break;
                     case kNoRetry:    break;
                 }
--- a/Replicator/Pusher.cc
+++ b/Replicator/Pusher.cc
@@ -627,7 +627,6 @@ namespace litecore { namespace repl {
         // immediate means I want to resend as soon as possible, bypassing another changes feed entry
         // (for example in the case of a failed delta merge)
         logInfo("%d documents failed to push and will be retried now", int(revsToRetry.size()));
-        _caughtUp = false;
         if (immediate) {
             for (const auto& revToRetry : revsToRetry) {
                 _pushingDocs.insert({revToRetry->docID, revToRetry});
@@ -635,6 +634,7 @@ namespace litecore { namespace repl {
             }
             _revQueue.insert(_revQueue.begin(), revsToRetry.begin(), revsToRetry.end());
         } else {
+            _caughtUp = false;
             ChangesFeed::Changes changes = {};
             changes.revs = move(revsToRetry);
             changes.lastSequence = _lastSequenceRead;
--- a/Replicator/tests/ReplicatorWalrusTest.cc
+++ b/Replicator/tests/ReplicatorWalrusTest.cc
@@ -675,6 +675,166 @@ TEST_CASE_METHOD(ReplicatorWalrusTest, "Pull iTunes deltas from SG", "[.SyncServ
           timeWithDelta, timeWithoutDelta, timeWithoutDelta/timeWithDelta);
 }
 
+
+TEST_CASE_METHOD(ReplicatorWalrusTest, "Pull deltas with filter from SG", "[.SyncServerWalrus][Delta]") {
+    static constexpr int kNumDocs = 10, kNumProps = 100;
+    flushScratchDatabase();
+    _logRemoteRequests = false;
+
+    // -------- Populating local db --------
+    auto populateDB = [&]() {
+        TransactionHelper t(db);
+        std::srand(123456); // start random() sequence at a known place
+        for (int docNo = 0; docNo < kNumDocs; ++docNo) {
+            char docID[kDocBufSize];
+            snprintf(docID, kDocBufSize, "doc-%03d", docNo);
+            Encoder enc(c4db_createFleeceEncoder(db));
+            enc.beginDict();
+            for (int p = 0; p < kNumProps; ++p) {
+                enc.writeKey(format("field%03d", p));
+                enc.writeInt(std::rand());
+            }
+            enc.endDict();
+            alloc_slice body = enc.finish();
+            string revID = createNewRev(db, slice(docID), body);
+        }
+    };
+    populateDB();
+
+    // -------- Pushing to SG --------"
+    replicate(kC4OneShot, kC4Disabled);
+
+    // -------- Updating docs on SG --------
+    {
+        JSONEncoder enc;
+        enc.beginDict();
+        enc.writeKey("docs"_sl);
+        enc.beginArray();
+        for (int docNo = 0; docNo < kNumDocs; ++docNo) {
+            char docID[kDocBufSize];
+            snprintf(docID, kDocBufSize, "doc-%03d", docNo);
+            C4Error error;
+            c4::ref<C4Document> doc = c4doc_get(db, slice(docID), false, ERROR_INFO(error));
+            REQUIRE(doc);
+            Dict props = c4doc_getProperties(doc);
+
+            enc.beginDict();
+            enc.writeKey("_id"_sl);
+            enc.writeString(docID);
+            enc.writeKey("_rev"_sl);
+            enc.writeString(doc->revID);
+            for (Dict::iterator i(props); i; ++i) {
+                enc.writeKey(i.keyString());
+                auto value = i.value().asInt();
+                if (RandomNumber() % 2 == 0)
+                    value = RandomNumber();
+                enc.writeInt(value);
+            }
+            enc.endDict();
+        }
+        enc.endArray();
+        enc.endDict();
+        _sg.sendRemoteRequest("POST", "_bulk_docs", enc.finish(), false, HTTPStatus::Created);
+    }
+
+    // -------- Repopulating local db --------
+    deleteAndRecreateDB();
+    populateDB();
+
+    // Setup pull filter:
+    _pullFilter = [](C4String collectionName, C4String docID, C4String revID,
+                     C4RevisionFlags flags, FLDict flbody, void *context) {
+        return true;
+    };
+
+    // -------- Pulling changes from SG --------
+    replicate(kC4Disabled, kC4OneShot);
+
+    // Verify
+    int n = 0;
+    C4Error error;
+    c4::ref<C4DocEnumerator> e = c4db_enumerateAllDocs(db, nullptr, ERROR_INFO(error));
+    REQUIRE(e);
+    while (c4enum_next(e, ERROR_INFO(error))) {
+        CHECK(error.code == 0);
+        C4DocumentInfo info;
+        c4enum_getDocumentInfo(e, &info);
+        CHECK(slice(info.docID).hasPrefix("doc-"_sl));
+        CHECK(slice(info.revID).hasPrefix("2-"_sl));
+        ++n;
+    }
+    CHECK(n == kNumDocs);
+}
+
+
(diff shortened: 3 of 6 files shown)


Replicator/Pusher+Revs.cc
```diff
@@ -240,7 +240,12 @@ namespace litecore::repl {
                 }
                 doneWithRev(rev, completed, synced);
                 switch (retry) {
-                    case kRetryNow:   retryRevs({rev}, true); break;
+                    case kRetryNow:
+                        rev->error.code = 0;
+                        rev->isWarning = 0;
+                        rev->errorIsTransient = false;
+                        retryRevs({rev}, true);
+                        break;
```
pasin, line 248: This issue is not described in the CBSE ticket (Only pull case). Can you provide a bit info about the behavior before the change?

Replicator/Pusher.cc
```diff
@@ -627,14 +627,14 @@ namespace litecore { namespace repl {
...
         logInfo("%d documents failed to push and will be retried now", int(revsToRetry.size()));
-        _caughtUp = false;
         if (immediate) {
             for (const auto& revToRetry : revsToRetry) {
                 _pushingDocs.insert({revToRetry->docID, revToRetry});
                 addProgress({0, revToRetry->bodySize});
             }
             _revQueue.insert(_revQueue.begin(), revsToRetry.begin(), revsToRetry.end());
         } else {
+            _caughtUp = false;
```
pasin, line 637: Can you describe about this fix as well?

Replicator/Pusher+Revs.cc
jianminzhao: Without this change, because we use the same pointer to the previous rev as we retry, if don't clear the error, after successful retry, the error is still stuck, and we won't get the completed doc count correct. There is a minor issue here -- I meant to point out after the build succeeds and invite reviews -- that by clearing the error, before it goes to the docEnded callback, the error of the first try will be cleared as well. 
Ideally, we probably should make a clone of the rev for the retry, or make a clone to send to docEnded, but this will defeat our super-efficient way passing DocumentEnded to the callback.
I feel lose an error to the callback is not big thing, because it's pending retry. If retry fails again, it will be failed doc to the docEnded callback.

Replicator/Pusher.cc
jianminzhao: In immediate case, we take them directly to _revQueue, not going through the changes feed, and _caughtUp describes whether the changes feed has all ben taken out. That is, immediate case does not affect _caughtUp.

jianminzhao: This is the cause of hanging Push. It does not contributing to the CBSE case, which is Pull only.

Replicator/Pusher+Revs.cc
pasin: > I feel lose an error to the callback is not big thing, because it's pending retry. If retry fails again, it will be failed doc to the docEnded callback.

Not sure I understand the behavior correctly. So the first docEnd of the failure before retry with be shown as success or the docEnd will not be sent before the retry result?

Replicator/Pusher.cc
pasin: Got it. Thanks!

Replicator/Pusher+Revs.cc
callumbirks: It sounds to me like: if first attempt is fail, it will be retried. Because we now clear error on retry, the error from first fail will not be shown. So if the second attempt succeeds, we will never know the error for the first attempt. If the second attempt fails, we will get that error.

callumbirks:
I assume, the tests you've added, will hang without your change? And your change fixes this hang? If so, LGTM.

Replicator/Pusher+Revs.cc
bmeike: Is there a CBL ticket for this change?  If not. please create one before committing this

Replicator/Pusher.cc
bmeike: If there is not a CBL ticket for this change, please create one before this is committed

Replicator/tests/data/cbl-4418-inject-error-pull.patch
```diff
@@ -0,0 +1,25 @@
+diff --git a/Replicator/DBAccess.cc b/Replicator/DBAccess.cc
+index 9efa7493f..46d53482f 100644
+--- a/Replicator/DBAccess.cc
++++ b/Replicator/DBAccess.cc
+@@ -364,6 +364,12 @@ namespace litecore { namespace repl {
+         
+         Doc result;
+         FLError flErr;
++
++        if (string(doc->docID()) == "doc-001") {
```
bmeike, line 10: What the heck is this?

Replicator/tests/data/cbl-4418-inject-error-push.patch
```diff
@@ -0,0 +1,17 @@
+diff --git a/Replicator/Pusher+Revs.cc b/Replicator/Pusher+Revs.cc
+index f27262189..b7f598cc8 100644
+--- a/Replicator/Pusher+Revs.cc
++++ b/Replicator/Pusher+Revs.cc
+@@ -336,6 +336,12 @@ namespace litecore::repl {
+                        nuu.size - delta.size,
+                        SPLAT(old), SPLAT(nuu), SPLAT(delta));
+         }
++        if (string(doc->docID()) == "doc-001") {
```
bmeike, line 9: ???

bmeike (changes requested):
There are changes in this commit that do not appear to be motivated by any CBL ticket.
1) you can't do that.
2) please, in the future, submit fixes to release branch one PR per CBL ticket
3) What is the check for docID "doc-001" for?

Replicator/Pusher+Revs.cc
jianminzhao: That's part of the thing I need to adjust to fix the problem in the ticket. Otherwise, the bug will show in the other direction, namely the Push.

jianminzhao: As to @pasin 's question, @callumbirks 's understanding is correct. Basically this what happens
1. First attempt of "rev" with delta  received response from SG that shows error.
2. We push the ReplicatedRev object to docEnded queue.
3. Meanwhile, afterwards, we decided to retry without delta, using the same ReplicatedRev object, which has the error fields assigned.
4. As is done in the PR, I cleared the errors so that second attempt starts afresh.
5. Asynchronously, when we get to report status, we pop the entries from docEnded. At this time, the ReplicatedRev object has clear error fields due to (4).
If the info about the error of the first attempt is important, I probably can do something. For example, not clear the error until the response of the second attempt arrives.

Replicator/Pusher.cc
jianminzhao: This fix only affects Push side. The original CBSE is related to Pull. That is, it won't affect CBSE per se. However, without the fix, our localReplicator test will fail, because in this case, CBL plays both sides of Push and Pull.

Replicator/tests/data/cbl-4418-inject-error-pull.patch
jianminzhao: This file won't go to the source. It is a patch file. It can help to inject artificial delta error to test the code added by this PR. It's explained in the relevant test.

Replicator/tests/data/cbl-4418-inject-error-push.patch
jianminzhao: Same as above.

Replicator/Pusher+Revs.cc
pasin: From the API point of view, the document ended shouldn't report as success if it doesn't. Even better, if it doesn't report until retry (This is probably out of scope of this PR).

jianminzhao:
@callumbirks the tests I added will pass because it's not easy to have delta error. It also means they won't test the new code. The two patch files can help to inject errors. After applying the patches, the tests will fail without the new code in the PR. A year from now, if CBSE comes with similar problem, we could apply the patches and run tests, hoping to find where the regression occurs.

jianminzhao:
If the patch files are only confusing, I can take them out. The tests as they stand now, they will be meaningful there were delta errors on that that purposefully simple document body. If that should happen, then they would hide the error. Quite dilemma.

jianminzhao:
After talking with @pasin, I am going to do two things: 
1. remove the patch files from test/data to the CBL ticket.
2. don't send ReplicateRev to docEnded if the rev is going to be retried immediately.

Replicator/tests/ReplicatorWalrusTest.cc
```diff
@@ -675,6 +675,179 @@ TEST_CASE_METHOD(ReplicatorWalrusTest, "Pull iTunes deltas from SG", "[.SyncServ
...
+    // Setup pull filter:
+    _pullFilter = [](C4String collectionName, C4String docID, C4String revID,
+                     C4RevisionFlags flags, FLDict flbody, void *context) {
+        return true;
+    };
+
+    // -------- Pulling changes from SG --------
+    // CBL-4418: this test is interesting only if the puller finds an error of
+    // "Invalid delta". This error is not common. In the attachments of the ticket,
+    // a patch file, "cbl-4418-inject-error-pull.patch," can be found.
```
borrrden, line 753: Can it?  Wasn't this removed?  If so, is this test providing anything?

jianminzhao: I have't put them there yet. I thought to put them there after the PR is approved. Just put them there. Can always remove them if it's not a good idea.
Yes, this is a dilemma. I cannot find a way to generate the error in the test itself. With patch files, we can at least reproduce the current state sometime in the future.

borrrden: I don't much like adding tests that do nothing.  If they rely on the patch, then they are not useful in their current state and should be added along with the patch if and when it is added.

jianminzhao: So, you are suggesting to put the test cases and the code that inject the errors together into a patch and save it in the CBL ticket?

borrrden: I'm not sure that is a good long term solution, since what happens when the patch becomes unable to be applied?  I was thinking that if you had a patch you could add the test cases later and commit the patch file along side of it.  That also has the same problem though.  Could you create a special compile mode for this?  That way it won't suffer from patch rot and you can conditionally compile in the tests using the same define.

jianminzhao: Store the (combined) patch file in the Ticket has an advantage that we can alway replay the test, since the ticket also has the commit hash. To replay in the future, simply rollback the commit and apply the patch.
What do you think?

jianminzhao:
I removed the tests from the commit and made self-contained patch, cbl-4418-test.patch, that includes the instrumental code to inject the delta errors and the tests that verify that errors are handled properly.
The patch is attached to the ticket, CBL-4418, for your review.

bmeike (changes requested):

jianminzhao:
I made a new ticket, CBL-4448 to cover both pusher and puller case. CBL-4418 would resolved as a duplicate.

bmeike:
I don't, at all, understand why you created a new ticket... but that works.

bmeike (approved):

borrrden:
What happens when the patch becomes unable to be applied?  Are we going to be forced to roll back to this commit every time?  At that point it is no different than a branch and it seems like a questionable thing to do.  I still like the idea of `ifdef` to enable and disable this when needed.

jianminzhao:
This is my thoughts. If the patch is too old to apply, we can create a branch off the commit in the ticket, applying the patch, and do a manual cherry-pick. As you pointed out, it's like a branch but, with the patch, we can generate on-demand. At least we will have the factual context when the problem is taken care of.
Putting #ifdef in the main source basically will force us to pay attention to the #if'ed part whenever we will rearrange the code that would make the patch invalid. Of course it has advantage that the test can be run on-demand anytime, but realistically, will we ever care to run the test unprovokedly?
Of course, just my 2-cent thoughts.

PR #1756, opened 2023-04-21
Fix some weird formatting in public headers

borrrden:
.NET uses a script to parse the headers and generate its binding functions programtically.  Stuff like having random things on the same line, or odd formatting, makes that a lot harder.  These changes fix some of those locations.

### DIFF
--- a/C/include/c4Certificate.h
+++ b/C/include/c4Certificate.h
@@ -271,7 +271,7 @@ CBL_CORE_API bool c4keypair_removePersistent(C4KeyPair*, C4Error* C4NULLABLE out
     @param outError  On failure, the error info will be stored here.
     @return  The key object, or NULL on failure. */
 CBL_CORE_API C4KeyPair* c4keypair_fromExternal(C4KeyPairAlgorithm algorithm, size_t keySizeInBits, void* externalKey,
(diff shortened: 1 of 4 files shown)


callumbirks:
If you want to make sure that the formatter never touches those lines, you can place `// clang-format off` and `// clang-format on` around them. Give me a min and I'll figure out how to get the Linter to ignore them too.

callumbirks:
Okay nothing easy I can see on https://github.com/jidicula/clang-format-action. But come to think of it, I did already use `// clang-format off` in some places, and I don't think it's complained about that. So those sections are probably ignored for linting too.

borrrden:
The clang format off and on didn't seem to help @callumbirks .  Do you have any idea why?

callumbirks:
From running Clang-Format locally, it seems like it wants to format your `clang-format` comments 😆 
I'm gonna guess that's because they're both within the structs. I would suggest either going with the formatted comments, or applying your own formatting to the whole of each struct, and wrapping the whole structs in the `clang-format` comments.

jianminzhao (approved):

callumbirks (approved):

PR #1736, opened 2023-03-30
CBL-4375: Opening the upgraded database from 2.8 to 3.1 is slow

jianminzhao:
Compared with 2.8, we added a process of startHousekeeping to keep track of expired documents. To do it, we need to find out the closest exipration time, by SQL query, SELECT MIN(expiration) from kv_default". It turned out it's quite expensive as most docs do not have expiration field (= SQLite NULL). By passing this fact to the query, we got return nearly instantly.

Ported from release/lithium, CBL 4325

PR #1757, opened 2023-04-21
CBL-4326: Opening the upgraded database from 2.8 to 3.1 is slow (#1736)

jianminzhao:
Compared with 2.8, we added a process of startHousekeeping to keep track of expired documents. To do it, we need to find out the closest exipration time, by SQL query, SELECT MIN(expiration) from kv_default". It turned out it's quite expensive as most docs do not have expiration field (= SQLite NULL). By passing this fact to the query, we got return nearly instantly.

Ported from release/lithium, CBL 4375


pasin (approved):

pasin (approved):

borrrden (approved):
Not a huge fan of leaving the link to the PR for the other branch in the PR message....

PR #1758, opened 2023-04-21
CBL-4100: Add replicator tests to test with SG 3.0

callumbirks:
Add tests for specifically testing SG3.0.
Add an SG version check to SG tests, to avoid testing against the wrong version. 
Improve sg_setup script.
Add SG3.0-specific setup for Collection test utils.

### DIFF
--- a/Replicator/tests/ReplicatorCollectionSGTest.hh
+++ b/Replicator/tests/ReplicatorCollectionSGTest.hh
@@ -31,21 +31,11 @@ static constexpr C4CollectionSpec Tulips    = {TulipsName, FlowersScopeName};
 static constexpr C4CollectionSpec Lavenders = {LavenderName, FlowersScopeName};
 static constexpr C4CollectionSpec Default   = kC4DefaultCollectionSpec;
 
-#ifdef COUCHBASE_ENTERPRISE
-static C4SliceResult propEncryptor(void* ctx, C4CollectionSpec spec, C4String docID, FLDict properties,
-                                   C4String keyPath, C4Slice input, C4StringResult* outAlgorithm,
-                                   C4StringResult* outKeyID, C4Error* outError);
-
-static C4SliceResult propDecryptor(void* ctx, C4CollectionSpec spec, C4String docID, FLDict properties,
-                                   C4String keyPath, C4Slice input, C4String algorithm, C4String keyID,
-                                   C4Error* outError);
-#endif
-
 static constexpr const char* kTestUserName = "sguser";
 
 class ReplicatorCollectionSGTest : public ReplicatorAPITest {
   public:
-    ReplicatorCollectionSGTest() : ReplicatorAPITest() {
+    ReplicatorCollectionSGTest() : ReplicatorAPITest(), _expectedSGVersion{"3.1", ""} {
         _sg.pinnedCert = C4Test::readFile("Replicator/tests/data/cert/cert.pem");
         if ( getenv("NOTLS") ) {
             _sg.address = {kC4Replicator2Scheme, C4STR("localhost"), 4984};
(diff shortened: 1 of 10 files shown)


Replicator/tests/ReplicatorCollectionSGTest.hh
```diff
@@ -65,6 +65,11 @@ class ReplicatorCollectionSGTest : public ReplicatorAPITest {
         }
     }
 
+    //    virtual inline bool sgVersionCheck() const {
+    //        alloc_slice serverName = _sg.getServerName();
+    //        return serverName >= "Couchbase Sync Gateway/3.1";
+    //    }
```
borrrden, line 71: I don't really like leaving commented code around

Replicator/tests/SG.cc
```diff
@@ -116,6 +117,9 @@ bool SG::deleteUser(const string& username) const {
 
 bool SG::assignUserChannel(const std::string& username, const std::vector<C4CollectionSpec>& collectionSpecs,
                            const std::vector<std::string>& channelIDs) const {
+    if ( getServerName() < "Couchbase Sync Gateway/3.1" ) return assignUserChannelOld(username, channelIDs);
```
borrrden, line 120: Can you confirm this is intentional and make a comment saying why it works?  Is this comparing the bytes of the two slices by ASCII value?

borrrden (changes requested):

Replicator/tests/ReplicatorCollectionSGTest.hh
callumbirks: Sorry! I thought I found them all

Replicator/tests/SG.cc
callumbirks: I think so. I've done a lot of testing with it, so I know it works, but I don't entirely know what it's doing. I'll try and look into it.

callumbirks: Yeah, under the hood, this does a `memcmp` of the two slices. So I'm gonna say that'll be the ASCII bytes. Maybe some weird scenarios where this would say the version is correct when it isn't / vice versa, but I think it'll do for now. I might add a log call in, to say ~"assigning user channels for SG < 3.1", so if it becomes a problem at any point it can be found easily.

Replicator/tests/ReplicatorCollectionSGTest.hh
```diff
@@ -65,6 +65,11 @@ class ReplicatorCollectionSGTest : public ReplicatorAPITest {
         }
     }
 
+    //    virtual inline bool sgVersionCheck() const {
```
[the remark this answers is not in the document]
callumbirks: Forgot to remove them again 😅

Replicator/tests/data/sg_setup.sh
```diff
@@ -8,41 +8,99 @@
...
+done
+
+
+# Define the REST request to create the DB
+if [ -z $DEFAULT_COLL ]; then
+
+CURL_DB="curl -kL -X PUT '$PROTOCOL://localhost:4985/scratch/' \
+ -H 'Content-Type: application/json' \
+ -H 'Authorization: Basic QWRtaW5pc3RyYXRvcjpwYXNzd29yZA==' \
+ --data-raw '{\"num_index_replicas\": 0, \"bucket\": \"$BUCKET_NAME\", \"scopes\": {\"flowers\": {\"collections\":{\"roses\":{}, \"tulips\":{}, \"lavenders\":{}}}}}'"
```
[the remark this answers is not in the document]
callumbirks: Yes it does

pasin (approved):

Replicator/tests/SG.cc
borrrden: It will be ok until we get to version 3.10, in which case it will sort before 3.2 ;)

borrrden (approved):

PR #1762, opened 2023-04-26
CBL-4462: sync function missing from sg_setup script

callumbirks:
Add sync functions to sg_setup script.
Remove assertions from SGTestUser because some tests caused it to fail. 
Change test tags of a test that shouldn't be grouped with the other 'ReplicatorCollectionTest's.

### DIFF
--- a/Replicator/tests/SGTestUser.hh
+++ b/Replicator/tests/SGTestUser.hh
@@ -24,6 +24,7 @@
 #include "SG.hh"
 #include "Error.hh"
 #include "HTTPLogic.hh"
+#include "c4Private.h"
 
 class SG::TestUser {
   public:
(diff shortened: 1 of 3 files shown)


Replicator/tests/SGTestUser.hh
```diff
@@ -24,6 +24,7 @@
 #include "SG.hh"
 #include "Error.hh"
 #include "HTTPLogic.hh"
+#include "c4Private.h"
```
jianminzhao, line 27: Why is this needed now?

jianminzhao (approved):

Replicator/tests/SGTestUser.hh
callumbirks: Hm, good point, it's probably not

PR #1765, opened 2023-05-02
CBL-4101: adapting tests based on SG 3.0 of walrus mode to also work …

jianminzhao:
…with SG 3.1 backed by CB server.

When Lithium (aka v3.0) was released, all replicator tests were run against SG of the Walrus mode. With v3.1, we started to use SG backed up by actual CB servers. We ported tests in release/3.1, and the associated test framework, to lithium and put them in ReplicatorSGTest, and moved original tests based on Walrus to ReplicatorWalrusTest. This commit adapts the original tests of Walrus to also work with SG(3.1) backed by CB server.


jianminzhao:
This PR only affects sources in test folder, Replicator/tests.

Replicator/tests/ReplicatorWalrusTest.cc
```diff
@@ -79,11 +98,65 @@ class ReplicatorWalrusTest : public ReplicatorAPITest {
...
     }
 
-};
+    enum AuthType {
+        kAuthNone,
+        kAuthBody,
+        kAuthHeader
+    };
 
+    void notWalrus(AuthType authType =kAuthNone) {
```
borrrden, line 107: This is pretty weak function naming

borrrden (approved):

Replicator/tests/ReplicatorWalrusTest.cc
jianminzhao: Totally agreed. It's function is to make up the difference that, for non-walrus, we have to supply credentials, whereas in the walrus mode, database "scratch" is in the guest mode.
What about, "withAuthentication(AuthType)" ?

jianminzhao: Changed function name to withServerBackedSG()

PR #1766, opened 2023-05-02
CBL-4024 : Update BuiltInWebSocket to do preemptive auth by default

jianminzhao:
* Updated BuiltInWebSocket and HTTPLogic to do preemptive auth by default with an option for doing challenge based auth.

* This change is required by CBL-C to do basic auth preemptively to avoid the issue that users can get authenticated as a GUEST user when they don't specify credentials.

### DIFF
--- a/Networking/WebSockets/BuiltInWebSocket.cc
+++ b/Networking/WebSockets/BuiltInWebSocket.cc
@@ -192,6 +192,14 @@ namespace litecore { namespace websocket {
                                         "Invalid/unsupported proxy settings"_sl));
             return nullptr;
         }
+        
+        if (authType == slice(kC4AuthTypeBasic)) {
+            bool enableChallengeAuth = authDict[kC4ReplicatorAuthEnableChallengeAuth].asBool();
+            logic.enableChallengeAuth(enableChallengeAuth);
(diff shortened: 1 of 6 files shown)


jianminzhao:
Note: ReplicatorWalrusTest.cc is manually merged.

bmeike (approved):
I would appreciate it if you would pay a little bit more attention to the requirement that changes to release branches be "minimal".

jianminzhao:
@bmeike the main part of this PR is to support CBL-C  the feature targeted for 3.0.12. As for many scattered changes, they are in the test area, not to be built to Core's artifacts.

Networking/WebSockets/BuiltInWebSocket.cc
```diff
@@ -219,13 +227,14 @@ namespace litecore { namespace websocket {
                 case HTTPLogic::kContinue:
                     break; // Will continue with the same socket (after connecting to a proxy)
                 case HTTPLogic::kAuthenticate: {
+                    // If there is no credentials aet OR the authentication failed after responding
```
callumbirks, line 230: nitpick, typo "aet" instead of "set" 😄

pasin (approved):

PR #1767, opened 2023-05-04
CBL-4199: Remove deprecated pre-collection API usages

callumbirks:
Replaced usages of deprecated API with collection API.
Added some additional API to replace deprecated API used only by tests, there are only used by tests and not exposed publicly.
Ran all C4Tests and CppTests, all passing.

### DIFF
--- a/C/c4CAPI.cc
+++ b/C/c4CAPI.cc
@@ -492,6 +492,18 @@ bool c4db_findDocAncestors(C4Database* database, unsigned numDocs, unsigned maxA
     });
 }
 
+// only used by tests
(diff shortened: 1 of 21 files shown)


C/c4CAPI.cc
```diff
@@ -492,6 +492,18 @@ bool c4db_findDocAncestors(C4Database* database, unsigned numDocs, unsigned maxA
     });
 }
 
+// only used by tests
```
jianminzhao, line 495: Perhaps more accurate is "declared in c4Private.h and meant to be used by tests."

jianminzhao (approved):

borrrden (approved):

pasin (approved):

PR #1768, opened 2023-05-05
Remove bitcode from iOS builds

borrrden:


borrrden:
Need some attention from @Pasin first for Xcode stuff.  Perhaps an update is all that is needed but somehow bitcode is getting forced on for the overall build, which then complains about the lack of bitcode for the overall build....

borrrden:
Actually it looks like upgrading Xcode will do the trick here.  I confirmed things build fine on Xcode 14.2 without any more changes

pasin (approved):
When using XCode 14, this will be perfectly correct.

pasin (approved):
When using XCode 14, this will be perfectly correct.

jianminzhao (approved):

PR #1769, opened 2023-05-09
CBL-4507: Remodel revidBuffer to use composition rather than inheritance

callumbirks:
Remodel revidBuffer to use composition.
Address some small warnings in the revID.hh/.cc files.

See the [Jira issue](https://issues.couchbase.com/browse/CBL-4507) for the motivation behind this.

### DIFF
--- a/C/c4Document.cc
+++ b/C/c4Document.cc
@@ -277,14 +277,14 @@ bool C4Document::equalRevIDs(slice rev1, slice rev2) noexcept {
     try {
         if ( rev1 == rev2 ) return true;
         revidBuffer buf1, buf2;
-        return buf1.tryParse(rev1) && buf2.tryParse(rev2) && buf1.isEquivalentTo(buf2);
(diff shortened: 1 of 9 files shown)


callumbirks:
Linux build keeps failing on:
`couchbase-lite-core/LiteCore/RevTrees/RevID.cc:169:41: error: ‘digittoint’ was not declared in this scope`
It's literally a built in C function, and I've included the header where it's defined. Any clues?

borrrden:
> It's literally a built in C function

From your point of view maybe ;).  It's defined in the BSD spec which means only BSD variants contain it (including Apple)

For the other platforms Core has one written in StringUtil.hh

callumbirks:
Interesting, Xcode build fails on `VectorRecordTest.cc`. That file is not included in the cmake builds.

pasin (approved):

borrrden (approved):

PR #1773, opened 2023-05-12
CBL-4224: Analysis Warnings (4/13) Crypto/ + LiteCore/Database

callumbirks:
Notable changes
- `SubjectAltNames` is adjusted to use composition rather than inheriting from `vector`, which was causing a few issues.
- Many `operator` functions which should have been marked explicit now are (This means it will now be necessary to cast those types explicitly).
    - In one place, I added in a helper constructor (`Cert` class) that does the explicit type casting, so the constructor can be called the same way still.
- An `std::bind` in Housekeeper changed to a lambda func. They both produce the same functionality, but the `std::bind` is harder to read and increases the binary size produced because of unnecessary type information it creates.

On top of the usual changes:
- Optimizing includes
- `std::move` calls are now qualified
- Lots of parameters now using `const&` or `std::move` rather than copying
- Deleted member functions moved to public scope of classes
- Concatenating namespaces
- Marking functions as `[[nodiscard]]`
- Removing `virtual` from functions which are already marked `override`
- Performing `static_cast` where necessary to avoid undefined behaviour with implicit casts (Usually various size types with mixed signs)

### DIFF
--- a/Crypto/Certificate.hh
+++ b/Crypto/Certificate.hh
@@ -85,17 +85,19 @@ namespace litecore::crypto {
         [[nodiscard]] fleece::alloc_slice encode() const;
 
         [[nodiscard]] bool empty() const { return _names.empty(); }
-       //'''/;;;;''''''
+
+        //'''/;;;;''''''
         [[nodiscard]] size_t size() const { return _names.size(); }
 
-        template <typename ... Args>
-        void emplace_back(Args && ... args) {
+        template <typename... Args>
+        void emplace_back(Args&&... args) {
             _names.emplace_back(std::forward<Args>(args)...);
         }
 
         fleece::alloc_slice operator[](Tag) const;
 
         const SubjectAltName& operator[](size_t i) const { return _names[i]; }
+
       private:
         std::vector<SubjectAltName> _names;
     };
--- a/Crypto/Certificate.hh
+++ b/Crypto/Certificate.hh
@@ -12,11 +12,9 @@
 
 #pragma once
 #include "PublicKey.hh"
-#include "fleece/function_ref.hh"
 #include <initializer_list>
 #include <memory>
 #include <optional>
-#include <time.h>
 #include <utility>
 #include <vector>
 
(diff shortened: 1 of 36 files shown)


pasin (approved):

Crypto/Certificate.hh
```diff
@@ -84,11 +82,24 @@ namespace litecore { namespace crypto {
         SubjectAltNames() = default;
         explicit SubjectAltNames(::mbedtls_asn1_sequence*);
 
-        fleece::alloc_slice encode() const;
+        [[nodiscard]] fleece::alloc_slice encode() const;
+
+        [[nodiscard]] bool empty() const { return _names.empty(); }
+
+        //'''/;;;;''''''
```
borrrden, line 89: ???

borrrden (approved):

jianminzhao:
This are mass changes.
1. Regarding std::function vs lambda, in the end, we are keeping the callback as std::function. The lambda passed to Housekeeping will be freed afterwards. 
2. I am not clear of the reason behind inheritance vs composition. I think we have multiple uses of inheritance pattern.

callumbirks:
> This are mass changes.
> 
>     1. Regarding std::function vs lambda, in the end, we are keeping the callback as std::function. The lambda passed to Housekeeping will be freed afterwards.
> 
>     2. I am not clear of the reason behind inheritance vs composition. I think we have multiple uses of inheritance pattern.

The lambda used in that way has the same effect as `std::bind`, just the code is a lot cleaner.
There were warnings caused by the inheritance, as the derived classes were only being used to add functionality to the base class, and mostly the derived class was being discarded in order to use the base class. 
i.e. `revidBuffer` was just used to hold a buffer for revid and to parse ASCII strings to binary revid, then the `revidBuffer` class was passed as a revid (discarding the buffer). In our use case, it did not seem to be dangerous, but using composition achieves the same result without implicitly discarding anything.

jianminzhao (approved):

PR #1774, opened 2023-05-12
CBL-3903: Assertion failure when stopping replicator while replicator…

jianminzhao:
… is connecting

### DIFF
--- a/Networking/BLIP/LoopbackProvider.hh
+++ b/Networking/BLIP/LoopbackProvider.hh
@@ -63,7 +63,7 @@ namespace litecore { namespace websocket {
         }
 
         virtual void connect() override {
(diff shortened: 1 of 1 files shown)


pasin (commented):
Can you explain why the original assertion (_driver->_peer) in the connection() function is not correct sometime and why move it to _connect() solves the problem?

jianminzhao:
This is because connect() is called in the replicator's worker thread, and c4repl_stop() called from main thread. That means, WebSocket::connect() and _closed run concurrently. And, _peer is reset in _closed. Therefore, we cannot assert in connect(). _drive is ref counted. So, we assert _driver but not _peer.

pasin (approved):

borrrden (approved):

PR #1779, opened 2023-05-16
CBL-4499: Replicator may get stuck when there is an error of "Invalid…

jianminzhao:
… delta"

Replicator is stuck in busy state when there is an error thrown while applying delta to create full fleece doc

This PR does not fix the delta error. (The reason of the delta errors as shown in the ticket is still unknown.) Instead, I try to fix the problem of "stuck in busy state." In fact, our code already has paths to deal with the possible delta error. After injecting delta errors for push and pull, bugs emerged as the ticket describes.

After the fix, the delta error will trigger the second round of "rev" message which uses full rev as opposed to the delta rev.

### DIFF
--- a/Replicator/IncomingRev.cc
+++ b/Replicator/IncomingRev.cc
@@ -166,12 +166,10 @@ namespace litecore { namespace repl {
                     if ( _options->noIncomingConflicts() ) err = {WebSocketDomain, 409};
                     else
                         err = C4Error::printf(LiteCoreDomain, kC4ErrorDeltaBaseUnknown,
-                                              "Couldn't apply delta: Don't have body of '%.*s' #%.*s", SPLAT(_rev->docID),
-                                              SPLAT(_rev->deltaSrcRevID));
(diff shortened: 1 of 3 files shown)


jianminzhao:
This is backported from [CBL-4448](https://issues.couchbase.com/browse/CBL-4448)
Delta errors are rare; I have to inject fake errors to the main source in quite coarse fashion. To review how the code is tested, you may apply the patch attached in the ticket, CBL-4499, on top of this commit.

borrrden:
Nobody is going to remember that this patch is in the ticket.  This is a really weird way to do things....

borrrden (approved):

jianminzhao:
I agree. Since this is main dev branch, we probably should do it a more systematic way.
How about define a project-wise DEFINE, like UNIT_TEST, which will be defined in the test schemes, both in Xcode and cmake. In the main source, we include the ad-hoc and ugly code in #ifdef UNIT_TEST.
If this is good idea, I can create a ticket.

pasin (approved):

PR #1780, opened 2023-05-16
CBL-4533 : Add docker compose running CBS and SGs for tests

pasin:
* The docker compose will start one CBS and two SGs : 3.1 SG with flowers collection configured (`:4984/scratch`) and legacy 3.0 SG with default collection configured (`:4884/scratch-30`). Found that when connecting two SGs with the same database name to the same CB Server, SG could get confused and result to collection or namespace not found error.

* The SG's certs are copied from the top level `cert` directory to `docker/sg/cert` directory. This is mandatory as the docker cannot copy anything from the parent directory, and symlink is not supported either. So I made the top level `cert` directory a symlink of `docker/sg/cert` instead.

* The default SG binaries set in docker-compose.yml are ARM64 based and the SSL is enabled. This could be adjusted. To reconfigure without changing docker-compose.yml, create .env with SG_DEB and SSL properties.

* Tested by running all tests with "[.SyncServerCollection],[.SyncServer30]" tags. And with the docker compose, we could setup GH Action to run [.SyncServerCollection],[.SyncServer30] tests without much of the work (Not in this PR).

* Cherry-picked @callumbirks 's 0706d9745453839b9c5f9f09ead634b57226743b commit from the docker-fix branch that updates the test code to use port 4884 for SyncServer30 tests.


Replicator/tests/data/docker/README.md
```diff
@@ -0,0 +1,69 @@
...
+ #### Sync Function
+
+ Use default sync function which is as follows
+ ```JS
+    function (doc, oldDoc) {
+        channel(doc.channels); 
+    }
+ ``` 
+
+The Admin Credentials of both Sync Gateways are `admin/password` or `Administrtor/password`.
```
borrrden, line 41: typo in second credentials?

Replicator/tests/data/docker/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+The `docker-compose.yml` has 3 environment variables for configuration.
+
+ | Variable      |   Description  |
+ | ------------- | -------------- |
+ | SG_DEB        | Sync Gateway 3.1+ deb file URL. Default is SG 3.1.0 Ubuntu ARM64 URL.      |
+ | SG_LEGACY_DEB | Sync Gateway 3.0 deb file URL. Default is SG 3.0.7 Ubuntu ARM64 URL.       |
+ | SSL           | Boolean flag to configure Sync Gateway for SSL endpoints. Default is true. |
+
+ Note: If you are using Mac x86-64, you must configure SG_DEB and SG_LEGACY_DEB to use x86_64 versions.
```
borrrden, line 53: This can be fairly simply automated.  See https://stackoverflow.com/a/70369688/1155387

borrrden (approved):

Replicator/tests/data/docker/README.md
```diff
@@ -0,0 +1,69 @@
...
+ ``` 
+
+### 2. Legacy Sync Gateway for .SyncServer30 Tests
+
+ | Config      | Value       |
+ | ----------- | ----------- |
+ | Database    | scratch-30  |
+ | Port        | 4884        |
+ | Admin Port  | 4885        |
+ | Collections | default     |
```
jianminzhao, line 30: Does SB 3.0 know Collections?

pasin: No, it's just the name in the table here to make it consistent with the other table. I could remote it if that is confused.

Replicator/tests/data/docker/README.md
pasin: Yes, will fix.

Replicator/tests/data/docker/README.md
pasin: Currently the URL is set from docker-compose though. Maybe having the env to supply both and docker can pick one based on the architecture.

Replicator/tests/data/docker/README.md
jianminzhao: I think it's confusing. Perhaps put a note there in lieu of "default", something "N/A"

jianminzhao (approved):

PR #1781, opened 2023-05-17
CBL-4226: Analysis Warnings (6/13) LiteCore/Support

callumbirks:
## Notable changes
- `error::convertException()` re-arranged to address the issues around `std::invalid_argument` and `std::domain_error` inheriting from `std::logic_error`.
- Removed `Actor` functions; `scheduler()` `setScheduler()` and `startScheduler()` as they did nothing and were unused
- One of the `Any` constructors was potentially hiding the copy and move constructors. I've modified it to only allow types that aren't `Any`.
- `error::assertionFailed` modified to throw anonymous error value rather than local. This is part of the C++ Core guidelines and exists because of dangers around the lifetime of the local error variable.

Some changes are very spread across Core, mostly small changes that had to be made to address side-effects of optimizing imports or making functions explicit that should be explicit.

And, the usual changes.

### DIFF
--- a/LiteCore/Support/Any.hh
+++ b/LiteCore/Support/Any.hh
@@ -44,25 +44,29 @@ namespace litecore {
     using StorageType = typename std::decay<T>::type;
 
     struct Any {
-        bool isNull() const { return _ptr == nullptr; }
+        [[nodiscard]] bool isNull() const { return _ptr == nullptr; }
 
-        bool isNotNull() const { return _ptr != nullptr; }
+        [[nodiscard]] bool isNotNull() const { return _ptr != nullptr; }
 
         Any() : _ptr(nullptr) {}
 
         Any(Any& that) : _ptr(that.clone()) {}
 
-        Any(Any&& that) : _ptr(that._ptr) { that._ptr = nullptr; }
+        Any(Any&& that) noexcept : _ptr(that._ptr) { that._ptr = nullptr; }
 
         Any(const Any& that) : _ptr(that.clone()) {}
 
-        Any(const Any&& that) : _ptr(that.clone()) {}
+        Any(const Any&& that) noexcept : _ptr(that.clone()) {}
 
-        template <typename U>
+        // Making the constructors in this class explicit breaks half the code
+        // NOLINTBEGIN(google-explicit-constructor)
+
+        // Only enable for types that aren't `Any` to avoid the compiler choosing this over copy or move
+        template <typename U, typename = std::enable_if_t<!std::is_same_v<Any, std::decay_t<U>>>>
         Any(U&& value) : _ptr(new Derived<StorageType<U>>(std::forward<U>(value))) {}
 
         template <class U>
-        bool is() const {
+        [[nodiscard]] bool is() const {
             auto derived = getDerived<U>(false);
 
             return derived != nullptr;
@@ -87,8 +91,10 @@ namespace litecore {
             return as<StorageType<U>>();
         }
 
+        // NOLINTEND(google-explicit-constructor)
+
         template <class U>
-        operator const U() const {
+        explicit operator U() const {
             return as<const StorageType<U>>();
         }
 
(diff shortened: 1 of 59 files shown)


LiteCore/Support/LogDecoder.cc
```diff
@@ -218,8 +217,8 @@ namespace litecore {
                         case 'd':
                         case 'i':
                             {
-                                bool    negative = _in.get() > 0;
-                                int64_t param    = readUVarInt();
+                                bool negative = _in.get() > 0;
+                                auto param    = static_cast<int64_t>(readUVarInt());
```
borrrden, line 221: Are these in need of narrow_cast?

pasin (approved):

LiteCore/Support/LogDecoder.cc
jianminzhao: Ya, narrow_cast will do runtime check. Don't know what static_cast will do to convert unsigned to signed. How will it be different from plain assignment?

callumbirks: Plain assignment (implicit cast) is platform-dependent. I can change it to narrow_cast if we think that is more suitable.

jianminzhao: My understanding is implicit cast and static_cast do the same: keep all the bits unchanged. All platforms use 2's complement to represent negative.

callumbirks: From LLVM's documentation:
> The C/C++ standard does not mandate two’s complement for signed integers, and so the compiler is free to define what the semantics are for converting an unsigned integer to signed integer. Clang’s implementation uses the two’s complement format.

callumbirks: And yeah, implicit cast and static_cast do the same thing. Just implicit cast means it's not obvious that you intended to cast between signed/unsigned and therefore you aren't necessarily aware of the dangers. Whereas static_cast makes it explicit that you understand you are converting between signed/unsigned and accept the dangers.

borrrden: Basically static cast only does some trivial check that you are assigning to compatible types.  For example you cannot static cast from int to string.  It's marginally better than implicit cast (which in C++ is called reinterpret_cast) which literally just takes the size of the underlying struct or pointer and reads it as another.

narrow_cast in release mode is just static_cast, but in debug mode it checks that the result is not out of bounds.

borrrden (commented):
You have reviewed the other locations of static_cast for narrow_cast possibility?

LiteCore/Support/Channel.cc
```diff
@@ -10,9 +10,11 @@
 // the file licenses/APL2.txt.
 //
 
-#include "Channel.hh"
-#include "Error.hh"
+#if 0
```
jianminzhao, line 13: Why to keep it?

LiteCore/Support/Logging.cc
```diff
@@ -415,10 +416,12 @@ namespace litecore {
         va_end(args);
     }
 
+    // Can't make this function static, it breaks the usage of __printflike.
+    // NOLINTBEGIN(readability-convert-member-functions-to-static)
     // Must have sLogMutex held
     void LogDomain::dylog(LogLevel level, const char* domain, unsigned objRef, const char* fmt, va_list args) {
         auto     obj = getObject(objRef);
-        uint64_t pos = 0;
+        uint64_t pos;
```
jianminzhao, line 424: Why not to initilize?

jianminzhao (approved):

LiteCore/Support/Logging.cc
callumbirks: Because the variable is assigned a value in either code path - so it will always be initialized by the time it's used. Initializing here has no effect.

callumbirks:
> You have reviewed the other locations of static_cast for narrow_cast possibility?

@borrrden In that file, yes - across the other changes in the PR, no. I wasn't sure how to search for it while limiting scope to changes I've made in this PR. I can give it another try - and update the PR if I'm successful.

LiteCore/Support/Any.hh
```diff
@@ -121,40 +127,42 @@ namespace litecore {
...
             ;
-            virtual Base* clone() const = 0;
+            [[nodiscard]] virtual Base* clone() const = 0;
         };
 
         template <typename T>
         struct Derived : Base {
-            template <typename U>
-            Derived(U&& value_) : value(std::forward<U>(value_)) {}
+            template <typename U, typename = std::enable_if_t<!std::is_same_v<Derived, std::decay_t<U>>>>
```
borrrden, line 143: TIL about `std::decay` wow!

LiteCore/Support/Channel.cc
borrrden: I am universally in favor of not committing commented or ifdef'd code.  If it is needed again, git has it in its history.  Perhaps the next part of this PR can remove it.

borrrden (approved):

PR #1782, opened 2023-05-18
CBL-4531 : Update PublicKey+Apple to use non deprecated keychain APIs

pasin:
* Ported 539b8084786b09d6d99f3799a7738ad14aaadf22 and 52d835eb14539813b699f483d3a9afc01a25c0c0 from master / 3.1 branch.
* Plus, Removed base SDK check as it's not necessary and corrected issuer and serial number data type.

### DIFF
--- a/Crypto/PublicKey+Apple.mm
+++ b/Crypto/PublicKey+Apple.mm
@@ -340,12 +340,26 @@ namespace litecore { namespace crypto {
                 (id)kSecAttrIsPermanent:    @YES,
                 (id)kSecAttrLabel:          @(timestr),
             };
-            SecKeyRef publicKey, privateKey;
-            ++gC4ExpectExceptions;
-            OSStatus err = SecKeyGeneratePair((CFDictionaryRef)params, &publicKey, &privateKey);
-            --gC4ExpectExceptions;
-            checkOSStatus(err, "SecKeyGeneratePair", "Couldn't create a private key");
-
+            
+            SecKeyRef publicKey = NULL, privateKey = NULL;
+            if (@available(macOS 10.12, iOS 10.0, *)) {
+                CFErrorRef error;
+                ++gC4ExpectExceptions;
+                privateKey = SecKeyCreateRandomKey((CFDictionaryRef)params, &error);
+                --gC4ExpectExceptions;
+                if (!privateKey) {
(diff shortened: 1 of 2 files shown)


bmeike (approved):
Would it be worthwhile to create two subclasses, one for 12/15 and greater, one for the rest?  That might save those if-then-else paths...

Crypto/PublicKey+Apple.mm
```diff
@@ -340,12 +340,26 @@ virtual int _sign(int/*mbedtls_md_type_t*/ mbedDigestAlgorithm,
...
-            ++gC4ExpectExceptions;
-            OSStatus err = SecKeyGeneratePair((CFDictionaryRef)params, &publicKey, &privateKey);
-            --gC4ExpectExceptions;
-            checkOSStatus(err, "SecKeyGeneratePair", "Couldn't create a private key");
-
+            
+            SecKeyRef publicKey = NULL, privateKey = NULL;
+            if (@available(macOS 10.12, iOS 10.0, *)) {
+                CFErrorRef error;
+                ++gC4ExpectExceptions;
```
jianminzhao, line 347: gC4ExpectException is, I think, intended for tests Why here?

pasin: Those are ported from the master branch. I do not know exactly but there are multiple places that use that in the file. My best guess is that Apple Keychain internally throws exception and without that all tests that use keychain could be annoyingly stopped when the exception was thrown internally. 

If the recommendation is to not using those, that should be done in a separate tickets in all places in the file.

pasin:
> Would it be worthwhile to create two subclasses, one for 12/15 and greater, one for the rest? That might save those if-then-else paths...

This is a good suggestion, but I'm not sure if it's worth to do or not as they will be 3 subclasses as of now and could be more in the future as Apple keeps changing this. Also this is too big of the change to the release branch.

Crypto/PublicKey+Apple.mm
pasin: I created CBL-4532.

jianminzhao (approved):

bmeike:
Great point about too big a change.  I hadn't noticed that this targeted a release branch.

PR #1783, opened 2023-05-18
CBL-4536: Share `DBAccess` of `Replicator` with `BuiltinWebSocket`

blaugold:
<strike>
This change opens a dedicated `C4Database` for the `BuiltinWebSocket` so that it can safely use it in the thread that is launched to establish a connection.
</strike>

This change shares the thread safe `DBAccess` of `Replicator` with `BuiltinWebSocket`.

Previously, the same `C4Database` that was used by an app was also used by the `BuiltinWebSocket`. This left the possibility of unsafe concurrent access from two threads, which could leave the database in an inconsistent state, rendering it unusable.

### DIFF
--- a/Replicator/Replicator.cc
+++ b/Replicator/Replicator.cc
@@ -58,12 +58,9 @@ namespace litecore { namespace repl {
         return result.str();
     }
 
-    shared_ptr<DBAccess> Replicator::createDBAccess(C4Database* db, Options* options) {
-        return make_shared<DBAccess>(db, options->properties["disable_blob_support"_sl].asBool());
-    }
-
     Replicator::Replicator(C4Database* db, websocket::WebSocket* webSocket, Delegate& delegate, Options* options)
-        : Replicator(createDBAccess(db, options), webSocket, delegate, options) {}
+        : Replicator(make_shared<DBAccess>(db, options->properties["disable_blob_support"_sl].asBool()), webSocket,
+                     delegate, options) {}
 
     Replicator::Replicator(shared_ptr<DBAccess> db, websocket::WebSocket* webSocket, Delegate& delegate,
                            Options* options)
--- a/Replicator/Replicator.cc
+++ b/Replicator/Replicator.cc
@@ -58,10 +58,16 @@ namespace litecore { namespace repl {
         return result.str();
     }
 
+    shared_ptr<DBAccess> Replicator::createDBAccess(C4Database* db, Options* options) {
+        return make_shared<DBAccess>(db, options->properties["disable_blob_support"_sl].asBool());
+    }
+
     Replicator::Replicator(C4Database* db, websocket::WebSocket* webSocket, Delegate& delegate, Options* options)
-        : Worker(new Connection(webSocket, options->properties, {}), nullptr, options,
-                 make_shared<DBAccess>(db, options->properties["disable_blob_support"_sl].asBool()), "Repl",
-                 kNotCollectionIndex)
+        : Replicator(createDBAccess(db, options), webSocket, delegate, options) {}
+
+    Replicator::Replicator(shared_ptr<DBAccess> db, websocket::WebSocket* webSocket, Delegate& delegate,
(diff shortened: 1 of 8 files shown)


jianminzhao:
Good catch! The BuiltInWebSocket only uses the db for connecting and has incur the cost of concurrency. It would be more efficient to be able to share db, shared_ptr<DBAccess>, with the Replicator. But current Replicator API does not take it.

jianminzhao:
@snej what do you think?

snej:
I was actually surprised to see that BuiltinWebSocket uses C4Database at all -- it seems like a layering problem. The only thing it uses the database for is getting and setting cookies. It seems a waste to open a new db handle (with megabytes of block cache) just for that.

The C4RemoteReplicator already opens another db handle for the Replicator to use. This one should be fine to use because the WebSocket only uses it before the connection is established, and the replicator only uses it afterwards. So in C4RemoteReplicator.hh:

```c++
            auto webSocket     = CreateWebSocket(_url, socketOptions(), _database, _socketFactory);
            auto dbOpenedAgain = _database->openAgain();
```

How about reversing those two lines and passing dbOpenedAgain to CreateWebSocket?

snej (changes requested):
(see comment above)

jianminzhao:
That is my first reaction on the matter. But reading more carefully, I found Replicator actually start to use database after threading out to establish the connection. So, there is a potential racing.
If we could have the following 
    Replicator::Replicator(C4Database* db,
                           websocket::WebSocket *webSocket,
                           Delegate &delegate,
                           Options options)
to take "shared_ptr<DBAccess>" in place of the first argument, then we can safely share it with WebSocket.

pasin:
Please note that CBL-4529 is created for release/lithium branch. For this PR, we will need another ticket for reference.

pasin:
I have cloned the ticket into CBL-4536 for master branch.

blaugold:
Thanks for the feedback. I've changed the PR to share the `DBAccess` of `Replicator` with `BuiltinWebSocket`, as suggested by @jianminzhao.

@pasin Thanks for linking the correct ticket.

Replicator/Replicator.hh
```diff
@@ -43,7 +43,10 @@ namespace litecore { namespace repl {
         using CloseStatus = blip::Connection::CloseStatus;
         using Options     = litecore::repl::Options;
 
+        static shared_ptr<DBAccess> createDBAccess(C4Database* NONNULL, Options* NONNULL);
```
jianminzhao, line 46: It appears to me little extraneous:  createDBAccess.  vs make_shared<DBAccess>. The latter is more straightforward, saving me a "go to definition" as I encounter it. 
Otherwise, very good.

blaugold: I pulled `createDBAccess` out because we now create `DBAccess`, based on `Options`, in two different places. 

1. In the preexisting `Replicator` constructor, that takes a `C4Database`. It's convenient to keep that because it is used in tests.
2. In `C4RemoteReplicator::createReplicator`.

I wanted to avoid the duplication, but please let me know if I should create `DBAccess` with `make_shared` in both places.

jianminzhao: Of course, we don't want to remove the old API. I meant to compare make_shared<DBAccess>() with createDBAccess(). They have the same argument signature and users supply same arguments.

Replicator/Replicator.cc
```diff
@@ -74,7 +77,7 @@ namespace litecore { namespace repl {
             //                    : all collections are passive.
             _options->verify();
 
-            _loggingID  = string(db->getPath()) + " " + _loggingID;
+            _loggingID  = string(db->useLocked()->getPath()) + " " + _loggingID;
```
jianminzhao, line 80: good catch

jianminzhao (approved):

snej (approved):
Looks good to me. Thanks!

PR #1784, opened 2023-05-19
CBL-4227: Analysis Warnings (7/13) LiteCore/RevTrees

callumbirks:
No notable changes in this one.
Also a small addition to gitignore for CLion config files (I didn't feel like this justified a ticket).

### DIFF
--- a/LiteCore/RevTrees/RawRevTree.cc
+++ b/LiteCore/RevTrees/RawRevTree.cc
@@ -14,6 +14,7 @@
 #include "RevTree.hh"
 #include "Error.hh"
 #include "varint.hh"
+#include "NumConversion.hh"
 
 using namespace std;
 using namespace fleece;
@@ -45,7 +46,7 @@ namespace litecore {
 
     std::deque<Rev> RawRevision::decodeTree(slice raw_tree, RevTree::RemoteRevMap& remoteMap,
                                             std::vector<const Rev*>& rejectedRevs, RevTree* owner, sequence_t curSeq) {
-        const RawRevision* rawRev = (const RawRevision*)raw_tree.buf;
+        auto rawRev = (const RawRevision*)raw_tree.buf;
         if ( fleece::endian::dec32(rawRev->size_BE) > raw_tree.size )
             error::_throw(error::CorruptRevisionData, "RawRevision decodeTree binary error");
         unsigned count = rawRev->count();
@@ -103,7 +104,7 @@ namespace litecore {
         alloc_slice result(totalSize);
 
         // Write the raw revs:
-        RawRevision* dst = (RawRevision*)result.buf;
+        auto dst = (RawRevision*)result.buf;
         for ( Rev* src : revs ) { dst = dst->copyFrom(*src); }
         dst->size_BE = endian::enc32(0);  // write trailing 0 size marker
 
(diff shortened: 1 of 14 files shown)


jianminzhao (approved):

LiteCore/RevTrees/VectorRecord.cc
```diff
@@ -276,7 +275,7 @@ namespace litecore {
                 revDict[kRevIDKey].setData(newRev.revID);
                 _changed = true;
             }
-            if ( newRev.properties != revDict[kRevPropertiesKey] ) {
+            if ( newRev.properties != revDict.get(kRevPropertiesKey) ) {
```
pasin, line 278: Wondering why using subscript is not recommended?

pasin (approved):

LiteCore/RevTrees/RawRevTree.cc
```diff
@@ -45,7 +46,7 @@ namespace litecore {
 
     std::deque<Rev> RawRevision::decodeTree(slice raw_tree, RevTree::RemoteRevMap& remoteMap,
                                             std::vector<const Rev*>& rejectedRevs, RevTree* owner, sequence_t curSeq) {
-        const RawRevision* rawRev = (const RawRevision*)raw_tree.buf;
+        auto rawRev = (const RawRevision*)raw_tree.buf;
```
borrrden, line 49: const auto ?  I don't actually know what rules apply to auto when casting explicitly to a const, but `const auto` is a thing.

LiteCore/RevTrees/RevTree.cc
```diff
@@ -569,10 +569,7 @@ namespace litecore {
     }
 
     bool RevTree::hasNewRevisions() const {
-        for ( Rev* rev : _revs ) {
-            if ( rev->isNew() || rev->sequence == 0_seq ) return true;
-        }
-        return false;
+        return std::any_of(_revs.begin(), _revs.end(), [](Rev* rev) { return rev->isNew() || rev->sequence == 0_seq; });
```
borrrden, line 572: I'm interested in the clang analysis justification for this.  It doesn't seem any more readable than the former, so does it have some performance gains or something?

LiteCore/RevTrees/RevTree.hh
```diff
@@ -26,36 +25,36 @@ namespace litecore {
...
+        [[nodiscard]] bool isConflict() const FLPURE { return (flags & kIsConflict) != 0; }
 
-        bool isClosed() const FLPURE { return (flags & kClosed) != 0; }
+        [[nodiscard]] bool isClosed() const FLPURE { return (flags & kClosed) != 0; }
 
-        bool keepBody() const FLPURE { return (flags & kKeepBody) != 0; }
+        [[nodiscard]] bool keepBody() const FLPURE { return (flags & kKeepBody) != 0; }
 
-        bool isActive() const FLPURE;
+        [[nodiscard]] bool isActive() const FLPURE;
```
borrrden, line 51: Can't we just make this the default somehow? 😝

borrrden (approved):

LiteCore/RevTrees/VectorRecord.cc
callumbirks: Because fleece::MutableDict subscript operator has a different return type than expected. While fleece::Dict `[]` has a return of `fleece::Value`, and the `get` methods from both Dict and MutableDict return a `fleece::Value`, the MutableDict `[]` returns a `keyref<MutableDict,slice>`. 

The `keyref` type does inherit from `Value`, but has some additional state. Here, it is just used as a `Value`, so `.get` is more suitable.

LiteCore/RevTrees/RawRevTree.cc
callumbirks: The variable still has type `const RawRevision*`. I think using `const auto` can add const to a value you don't want to change, but using `auto` for a type that is already const can't take the const-ness away.

LiteCore/RevTrees/RevTree.hh
callumbirks: That is something I wonder myself. I'll do some research on it. At first I avoided the warning specifically because of these. I hate putting `[[nodiscard]]` next to virtually every bloody function.

LiteCore/RevTrees/RevTree.cc
callumbirks: LLVM's justification for preferring this is lacking. As far as I know it's about the same performance. To me, this makes the code more verbal on what exactly it is doing - the reader can see `any_of`, and a lambda function.

PR #1787, opened 2023-05-23
CBL-4544: Add Github Action workflow to run tests with SG

callumbirks:
Testing this using this PR. This relies on the change in CBL-4545 PR, so that change is also included here.


callumbirks:
I just thought. We probably don't want this to run on lithium, or any PRs targeting it. No idea how to do that.

.github/workflows/replication.yml
```diff
@@ -0,0 +1,93 @@
...
+      run: |
+        pushd LiteCore/tests
+        ./CppTests -r quiet "[.SyncServer30]"
+        popd
+
+    - name: Stop Docker containers
+      if: always()
+      run: |
+        pushd Replicator/tests/data/docker
+        docker-compose down
```
pasin, line 90: docker compose down

pasin (approved):
The workflow will be run only on the branch that has the workflow yaml file so with only this change, only PR or commit to the master branch will trigger the work flow.

Have one comment about docket-compose down. The rest looks good to me.

callumbirks:
> The workflow will be run only on the branch that has the workflow yaml file so with only this change, only PR or commit to the master branch will trigger the work flow.
> 
> Have one comment about docker-compose down. The rest looks good to me.

Of course yeah, I don't know why I didn't think about that 🙂

borrrden (commented):
What is the story on the validation failure?

pasin:
The failure is a flaky test, but it's not related to this PR.

borrrden (approved):

PR #1793, opened 2023-05-26
CBL-4050: Upgrade to SQLite 3.40.1

jianminzhao:
In order to build it with Xcode 14, I need set some flags in Xcode configs.


jenkins/couchbase-lite-core-black-duck-manifest.yaml
```diff
@@ -0,0 +1,51 @@
...
+    src-path: vendor/fleece/vendor/jsonsl
+    parent-repo: vendor/fleece
+  libb64:
+    bd-id: 574849b1-902f-4367-a686-ca11c8815ba4
+    versions: [ 1.2.1 ]
+    src-path: vendor/fleece/vendor/libb64
+    parent-repo: vendor/fleece
+  libicu:
+    bd-id: a7441c50-9be3-493b-82ed-19666236acef
+    versions: [ 71.1 ]
```
borrrden, line 17: I'm not sure this is correct for Hydrogen....

jianminzhao: Yes. Thanks for catching it up. I cannot verify it from the source. How to find/decide it?

borrrden: The old record came from the build repo, and you can see the recorded version here https://github.com/couchbase/build/blob/master/scripts/jenkins/lite-core/manifest.cmake

jianminzhao: The cmake file you pointed to above has version 54.1. Is that cmake's way obsoleted? Going forward, how can I find it.

PR #1794, opened 2023-05-30
CBL-4411: Enhance checkpoint resolution algorithm when local and remo…

jianminzhao:
…te checkpoint are mismatched

### DIFF
--- a/Replicator/Checkpoint.cc
+++ b/Replicator/Checkpoint.cc
@@ -102,21 +102,43 @@ namespace litecore { namespace repl {
     }
 
 
-    bool Checkpoint::validateWith(const Checkpoint &remoteSequences) {
+    bool Checkpoint::validateWith(const Checkpoint& remoteSequences) {
+        // If _completed or _remote changes in any way because of this method, this method must
+        // return false.  Currently the only way this remains true is if neither of the below if
+        // blocks are entered, or if we ignore the fact that the only difference in the local
+        // and remote checkpoint documents are that the INTEGRAL (i.e. NOT a backfill checkpoint)
+        // remote sequence on the local side is older than on the remote side and all we need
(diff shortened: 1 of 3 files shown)


Replicator/Checkpoint.cc
```diff
@@ -102,21 +102,43 @@ namespace litecore { namespace repl {
...
         if (_completed != remoteSequences._completed) {
-            LogTo(SyncLog, "Local sequence mismatch: I had completed: %s, remote had %s",
-                  _completed.to_string().c_str(),
-                  remoteSequences._completed.to_string().c_str());
-            resetLocal();
+            LogTo(SyncLog, "Local sequence mismatch: I had completed: %s, remote had %s.",
+                _completed.to_string().c_str(),
+                remoteSequences._completed.to_string().c_str());
+            LogTo(SyncLog, "Rolling back to a failsafe, some redundant changes may be proposed...");
+            _completed = SequenceSet::intersection(_completed, remoteSequences._completed);
```
jianminzhao, line 119: I don't know the exact rational behind the intersections. My understanding is, the remote checkpoint we fetched from the remote was sent by us in the past session. We could get to a benign mismatch if the the last setCheckpoint dd make it to the remote (otherwise, the sync state is sound). In this case, local _complete would properly contain remoteSequences._completed. It sounds reasonable to use the previous sync-point as stored represented in the remote. I am afraid arbitrary mismatch (intersection) may be a sign of some kind of corruption.

snej (approved):

Replicator/Checkpoint.cc
borrrden: There is a third case of pure checkpoint corruption, usually by using different local databases with the same UUID inside (i.e. not using the copy API)

borrrden (approved):

PR #1798, opened 2023-06-01
CBL-4532: Remove gC4ExpectExceptions usage from PublicKey+Apple

jianminzhao:
gC4ExpectExceptions is for the test code.


pasin:
I will need to run URLEndpointTests from iOS to see the impact as I'm not sure if I can always set ExpectException where the code is run in the test.

borrrden (approved):
Conditional on Pasin's findings

pasin (approved):
Good to go.

PR #1800, opened 2023-06-05
CBL-4414: Enhance checkpoint resolution algorithm when local and remo…

jianminzhao:
…te checkpoint are mismatched.


bmeike (approved):

borrrden (approved):
It looks like yes, but this includes the subsequent fix that you pointed out, for the failing test right?

pasin (approved):

PR #1801, opened 2023-06-05
CBL-4445: Replicator may get stuck when there is an error of "Invalid…

jianminzhao:
… delta"

Replicator is stuck in busy state when there is an error thrown while applying delta to create full fleece doc This PR does not fix the delta error. (The reason of the delta errors as shown in the ticket is still unknown.) Instead, I try to fix the problem of "stuck in busy state." In fact, our code already has paths to deal with the possible delta error. After injecting delta errors for push and pull, bugs emerged as the ticket describes. After the fix, the delta error will trigger the second round of "rev" message which uses full rev as opposed to the delta rev.

### DIFF
--- a/Replicator/IncomingRev.cc
+++ b/Replicator/IncomingRev.cc
@@ -174,19 +174,19 @@ namespace litecore { namespace repl {
             // It's a delta, but we need the entire document body now because either it has to be
             // passed to the validation function, it may contain new blobs to download, or it may
             // have properties to decrypt.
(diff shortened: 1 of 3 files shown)


pasin (approved):

borrrden:
Don't just say "ported from lithium", give us the commit so we can compare.

borrrden (approved):
Looks identical to the change it was ported from, weird that any manual merging was needed.

PR #1804, opened 2023-06-06
CBL-4230: Analysis Warnings (10/13) Networking/HTTP + Networking/WebSockets

callumbirks:
The only major change is improving `WebSocketProtocol.hh` by replacing `char` with `std::byte`, and a few other type changes in the file. The underlying type of `char` and `std::byte` is essentially the same, but `std::byte` is both more type-safe and also a lot clearer on what it's representing.


Networking/WebSockets/WebSocketProtocol.hh
```diff
@@ -337,25 +332,28 @@ namespace uWS {
...
-                char* start = dst + headerLength;
-                char* stop  = start + length;
-                int   i     = 0;
-                while ( start != stop ) { (*start++) ^= mask[i++ % 4]; }
+                uint32_t mask_i32 = *reinterpret_cast<uint32_t*>(mask);
+
+                auto*     start = reinterpret_cast<uint32_t*>(dst + headerLength);
+                uint32_t* end   = start + length / 4;
+
+                while ( start != end ) { *start++ ^= mask_i32; }
```
pasin, line 340: Additional logic? Is this needed?

Networking/WebSockets/WebSocketProtocol.hh
```diff
@@ -303,28 +298,28 @@ namespace uWS {
...
             }
 
             int flags = 0;
-            dst[0]    = char((flags & SND_NO_FIN ? 0 : 128) | (compressed ? SND_COMPRESSED : 0));
-            if ( !(flags & SND_CONTINUATION) ) { dst[0] |= opCode; }
+            dst[0]    = (std::byte)((flags & SND_NO_FIN ? 0 : 128) | (compressed ? SND_COMPRESSED : 0));
+            if ( !(flags & SND_CONTINUATION) ) { dst[0] |= (std::byte)opCode; }
 
-            char mask[4];
+            std::byte mask[4];
```
borrrden, line 322: This looks weird before, and still weird now....does anything actually set these values or are they just...garbage?

Networking/WebSockets/WebSocketProtocol.hh
```diff
@@ -303,28 +298,28 @@ namespace uWS {
             return 0;
         }
 
-        static inline size_t formatMessage(char* dst, const char* src, size_t length, OpCode opCode,
+        static inline size_t formatMessage(std::byte* dst, const char* src, size_t length, OpCode opCode,
```
borrrden, line 301: If you are changing to std::byte, why is src still const char*?

Networking/WebSockets/WebSocketProtocol.hh
borrrden: It seems to be addressing the comment "This is not optimal" by making it a bit more optimal.

borrrden (approved):

jianminzhao (approved):

PR #1806, opened 2023-06-07
XCode 14.3 Error : variable 't' set but not used in mbedtls's bignum.c

callumbirks:
Updated mbedtls commit


callumbirks:
What does the blackduck error mean? I still don't understand what that actually is

pasin (approved):

PR #1809, opened 2023-06-09
CBL-4546: Mix code for unit tests into the main source

jianminzhao:
For Xcode, We added a new build configuration, Test_EE. This configuration is based on Debug_EE, and the only thing that is added to Test_EE is a preprocessor DEFINE, LITECORE_CPPTEST.

For cmake, we added a new target, LiteCoreObjectsCppTest. This differs from LiteCoreObjects only by addintional DEFINE, LITECORE_CPPTES.


CMakeLists.txt
```diff
@@ -196,12 +196,22 @@ target_compile_definitions(
...
     LITECORE_IMPL
     LITECORE_CPP_API=1
     HAS_UNCAUGHT_EXCEPTIONS # date.h use std::uncaught_exceptions instead of std::uncaught_exception
 )
+target_compile_definitions(
+    LiteCoreObjects PRIVATE
+    ${LiteCoreObjectsDefines}
+)
+target_compile_definitions(
+    LiteCoreObjectsCppTest PRIVATE
```
borrrden, line 210: This is quite a redundant way to do this.  There is no need to create an object project here.  The reason it exists for LiteCore is so that the files can be built once and then compiled into both a static and shared library.  I think the proper approach here is to add the define to CppTests, and then add the source files directly (ALL_SRC_FILES) alongside the other test source files

borrrden (changes requested):

CMakeLists.txt
jianminzhao: Are you saying, CppTests won't link to LiteCoreStatic lib? Instead, it will link all object files built from the same sources of LiteCoreObjects?

borrrden: No, it will just include the same sources in its own source list.  That's what it is doing now with this PR, but with the extra step of adding in a new target called LiteCoreObjectsCppTest to prebuild them.

jianminzhao: I tried the following: added all sources of LiteCoreObjects to target CppTests, appended Defines, libraries of LiteCoreObjects to CppTests. I am not sure if this is what you meant but this does not work. The reason is that, as we build LiteCoreObjects for LiteCore target, we use LITECORE_CPP_API=1 but not use as we build test executable that links to the LiteCore lib. 
My feeling is having a separate LiteCoreObjectsCppTest is easier maintain the parity between the LiteCore lib used by the platforms and the lib used by CppTests.

borrrden (approved):

PR #1810, opened 2023-06-09
CBL-4231: Analysis Warnings (11/13) Replicator/

callumbirks:
No major changes.


Networking/WebSockets/WebSocketProtocol.hh
```diff
@@ -337,7 +337,13 @@ namespace uWS {
...
                 uint32_t* end   = start + length / 4;
 
-                while ( start != end ) { *start++ ^= mask_i32; }
+                while ( start != end ) {
+                    uint32_t chunk;
+                    std::memcpy(&chunk, start, sizeof(chunk));
+                    chunk ^= mask_i32;
+                    std::memcpy(start, &chunk, sizeof(chunk));
+                    start++;
+                }
```
jianminzhao, line 347: Didn't it cause Xcode warning?

C/Cpp_include/c4Replicator.hh
```diff
@@ -34,15 +34,15 @@ struct C4Replicator
...
+    bool         retry() const;
 
     virtual void stopCallbacks() noexcept = 0;
 
     virtual void setHostReachable(bool) noexcept {}
 
     virtual void setSuspended(bool) noexcept = 0;
 
-    void         setOptions(slice optionsDictFleece);
+    void         setOptions(slice optionsDictFleece) const;
```
jianminzhao, line 45: Weird, why is it constant set method?

Networking/WebSockets/WebSocketProtocol.hh
callumbirks: If you mean the run-time memory warning which we discussed - this fixes that. Per Apple's documentation, `std::memcpy` automatically resolves the memory alignment issues which can be hit by plain assignment, so this is their recommended solution.

jianminzhao: I see.

C/Cpp_include/c4Replicator.hh
callumbirks: I would guess because the method doesn't modify anything that's part of C4Replicator, instead it converts `this` to `C4ReplicatorImpl` (which inherits from `C4Replicator`) and modifies a variable of that.

jianminzhao: Logically, it's a set-method. Why do you want to change it to const method, witch actually would change the underlying replicator?

callumbirks: This was just as recommended by Clang-Tidy. To the compiler, this is a const method, because of the weird way we are handling the Replicator / ReplicatorImpl. I agree though, logically it makes more sense of non-const. I can change it.

jianminzhao: I guess it's because of this, 
static C4ReplicatorImpl* asInternal(const C4Replicator* repl) { return (C4ReplicatorImpl*)repl; }
which turned a const pointer to non-const pointer. Perhaps it should return const and add another overload
static C4ReplicatorImpl* asInternalC4Replicator* repl) { return (C4ReplicatorImpl*)repl; }

callumbirks: I don't know, I don't want to change too much as I don't understand that too well. For now, I've reverted the `const` of this method.

jianminzhao (approved):

Networking/WebSockets/WebSocketProtocol.hh
borrrden: I cannot believe someone would unironically suggest this as a solution....adding two more memory copies?!

Replicator/Replicator.cc
```diff
@@ -428,7 +429,7 @@ namespace litecore { namespace repl {
         Worker::onError(error);
         for ( const StoppingErrorEntry& stoppingErr : StoppingErrors ) {
             if ( stoppingErr.err == error ) {
-                string message = error.description().c_str();
+                string message = error.description();
```
borrrden, line 432: Wow, good catch by clang!

Replicator/Replicator.cc
```diff
@@ -695,14 +695,14 @@ namespace litecore { namespace repl {
         if ( _connectionState != Connection::kConnected )
             return;  // Not ready yet; Will be called again from _onConnect.
 
-        for ( int i = 0; i < _subRepls.size(); i++ ) {
-            if ( !_subRepls[i].remoteCheckpointDocID )
-                _subRepls[i].remoteCheckpointDocID = _subRepls[i].checkpointer->initialCheckpointID();
+        for ( auto& _subRepl : _subRepls ) {
```
borrrden, line 698: `const auto&` ?  If the checkpoint ID methods are not const, though, then nevermind.

borrrden (changes requested):
I must protest this memcpy change....if there is a problem with unaligned access, then perhaps you can check the memory address first.  If it's a multiple of 4 then you can proceed with the non memcpy version (and in general I think this will most likely be the case).  I also think this qualifies as a significant change (vs the notes saying no major changes)

Replicator/Replicator.cc
callumbirks: It can't be because of the assignment on line 700

Networking/WebSockets/WebSocketProtocol.hh
```diff
@@ -332,23 +333,29 @@ namespace uWS {
...
             if ( !isServer ) {
-                uint32_t mask_i32 = *reinterpret_cast<uint32_t*>(mask);
+                std::byte* start_byte = dst + headerLength;
+                std::byte* end_byte   = start_byte + length;
+                size_t     i          = 0;
 
-                auto*     start = reinterpret_cast<uint32_t*>(dst + headerLength);
-                uint32_t* end   = start + length / 4;
+                // Handle first x amount of bytes individually until we are aligned with the 4-byte alignment
+                for ( ; reinterpret_cast<uintptr_t>(start_byte) % 4 != 0 && start_byte != end_byte; ++i ) {
```
borrrden, line 341: Small improvement here:  The standard way to check for alignment rather than `% 4` is `& 3`.

Networking/WebSockets/WebSocketProtocol.hh
```diff
@@ -332,23 +333,29 @@ namespace uWS {
             memcpy(dst + headerLength, src, length);
 
             if ( !isServer ) {
-                uint32_t mask_i32 = *reinterpret_cast<uint32_t*>(mask);
+                std::byte* start_byte = dst + headerLength;
+                std::byte* end_byte   = start_byte + length;
+                size_t     i          = 0;
```
borrrden, line 338: In cases like these where a loop variable continues to be used outside of its scope it's better to pick a name that explains what it is used for.  So instead of `i` something like `start_offset` would be better.

Networking/WebSockets/WebSocketProtocol.hh
```diff
@@ -332,23 +333,29 @@ namespace uWS {
...
+                for ( ; reinterpret_cast<uintptr_t>(start_byte) % 4 != 0 && start_byte != end_byte; ++i ) {
+                    *start_byte++ ^= mask[i % 4];
+                }
 
+                // Process majority of bytes in chunks of 4
+                auto*     start = reinterpret_cast<uint32_t*>(start_byte);
+                uint32_t* end   = start + ((end_byte - start_byte) / 4);
+                // Construct mask_i32 such that it's a rotation of 'mask' following on from the first loop
+                uint32_t mask_i32 = 0;
+                for ( int j = 0; j < 4; ++j ) { mask_i32 |= (static_cast<uint32_t>(mask[(i + j) % 4]) << (8 * j)); }
```
borrrden, line 350: This concerns me slightly because of the assumed endianness here, but if the tests pass that is probably good enough since I think all of our platforms have the same endianness.  Was this a consideration here?

borrrden (changes requested):

Networking/WebSockets/WebSocketProtocol.hh
callumbirks: Something I considered, but have no idea how to handle, so I'm glad you brought it up. Is there a good standard way we have to handle cases like this?

Networking/WebSockets/WebSocketProtocol.hh
```diff
@@ -319,11 +321,14 @@ namespace uWS {
...
             if ( !(flags & SND_CONTINUATION) ) { dst[0] |= (std::byte)opCode; }
 
-            std::byte mask[4];
+            std::array<std::byte, 4> mask{};
+
             if ( !isServer ) {
                 ((uint8_t*)dst)[1] |= 0x80;
-                uint32_t random = litecore::RandomNumber();
-                memcpy(mask, &random, 4);
+                uint32_t  random = litecore::RandomNumber();
```
jianminzhao, line 328: What's the motivation of this? To avoid endianness?

Networking/WebSockets/WebSocketProtocol.hh
```diff
@@ -332,28 +337,28 @@ namespace uWS {
...
 
-                auto*     start = reinterpret_cast<uint32_t*>(dst + headerLength);
-                uint32_t* end   = start + length / 4;
+                // Handle first x amount of bytes individually until we are aligned with the 4-byte alignment
+                for ( ; (reinterpret_cast<uintptr_t>(start_byte) & 3) != 0 && start_byte != end_byte; ++offset ) {
+                    *start_byte++ ^= mask[offset % 4];
+                }
 
+                // Rotate mask by the offset we reached in the first loop to become memory-aligned
+                std::rotate(mask.begin(), mask.begin() + offset, mask.end());
```
jianminzhao, line 350: Is offset guaranteed to be < 4?

Networking/WebSockets/WebSocketProtocol.hh
```diff
@@ -332,28 +337,28 @@ namespace uWS {
...
+
+                // Process majority of bytes in chunks of 4 (until end or < 4 bytes away from end)
+                auto*     start = reinterpret_cast<uint32_t*>(start_byte);
+                uint32_t* end   = start + ((end_byte - start_byte) / 4);
+                uint32_t  mask_i32{};
+                std::memcpy(&mask_i32, mask.data(), 4);
                 while ( start != end ) { *start++ ^= mask_i32; }
```
jianminzhao, line 357: It's so complicated. mask has been adjusted for endianness, etc, to reduce the time by factor of 4. How to get it fully tested?

Networking/WebSockets/WebSocketProtocol.hh
borrrden: With a 4-byte alignment, the maximum mathematically possible offset is 3.  If it is offset by 4, it is once again aligned.

Networking/WebSockets/WebSocketProtocol.hh
borrrden: This seems like a side effect of using std::rotate.  This function can only operate on std containers, which means std::array in this case.  Though I'd prefer the use of a slice, and `SecureRandomize` here rather than a memcpy.

borrrden (changes requested):
I believe this would be a cleaner way to accomplish the random mask:

```
std::array<std::byte, 4> mask{};
fleece::mutable_slice maskSlice(mask.data(), 4);
litecore::SecureRandomize(maskSlice);
```

Could you give this a try?

Networking/WebSockets/WebSocketProtocol.hh
callumbirks: WebSocket has built-in "testing", which I discovered during the painful process of getting all this working. If the data received through WebSocket isn't perfect, it will give some warning. A couple examples were "Frame too small" or "Invalid checksum", and there were a few more. I think the main method I used for testing was "P2P Sync" test.

Networking/WebSockets/WebSocketProtocol.hh
jianminzhao: I referred to "offset". It advances with *start_byte++ . Should it modulo 4, %= 4 inside std::rotate?

callumbirks: It will never be greater than 3, because of the conditions it is set under. It is incremented until start_byte % 4 == 0 (equivalent to start_byte & 3), so it will be 0, 1, 2 or 3.

jianminzhao: Okay, so, '%' in "mask[offset % 4]" is not necessary.

callumbirks: Yeah, you're right. Think it was just leftover from copying the final loop

borrrden (approved):

PR #1815, opened 2023-06-17
CBL-4623: Use FTS match() in the WHERE clause of LEFT OUTER JOINS Not…

jianminzhao:
… Returning Correct Result

When doing self join, we need to remember the database alias assolicated with the use of FTS table.

### DIFF
--- a/LiteCore/Query/QueryParser.cc
+++ b/LiteCore/Query/QueryParser.cc
@@ -2048,7 +2048,7 @@ namespace {
         });
     }
 
-    // Returns the FTS table name given the LHS of a MATCH expression.
+    // Returns the pair of the FTS table name and database alias given the LHS of a MATCH expression.
     pair<string, string> QueryParser::FTSTableName(const Value *key) const {
         Path keyPath(requiredString(key, "left-hand side of MATCH expression"));
         // Path to FTS table has at most two components: [collectionAlias .] IndexName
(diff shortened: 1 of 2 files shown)


jianminzhao:
This has been tested with cblite and I checked the result with Pasin.

jianminzhao:
I will add unit teat when porting it to the master branch.

LiteCore/Query/QueryParser.hh
```diff
@@ -257,6 +257,7 @@ namespace litecore {
         map<string, string> _indexJoinTables;    // index table name --> alias
         set<string> _kvTables;                   // Collection tables referenced in this query
         vector<string> _ftsTables;               // FTS virtual tables being used
+        std::multimap<string, string> _ftsTableAliases;
```
pasin, line 260: Can you add a description comment about this map?

LiteCore/Query/QueryParser.cc
```diff
@@ -2042,7 +2049,7 @@ namespace {
     }
 
     // Returns the FTS table name given the LHS of a MATCH expression.
-    string QueryParser::FTSTableName(const Value *key) const {
+    pair<string, string> QueryParser::FTSTableName(const Value *key) const {
```
pasin, line 2052: The comment description about needs to update as the return type is changed?

pasin (changes requested):
I'm not familiar with this code but I think the fix would work. Just a couple suggestions to provide comment in the code.

borrrden (approved):

pasin (approved):

LiteCore/Query/QueryParser.cc
jianminzhao: done

bmeike (commented):
✅

PR #1818, opened 2023-06-20
CBL-4625: Set macOS target version to 10.14

jianminzhao:
CBL-4631: Update iOS target version to 11.0
And updated Fleece's target versions


pasin (approved):

bmeike (commented):
✅

PR #1819, opened 2023-06-21
Fixing a test failure on Win64.

jianminzhao:


jianminzhao:
I verified the fix in a toy build, 3.2.0-50705

borrrden (approved):
This is basically just a reversal of a previous change with no indication of how it will go forward.  @callumbirks Please take note of this and work out any conflicts with your current branch (and try to diagnose why this is happening in official builds...)

PR #1821, opened 2023-06-23
CBL-4624: Mark return types nullable to avoid Clang UBSan warnings

snej:
I hit a couple of UBSan `_Nonnull binding to null pointer` warnings while running the unit tests with Xcode 14.3. They all stem from API functions that can return NULL but whose return types are not marked `C4NULLABLE`. I've fixed these and other nearby ones I found.

### DIFF
--- a/C/c4CAPI.cc
+++ b/C/c4CAPI.cc
@@ -204,7 +204,7 @@ bool c4db_hasScope(C4Database* db, C4String name) noexcept {
 
 C4Collection* C4NULLABLE c4db_getCollection(C4Database* db, C4CollectionSpec spec,
                                             C4Error* C4NULLABLE outError) noexcept {
-    return tryCatch<C4Collection*>(outError, [&] { return db->getCollection(spec); });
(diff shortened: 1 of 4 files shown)


borrrden:
This failed our new code style validation job, unfortunately.  You should be able to use clang tidy here to figure out what it wants.  Especially since this changes C header formatting I'm worried it will break the .NET generation again.

borrrden (changes requested):
Let's address the linting job failure

snej:
Other than inserting some whitespace, the only header changes are adding the `C4NULLABLE` token after some return types. That token already exists on some return types, so I don't think this will cause trouble for .NET.

borrrden:
I looked at it and I don't think so either.  API is handled in a more resilient way than struct analysis, which is the pain I was remembering (getting two variables on the same line in a struct or enum, etc)

callumbirks:
I've already filed a ticket for this in Jira, [CBL-4624](https://issues.couchbase.com/browse/CBL-4624), so I guess this can close that.

borrrden (approved):

PR #1823, opened 2023-06-26
CBL-4232: Analysis Warnings (12/13) Replicator/tests

callumbirks:
No major changes.
Removed one warning from the `.clang-tidy` configuration file ([bugprone-lambda-function-name](https://clang.llvm.org/extra/clang-tidy/checks/bugprone/lambda-function-name.html)) because it results in a lot of warnings (from any call to the `Assert()` macro), but this is only coming from tests, and I don't think it's too concerning (we can still debug just as easily from the messages output by the `Assert` function).


jianminzhao:
Does our current `Assert` suffer from [bugprone-lambda-function-name](https://clang.llvm.org/extra/clang-tidy/checks/bugprone/lambda-function-name.html)? That is, it won't give correct place where Assert takes place?

callumbirks:
I believe that can happen, where it prints `operator()` rather than the function name, not sure in what circumstances. But our `Assert` function also prints file name and line number, so it's not a huge problem if it does occur.

Replicator/tests/CookieStoreTest.cc
```diff
@@ -183,9 +183,9 @@ TEST_CASE("Cookie Parser Failure", "[cookies]") {
             "name=value; Max-Age=z7",
             "name=value; Max-Age=",
     };
-    for ( int i = 0; i < sizeof(badCookies) / sizeof(badCookies[0]); ++i ) {
-        INFO("Checking " << badCookies[i]);
-        Cookie c(badCookies[i], "example.com", "/");
+    for ( auto& badCookie : badCookies ) {
```
jianminzhao, line 186: consider using "const auto&"

Replicator/tests/ReplParams.cc
```diff
@@ -48,7 +48,7 @@ void ReplParams::addCollections(const std::vector<C4ReplicationCollection>& coll
 
 AllocedDict ReplParams::setOptions(const AllocedDict& params, const AllocedDict& options) {
     AllocedDict result = params;
-    for ( Dict::iterator i(options); i; ++i ) {
+    for ( Dict::iterator i(options.asDict()); i; ++i ) {
         result = repl::Options::updateProperties(result, i.keyString(), i.value());
```
jianminzhao, line 52: What's the incentive of this? `i(options)` is very popular idiom in the code base.

jianminzhao (approved):

Replicator/tests/ReplicatorAPITest.cc
```diff
@@ -618,7 +614,6 @@ TEST_CASE_METHOD(ReplicatorAPITest, "Pending Document IDs Non-Existent Collectio
 
     c4repl_start(_repl, false);
     REQUIRE_BEFORE(5s, c4repl_getStatus(_repl).level == kC4Stopped);
-    encodedDocIDs = c4repl_getPendingDocIDs(_repl, Republic, &err);
```
borrrden, line 621: How is this still going to pass?  The next line is checking the result of this error (perhaps it should also be checking the result of `encodedDocIDs`?)

Replicator/tests/ReplicatorCollectionSGTest.cc
```diff
@@ -915,7 +915,7 @@ TEST_CASE_METHOD(ReplicatorCollectionSGTest, "Auto Purge Enabled - w/ and w/o Fi
             pullFilterPurge = 0;
         }
 
-        unsigned collIndex;
+        unsigned collIndex = -1;
```
borrrden, line 918: Kind of interesting that this wouldn't set of some kind of signed/unsigned mismatch...

borrrden (changes requested):

Replicator/tests/ReplParams.cc
callumbirks: `i(options)` causes object slicing because the `Dict::iterator()` expects a `Dict`, but here we are passing `AllocedDict`

borrrden (approved):

Replicator/tests/ReplParams.cc
callumbirks: The issue is `Dict::iterator i(options)` which takes an `AllocedDict` as a `Dict`, this causes object slicing because the additional state that `AllocedDict` holds is implicitly discarded.

PR #1824, opened 2023-06-26
CBL-4253: SQL++, Allow to refer to a property with full collection name

jianminzhao:
Extend SQL++ to allow,
    SELECT store.customers.name FROM store.customers
    SELECT store.customers.name, store2.customers.name FROM store.customers JOIN store2.customers ON store.customers.name = store2.customers.name
in addition to currently legit statements:
    SELECT name FROM store.customers
    SELECT customers.name FROM store.customers

With this commit, the database alias can be backquoted.

### DIFF
--- a/C/tests/c4QueryTest.cc
+++ b/C/tests/c4QueryTest.cc
@@ -1180,7 +1180,7 @@ TEST_CASE_METHOD(CollectionTest, "C4Query collections", "[Query][C]") {
     compileSelect(json5("{WHAT: ['.'], FROM: [{COLLECTION:'Widgets'}]}"));
     checkColumnTitles({"Widgets"});
     compileSelect(json5("{WHAT: ['.'], FROM: [{COLLECTION:'nested', SCOPE: 'small'}]}"));
-    checkColumnTitles({"nested"});
+    checkColumnTitles({"small.nested"});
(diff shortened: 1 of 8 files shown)


jianminzhao:
[Notes] The objective is to allow to select property by collection path. Currently, can say,
`SELECT collection.prop FROM scope.collection` -- (1)
The SQL++ parser will turn it to something like, 
`SELECT collection.prop FROM kv_.scope.collection AS collection`. -- (2)
That is, we use "collection" as a db alias. With this approach, we cannot parse,
`SELECT scope.collection.prop FROM scope.collection`.   -- (3)
because, currently the lexer only admit simple identity as an alias. Our N1QL grammar actually allows quoting -- for example -- and we say 
`SELECT collection.prop FROM scope.collection AS \`scope.collection\`` -- (4)
however, the lexer is not handling the quoting properly. This commit attempts to handle it property from lexer through the QueryParser. After it becomes formally supported, we will parse (1) into,
`SELECT collection.prop FROM kv_.scope.collection AS \`scope.collection\`` -- (5)
and (3), into
`SELECT scope.collection.prop FROM kv._scope.collection AS \`scope.collection\`` -- (6)
That is, if provided in the N1QL statement, we will always use (properly quoted) collection path as the database alias.

borrrden (approved):

pasin (approved):

PR #1825, opened 2023-06-27
CBL-4600: Doc update c4repl_start

jianminzhao:
Added the following
```
\note This function may perform action in the calling thread that requires the transaction from the database
              that was passed to the replicator instance, \p repl.
```
I merely note the fact that is critical to the user I assume that general prerequisite knowledge in using the library that they cannot open a transaction inside a transaction.

### DIFF
--- a/C/include/c4Replicator.h
+++ b/C/include/c4Replicator.h
@@ -77,6 +77,9 @@ CBL_CORE_API C4Replicator* c4repl_newWithSocket(C4Database* db, C4Socket* openSo
 
 /** Tells a replicator to start. Ignored if it's not in the Stopped state.
         \note This function is thread-safe.
+        \note This function may perform action in the calling thread that requires the transaction from the database
+              that was passed to the replicator instance, \p repl.
+
         @param repl  The C4Replicator instance.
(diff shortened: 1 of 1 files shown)


borrrden (changes requested):
This is kind of a word salad.  It's hard to follow even if you know the situation, and I think it will be impossible to follow if you don't.  Try to be clear about what to avoid.

velicuvlad:
Maybe something like `If this function is called while a transaction is open on the same database it will deadlock, as it trying to acquire the same database` ... maybe worth specifying that this is considered transaction as well (starting the repl)...?

jianminzhao:
Simplified wording. Thanks to @velicuvlad for the suggestion. I avoided emphasizing the "same thread," because it makes the doc unnecessary complicated and why people want to do transaction in another thread while calling this function. If you think we should limit the "Do not" to the "same thread," I can change it.

borrrden (approved):

pasin (approved):

callumbirks (approved):

velicuvlad (approved):

PR #1828, opened 2023-06-29
CBL-3836: Corrupt Revision Data error when saving documents

jianminzhao:
Added sanity check at the time we encode the rev tree to binary. I am not lifting the limit of the number revs currently in place. With this check, the "error::CorruptRevisionData" should indicate real corruption.

### DIFF
--- a/LiteCore/RevTrees/RawRevTree.cc
+++ b/LiteCore/RevTrees/RawRevTree.cc
@@ -132,11 +132,12 @@ namespace litecore {
 
         Assert(entry == (const void*)result.end());
         // Sanity check:
-        auto rawRev = (const RawRevision*)result.buf;
-        unsigned count = rawRev->count();
+        auto     rawRev = (const RawRevision*)result.buf;
+        unsigned count  = rawRev->count();
         // c.f. RawRevision::decodeTree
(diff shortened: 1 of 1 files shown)


borrrden:
How did you come to the conclusion that this is the cause of the problem?

jianminzhao:
I don't have direct evidence of corruption, but would like to note, A. without the assertion that the encode won't put the amount of revs within the limit, we cannot conclude, as error message asserts as corruption. B. the original customer returned our query (via Priya) that "My guess is that this number is out of bounds because revisions were never pruned without [one of the fixes](https://github.com/couchbase/couchbase-lite-core/commit/7cb516b9f7c030ba8851178b71bcae93fafa38f3)included in the coming release. Inspecting the database shows that the document carries 2202890 bytes in its "extra" column and has the sequence number 65762."
With this commit, 1. we will error out before it is encode, and 2. if still get the corruption error, it is more likely a problem of corruption.

borrrden (approved):
Ok, but this is going to close the ticket when the resolution is unsure.  I hope there is a way to not lose track of it if it happens again.

velicuvlad (approved):

PR #1832, opened 2023-07-05
CBL-4638: Decoded timestamps appear incorrect

jianminzhao:
1. There is no change in the binary loggers.
2. In the header of the decoded log file, we converted the timestamp into the UTC time, instead of the local time, to make it independent of where the binary is decoded.
3. Fixed mistaken applicatino of the time zone offset when we decode the timestamp into the local time for each line of the log.
4. Informally, we insert the timestamp to the name of log file. Meanwhile, as noted by above (2), the header of the log file includes the timestamp. We added test to ensure they are consistent.

### DIFF
--- a/LiteCore/tests/LogEncoderTest.cc
+++ b/LiteCore/tests/LogEncoderTest.cc
@@ -95,8 +95,7 @@ TEST_CASE("LogEncoder formatting", "[Log]") {
     // We insert timestamp in milliseconds (w.r.t. UTC) in the path to the binary log files.
     // We also add the timestamp inside the log. When decoded to string, it is
     // represented as UTC time, like, "Monday 2023-07-03T19:25:01Z"
-    // We want to ensure they are consistent. The timestamp prepended to each line is in the local time
-    // where the binary is decoded.
+    // We want to ensure they are consistent.
     regex  catchUTCTimeTag{"^" TIMESTAMP "---- Logging begins on (" DATESTAMP ")"};
     smatch m;
     REQUIRE(regex_search(result, m, catchUTCTimeTag));
--- a/LiteCore/tests/LogEncoderTest.cc
+++ b/LiteCore/tests/LogEncoderTest.cc
@@ -15,6 +15,7 @@
 #include "LogDecoder.hh"
 #include "LiteCoreTest.hh"
 #include "StringUtil.hh"
+#include "ParseDate.hh"
 #include "fleece/PlatformCompat.hh"
 #include <regex>
 #include <sstream>
@@ -22,8 +23,9 @@
 
 using namespace std;
 
-#define DATESTAMP "\\w+, \\d{2}/\\d{2}/\\d{2}"
-#define TIMESTAMP "\\d{2}:\\d{2}:\\d{2}\\.\\d{6}\\| "
+#define DATESTAMP     "\\w+, \\d{2}/\\d{2}/\\d{2}"
+#define DATESTAMP_UTC "\\w+ \\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}Z"
+#define TIMESTAMP     "\\d{2}:\\d{2}:\\d{2}\\.\\d{6}\\| "
 
 constexpr size_t kFolderBufSize = 64;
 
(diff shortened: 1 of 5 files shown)


LiteCore/tests/LogEncoderTest.cc
```diff
@@ -77,13 +83,42 @@ TEST_CASE("LogEncoder formatting", "[Log]") {
...
             "String is 'C string', slice is 'hello' \\(hex 68656c6c6f\\)\\n");
     CHECK(regex_match(result, expected));
+
+#ifdef LITECORE_CPPTEST
+    // We insert timestamp in milliseconds (w.r.t. UTC) in the path to the binary log files.
+    // We also add the timestamp inside the log. When decoded to string, it is
+    // represented as UTC time, like, "Monday 2023-07-03T19:25:01Z"
+    // We want to ensure they are consistent. The timestamp prepended to each line is in the local time
+    // where the binary is decoded.
+    regex  catchUTCTimeTag{"^" TIMESTAMP "---- Logging begins on (" DATESTAMP_UTC ")"};
```
borrrden, line 100: Why do we need to log both times?  If you want to use UTC why not just use UTC?  Otherwise things are going to get confusing when the timestamps of each line don't match well with the "begin"

borrrden (changes requested):

LiteCore/tests/LogEncoderTest.cc
jianminzhao: I thought we preferred local time but felt we'd better to have at least one invariant time in the decoded file to ensure some certainty. We kind of took pain to display the local time throughout the logs. Since per-line timestamps also show in the console, probably local time appears more human-friendly. Actually, the UTC time only show in log content. All lines are consistently prefixed with local time, like this,
`20:49:23.000000| ---- Logging begins on Friday 2023-07-07T03:49:23Z ----`
Which is the extra line logged by the logger.

borrrden: If the logs are truly showing the time in the timezone where the logs are decoded, then I feel it is quite useless.  We might as well universally use UTC.  In this case, it says logging beginning on Friday the 7th, but is the local time actually Friday the 7th?  It requires useless thought power to figure out the answer to that question.  In this case, it is either 7 hours behind or 17 hours ahead so it must be that it is actually Thursday the 6th.  What benefit is there to this mismatch?  I'd say absolutely nothing.

jianminzhao: What do you think about the time showed in the console? Should we use the UTC time only in the decoded file, and local time on the console?

borrrden: I forget, do console logs actually make it back to the platform console?  Or are they just showing up in the LiteCore tests?

jianminzhao: Good question. I have to check it out. It appears that timestamp is prepended by the callback. That is, if the platform does not supply the callback, the default callback in the Core is responsible. For Android, if not provided by the platform, we will call ndk's __android_log_vprint.

jianminzhao: Changed all timestamps in the decoded file with UTC time,
`18:46:56.000000Z| ---- Logging begins on Monday 2023-07-10T18:46:56Z ----`
`18:46:56.924683Z| Unsigned 1234567890, Long 2345678901, LongLong 123456789123456789, Size abcdabcd, Pointer 0x7fff5fbc`
`18:46:56.924729Z| Int -1234567890, Long -234567890, LongLong -123456789123456789, Size -1234567890, Char @`
`18:46:56.924750Z| Int 1234567890, Long 234567890, LongLong 123456789123456789, Size 1234567890, Char @`
`18:46:56.924762Z| String is 'C string', slice is 'hello' (hex 68656c6c6f)`
And on the console. we still use the local time, 
`12:42:07.490396| WARNING: Invalid LiteCore query: Unknown function 'poop'`
`12:42:07.523654| WARNING: Invalid LiteCore query: Too few arguments for function 'power'`
I am in time zone of UTC-7.

borrrden (approved):

PR #1835, opened 2023-07-10
CBL-4530: Remove actor::Observer and actor::Property

callumbirks:
These classes/files are unused. Spoke with Jens and he is okay with the removal of these files.


callumbirks:
@jianminzhao Thanks :) '3' and '5' the wrong way round

borrrden (approved):

PR #1836, opened 2023-07-11
CBL-4704: Add `next/*` branches to GH Actions workflows

callumbirks:
I've also added the same branch protection rules that `release/*` branches have to the `next/*` branches. (Requires pull request, requires approval, cannot be bypassed)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,6 +14,7 @@ on:
       - master
       - dev
       - 'release/*'
+      - 'next/*'
   pull_request:
     branches:
       - '**'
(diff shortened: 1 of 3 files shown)


jianminzhao (approved):

borrrden:
Will this then mean that we will need an extra redundant review when merging next into release?

borrrden (commented):
Where is the corresponding EE PR?  This is going to mess with the checkout process since the requirement is for a matching base branch to exist in EE when making a PR.

jianminzhao:
We decided not to use this workflow.

PR #1837, opened 2023-07-11
CBL-4687: Add Test_CE to Xcode configuration for unit tests

jianminzhao:
GitHub action, "Xcode Build," builds scheme "LiteCore C++ Tests." We will build it with this new config, Test_CE, which is duplicated from Debug but has the c++ preprocessor, LITECORE_CPPTEST, defined.


borrrden (approved):
What happened to the one I already approved?

jianminzhao:
That one is in fleece. When it builds LiteCore with a configuration, Test_CE, in particular, Xcode will build with the same config. Otherwise, I don't know exact rule, it won't get DEBUG, etc, that the Debug config includes. This will cause runtime failure since LiteCore is basically in Debug config. In fleece, Test_CE is exact copy of Debug.

PR #1838, opened 2023-07-12
CBL-127: Conflict error in CBL P2P push replication when using differ…

jianminzhao:
…ent target database UID

The problem has been taken care of by the new property, "conflictIncludesRev", introduced by the resolution of CBL 2637, to the BLIP message "proposeChanges". We added test case to verify it.

### DIFF
--- a/Replicator/tests/ReplicatorLoopbackTest.cc
+++ b/Replicator/tests/ReplicatorLoopbackTest.cc
@@ -1928,6 +1928,7 @@ TEST_CASE_METHOD(ReplicatorLoopbackTest, "Replicate Encrypted Properties", "[Pus
         CHECK(clear == "\"123-45-6789\"");
     }
 }
+#endif  // COUCHBASE_ENTERPRISE
 
 TEST_CASE_METHOD(ReplicatorLoopbackTest, "Replication Collections Must Match", "[Push][Pull][Sync]") {
     Options opts       = GENERATE_COPY(Options::pushing(kC4OneShot, _collSpec), Options::pulling(kC4OneShot, _collSpec),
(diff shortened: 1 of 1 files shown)


callumbirks (approved):

jianminzhao:
I originally tried to verify that the test hit the targeted code path by the following means,
    {
        catchLog::savedCB       = c4log_getCallback();
        C4LogLevel savedCBLevel = c4log_callbackLevel();
        catchLog::logs.reset(new vector<string>);
        c4log_writeToCallback(kC4LogInfo, catchLog::logCB, true);

        DEFER {
            catchLog::logs.reset(nullptr);
            c4log_writeToCallback(savedCBLevel, catchLog::savedCB, false);
            catchLog::savedCB = nullptr;
        };

        ......

        std::regex regex{R"(Rejecting proposed change .* \(status 409; current rev is .*\))"};
        bool       found409 = false;
        for ( auto& s : *catchLog::logs.get() ) {
            std::smatch sm;
            if ( std::regex_search(s, sm, regex) ) { found409 = true; }
        }
        if ( toDifferentTarget ) {
            // Pusher receives 409 error with different target. It then send a new "proposedChange" as
            // it recognizes the revision that the error response includes and ends up with the same
            // result.
            CHECK(found409);
        } else {
            CHECK_FALSE(found409);
        }
    }
but Windows does not like it.

borrrden (approved):
I'm assuming you observed this new message happen, or saw this test fail in an older commit.

jianminzhao:
yes, saw it failed a couple times before I decided Windows just did not like it. My request of VS 17 2022 is still pending at Zendesk, and I cannot try it on my local machine.

borrrden:
I meant the new BLIP message, not the error message.

jianminzhao:
For the blip message, I observed in Xcode by manually turning the BLIPMessage log to Debug

PR #1841, opened 2023-07-17
CBL-4331: Remove C4QueryOptions parameter from c4query_run

callumbirks:
⚠️ API has changed for platforms.
Behaviour does not change, as the parameter is not used internally.
All platforms are only using default parameter, so is an easy change for the platforms.

### DIFF
--- a/C/c4CAPI.cc
+++ b/C/c4CAPI.cc
@@ -960,8 +960,7 @@ void c4query_setParameters(C4Query* query, C4String encodedParameters) noexcept 
     query->setParameters(encodedParameters);
 }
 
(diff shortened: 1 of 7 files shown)


jianminzhao:
"All platforms are only using default parameter" -- even if they use non-default, it is disregarded, right?

jianminzhao (approved):

callumbirks:
> "All platforms are only using default parameter" -- even if they use non-default, it is disregarded, right?

@jianminzhao Yeah - I point it out just to show it's a pretty easy fix for everyone when we change the API.

borrrden (approved):
"API has changed" is not a good enough message, and it needs to be in the COMMIT message in the final commit.  It should detail what needs to change from the API consumer point of view.  Please do this when squashing, along with the warn symbol.

PR #1843, opened 2023-07-18
CBL-4719: ReplicatorAPITest::options() Set proxy settings from _sg.proxy

callumbirks:
I think I've set the proxy options appropriately, but I don't believe we have any proxy tests? So it might be good to add some alongside this PR, if possible.

### DIFF
--- a/Replicator/tests/ReplicatorAPITest.hh
+++ b/Replicator/tests/ReplicatorAPITest.hh
@@ -142,7 +142,7 @@ class ReplicatorAPITest : public C4Test {
                     proxyType = slice(kC4ProxyTypeHTTP);
                 case ProxyType::HTTPS:
                     proxyType = slice(kC4ProxyTypeHTTPS);
-                // TODO: Once SOCKS is implemented (CBL-4715)
+                // If we ever support SOCKS
                 // case ProxyType::SOCKS:
                 //     proxyType = slice(kC4ProxyTypeSOCKS);
                 default:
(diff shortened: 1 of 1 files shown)


jianminzhao:
Where is this used? Do you put it up based on doc or current use in the code base? Adding a comment will be helpful.

Replicator/tests/ReplicatorAPITest.hh
```diff
@@ -134,7 +134,43 @@ class ReplicatorAPITest : public C4Test {
...
-        // TODO: Set proxy settings from _proxy
+        if ( _sg.proxy ) {
+            enc.writeKey(C4STR(kC4ReplicatorProxyType));
+            slice proxyType;
+            switch ( _sg.proxy->type ) {
+                case ProxyType::HTTP:
+                    proxyType = slice(kC4ProxyTypeHTTP);
+                case ProxyType::HTTPS:
+                    proxyType = slice(kC4ProxyTypeHTTPS);
+                // TODO: Once SOCKS is implemented (CBL-4715)
```
borrrden, line 145: Let's not :D

borrrden (approved):

callumbirks:
> Where is this used? Do you put it up based on doc or current use in the code base? Adding a comment will be helpful.

It's not currently used, as it doesn't seem we have any tests for proxy. But if we do add any proxy tests, they _should_ ™  just work, as this will set that up in `ReplicatorAPITest` constructor based on what's been put in `_sg` (SG connection settings).

pasin (approved):

jianminzhao (approved):

PR #1846, opened 2023-07-22
CBL-4639: Use FTS match() in the WHERE clause of LEFT OUTER JOINS not…

jianminzhao:
… returning correct result.

When doing self join, we need to remember the database alias associated with the use of FTS table.


pasin (approved):
Is this a porting ticket?

PR #1849, opened 2023-07-25
CBL-4751: Putting more logs for error 1002 (protocol error).

jianminzhao:


Networking/WebSockets/WebSocketProtocol.hh
```diff
@@ -413,7 +413,7 @@ public:
                 // invalid reserved bits / invalid opcodes / invalid control frames / set compressed frame
                 if ((rsv1(frame) && !setCompressed(user)) || rsv23(frame) || (getOpCode(frame) > 2 && getOpCode(frame) < 8) ||
                     getOpCode(frame) > 10 || (getOpCode(frame) > 2 && (!isFin(frame) || payloadLength(frame) > 125))) {
-                    forceClose(user);
+                    forceClose(user, "cs-1");
```
pasin, line 416: Can we provide the reason along with the some properties of the frame such as reserved bits, opcodes, control frames, compression, and payload length?

Networking/WebSockets/WebSocketProtocol.hh
```diff
@@ -160,18 +160,18 @@ private:
     inline bool consumeMessage(T payLength, char *&src, unsigned int &length, frameFormat frame, void *user) {
         if (getOpCode(frame)) {
             if (opStack == 1 || (!lastFin && getOpCode(frame) < 2)) {
-                forceClose(user);
+                forceClose(user, "cm-1");
```
pasin, line 163: Can we provide the reason along with the some properties of the frame such as reserved bits, opcodes, control frames, compression, and payload length? This info could be helpful to the customer to diagnostic the problems on their side.

pasin: Also any change to this file should be marked as WebSocketProtocol.hh is from a 3rd party so if we need to update this file, we know what we have made change to the file.

pasin (changes requested):

pasin:
PR validation requires release/3.0.12.1 branch to exist in EE repo.

Networking/WebSockets/WebSocketProtocol.hh
jianminzhao: Ah, this is subtle: an external header file inside our directory. Let me try to organize it a bit.

Networking/WebSockets/WebSocketProtocol.hh
jianminzhao: Good point

Networking/WebSockets/WebSocketProtocol.hh
```diff
@@ -160,18 +167,34 @@ private:
     inline bool consumeMessage(T payLength, char *&src, unsigned int &length, frameFormat frame, void *user) {
         if (getOpCode(frame)) {
             if (opStack == 1 || (!lastFin && getOpCode(frame) < 2)) {
-                forceClose(user, "cm-1");
+#ifdef COUCHBASE_forceClose
+                std::stringstream ss;
+                ss << "[opStack=1,lastFin=" << lastFin << ",frame=" <<frame << "]";
```
pasin, line 172: * output opStack instead of 1 as it may not be one.
* (format) Add a space between << and frame.

Networking/WebSockets/WebSocketProtocol.hh
```diff
@@ -413,7 +436,13 @@ public:
                 // invalid reserved bits / invalid opcodes / invalid control frames / set compressed frame
                 if ((rsv1(frame) && !setCompressed(user)) || rsv23(frame) || (getOpCode(frame) > 2 && getOpCode(frame) < 8) ||
                     getOpCode(frame) > 10 || (getOpCode(frame) > 2 && (!isFin(frame) || payloadLength(frame) > 125))) {
-                    forceClose(user, "cs-1");
+#ifdef COUCHBASE_forceClose
+                    std::stringstream ss;
+                    ss << "[frame=" << frame << "]";
```
pasin, line 441: What will be printed as frame here? I think we should print out the value that we could track back what is failing here.

Networking/WebSockets/WebSocketProtocol.hh
pasin: I'm not sure what will be printed as `frame`, at least `opcode` should be printed.

Networking/WebSockets/WebSocketProtocol.hh
```diff
@@ -160,18 +167,34 @@ private:
...
+#else
+                forceClose(user);
+#endif
                 return true;
             }
             opCode[(unsigned char) ++opStack] = (OpCode) getOpCode(frame);
         } else if (opStack == -1) {
-            forceClose(user, "cm-2");
+#ifdef COUCHBASE_forceClose
+            forceClose(user, "[opStack=-1]");
```
pasin, line 182: Can we print opcode as well?

pasin (changes requested):

Networking/WebSockets/WebSocketProtocol.hh
```diff
@@ -160,18 +167,34 @@ private:
...
+#endif
             return true;
         }
         lastFin = isFin(frame);
 
         if (payLength > SIZE_MAX || refusePayloadLength(user, (int)payLength)) {
-            forceClose(user, "cm-3");
+#ifdef COUCHBASE_forceClose
+            std::stringstream ss;
+            ss << "[payLength=" << payLength << "]";
```
pasin, line 195: Can you print frame as well?

pasin (approved):

PR #1859, opened 2023-08-04
CBL-4506: Replicator starts up slow for big database

jianminzhao:
The underlying cause has been fixed in #711f886899b23b. We added a test case.

### DIFF
--- a/C/tests/c4DatabaseTest.cc
+++ b/C/tests/c4DatabaseTest.cc
@@ -16,8 +16,11 @@
 #include "c4IndexTypes.h"
 #include "c4Query.h"
 #include "c4Collection.h"
+#include "Error.hh"
 #include "FilePath.hh"
 #include "SecureRandomize.hh"
+#include "StringUtil.hh"
+#include "Stopwatch.hh"
 #include <cmath>
 #include <cerrno>
 #include <iostream>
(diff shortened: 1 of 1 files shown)


C/tests/c4DatabaseTest.cc
```diff
@@ -1011,6 +1014,69 @@ static void testOpeningOlderDBFixture(const string& dbPath, C4DatabaseFlags with
...
+    double elapsedUnsorted = doIt(false);
+    double elapsedSorted   = doIt(true);
+    cout << "Enum with conflicted/sorted takes " << elapsedSorted << " sec, Enum With conflicted/unsorted takes "
+         << elapsedUnsorted << "sec, ratio = " << elapsedSorted / elapsedUnsorted << endl;
+    auto config = c4db_getConfig2(db);
+    if ( config->encryptionKey.algorithm != kC4EncryptionNone ) {
+        // Encrypted case is not stable.
+        return;
+    }
+    CHECK(elapsedSorted / elapsedUnsorted > 5);
```
pasin, line 1077: Could this be flaky when running on some machines?

jianminzhao: I picked number, 5, quite conservatively. On my machine, the ratio is always > 10, but I found over the weekend that jenkins' windows build may be smaller. The theory is, the more docs the db has, the higher the ratio, I don't want to put too much time on this test. Perhaps this test does not belong here, but I wanted to set a guard here lest in the future we somehow we make changes to cause the problem reappear.

borrrden: It sounds like a performance test, perhaps it should be run as part of the perf suite for showfast?

borrrden (approved):

C/tests/c4DatabaseTest.cc
jianminzhao: Definitely. I should get familiar with that.

PR #1866, opened 2023-08-11
Fix: Upgrade to VV setting incorrect version(s)

snej:
The code that assigns a version-vector to a doc during DB upgrade was left unfinished in my earlier commits. I saw today it isn't creating valid timestamps, and fixed it:

* The latest replicated revision now becomes a version whose timestamp is the minimum allowable value plus its old generation number. This is a valid timestamp, and provides an easy way to recover the generation number for e.g. display in `cblite`.
* If the current revision has not been replicated, it becomes a new version tagged with the "me" source ID. Its timestamp now uses the current logical clock.

We may refine this further during development, especially once SG comes aboard, but this update at least doesn't create illegal versions with too-small timestamps!


borrrden (approved):
PR validation failure was simply a timeout on the job

jianminzhao (approved):

PR #1868, opened 2023-08-16
CBL-4824: Move fixes for 3.1.2 from next/ to release/

callumbirks:
Just did a `git merge` and manually checked out the right commits in submodules


callumbirks:
Opened Fleece PR: https://github.com/couchbase/fleece/pull/187
Looks like that will need to be merged and the fleece submodule on this branch updated to point at `release/3.1` once it is merged.

borrrden (approved):
Once the fleece PR is fixed this seems fine.  There is one ticket that was marked as fixed in 3.1.2 as opposed to 3.1.x though, is that still the case?

pasin (approved):

jianminzhao:
How are we going to manage the relationships between cbl tickets and commits? How many tickets would this commit cover? How are we going to come back to search the commit responsible for a fix of ticket?

borrrden:
The squash commit could simply contain all the ticket numbers, or we could use one of the other merge methods to preserve the individual commits (preferably rebase)

jianminzhao:
Since all commits in next are finished ones, as opposed to work-in-progress, I would prefer to preserve all them with rebase rather than squash.

bmeike:
lol!   @jianminzhao is now in favor of rebase!!!

jianminzhao:
@bmeike this is a different situation. We have, say, 3 commits in the next branch. They represent finished works for respective CBL tickets. I would like to preserve the individual commits. Also I don't there will be significant conflicts from commit to commit. Rebase will keep the original separate commits whereas squash will combine all the commit messages. We will not be able to isolate to a particular commit in the future. 
In this particular case, I rather to have (normal) merge than squash merge. My key motivation is to allow in the future replay the fix that we put it in now.

jianminzhao (approved):

borrrden (approved):

PR #1869, opened 2023-08-18
CBL-4812: Attachments/Blobs got deleted after compaction&re-sync

jianminzhao:
For review: includes test code,

THIS IS NOT CHEC-IN-ABLE.

### DIFF
--- a/LiteCore/Database/TreeDocument.cc
+++ b/LiteCore/Database/TreeDocument.cc
@@ -79,6 +79,9 @@ namespace litecore {
 
             initRevID();
             selectCurrentRevision();
+#ifdef CBL_4812_TEST
+            _revTree._dontRemoveNonLeafBodies = (_collection->getDatabase()->getName() == "cbl_core_test");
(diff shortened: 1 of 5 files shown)


jianminzhao:
This is not checkin-able.
1. I used loopback replicator test. In the test, db plays the role of server. In the server, I believe they won't remove the bodies non-leaf bodies. Without the parent body, server cannot generate the delta/
2. In the test, there are two places to check the document flag for "has attachment", after two pulls of delta. For each delta, I did not touch the attachment part. So, the delta our passive replicator simply does not include "_attachments" at all. In customers log, SG included _attachements with two unchanged fields, but not "digest" properties which is cause of failure. In theory, delta does not have to include _attachments at all if there is not change. So, the fix simply assume 
4. _mayContainBlobs to be true if it's delta src. After all, it's "may contain," right.
5. The first check in the test, that is, after first delta pull, always succeeds. The second check fails without the above fix.

PR #1870, opened 2023-08-21
CBL-4812: Attachments/Blobs got deleted after compaction&re-sync

jianminzhao:
As we decide whether an attachment is referenced by a document, we first check whether the document flag, kHasAttachment, is set. This flag may be wrong when we pull a new revision by the delta body, particularly if the delta does not include blob's references but the delta source does. In this case, we must check it after the delta is applied. We fixed the bug by adding a parameter to C4DocDeltaApplier so that the callback can update the revision flag after the delta is applied.

### DIFF
--- a/Replicator/Inserter.cc
+++ b/Replicator/Inserter.cc
@@ -136,10 +136,10 @@ namespace litecore { namespace repl {
                     // If this is a delta, put the JSON delta in the put-request:
                     bodyForDB = move(rev->deltaSrc);
                     put.deltaSourceRevID = rev->deltaSrcRevID;
-                    put.deltaCB = [](void *context, C4Document *doc,
-                                     C4Slice delta, C4Error *outError) -> C4SliceResult {
+                    put.deltaCB = [](void *context, C4Document *doc, C4Slice delta,
+                                     C4RevisionFlags *revFlags, C4Error *outError) -> C4SliceResult {
                         try {
-                            return ((Inserter*)context)->applyDeltaCallback(doc, delta, outError);
+                            return ((Inserter*)context)->applyDeltaCallback(doc, delta, revFlags, outError);
                         } catch (...) {
                             *outError = C4Error::fromCurrentException();
                             return {};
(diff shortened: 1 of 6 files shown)


Replicator/Inserter.cc
```diff
@@ -191,25 +191,47 @@ namespace litecore { namespace repl {
...
+                root = doc.asDict();
+            }
+            // This is not optimal. Consider to add a method to DBAccess, hasBlobReferences
+            _db->findBlobReferences(root, true, [&](FLDeepIterator, Dict, const C4BlobKey&) {
+                hasAttachments = true;
+            });
+            if ( hasAttachments && !(*revFlags & kRevHasAttachments) ) {
+                *revFlags |= kRevHasAttachments;
+            } else if ( !hasAttachments && (*revFlags & kRevHasAttachments) ) {
+                // This shouldn't happen
```
borrrden, line 230: Why shouldn't it happen?  What if all the attachments were removed by the incoming rev?

Replicator/Inserter.cc
```diff
@@ -191,25 +191,47 @@ namespace litecore { namespace repl {
     // Callback from c4doc_put() that applies a delta, during _insertRevisionsNow()
     C4SliceResult Inserter::applyDeltaCallback(C4Document *c4doc,
                                                C4Slice deltaJSON,
+                                               C4RevisionFlags *revFlags,
                                                C4Error *outError)
     {
         Doc doc = _db->applyDelta(c4doc, deltaJSON, true);
         alloc_slice body = doc.allocedData();
+        Dict root = doc.root().asDict();
+        FLSharedKeys sk = nullptr;
```
borrrden, line 200: Can't this be declared closer to line 209?  Where else is it used?

Replicator/Inserter.cc
jianminzhao: As per my current reading, there are only two places where the flag is set. 1. when _mayContainBlobs == true. In this case. we apply the delta first. Then, we read in the doc to find all blobs. That is, we look at the final whole document body of the rev to be put nto the store. 2. when _mayContainBlobs == false, in this case we also read the doc after the delta applied. Thees two paths do not intersect. Unless the new rev to be inserted is initialized with the flag being set, we cannot have this case. My current understanding is, the attachments are not inherited from the parent revision. Like other properties, they have to be in the body of the new revision. To remove it, we simply not include that attachment in the new revision body.
I could be wrong, so it is a DebugAssert. Won't affect the final released program. Only I think it would never happen.

Replicator/Inserter.cc
jianminzhao: My intention is to not get it again if the body has changed. It was used to generate the changed body, and I can use to generate the new Doc.

Replicator/Inserter.cc
borrrden: So you mean to say that these flags are not initialized to what the parent revision flags were?  If so I can see the reasoning.

Replicator/Inserter.cc
borrrden: I missed where it was used on line 220, now I see it.

borrrden (approved):

pasin (approved):

Replicator/Inserter.cc
```diff
@@ -191,25 +191,47 @@ namespace litecore { namespace repl {
...
         }
+        if ( body && revFlags != nullptr ) {
+            bool hasAttachments = false;
+            if ( bodyChanged ) {
+                doc = Doc(body, kFLTrusted, sk);
+                root = doc.asDict();
+            }
+            // This is not optimal. Consider to add a method to DBAccess, hasBlobReferences
+            _db->findBlobReferences(root, true, [&](FLDeepIterator, Dict, const C4BlobKey&) {
+                hasAttachments = true;
```
pasin, line 225: I think this will be called multiple times with all blob references? 

I did a quick search and found `C4Blob::dictContainsBlobs(FLDict)` which could be better?

PR #1871, opened 2023-08-22
CBL-4382: Test "Database Upgrade From 2.7 to Version Vectors" Fails

jianminzhao:
Current, the signature KeyStore::set only permits insertion of new records, and hence must pick a new sequence (updateSequence). This is right for the top leve store, namely BothKeyStore. However, to move a record from sub-store to the other one, we delete it from one and intert it to the other one while keeping the sequence. The sub-store, or SQLiteKeyStore, also inherits from KeyStore. This is why we change the signature and make updateSequence and insert seaparate and ensure the integrity at appropriate places.

### DIFF
--- a/C/tests/c4DatabaseTest.cc
+++ b/C/tests/c4DatabaseTest.cc
@@ -1232,7 +1232,7 @@ TEST_CASE("Database Upgrade From 2.7", "[Database][Upgrade][C]") {
     //    testOpeningOlderDBFixture("upgrade_2.7.cblite2", kC4DB_ReadOnly);
 }
 
-// This one is failing due to CBL-4382
+// This one is failing due to CBL-4840
 TEST_CASE("Database Upgrade From 2.7 to Version Vectors", "[.failing][Database][Upgrade][C]") {
     testOpeningOlderDBFixture("upgrade_2.7.cblite2", kC4DB_VersionVectors);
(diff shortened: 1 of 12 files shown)


snej (changes requested):
see above ^

jianminzhao:
"The fix seems more elaborate than is called for. Fundamentally, the set() method just needs another boolean parameter insert that forces a new record to be created even with a nonzero sequence. (Right?)"
-- yes

"First, since it's never valid to store a record with a 0 sequence, you can treat that case as though insert were true. That means most of the callers don't have to worry about whether or not to set insert"
-- the params maker does exactly this: it sets Insert to true of rec.sequence == 0. In fact, there is no place in the code that directly sets Insert except for the one place in BothKeyStore::set. 

I assumed you meant to replace the current parameter updateSequence with an enum flag, right? 
I believe std::bitset is nothing more than an unsigned or enum. It just does the bitwise operations for us. But I am open to simply use an enum class. The code will look very similar except, as you mentioned, cast  of result of bitwise or.

jianminzhao:
I meant SetParams will become enum class SetParams : unsigned { UpdateSeuqnece, Insert }
Or, do you have a suggestion for the name?

snej (approved):

PR #1872, opened 2023-08-22
CBL-4811: Print UTC time in log file with plain text

jianminzhao:


borrrden:
Does this add a `Z` to the timestamp?  If not, can it?  That's what my tool uses to determine local vs UTC

jianminzhao:
Yes, it has Z at the end.

jianminzhao:
Without end ing Z, it means local time.

borrrden (approved):

PR #1874, opened 2023-08-28
CBL-4495: Remove bitcode from iOS builds

jianminzhao:
Portd from release/3.1 #bc52ab02


pasin (approved):

borrrden (approved):
Not sure what the issue was, looks like ENABLE_BITCODE=0 causes an issue here?  Can you verify that no bitcode it being generated?

bmeike (commented):
✅

jianminzhao:
The problem with my first attempt is a mistake with an extra quote.  How to verify bitcode? build an ios target?

PR #1877, opened 2023-08-30
CBL-4789: Properly handle continuation frames from websockets (#1860)

jianminzhao:
remainingBytes is not a reliable source of information about how big the final buffer is going to be, so it needs to be resized as needed.


jianminzhao:
Ported from lithium with little manual merging.

borrrden (approved):
Assuming this logic is taken as-is from master

pasin (approved):

PR #1879, opened 2023-09-01
CBL-4698: Upgrade to SQLite 3.43.0

jianminzhao:


jianminzhao:
The failure of the blackduck manifest is due to sqlitecpp. Don't how to make it work. Do I need to bump its version as well?
@borrrden

borrrden (approved):
Nice idea for the validation fix

PR #1881, opened 2023-09-14
CBL-4898: Handling blobs that reference attachments that are already …

jianminzhao:
…deleted in the remote.

If the rev pulled down from SG includes top level property "_attatchments", it should include all attachments that exist in the SG for the current revision. If a blob in the revision's body references an attachment that is not in "_attachments," it is bound for "getAttachment" to fail. We try to detect this case and, instead of handling the bound error response, err our within the Core library.

### DIFF
--- a/Replicator/IncomingRev+Blobs.cc
+++ b/Replicator/IncomingRev+Blobs.cc
@@ -28,12 +28,34 @@ namespace litecore { namespace repl {
     static std::atomic_int sMaxOpenWriters {0};
 #endif
 
+//#define FETCH_DANGLING_BLOBS
+
     // Looks for another blob to download; when they're all done, finishes up the revision.
     void IncomingRev::fetchNextBlob() {
+        // Invariant: _danglingBlobBegin == _pendingBlobs.begin() + n, for some n where 0 <= n && n <= _pendingBlobs.size()
+#ifdef FETCH_DANGLING_BLOBS
         while (_blob != _pendingBlobs.end()) {
+#else
+        while (_blob != _danglingBlobBegin) {
+#endif
             if (startBlob())
                 return;
             ++_blob;
+        }
+
+        if (_blob != _pendingBlobs.end()) {
+            string errmsg = "The rev(" + _rev->docID.asString() + "/" + _rev->revID.asString() + ") contains blobs (";
+            bool first = true;
+            for (std::vector<PendingBlob>::const_iterator iter = _blob; iter != _pendingBlobs.end(); ++iter) {
+                if (!first) errmsg += ",";
+                errmsg += iter->key.digestString();
+                first = false;
+            }
+            errmsg += ") that are not in the remote.";
+            C4Error c4Error = C4Error::make(LiteCoreDomain, kC4ErrorNotFound, slice(errmsg));
+            logError("Error: %s", c4Error.description().c_str());
+            failWithError(c4Error);
+            return;
         }
 
         // All blobs completed, now finish:
--- a/Replicator/IncomingRev.cc
+++ b/Replicator/IncomingRev.cc
@@ -51,8 +51,7 @@ namespace litecore { namespace repl {
         _parent = _puller;  // Necessary because Worker clears _parent when first completed
         _provisionallyInserted = false;
         DebugAssert(_pendingCallbacks == 0 && !_writer && _pendingBlobs.empty());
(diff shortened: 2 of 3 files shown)


jianminzhao:
Author's Notes:
1. I refactored the code so that we can manage the blobs that reference attachments that do not exist in the SG.
2. If FETCH_DANGLING_BLOBS is defined, the puller should behave the same. Otherwise, we will not try to fetch the attachment that does not exist in _attatchments and err locally.
3. If defined, we have following relevant logs:
  - BLIP {
	Profile: getAttachment
	digest: sha1-ERWD9RaGBqLSWOQ+96TZ6Kisjck=
	docID: att1
        compress: true }
 - BLIP response  {
	Error-Domain: HTTP
	Error-Code: 403
	BODY: Attachment's doc not being synced }
 - Log [Sync] ERROR: {IncomingRev#22} Got error response: HTTP 403 'Attachment's doc not being synced'
 - Log [Sync] WARNING: {IncomingRev#22} failed with error: WebSocket error 403, "Attachment's doc not being synced"
 - BLIP response to the "rev" message {
	Error-Domain: HTTP
	Error-Code: 403
	BODY: Attachment's doc not being synced }
4. If not defined,
 - There is no the first BLIP message, "getAttachment,"  appeared in (3)
 - Log ERROR: {IncomingRev#22} Error: LiteCore NotFound, "The rev(att1/2-8954a50e4991fad3f861d03073a5989a) contains blobs (sha1-ERWD9RaGBqLSWOQ+96TZ6Kisjck=) that are not in the remote."
 - Log WARNING: {IncomingRev#22} failed with error: LiteCore NotFound, "The rev(att1/2-8954a50e4991fad3f861d03073a5989a) contains blobs (sha1-ERWD9RaGBqLSWOQ+96TZ6Kisjck=) that are not in the remote." // this corresponds to the WARNING of (3)
 - BLIP response to the "rev" message {
	Error-Domain: LiteCore
	Error-Code: 7
	BODY: The rev(att1/2-8954a50e4991fad3f861d03073a5989a) contains blobs (sha1-ERWD9RaGBqLSWOQ+96TZ6Kisjck=) that are not in the remote. }
5. I did the test with SG 3.0 walrus, because it's faster. Still walrus tests are not working in 3.1. I need to fiddle a little. So the test is not part of the PR. The test goes as follows,
 - push rev 1 with 3 attachments, 3 blobs
 - push rev 2 with 2 attachments, same 3 blobs
 - delete local db
 - pull 
6. Please provide your suggestion of the error message, "The rev(att1/2-8954a50e4991fad3f861d03073a5989a) contains blobs (sha1-ERWD9RaGBqLSWOQ+96TZ6Kisjck=) that are not in the remote."

Replicator/IncomingRev.cc
```diff
@@ -273,16 +287,31 @@ namespace litecore { namespace repl {
...
+            }
+            for (const auto& b: dangled) {
+                _pendingBlobs.push_back(b);
+            }
+            _blob = _pendingBlobs.begin();
+            _danglingBlobBegin = _blob + (_pendingBlobs.size() - dangled.size());
         }
+
+        Assert(!_pendingBlobs.empty() || _blob == _pendingBlobs.end());
+        fetchNextBlob();
```
pasin, line 314: If there are any _dangling blobs, can we just fail here without trying to fetch any blobs?

Replicator/IncomingRev+Blobs.cc
```diff
@@ -28,14 +28,36 @@ namespace litecore { namespace repl {
     static std::atomic_int sMaxOpenWriters {0};
 #endif
 
+//#define FETCH_DANGLING_BLOBS
```
pasin, line 31: I don't feel like we need this to rollback to the previous logic.

Replicator/IncomingRev.cc
jianminzhao: That will be much easier.

Replicator/IncomingRev+Blobs.cc
jianminzhao: okay, I will deal with in the final version.

bmeike (changes requested):
I don't want to see this PR go through until that 6 case table is okayed by everybody who understands what's going on.

jianminzhao:
Redid in #1885

PR #1884, opened 2023-09-21
CBL-4912: Handling blobs that reference attachments that are already …

jianminzhao:
…deleted in the remote.

If the rev pulled down from SG includes top level property "_attatchments", it should include all attachments that exist in the SG for the current revision. If a blob in the revision's body references an attachment that is not in "_attachments," "getAttachment" is bound to fail. We try to detect this case and, instead of handling the error response from SG, we err out in CBL.

The above solution may result in different behavior whether the delta sync is enabled or not. It is due to false detection whether the delta revision can affect blobs without checking with the current revision. We fixed it.

### DIFF
--- a/Replicator/IncomingRev+Blobs.cc
+++ b/Replicator/IncomingRev+Blobs.cc
@@ -30,31 +30,10 @@ namespace litecore { namespace repl {
 
     // Looks for another blob to download; when they're all done, finishes up the revision.
     void IncomingRev::fetchNextBlob() {
-        // Invariant: _danglingBlobBegin == _pendingBlobs.begin() + n, for some n where 0 <= n && n <= _pendingBlobs.size()
-        while (_blob != _danglingBlobBegin) {
+        while (_blob != _pendingBlobs.end()) {
             if (startBlob())
                 return;
             ++_blob;
-        }
-
-        if (_blob != _pendingBlobs.end()) {
-            bool plural = (_pendingBlobs.end() - _danglingBlobBegin) > 1;
-            string errmsg = "There ";
-            if (plural) {
-                errmsg += "are no contents for the blobs with digests ";
-            } else {
-                errmsg += "is no content for the blob with digest ";
-            }
-            bool first = true;
-            for (std::vector<PendingBlob>::const_iterator iter = _blob; iter != _pendingBlobs.end(); ++iter) {
-                if (!first) errmsg += ", ";
-                errmsg += iter->key.digestString();
-                first = false;
-            }
-            errmsg += " in the attachments for document " + _rev->docID.asString();
-            C4Error c4Error = C4Error::make(LiteCoreDomain, kC4ErrorNotFound, slice(errmsg));
-            failWithError(c4Error);
-            return;
         }
 
         // All blobs completed, now finish:
--- a/Replicator/IncomingRev+Blobs.cc
+++ b/Replicator/IncomingRev+Blobs.cc
@@ -30,10 +30,31 @@ namespace litecore { namespace repl {
 
     // Looks for another blob to download; when they're all done, finishes up the revision.
     void IncomingRev::fetchNextBlob() {
-        while (_blob != _pendingBlobs.end()) {
+        // Invariant: _danglingBlobBegin == _pendingBlobs.begin() + n, for some n where 0 <= n && n <= _pendingBlobs.size()
+        while (_blob != _danglingBlobBegin) {
             if (startBlob())
                 return;
             ++_blob;
+        }
+
+        if (_blob != _pendingBlobs.end()) {
+            bool plural = (_pendingBlobs.end() - _danglingBlobBegin) > 1;
+            string errmsg = "There ";
+            if (plural) {
+                errmsg += "are no contents for the blobs with digests ";
+            } else {
+                errmsg += "is no content for the blob with digest ";
+            }
+            bool first = true;
+            for (std::vector<PendingBlob>::const_iterator iter = _blob; iter != _pendingBlobs.end(); ++iter) {
+                if (!first) errmsg += ", ";
+                errmsg += iter->key.digestString();
+                first = false;
+            }
+            errmsg += " in the attachments for document " + _rev->docID.asString();
+            C4Error c4Error = C4Error::make(LiteCoreDomain, kC4ErrorNotFound, slice(errmsg));
+            failWithError(c4Error);
+            return;
         }
 
         // All blobs completed, now finish:
--- a/Replicator/IncomingRev.cc
+++ b/Replicator/IncomingRev.cc
@@ -20,6 +20,7 @@
 #include "c4Document.hh"
 #include "Instrumentation.hh"
 #include "BLIP.hh"
+#include "FleeceImpl.hh"
 #include "fleece/Mutable.hh"
 #include <atomic>
 #include <deque>
@@ -51,8 +52,7 @@ namespace litecore { namespace repl {
         _parent = _puller;  // Necessary because Worker clears _parent when first completed
         _provisionallyInserted = false;
         DebugAssert(_pendingCallbacks == 0 && !_writer && _pendingBlobs.empty());
-        _blob = _pendingBlobs.end();
-
+        _danglingBlobBegin = _blob = _pendingBlobs.end();
     }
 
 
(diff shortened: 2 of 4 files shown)


jianminzhao:
Relevant flow of events with delta revision
- receive rev request from SG:
{
	Profile: rev
	id: att1
	rev: 2-8954a50e4991fad3f861d03073a5989a
	history: 1-b98a25d09a549dc2f68ac7b6a1acaf4da55e0f0d
	deltaSrc: 1-b98a25d09a549dc2f68ac7b6a1acaf4da55e0f0d
	sequence: 3
	Content-Type: application/json
	BODY: {"_attachments":{"blob_/attached/0":[],"blob_/attached/1":{"length":10,"revpos":1},"blob_/attached/2":{"length":0,"revpos":1}}} }
- Error generated by CBL, LiteCore Not Found
[Sync]: {IncomingRev#42}==> litecore::repl::IncomingRev {Repl#37} @0x614000020650
08:16:15.320469| [Sync] ERROR: {IncomingRev#42} failed with error: LiteCore NotFound, "There is no content for the blob with digest sha1-ERWD9RaGBqLSWOQ+96TZ6Kisjck= in the attachments for document att1"
- response to SG's rev request
{
	Error-Domain: LiteCore
	Error-Code: 7
	BODY: There is no content for the blob with digest sha1-ERWD9RaGBqLSWOQ+96TZ6Kisjck= in the attachments for document att1 }

jianminzhao:
Windows timed out twice but I think it's reviewable.

Replicator/IncomingRev+Blobs.cc
```diff
@@ -30,12 +30,33 @@ namespace litecore { namespace repl {
 
     // Looks for another blob to download; when they're all done, finishes up the revision.
     void IncomingRev::fetchNextBlob() {
-        while (_blob != _pendingBlobs.end()) {
+        // Invariant: _danglingBlobBegin == _pendingBlobs.begin() + n, for some n where 0 <= n && n <= _pendingBlobs.size()
```
bmeike, line 33: If this is an invariant, why is it not an assertion?

bmeike (commented):
What is the code coverage for this new code.  It is quite complex and seems fairly risky....

Replicator/IncomingRev+Blobs.cc
jianminzhao: I sometimes put simple assertion or invariant in the comment to help understand the code and easily verified, to avoid overcrowd the runtime assertion.

jianminzhao:
Code coverage is by the new test.

Replicator/IncomingRev.cc
```diff
@@ -119,7 +119,66 @@ namespace litecore { namespace repl {
         if (_revMessage->noReply())
             _revMessage = nullptr;
 
-        _mayContainBlobs = jsonBody.containsBytes("\"digest\""_sl);
+        auto checkBlob = [](bool isDelta, alloc_slice jsonBody) -> bool {
```
pasin, line 122: What about just `jsonBody.containsBytes("\"digest\""_sl)) || jsonBody.containsBytes("\"_attachments\""_sl))`?

Replicator/IncomingRev.cc
```diff
@@ -273,16 +347,31 @@ namespace litecore { namespace repl {
...
+                    ++iter;
+                }
+            }
+            for (const auto& b: dangled) {
+                _pendingBlobs.push_back(b);
+            }
+            _blob = _pendingBlobs.begin();
+            _danglingBlobBegin = _blob + (_pendingBlobs.size() - dangled.size());
         }
+
```
pasin, line 372: Same as my comment on the previous PR that I think we don't need to fetch any blobs if we found a dangling blob. Then the logic to move / append the dangling blobs to the end of pending blobs will not be needed as well.

Replicator/IncomingRev.cc
```diff
@@ -237,6 +296,21 @@ namespace litecore { namespace repl {
...
+            for (Dict::iterator it(attachments); it; ++it) {
+                if (Dict v = it.value().asDict(); v) {
+                    auto digest = v[C4Blob::kDigestProperty];
+                    if (digest.asString()) {
+                        attachmentsFromSG->emplace(digest.asString());
+                    }
+                }
+            }
+        }
+
```
pasin, line 313: For the case of using delta sync and that the customer only deletes an attachment in _attachments but not its corresponding blob in the body, will `_db->applyDelta(_rev->docID, _rev->deltaSrcRevID, jsonBody);` synthesizes _attachments correctly by not adding the attachment back to the _attachments?

pasin: For the case of using **delta sync** and when the customer only deleted an attachment in _attachments but not its corresponding blob in the body.

Will `C4Document::encodeStrippingOldMetaProperties(root, sk)` in line 320 add the attachment back in the _attachments?

pasin: I think it will not otherwise the test will not pass but I'm asksing to make sure that we are synthezing _attachments correctly.

Replicator/IncomingRev.cc
jianminzhao: Well, this may give incorrect true. In theory, _mayContainBlobs being true is always right, because it says MAY, and the consequence is applying the delta earlier than necessary. In the other day when we were solving the other cbse that deleted attachment by db compact, I suggested to treat it alwasy true. In that case, we had delta _attachments actually with no changes.

Replicator/IncomingRev.cc
jianminzhao: yes, this can simplified. I can simply not call startNextBlob().
Should I do it?

Replicator/IncomingRev.cc
jianminzhao: Apply delta will not put the blob back to _attachments. However, if we save it to db, we won't save _attachments, and yes, next time as we synthesize _attachments, it will be put back.That's why I said if we let it succeed, next time push will push the attachment to SG

Replicator/IncomingRev.cc
jianminzhao: I think it makes sense and did it.

Replicator/IncomingRev.cc
jianminzhao: It goes like, we synthesize the base of the delta, namely the current revision, and get _attachments before the delta. If the current revision is in the sound state, one of the attachment will be removed after applying the delta.

Replicator/IncomingRev.cc
```diff
@@ -277,6 +352,46 @@ namespace litecore { namespace repl {
...
+                } else {
+                    ++iter;
+                }
+            }
+            for (const auto& b: dangled) {
+                _pendingBlobs.push_back(b);
+            }
+            _blob = _pendingBlobs.begin();
+            danglingBlobBegin = _blob + (_pendingBlobs.size() - dangled.size());
+        }
```
pasin, line 372: We could simplify this logic by putting the dangled attachments in a separated vector.

Replicator/IncomingRev.cc
```diff
@@ -119,13 +120,73 @@ namespace litecore { namespace repl {
...
+                    // We already inspected _attachments.
+                    iter.skipChildren();
+                } else {
+                    // Other than _attachments
+
+                    if (Dict dict = iter.value().asDict(); dict) {
+                        if (C4Blob::isBlob(dict) && dict.get(C4Blob::kDigestProperty)) {
+                            return true;
+                        }
+                    } else if (Array arr = iter.value().asArray(); arr && arr.count() == 1) {
```
pasin, line 167: Do we need to look into the dictionary in the array here? Won't the `DeepIterator` iterate into the dictionaries in the array ?

Replicator/IncomingRev.cc
jianminzhao: done

Replicator/IncomingRev.cc
jianminzhao: Here I am looking for new blob that is added. The JSON diff syntax is
property: [ newValue ]
the bracket is an artificial marker that indicates the value inside is to replace the original value. Example would be
`{"_attachments":{"blob_/attached/0":[],"blob_/attached/1":[],"blob_/attached/2":[]},"attached":{"3-":[{"@type":"blob","content_type":"text/plain","digest":"sha1-3jmj7l5rSw0yVb/vlWAYkK/YBwk=","length":10}]}} }`
"attached" is an array, but we use key, "3-", meaning from index 3 to the end of the array, replaced by a new blob
The new blob is denoted in square bracket, [], namely, arr[0] in the Value

pasin (approved):

Replicator/IncomingRev.cc
jianminzhao: But I really don't need to do it here. Instead, I can, as you suggested, let the iterator run to the blob object and check it then.
Changed as suggested.

PR #1886, opened 2023-09-26
CBL-4944: Add additional delta sync logs for investigating the issue

jianminzhao:


Replicator/DBAccess.cc
```diff
@@ -245,13 +253,24 @@ namespace litecore { namespace repl {
             string attName = string("blob_") + string(path);
             enc.writeKey(slice(attName));
             enc.beginDict();
+            bool cbl_4944_findLength = false;
             for (Dict::iterator i(blob); i; ++i) {
                 slice key = i.keyString();
+                if (key == "length"_sl) {
+                    cbl_4944_findLength = true;
```
pasin, line 260: CBSE case has attachment length = 0 while it shouldn't be 0. The attachment length = 0 could be valid for empty blob but for the purpose of the investigation, we should log as well.

LiteCore/Database/LegacyAttachments.cc
```diff
@@ -59,6 +59,12 @@ namespace litecore { namespace legacy_attachments {
                 auto attachment = Value::asDict(i.value());
                 if (!attachment)
                     continue;
+                {
+                    // CBL-4944
+                    if (!attachment->get("length"_sl)) {
+                        C4Warn("Find attahcment with nil length in legacy_attachments::encodeStrippingOldMetaProperties: %s", attachment->toJSONString().c_str());
+                    }
```
pasin, line 66: CBSE case has attachment length = 0 while it shouldn't be 0. The attachment length = 0 could be valid for empty blob but for the purpose of the investigation, we should log as well.

LiteCore/Database/LegacyAttachments.cc
```diff
@@ -92,6 +98,14 @@ namespace litecore { namespace legacy_attachments {
         Encoder enc;
         enc.setSharedKeys(sk);
         enc.writeValue(root, [&](const Value *key, const Value *value) {
+            {
+                // CBL-4944
+                if (const Dict* valDict = Value::asDict(value); C4Blob::isBlob(FLDict(valDict))) {
+                    if (!valDict->get("length"_sl)) {
+                        C4Warn("Find blob with nil length in legacy_attachments::encodeStrippingOldMetaProperties: %s", valDict->toJSONString().c_str());
```
pasin, line 105: CBSE case has attachment length = 0 while it shouldn't be 0. The attachment length = 0 could be valid for empty blob but for the purpose of the investigation, we should log as well.

Replicator/DBAccess.cc
```diff
@@ -234,6 +234,14 @@ namespace litecore { namespace repl {
                 // TODO: Should skip this entry if a blob with the same digest exists
                 enc.writeKey(key);
                 enc.writeValue(i.value());
+                {
+                    // CBL-4944
+                    if (Value v = i.value(); v) {
+                        if (Dict d = v.asDict(); d && !d.get("length"_sl)) {
+                            warn("Find attachment with nil length in DBAccess::encodeRevWithLegacyAttachments, %s", d.toJSONString().c_str());
```
pasin, line 241: CBSE case has attachment length = 0 while it shouldn't be 0. The attachment length = 0 could be valid for empty blob but for the purpose of the investigation, we should log as well.

pasin (changes requested):
In encodeStrippingOldMetaProperties, can we log 

1. When the function returns {};

2. When enc.finish(); returns false?

pasin (commented):
Need to add log for the case of invalid delta as well.

PR #1888, opened 2023-09-28
CBL-4920: Incorrect max length for version vector ASCII

jianminzhao:


velicuvlad (commented):
Jens mentioned something about a case where writeAscii will have an extra semicolon at the end and advice to be increase to +3.

jianminzhao:
The exact number is, (_vers.size()  - 1) * (Version::kMaxASCIILength + 2) +  (Version::kMaxASCIILength + 1) or 
 _vers.size() * (Version::kMaxASCIILength + 2) - 1

velicuvlad (approved):

PR #1889, opened 2023-09-29
CBL-4838: Attachments/Blobs got deleted after compaction&re-sync

jianminzhao:
As we decide whether an attachment is referenced by a document, we first check whether the document flag, kHasAttachment, is set. This flag may be wrong when we pull a new revision by the delta body, particularly if the delta does not include blob's references but the delta source does. In this case, we must check it after the delta is applied. We fixed the bug by adding a parameter to C4DocDeltaApplier so that the callback can update the revision flag after the delta is applied.


jianminzhao:
This PR is ported from Lithium except for a new function, DBAccess::hasBlobReferences(), adapted from DBAccess::findBlobReferences() for the sake of efficiency.

callumbirks (approved):

PR #1895, opened 2023-10-19
Zero-cost version-vector database upgrade

snej:
Eliminates the cost of upgrading an existing db to version vectors. 

- Instead of converting all docs to version-vector format the first time the db is opened, it leaves them alone.
- A C4Document is opened using its existing versioning type, by default. But the new document flag kDocGetUpgraded causes its in-memory representation to be upgraded to use version vectors.
- A C4Document is upgraded on disk when it's saved (regardless of what flags it was loaded with.)
- The replicator loads docs using kDocGetUpgraded since it uses version vectors.

Most of the work here is to make the VectorRecord class -- which is always used with a VV-upgraded db -- able to load a rev-tree based doc, and by default keep its revisions identified by rev-tree IDs.

### DIFF
--- a/LiteCore/Database/DatabaseImpl+Upgrade.cc
+++ b/LiteCore/Database/DatabaseImpl+Upgrade.cc
@@ -24,12 +24,15 @@
 #include <utility>
 #include <vector>
 
+#define ENABLE_VV_UPGRADE 0
+
 namespace litecore {
     using namespace std;
 
     static constexpr const char* kNameOfVersioning[3] = {"v2.x rev-trees", "v3.x rev-trees", "version vectors"};
 
 
+#if ENABLE_VV_UPGRADE
     static void upgradeToVersionVectors(DatabaseImpl*, const Record&, RevTreeRecord&, ExclusiveTransaction&);
     static pair<alloc_slice, alloc_slice> upgradeRemoteRevs(DatabaseImpl*, Record, RevTreeRecord&,
                                                             alloc_slice currentVersion);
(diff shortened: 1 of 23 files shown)


LiteCore/Database/DatabaseImpl+Upgrade.cc
```diff
@@ -82,8 +92,10 @@ namespace litecore {
                     if ( isVersionVector(rec) ) continue;
                     RevTreeRecord revTree(defaultKeyStore(), rec);
                     if ( newVersioning == kC4VectorVersioning ) {
+#if ENABLE_VV_UPGRADE
```
jianminzhao, line 95: Do we need this since if !ENABLE_VV_UPGRADE, we already downgrade newVersioning from VectorVersioning?

jianminzhao:
A general question. Suppose this is a huge legacy database. In what circumstance would we avoid the time to upgrade to VV? Is it true that only things we may do to the db are queries?

snej:
In all circumstances. As I said, it defers upgrading documents until the first time they're updated.

LiteCore/Database/DatabaseImpl+Upgrade.cc
snej: It's necessary because otherwise the call to `upgradeToVersionVectors` would be a compile error, since the function is ifdef'd out.

jianminzhao:
From the users' standpoint, beside updating and replicating, what else could users do except for query? Without updating, there won't be need of replicating.

snej:
There is no change from the user's POV. The database is still upgraded, and they can do anything they want. What this PR does is avoid having to rewrite every document during the upgrade.

jianminzhao:
I see. This will cut down the time when upgrade takes place.

jianminzhao (approved):

PR #1900, opened 2023-10-24
CBL-4995: Log the reasons of purging documents.

jianminzhao:
Identifing 5 reasons:
1. API purge, c4coll_purgeDoc
2. API purgeExpired, c4coll_purgeExpiredDocs
3. HouseKeeper expiring documents
4. SG removing, revMsg["_removed"] == true
5. SG revoking, due to the parameters of the changes message.

### DIFF
--- a/C/c4CAPI.cc
+++ b/C/c4CAPI.cc
@@ -26,6 +26,7 @@
 #include "c4Private.h"
 #include "Delimiter.hh"
 #include "fleece/Mutable.hh"
+#include "StringUtil.hh"
 #include <sstream>
 
 using namespace std;
@@ -371,9 +372,11 @@ bool c4coll_purgeDoc(C4Collection *coll,
 {
     returnIfCollectionInvalid(coll, outError, false);
     try {
-        if (coll->purgeDocument(docID))
+        if (coll->purgeDocument(docID)) {
+            C4CollectionSpec spec = c4coll_getSpec(coll);
+            Log("API purge doc: %.*s.%.*s.%.*s", SPLAT(spec.scope), SPLAT(spec.name), SPLAT(docID));
             return true;
-        else
+        } else
             c4error_return(LiteCoreDomain, kC4ErrorNotFound, {}, outError);
     } catchError(outError)
     return false;
@@ -413,6 +416,8 @@ C4Timestamp c4coll_nextDocExpiration(C4Collection *coll) noexcept {
 
 int64_t c4coll_purgeExpiredDocs(C4Collection *coll, C4Error * C4NULLABLE outError) noexcept {
     returnIfCollectionInvalid(coll, outError, 0);
+    C4CollectionSpec spec = c4coll_getSpec(coll);
+    Log("API purgeExpired in collection: %.*s.%.*s", SPLAT(spec.scope), SPLAT(spec.name));
     return tryCatch<int64_t>(outError, [=]{
         return coll->purgeExpiredDocs();
     });
--- a/C/c4CAPI.cc
+++ b/C/c4CAPI.cc
@@ -374,7 +374,7 @@ bool c4coll_purgeDoc(C4Collection *coll,
     try {
         if (coll->purgeDocument(docID)) {
             C4CollectionSpec spec = c4coll_getSpec(coll);
-            Log("API purge doc: %.*s.%.*s.%.*s", SPLAT(spec.scope), SPLAT(spec.name), SPLAT(docID));
+            Log("Purge doc \"%.*s.%.*s.%.*s\"", SPLAT(spec.scope), SPLAT(spec.name), SPLAT(docID));
             return true;
         } else
             c4error_return(LiteCoreDomain, kC4ErrorNotFound, {}, outError);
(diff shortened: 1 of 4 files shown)


C/c4CAPI.cc
```diff
@@ -371,9 +372,11 @@ bool c4coll_purgeDoc(C4Collection *coll,
 {
     returnIfCollectionInvalid(coll, outError, false);
     try {
-        if (coll->purgeDocument(docID))
+        if (coll->purgeDocument(docID)) {
+            C4CollectionSpec spec = c4coll_getSpec(coll);
+            Log("API purge doc: %.*s.%.*s.%.*s", SPLAT(spec.scope), SPLAT(spec.name), SPLAT(docID));
```
pasin, line 377: It's kind of weird to have "API" in the log message. Can we have something like "Purge doc : scope.collection.id" or Purge doc 'scope.collection.id' instead?

C/c4CAPI.cc
```diff
@@ -413,6 +416,8 @@ C4Timestamp c4coll_nextDocExpiration(C4Collection *coll) noexcept {
 
 int64_t c4coll_purgeExpiredDocs(C4Collection *coll, C4Error * C4NULLABLE outError) noexcept {
     returnIfCollectionInvalid(coll, outError, 0);
+    C4CollectionSpec spec = c4coll_getSpec(coll);
+    Log("API purgeExpired in collection: %.*s.%.*s", SPLAT(spec.scope), SPLAT(spec.name));
```
pasin, line 420: Similar comment as above. I think we don't need "API" there. Probably "Purge expired docs in collection : scope.collection" or "Purge expired docs in collection 'scope.collection'".

Replicator/RevFinder.cc
```diff
@@ -220,7 +220,10 @@ namespace litecore::repl {
                 // If the removal flag is accompanyied by the deleted flag, we don't purge, c.f. above remark.
                 auto mode = (deletion < 4) ? RevocationMode::kRevokedAccess
                                            : RevocationMode::kRemovedFromChannel;
-                revoked.emplace_back(new RevToInsert(docID, revID, mode, getCollection()->getSpec(),
+                auto collSpec = getCollection()->getSpec();
+                logInfo("SG revoking: %.*s.%.*s.%.*s/%.*s with mode %u", SPLAT(collSpec.scope), SPLAT(collSpec.name), SPLAT(docID),
```
pasin, line 224: Maybe "SG revoked access to the doc : 'scope.collection.id/revid" or "SG revoked access to the doc 'scope.collection.id' at rev 'revid'"

Replicator/IncomingRev.cc
```diff
@@ -264,6 +264,8 @@ namespace litecore { namespace repl {
         // SG sends a fake revision with a "_removed":true property, to indicate that the doc is
         // no longer accessible (not in any channel the client has access to.)
         if (root["_removed"_sl].asBool()) {
+            logInfo("SG removing: %.*s.%.*s.%.*s/%.*s", SPLAT(_rev->collectionSpec.scope), SPLAT(_rev->collectionSpec.name),
+                    SPLAT(_rev->docID), SPLAT(_rev->revID));
```
pasin, line 268: "Received removed revision for doc : scope.collection.docid/revid"

pasin (changes requested):

bmeike (commented):
✅
Agree with Pasin about rewording the message

C/c4CAPI.cc
jianminzhao: I tried to assign reason of high level. "API Purge" means that it's directly called from Platform SDK. But I can do as you suggested. "API" kind of weired. We can always tell where it is called from reading the source.

Replicator/RevFinder.cc
jianminzhao: I would like to keep the "mode" to help debugging.

Replicator/IncomingRev.cc
jianminzhao: okay

C/c4CAPI.cc
jianminzhao: okay

Replicator/IncomingRev.cc
pasin: Can you change the removed doc to the removed revision? It's the term that we are using for this special rev.

pasin (changes requested):

Replicator/IncomingRev.cc
jianminzhao: yes

jianminzhao: Actually, this is a fake revision, 
        // SG sends a fake revision with a "_removed":true property, to indicate that the doc is
        // no longer accessible (not in any channel the client has access to.)
Revision is not important. Actually, the doc will be purged, not the revison.

jianminzhao: revision here is only a technical thing. We log it here to help debugging.

pasin: What we received is the revision from getRev so technically it's the rev. You can also be specific like "Received the fake removed revision ..."

pasin: OK, I'm fine either way actually.

pasin (approved):

PR #1905, opened 2023-11-08
CBL-5033: Puller revoked docs should queue with other revs

callumbirks:
Are we all satisfied with this solution? Should I add more tests for doc revocation, or do the tests already in `ReplicatorCollectionSGTest` suffice?

### DIFF
--- a/Replicator/IncomingRev.cc
+++ b/Replicator/IncomingRev.cc
@@ -22,6 +22,7 @@
 #include <atomic>
 #include <deque>
 #include <set>
+#include <utility>
 
 using namespace std;
 using namespace fleece;
@@ -49,13 +50,13 @@ namespace litecore::repl {
 
     // Read the 'rev' message, then parse either synchronously or asynchronously.
     // This runs on the caller's (Puller's) thread.
-    void IncomingRev::handleRev(blip::MessageIn* msg, uint64_t bodySize) {
+    void IncomingRev::handleRev(Retained<blip::MessageIn> msg, uint64_t bodySize) {
         reinitialize();
         _bodySize = bodySize;
 
         // Set up to handle the current message:
         DebugAssert(!_revMessage);
-        _revMessage         = msg;
+        _revMessage         = std::move(msg);
         _rev                = new RevToInsert(this, _revMessage->property("id"_sl), _revMessage->property("rev"_sl),
                                               _revMessage->property("history"_sl), _revMessage->boolProperty("deleted"_sl),
                                               _revMessage->boolProperty("noconflicts"_sl) || _options->noIncomingConflicts(),
--- a/Replicator/Puller.cc
+++ b/Replicator/Puller.cc
@@ -19,6 +19,7 @@
 
 #include "Puller.hh"
 #include "Replicator.hh"
+#include "ReplicatorTypes.hh"
 #include "RevFinder.hh"
 #include "Inserter.hh"
 #include "IncomingRev.hh"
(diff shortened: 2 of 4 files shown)


Replicator/IncomingRev.cc
```diff
@@ -49,13 +50,13 @@ namespace litecore::repl {
...
     // This runs on the caller's (Puller's) thread.
-    void IncomingRev::handleRev(blip::MessageIn* msg, uint64_t bodySize) {
+    void IncomingRev::handleRev(Retained<blip::MessageIn> msg, uint64_t bodySize) {
         reinitialize();
         _bodySize = bodySize;
 
         // Set up to handle the current message:
         DebugAssert(!_revMessage);
-        _revMessage         = msg;
+        _revMessage         = std::move(msg);
```
jianminzhao, line 59: Why move a ref counted reference?

Replicator/Puller.cc
```diff
@@ -207,13 +225,25 @@ namespace litecore::repl {
...
         while ( connected() && _activeIncomingRevs < tuning::kMaxActiveIncomingRevs
                 && _unfinishedIncomingRevs < tuning::kMaxIncomingRevs && !_waitingRevMessages.empty() ) {
-            auto msg = _waitingRevMessages.front();
-            _waitingRevMessages.pop_front();
-            if ( _waitingRevMessages.empty() ) {
-                Signpost::end(Signpost::revsBackPressure);
-                logVerbose("Back pressure ended for changes messages");
+            // Waiting Rev is either regular rev message (MessageIn) or revoked rev (RevToInsert)
+            auto waiting = _waitingRevMessages.front();
+            if ( waiting.index() == 0 ) {
```
jianminzhao, line 230: std::get_if  appears more readable.

Replicator/IncomingRev.cc
callumbirks: Because it still has memory allocated, and it is not used after here, so we move ownership. It is likely the compiler does this already as an optimisation, its just better to be sure.

Replicator/Puller.cc
callumbirks: Will take a look

callumbirks: Changed to use this - agree is more readable

jianminzhao:
I understand your approach. In general, however, there is a design question as whether it's desirable to add them to the same waiting queue. The revoked revs are determined at the "changes" messages, whereas current waiting queue "rev" messages. The important difference is that handling the revoked msg does not involve BLIP communications (completely performed locally), whereas the server will wait for the response of the rev messages. Overwhelming the queue with revoked msgs may delay the handling of the rev msgs. The approach of this PR may not change the current dynamics as shown in the CBSE. Of course, you properly decrease the waiting count. If this is all it's about, you don't need to have variant wait queue, but making sure to decrease the count properly.
Need to simulate the CBSE's case to see how flood of revoked revs affect the dynamics.

Replicator/IncomingRev.cc
jianminzhao: Without move, because you shared it with _revMessage, as the function returns, the ref count will be decreased and that's all. Retained only holds a pointer.

callumbirks:
@jianminzhao Actually, this should block the server less than before. Because before, the revoked revs don't wait in the queue, but still increase `_activeIncomingRevs`, thereby blocking the normal revs as they wait in the queue. But now, if we receive some revoked rev at a similar time to normal rev, the revoked rev will wait their turn in the queue.

Replicator/IncomingRev.cc
callumbirks: Which is still memory on the stack. A move is always better than a copy. A copy here is calling extra code for no reason (the `Retained` class' copy constructor) when the instance that we're copying will be destroyed straight after anyway.

callumbirks: This scenario is the exact type of scenario that is perfect for `std::move` (when a variable will no longer be used in one place, but will now be used in another, and it is more efficient to move ownership of the existing memory than to copy)

jianminzhao: I don't think move will affect the memory footprint.It merely saves a couple of counter arithmetics.

jianminzhao:
As I noted, they come from two difference BLIP messages. As I remember, revoked comes from "changes' message, which precedes the "rev" message. If the former takes up the front positions of the queue, the later ones still have to wait. I feel the current problem is that revoked only increases the counter but, probably, not decrease it.

Replicator/IncomingRev.cc
callumbirks: Right, so even if it "merely saves a couple of counter arithmetics", why not perform it? It doesn't cost us anything.

callumbirks: And, this code is called as part of `Retained` constructor:
```c++
    __hot void assignRef(RefCounted* &holder, RefCounted *newValue) noexcept {
        RefCounted *oldValue = holder;
        if (_usuallyTrue(newValue != oldValue)) {
            if (newValue) newValue->_retain();
            holder = newValue;
            if (oldValue) oldValue->_release();
        }
    }
```
A new pointer on the stack (`oldValue`), a pointer comparison, the arithmetic of `retain` and `release`, and copying a pointer value. All small operations, but unnecessary here, and add up on each call of `handleRev`

callumbirks:
Revoked rev does decrease the count. If you look at the code, `handleRevokedDoc()` just basically calls `insertRevision()`, the same as any other rev, which eventually leads to a call to `revWasProvisionallyHandled()`, which decreases the `_activeIncomingRevs`. The customer's problem who had 40k activeIncoming was not because the count wasn't decreasing, but actually because they had a problem which caused 40k revocations.

As far as
> revoked comes from "changes' message, which precedes the "rev" message. If the former takes up the front positions of the queue, the later ones still have to wait.
Yeah, I see what you mean, as the revoked precedes the actual revs. So, another solution. Perhaps regular revs should take precedence, and revoked revs should only be handled once the regular revs are done?

callumbirks:
After discussion with @jianminzhao : Going to close this PR for now. Essentially, this approach does not change the result when revoked all come in one big `changes` message, which is the more likely scenario.
Going to write some new tests for this, and use a new approach. Most likely approach: Use a separate counter for revoked revs so they do not increase `_activeIncomingRevs`, so they do not block regular revs (which is important as SG awaits response for regular revs).

PR #1908, opened 2023-11-13
CBL-5030: Opening db of v2 into db of VersionVector causes conversion…

jianminzhao:
… of all documents.

For the legacy database of huge size, we cannot afford to enumerate and convert all documents as we open it into the database of current versioning. Fortunately, we are able to handle documents of all earlier versions, specifically TreeVersioning_v2 and TreeVersioning, in the database of current VectorVersioning. The program will upgrade those documents to the current versioning at appropriate time on demand.


LiteCore/Database/DatabaseImpl+Upgrade.cc
```diff
@@ -110,6 +110,7 @@ namespace litecore {
             }
         }
         LogTo(DBLog, "\t%" PRIu64 " documents upgraded, now committing changes...", docCount);
+        DebugAssert(docCount == 0, "Should not upgrade all documents when openning the database.");
```
snej, line 113: Why is this here? The assertion will fail unless the database is empty or already uses version vectors.

Since you are commenting out the only call to this method, why not just delete it? Actually you can delete the whole source file, I think.

C/c4Database.cc
```diff
@@ -73,7 +73,7 @@ static void ensureConfigDirExists(const C4DatabaseConfig2& config) {
 
 static C4DatabaseConfig newToOldConfig(const C4DatabaseConfig2& config2) {
     return C4DatabaseConfig{config2.flags | kC4DB_AutoCompact, NULL,
-                            (config2.flags & kC4DB_VersionVectors) ? kC4VectorVersioning : kC4TreeVersioning_v2,
+                            (config2.flags & kC4DB_VersionVectors) ? kC4VectorVersioning : kC4TreeVersioning,
```
snej, line 76: This is really strange -- I did not think we were using kC4TreeVersioning_v2 at all anymore. Looks like I did this by accident back in commit ea82fb755 in 2021.

LiteCore/Database/DatabaseImpl.cc
```diff
@@ -193,8 +193,9 @@ namespace litecore {
 
         // Yup, mismatch confirmed, so deal with it:
         if ( versDoc.exists() ) {
-            // Existing db versioning does not match runtime config!
-            upgradeDocumentVersioning(curVersioning, newVersioning, transaction());
+            // Pass for existing database. We can handle documents of old versions in
```
snej, line 196: Why keep this? Just delete the lines.

snej (changes requested):

LiteCore/Database/DatabaseImpl+Upgrade.cc
jianminzhao: Yes, I can simply delete that function. Perhaps I first #if it out and think to move it to cblite as a tool to upgrade db.

LiteCore/Database/DatabaseImpl.cc
jianminzhao: okay, will do it.

jianminzhao:
We. have tests to open dbs of all legacy versions, tree_v2, tree.

LiteCore/Database/DatabaseImpl+Upgrade.cc
jianminzhao: Perhaps I can just delete it and file a ticket to move it to cblite that mentions the commit where the function is deleted.

snej (approved):

PR #1914, opened 2023-11-21
CBL-5033: Puller revoked docs should queue with other revs

callumbirks:
**Before this change:**
High number of revocations blocks queue of incoming revs, because revocations arrive in "changes" message before the rev message. Especially bad because SG is waiting for response to incoming revs.

**After this change:**
When "rev" messages arrive, they take priority in the queue over revocations. They only consider themselves for queue limit (_activeIncomingRevs), whereas revocations consider themselves *and* regular revs.
This *does* mean there could be a brief period of 200 active `IncomingRev` workers when "rev" arrives, as the queue transitions from revocations to revs.


callumbirks:
Test was passing on local, not sure why failing... A race condition?

pasin (commented):
There is `kMaxRevsBeingRequested` which is described in the code comment as :

```
Maximum desirable number of incoming `rev` messages that aren't being handled yet. Past this number, the puller will stop handling or responding to `changes` messages, to attempt to stop getting more `revs`.
```

I think we will also need to include revoke rev processing in `kMaxRevsBeingRequested` to control the flow of the changes as well. My question is that if that is done, will it minitage the concern when there are a lot revoked rev comming and that will prevent rev messages to wait to get processed for a long time if a single queue is used for both revs and revoked-revs?

callumbirks:
Right, I see what you mean. Because if we receive a high number of revocations, they are not considered for `kMaxRevsBeingRequested`, and Puller will still accept many rugular revs. Changing that (so that revoked are considered) shouldn't make a difference to the actual order of processing. I guess it will just mean that if there are many revoked, Puller may essentially "refuse" some later changes message. But if that is the existing behaviour with regular revs, I don't see a problem with expanding the condition to include revoked.

jianminzhao:
From the code, my understanding is that, kMaxRevsBeingRequested is to help moderate SG's flow. If SG's responses to Lite's response of changes messages, namely the rev messages, is too slow, we will hold up responding the changes messages. 
The problem we are addressing here is the the converse, namely, the prabably protracted delay in responding the rev messages. The model here is the following.
After handling a changes msg, 
1. putting n revoked revToInserts to the queue
2. sending response to SG from which we expect m rev msgs.
As (1, 2)* loops, depending how fast we handle the rev messages and how fast SG sends us the expected rev msgs, the queue will hold n1 + n2 + n3 revoked revs, and m insert revs where m < m1 + m2 + m3. The problem is, if m comes late, with regard to changes msgs, m will have to wait until all n's to be completed.

callumbirks:
@jianminzhao `kMaxRevsBeingRequested` *is* to moderate SG, but if SG is being too fast (or, faster than we can handle). This is the comment in the code:
```c++
    /* Maximum desirable number of incoming `rev` messages that aren't being handled yet.
            Past this number, the puller will stop handling or responding to `changes` messages,
            to attempt to stop getting more `revs`. */
    constexpr unsigned kMaxRevsBeingRequested = 200;
```

jianminzhao:
I'm little confused. This number is decreased in ```
void Puller::startIncomingRev(MessageIn* msg) {
        _revFinder->revReceived();
```
This is also where we increment the active counter. So "beingRequested" number grows if SG sends more changes msgs than we could handle, we will hold up the responding to the changes msgs. 
@callumbirks could you look in more detail and describe the problem we want to solve. 
Looks like "beingRequested" constraint can be reached either SG slow in sending the revs or we are too slow digesting them. Is it right?

callumbirks:
Ah yeah, I guess it can be also because of SG being slow. As it won't be decrease if SG is slow, but it will increase if we are slow.
The problem we are solving is aligning the handling of revoked revs to be more similar to regular revs, but also making sure the revoked have a lower priority. Apart from the SG awaiting a response, revoked are very similar to regular revs, and should be treated as such.
The `kMaxRevsBeingRequested` acts as a throttle to SG sending "changes" when we are already full of changes to process, so that should also consider revoked, as they are taking away compute time also.

callumbirks:
Just noticed some underflow assertion failures after going back over test logs for latest commit... not sure why it didn't cause test to fail

pasin:
> So I think Pasin's concern is that SG may still send us many messages when we already have many revoked queued up?

Yes. This will control the flow of the revoked revs + revs so that we don't have too many in the queue that we cannot keep the processing up.

callumbirks:
Don't understand why suddenly the submodules have changed...

callumbirks:
The latest changes implement the results of our meeting @jianminzhao @pasin. Now, the two active counts `_activeIncomingMessages` and `_activeIncomingRevoked`, and the queues related to them `_waitingRevMessages` and `_waitingRevoked` act independently from each other. This means at any time there could be up to 100 active revs and 100 active revoked. The `pullerHasCapacity()` also considers active revoked for its result.

The function `Puller::maybeStartIncomingRevs()` contains the logic for deciding to pick waiting revs or revoked from the queue. First, it adds regular revs from the queue to the active revs until they're at the limit, then adds revoked from the queue to the active revoked until they're at the limit.

pasin (approved):

Issue #101, opened 2017-04-24
LiteCoreServ expects a db for startup


raghusarangapani:
After compiling LiteCoreServ, I tried to start it up to use the REST interface. I ran into the issue as below:
```
» ./LiteCoreServ --port 51000 --dir ~/Desktop/testdb                                                                        
Sharing all databases in /Users/raghu.sarangapani/Desktop/testdb: 
Error: No databases found

 » ./LiteCoreServ --port 51000 --dir ~/Desktop/testdb --create
Sharing all databases in /Users/raghu.sarangapani/Desktop/testdb: 
Error: No databases found
```
LiteCoreServ is expecting a db to be present beforehand to start.

LiteCoreServ should be able to start without a db and user should be able to create one through PUT /db

Issue #36, opened 2016-12-08
Very poor performance on Android


borrrden:
I know it should be slower than than desktop, but this seems a bit harsh

Tested on Nexus 5 with custom API 24 OS:

> \#\# c4AllDocsPerformanceTest.cc:
	>>> AllDocsPerformance
            SQLite, rev-trees
LiteCore-C info: Created 100000 docs
LiteCore-C info: Enumerating 100000 docs took 2768.801 ms (0.028 ms/doc)
LiteCore-C info: Created 100000 docs
LiteCore-C info: Enumerating 100000 docs took 3701.565 ms (0.037 ms/doc)
\#\# c4PerfTest.cc:
	>>> Import geoblocks
LiteCore-C info: Reading Couchbase/example-datasets-master/IPRanges/geoblocks.json ...  
LiteCore-C info: 100000  
LiteCore-C WARNING: Stopping JSON import after 15.032 sec  
LiteCore-C info: Committing...
Importing took 270512.413 ms for 191000 docs (1416.295 us/doc, or 706 docs/sec)
Reading random docs took 5798.508 ms for 100000 docs (57.985 us/doc, or 17246 docs/sec)
LiteCore-C info: Reading Couchbase/example-datasets-master/IPRanges/geoblocks.json ...  
LiteCore-C info: 100000  
LiteCore-C WARNING: Stopping JSON import after 15.112 sec  
LiteCore-C info: Committing...
Importing took 137176.062 ms for 134000 docs (1023.702 us/doc, or 977 docs/sec)
Reading random docs took 6344.108 ms for 100000 docs (63.441 us/doc, or 15763 docs/sec)

Notes:  Only one CPU core being utilized, and 18MB of RAM peak

snej:
Are you sure this is an optimized build? What optimization level is the compiler using?

borrrden:
As is starting to become par for the course here, this only happens when using libstdc++.  libc++ has acceptable performance, but crashes in the Import Names test...here is a sample of the poor performance compilation:

[...]

PR #1915, opened 2023-11-27
CBL-5073: Update replicator to keep body only on the remote revision

jianminzhao:
As the puller sets the latest revision on remote, we should set the flag, KeepBody, on the new revision. This will incidentally clear the KeepBody flag on its ancestors.
Also, fixed a bug of failure to assign the remote DB ID for the passive replicator.

### DIFF
--- a/LiteCore/Database/TreeDocument.cc
+++ b/LiteCore/Database/TreeDocument.cc
@@ -583,6 +583,7 @@ namespace litecore {
                           SPLAT(newRev->revID.expanded()), effect);
                 }
                 _revTree.setLatestRevisionOnRemote(rq.remoteDBID, newRev);
+                _revTree.keepBody(newRev);
             }
 
             if (!saveNewRev(rq, newRev, (commonAncestor > 0 || rq.remoteDBID))) {
--- a/Replicator/Inserter.cc
+++ b/Replicator/Inserter.cc
@@ -146,16 +146,10 @@ namespace litecore { namespace repl {
                         }
                     };
                     put.deltaCBContext = this;
-                    // Preserve rev body as the source of a future delta I may push back:
-                    put.revFlags |= kRevKeepBody;
                 } else {
                     // If not a delta, encode doc body using database's real sharedKeys:
                     bodyForDB = _db->reEncodeForDatabase(rev->doc);
                     rev->doc = nullptr;
-                    // Preserve rev body as the source of a future delta I may push back:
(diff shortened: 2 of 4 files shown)


pasin (commented):
I'm testing this PR with the db from the customer using my own SG which has a different URL. Now the database size was reduced from 100M to 5M correctly after pulling the docs with modified data from SG. However, for the case of having multiple remote IDs, the body is kept on only one remote rev. Should the body be kepth on each remote rev? Here is an example rev tree:

Document "0465fa2e-6898-45f2-82cf-6396f9e7a8a0"
 * 1-f165f80b7c5fd0e69d1662e4af13ddaacb77c650     ----- #36  [REMOTE#1]
   * 2-af28150fe6819cb7e6aa504993ced7222d9c3335   ---KL #101, 977KB  [CURRENT]  [REMOTE#2]
[REMOTE#1] = ws://localhost:4984/billeo-db/_blipsync
[REMOTE#2] = ws://192.168.86.26:4984/posts/_blipsync

pasin:
Another thing, we are getting into an awkward situation we will do the 3.0.17 MP for the customer instead. This would mean that we may not be able to merge this PR after review is completed.

callumbirks:
@pasin From what I can find, the only assertion we have is that we won't *prune* (remove the revID from rev-tree) any current remote revs, nothing about keeping the body.
From `wiki/Revision-Trees.md`:
> The RevTree also keeps track of which revision is the last-known current revision on each _remote_ database that the local database has replicated with. This is needed for the “no-conflicts” mode of replication, where a document being pushed must identify its remote ancestor.

> The remote-revision property is logically a flag on a revision, but there can be an arbitrary number of these flags — one per remote database — so instead it’s represented as a mapping from remote IDs (small integers) to Revs. The Database class in turn manages a mapping from remote database URLs to these remote IDs.

> A Rev that is a remote revision will never be pruned, since the replicator needs to know its ID. Instead, revisions _above_ it will be pruned. This leads to gaps in generations: the parent of a generation-_n_ Rev may have a generation less than _n-1_. This shouldn’t be a problem.

jianminzhao:
For question r. multiple remotes in the same branch. Currently with RevTree, we keep the latest body that has been synced with a remote. So that case is "by-design." If we later pull from the earlier remote, the remote may give the delta src that we don't have. This is allowed by the protocol. We will request the remote to send a new rev with full body.

pasin (approved):
From meeting with @jianminzhao, this is already an existing behavior that we keep the body on one remote rev.

PR #1917, opened 2023-11-30
CBL-4913: Handling blobs that reference attachments that are already …

jianminzhao:
…deleted in SG. (#1885)

If the rev pulled down from SG includes top level property "_attatchments", it should include all attachments that exist in the SG for the current revision. If a blob in the revision's body references an attachment that is not in "_attachments," "getAttachment" is bound to fail. We try to detect this case and, instead of handling the error response from SG, we err out in CBL.

The above solution may result in different behavior whether the delta sync is enabled or not. It is due to false detection whether the delta revision can affect blobs without checking with the current revision. We fixed it.

### DIFF
--- a/Replicator/IncomingRev.cc
+++ b/Replicator/IncomingRev.cc
@@ -18,6 +18,7 @@
 #include "StringUtil.hh"
 #include "c4BlobStore.hh"
 #include "Instrumentation.hh"
+#include "FleeceImpl.hh"
 #include "fleece/Mutable.hh"
 #include <atomic>
 #include <deque>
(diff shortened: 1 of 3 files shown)


jianminzhao:
Mostly auto merged

pasin:
It's been a long time as I don't remember all of the details we have discussed about the solution. Is this PR ported from a branch or another PR? Can you point me to the original PR so that I can read all of the previous discussion?

jianminzhao:
Yes, from CBL-4898, or release/3.1#9756e2bdae8a7

pasin (approved):

PR #1919, opened 2023-12-06
CBL-5226: Update replicator to keep body only on the remote revision.

jianminzhao:


jianminzhao:
Because of significant difference between release/3.0 and release/3.1, I manually ported from release/3.1, 26eca4791a4

jianminzhao:
Also fixed a bug in the test fixture that kept all tests in [.SyncServer] from passing.

pasin (approved):

PR #1921, opened 2023-12-09
CBL-5216: Support String Concatenation Function in SQL++

jianminzhao:


LiteCore/tests/N1QLParserTest.cc
```diff
@@ -268,6 +268,9 @@ TEST_CASE_METHOD(N1QLParserTest, "N1QL functions", "[Query][N1QL][C]") {
 
     CHECK(translate("SELECT count(*)") == "{'WHAT':[['count()',['.']]]}");
     CHECK(translate("SELECT count(db.*)") == "{'WHAT':[['count()',['.db.']]]}");
+    CHECK(translate("SELECT concat(a, b)") == "{'WHAT':[['concat()',['.a'],['.b']]]}");
+    CHECK(translate("SELECT concat('hello', \"world\", ' ', concat(true, 123.45 , sin(1)))")
```
pasin, line 272: Can concat take more than 2 params? and What will be the result of `concat(true, 123.45 , sin(1))`?

jianminzhao: yes, it may take any number of arguments (s.t. SQLite's limitation of number of arguments). concat(true, 123.45. sin(1)) should yield to "true123.450.84147". The last segment, "0.84147" depends on actual precision used by the runtime library.

pasin (approved):

PR #1922, opened 2023-12-11
CBL-5254: Upsert performance is degraded when the number of docs is increase

callumbirks:
Create sequence index on datafile open rather than lazily with replicator start. This means c4doc_update calls won't be dastardly slow before first repl.


callumbirks:
Hm not sure why theres an extra commit in there, gonna have to reset the branch

C/tests/c4DatabaseTest.cc
```diff
@@ -1474,3 +1474,61 @@ N_WAY_TEST_CASE_METHOD(C4DatabaseTest, "Re-open database with an index", "[Datab
...
+    expr;                                                                      \
+    auto end = std::chrono::system_clock::now();                               \
+    return end - start;                                                        \
+  }()
+
+// CBL-5241 / CBL-5254
+// Before the fix associated with this ticket, c4doc_update performs a SCAN query 
+// when fetching rev-tree, due to the lack of sequence index. This means the update
+// would get increasingly slower for every doc existing in the DB.
+N_WAY_TEST_CASE_METHOD(C4DatabaseTest, "c4doc_update sequence index perf", "[Database][C]") {
```
pasin, line 1490: I don't think we should add the perf test in the test here. The perf test should be in c4PerfTest.cc here. 

The fix itself is good enough and you can create another ticket to put the test in c4PerfTest and also showfast as a doc update perferformance test.

pasin (changes requested):

pasin (changes requested):
When always creating the sequence index, shall we remove the code in the replicator that creates the sequence index?

LiteCore/Storage/SQLiteDataFile.cc
```diff
@@ -557,6 +557,11 @@ namespace litecore {
 
             keyStore->addExpiration();
             deletedStore->addExpiration();
+
+            // Add index on sequences
+            // Requiement for RevTreeRecord::read()
+            keyStore->createSequenceIndex();
+            deletedStore->createSequenceIndex();
```
pasin, line 564: Do we also need the sequence index for the deleted table? Basically, do we have query by sequence on the deleted table?

pasin:
PR Validation hung on `c4ThreadingTest.cc::Threading CreateVsEnumerate` test. Not sure if this is related to the PR.

LiteCore/Storage/SQLiteDataFile.cc
callumbirks: From the Profiling report, RevTreeRecord::read does call BothKeyStore::read, which proceeds to call read on the live and dead store. Based on the code in RevTreeRecord::read, I think those would both be read with Sequence.

callumbirks:
Based on a conversation I had with @snej, moved the index creation to the SQLiteKeyStore::read function.

jianminzhao:
Is the PR comment, "Create sequence index on datafile open rather than lazily with replicator start. This means c4doc_update calls won't be dastardly slow before first repl," still valid?

jianminzhao (approved):

jianminzhao:
Please update the commit comments before squash merge.

callumbirks:
> Is the PR comment, "Create sequence index on datafile open rather than lazily with replicator start. This means c4doc_update calls won't be dastardly slow before first repl," still valid?

Kind of. Technicallly now we create the index on the first read from the database. I will update the PR comment to make sure its accurate to the final solution

pasin (approved):

PR #1924, opened 2023-12-18
CBL-5125: Update replicator to keep body only on the remote revision

jianminzhao:
As the puller sets the latest revision on remote, we should set the flag, KeepBody, on the new revision. This will incidentally clear the KeepBody flag on its ancestors. Also, fixed a bug of failure to assign the remote DB ID for the passive replicator.


jianminzhao:
Ported from CBL-5073, release/3.1

pasin:
What happens with PR validation especially with c4ThreadingTest.cc:Threading CreateVsEnumerate test. That test is hanging everywhere.

callumbirks (approved):

PR #1932, opened 2024-01-24
Added clustering and PQ to vector index options [API change]

snej:
- Now supports Inverted Multi-Index clustering, which improves training speed a lot and can improve search quality
- Now supports PQ encoding, which shrinks the on-disk size, at the cost of search quality
- Overhauled the index spec struct; see c4IndexTypes.h.
- The vectorsearch extension changed the syntax of its config (the args to CREATE VIRTUAL TABLE) so this updates the generation of that string.

### DIFF
--- a/C/include/c4IndexTypes.h
+++ b/C/include/c4IndexTypes.h
@@ -31,28 +31,50 @@ typedef C4_ENUM(uint32_t, C4IndexType){
 };
 
 /** Distance metric to use in vector indexes. */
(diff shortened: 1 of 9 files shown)


pasin (approved):

pasin:
I saw that it's still a draft. Also C4VectorIndexOptions doesn't have dimensions yet.

pasin (approved):

PR #1936, opened 2024-01-30
CBL-5241: Upsert performance is degraded when the number of docs is increase

callumbirks:


LiteCore/Storage/SQLiteKeyStore.cc
```diff
@@ -231,6 +231,9 @@ namespace litecore {
     }
 
     bool SQLiteKeyStore::read(Record& rec, ReadBy by, ContentOption content) const {
+        //  This statement does nothing if the sequence index has already been created.
+        if ( by == ReadBy::Sequence ) ((SQLiteKeyStore*)(this))->createSequenceIndex();
```
snej, line 235: Please use const_cast.
```suggestion
        if ( by == ReadBy::Sequence ) const_cast<SQLiteKeyStore*>(this)->createSequenceIndex();
```

snej (changes requested):

LiteCore/Storage/SQLiteKeyStore.cc
callumbirks: 😮 I didn't even know that existed, have never needed to cast const to non-const before

pasin (approved):

PR #1939, opened 2024-02-06
CBL-5355: Performance test changes batch sizes and reevaluate default

callumbirks:
Add an option to Replicator config dict to configure the changes batch size. This will enable performance testing to find the best value.
Quote from ticket:
> It would be useful to expose that even as an unsupported/hidden replicator config option so we're able to tune the batch size with the perf testing team to determine a new default value.

### DIFF
--- a/Replicator/ReplicatorTuning.hh
+++ b/Replicator/ReplicatorTuning.hh
@@ -85,7 +85,7 @@ namespace litecore { namespace repl {
         constexpr unsigned kMaxRevBytesAwaitingReply = 2*1024*1024;
 
         /* Number of changes to send in one "changes" msg */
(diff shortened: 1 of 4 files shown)


Replicator/ReplicatorTuning.hh
```diff
@@ -85,7 +85,7 @@ namespace litecore { namespace repl {
         constexpr unsigned kMaxRevBytesAwaitingReply = 2*1024*1024;
 
         /* Number of changes to send in one "changes" msg */
-        constexpr unsigned kDefaultChangeBatchSize = 200;
+        //constexpr unsigned kDefaultChangeBatchSize = 200;
```
jianminzhao, line 88: Simply delete it if not used.

jianminzhao (approved):

PR #1940, opened 2024-02-07
CBL-5345: Test vector match with join

jianminzhao:


LiteCore/tests/VectorQueryTest.cc
```diff
@@ -122,4 +122,80 @@ N_WAY_TEST_CASE_METHOD(SIFTVectorQueryTest, "Query Vector Index", "[Query][.Vect
     Log("done");
 }
 
+N_WAY_TEST_CASE_METHOD(SIFTVectorQueryTest, "Query Vector Index with Join", "[Query][.VectorSearch]") {
```
borrrden, line 125: We had discussed leaving some top level comments about what specifically a test is doing.  I'd like to see that here.  Also I am not sure this is the most appropriate place for testing vector search.  It is another product separate from LiteCore.  That is, by definition, integration testing.

borrrden (changes requested):

LiteCore/tests/VectorQueryTest.cc
jianminzhao: LiteCore serves as a test bed for Vector Search. It is little awkward now for they can only tested now in my local machine where the extension is installed. It's not tested in any of the build pipeline. I am also thinking how to test it as CI (continuous integration). Before then, let's park the tests here. We are in similar position before as we test SG before we used docker in the GH pipeline.

borrrden (approved):

PR #1944, opened 2024-02-09
CBL-5372: Check version of vector search library

borrrden:
Using this will require vector search 1.0.0-12 or higher.


borrrden:
NOTE: The bad conditions such as not found / mismatched versions when specifying the extension directory are now thrown as exceptions so that db open will quickly fail.

bmeike (approved):
I can't say anything sensible about this.  Willing to try it out.

LiteCore/Storage/SQLiteDataFile.cc
```diff
@@ -189,25 +195,41 @@ namespace litecore {
 
     SQLiteDataFile::~SQLiteDataFile() { close(); }
 
+    // A lot of this logic can be turned into a reusable function later if needed
+    // for more extensions.
     static void LoadVectorSearchExtension(sqlite3* sqlite) {
 #ifdef COUCHBASE_ENTERPRISE
+        static const char* extensionName = "CouchbaseLiteVectorSearch";
```
jianminzhao, line 202: maybe constexpr

LiteCore/Storage/SQLiteDataFile.cc
```diff
@@ -189,25 +195,41 @@ namespace litecore {
...
         if ( sExtensionPath.empty() ) return;
 
         // First enable extension loading (for security reasons it's off by default):
         int rc = sqlite3_db_config(sqlite, SQLITE_DBCONFIG_ENABLE_LOAD_EXTENSION, 1, NULL);
         if ( rc != SQLITE_OK ) {
-            LogToAt(DBLog, Warning, "Unable to enable SQLite extension loading: err %d", rc);
-            return;
+            LogToAt(DBLog, Error, "Unable to enable SQLite extension loading: err %d", rc);
```
jianminzhao, line 208: I guess this the style you put forth as an one of the logging wishes, log and throw. But _throw will also log, right? and with the unique custom message, we can easily find the location where the error occurs, or adding __FILE__/__LINE__, right?

jianminzhao (approved):

LiteCore/Storage/SQLiteDataFile.cc
borrrden: This will make no difference since there is no evaluation that can be done on this at compile time.  It's memory address is used, which collapses it into a `static const char *` anyway,

LiteCore/Storage/SQLiteDataFile.cc
borrrden: `_throw` only logs if it is set to log and as for adding file and line, I think it would be more appropriate to get the macro itself to do it somehow (i.e. define the macro such that it takes a file and line number)

PR #1946, opened 2024-02-10
CBL-5344: Test vector_match with FTS match

jianminzhao:


pasin (commented):
We also need to test using vector_match() and FTS match() without doing join.

jianminzhao:
Do you mean vector_match and FTS match on the same collection, right?

pasin (approved):

PR #1949, opened 2024-02-13
Lazy vector index updating

snej:
Adds API that enables asynchronous computation and updating of a vector index. Useful when generating the vector for a document requires a network request or is slowing down document updates.

To create a "lazy" vector index that is updated asynchronously:
1. Set the (new) `C4VectorIndexOptions.lazy` property to `true`.
2. The `indexSpec` parameter -- the expression to be indexed -- should be not the vector itself but whatever value you need in order to compute or request the vector.

A lazy index doesn't update automatically when documents are created or updated. (But it does update when a document is deleted or purged, to remove the corresponding index entry.) If you query it after creating it, you won't get any results.

You (the app) decide when to update the index, probably either in response to a collection change notification or when about to query the index.

1. Call the new `C4Collection.getIndex` method to get a `C4Index` reference to the index. (It's best to keep this reference around instead of releasing it, otherwise it has to recompile some queries every time.)
2. Call `beginUpdate` on the index. This will return a `C4IndexUpdater` reference, or will return `NULL` if the index is already up-to-date.
3. The updater has a list of `FLValue`s which are the values of the expression you gave when you created the index. For each of these, create a vector somehow and call the `setVector` method to at the corresponding index.
4. When done, call the updater's `finish` method. This must be done in a transaction as it updates the database.
5. Finally release the updater; it cannot be reused.


C/Cpp_include/c4Index.hh
```diff
@@ -0,0 +1,92 @@
...
+    /// Sets the vector for the i'th value. If you don't call this, it's assumed there is no
+    /// vector, and any existing vector will be removed upon `finish`.
+    void setVectorAt(size_t i, const float* vector, size_t dimension);
+
+    /// Updates the index with the computed vectors, removes any index rows for which no vector
+    /// was given, and updates the index's latest sequence.
+    /// @note  Must be called in a database transaction.
+    /// @returns  True if the index is now completely up-to-date; false if there have been
+    ///           changes to the Collection since the LazyIndexUpdate was created.
+    bool finish();
```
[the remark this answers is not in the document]
pasin: When I didn't read the @note, I thought that calling beginUpdate() would also open a transaction and calling finish() would end the transaction(). and I wondered if there should be a parameter to commit or rollback or not. 

So does this mean that the updated vectors will be in the memory until calling finish? 
Also to cancel the updater, just not calling finish but release the updater?

C/Cpp_include/c4Index.hh
```diff
@@ -0,0 +1,92 @@
...
+    /// The number of vectors to compute.
+    size_t count() const;
+
+    /// Returns the i'th value to compute a vector from.
+    /// This is the value of the expression in the index spec.
+    FLValue valueAt(size_t i) const;
+
+    /// Sets the vector for the i'th value. If you don't call this, it's assumed there is no
+    /// vector, and any existing vector will be removed upon `finish`.
+    void setVectorAt(size_t i, const float* vector, size_t dimension);
```
pasin, line 72: So for the case that the remote ML model is offline (e.g. no internet connection) during the updating process,  do I understand correctly that users will not call this function so the docs will be included again next time?

pasin (commented):
Added another comment about the offline ML model case.

C/Cpp_include/c4Index.hh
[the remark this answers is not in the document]
snej: You're right, that's confusing. I'll clarify the comment.

Also, since this is an asynchronous API it can't hold a transaction open the whole time! `beginUpdate` does not open a transaction.

I can easily change `finish` so it wraps its work in a `C4Database::Transaction`.

> So does this mean that the updated vectors will be in the memory until calling finish?

Yes.

> Also to cancel the updater, just not calling finish but release the updater?

Yes.

LiteCore/Query/LazyIndex.cc
```diff
@@ -0,0 +1,207 @@
...
+        return FLValue(_enum->columns()[kValueCol]);
+    }
+
+    void LazyIndexUpdate::setVectorAt(size_t i, const float* vec, size_t dimension) {
+        AssertArg(i < _count);
+        AssertArg(vec != nullptr);
+        AssertArg(dimension > 0);
+        if ( _dimension == 0 ) {
+            _dimension = dimension;
+        } else if ( dimension != _dimension ) {
```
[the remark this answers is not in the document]
snej: Currently that information isn't available to CBLite after the index is created; it doesn't get saved in the `indexes` table.

C/Cpp_include/c4Index.hh
snej: This iteration of the implementation doesn't support temporarily skipping a document; it just uses a `lastSequence` number to remember which documents are up to date. I'll create another PR after this one to support it.

In this PR I'll add a `skipVectorAt()` method to make it explicit that a vector is being skipped.

snej:
Huh, the Jenkins build failed on Android:
```
couchbase-lite-core/LiteCore/Query/LazyIndex.cc:133:15: error: call to member function 'bind' is ambiguous
         _ins->bind(1, rowid);
```
Looks like that issue with `int64_t` being `long` on Linux instead of `long long` ... but why didn't the GitHub CI Ubuntu build fail with the same error?

C/include/c4Index.h
```diff
@@ -107,6 +107,39 @@ NODISCARD CBL_CORE_API bool c4db_deleteIndex(C4Database* database, C4String name
...
+
+/** Returns the i'th value to compute a vector from.
+    This is _not_ the entire document, just the value of the expression in the index spec. */
+CBL_CORE_API FLValue c4indexupdater_valueAt(C4IndexUpdater*, size_t i) C4API;
+
+/** Sets the vector for the i'th value. If you don't call this, it's assumed there is no
+    vector, and any existing vector will be removed upon `finish`. */
+CBL_CORE_API bool c4indexupdater_setVectorAt(C4IndexUpdater*, size_t i, const float vector[_Nonnull], size_t dimension,
+                                             C4Error* outError) C4API;
+
```
pasin, line 134: Missing `c4indexupdater_skipVectorAt() API`.

pasin (changes requested):

C/include/c4Index.h
snej: Added it!

pasin (approved):

PR #1950, opened 2024-02-13
CBL-5335: array_agg seem to fail under some circumstances

callumbirks:
Add test for array_agg failure.
Make sure scopes are constructed and stay alive for Query Fleece values.

Many thanks to @snej who found the exact issue and gave me the fix.


callumbirks:
Essentially, in some cases the Fleece values given to the Query don't keep their sharedkeys around. The `fleeceParam()` function constructed a scope, but it only lived for the duration of the function.
The function is replaced with an RAII class, so the scope (and sharedkeys) will be alive long enough to be used in the Query.

jianminzhao (approved):

PR #1952, opened 2024-02-16
Use NODISCARD to catch mistakes

snej:
- Updated Fleece to pick up commit 16704a69 "Added NODISCARD annotations to catch mistakes", which adds a `NODISCARD` attr, informally deprecates `MUST_USE_RESULT`, and annotates a bunch of functions with `NODISCARD` or `[[nodiscard]]`.
- Added `NODISCARD` to a bunch of C API functions where discarding the result would leak memory or ignore potential errors.
- Changed use of `MUST_USE_RESULT` to `NODISCARD` (in C) or `[[nodiscard]]` (in C++).
- Fixed compile errors caused by the above, mostly by wrapping with CHECK() or REQUIRE() in tests.


jianminzhao:
@snej are you saying ignoring returned alloc_slice would lead to memory leak? Is it definitive or probably?

C/include/c4Database.h
```diff
@@ -171,12 +173,12 @@ CBL_CORE_API bool c4db_maintenance(C4Database* database, C4MaintenanceType type,
...
 /** Begins a transaction.
         Transactions can nest; only the first call actually creates a database transaction. */
-CBL_CORE_API bool c4db_beginTransaction(C4Database* database, C4Error* C4NULLABLE outError) C4API;
+NODISCARD CBL_CORE_API bool c4db_beginTransaction(C4Database* database, C4Error* C4NULLABLE outError) C4API;
 
 /** Commits or aborts a transaction. If there have been multiple calls to beginTransaction, it
         takes the same number of calls to endTransaction to actually end the transaction; only the
         last one commits or aborts the database transaction. */
-CBL_CORE_API bool c4db_endTransaction(C4Database* database, bool commit, C4Error* C4NULLABLE outError) C4API;
+NODISCARD CBL_CORE_API bool c4db_endTransaction(C4Database* database, bool commit, C4Error* C4NULLABLE outError) C4API;
```
velicuvlad, line 181: CBL Swift complains about this NODISCARD on db_endTransaction. If removed, it builds just fine on both iOS and macOS

velicuvlad (changes requested):

C/include/c4Database.h
snej: That's strange, because it's syntactically just like many of the other changes, `NODISCARD CBL_CORE_API <type> ...`.
What is the exact error?

velicuvlad: fixed on platform by just casting void to the destructor as discussed

jianminzhao (approved):

PR #1959, opened 2024-02-24
CBL-5390: Add parentObj to class Logging

jianminzhao:
- Logging::setParentObjectRef(parentObjRef)
- For objects derived from Worker, set parent object according to the parent relationship of Worker.
- Applied parent object for a non-Worker (Connection) to Replicator.
- Includes day in the timestamp.
- Removed brackes of domain tag in the log, namely [Sync] to Sync.
- Disabled several tests that check the logs. They will be reinstated at the end of logging enhancement effort.

### DIFF
--- a/LiteCore/Support/LogEncoder.cc
+++ b/LiteCore/Support/LogEncoder.cc
@@ -67,7 +67,7 @@ namespace litecore {
 
 #pragma mark - LOGGING:
 
-    void LogEncoder::log(const char* domain, const std::map<unsigned, std::string>& objectMap, ObjectRef object,
+    void LogEncoder::log(const char* domain, const LogDomain::ObjectMap& objectMap, ObjectRef object,
                          const char* format, ...) {
         va_list args;
         va_start(args, format);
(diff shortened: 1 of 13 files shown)


jianminzhao:
Examples:
2024-02-23T18:43:18.062004 Sync Obj=/Repl#167/inserter#169/ {Coll#0} Inserted   6 revs in  49.26ms (  122/sec) of which 65.2% was commit
2024-02-23T18:43:18.244072 BLIP Obj=/Repl#164/Connection#163/ Closing with code 1000, msg ''
(Connection is a Worker)
2024-02-23T18:43:18.783885 Sync Obj=/Repl#210/pusher#211/ {Coll#0} Peer is pulling changes from seq #0

borrrden:
Is this UTC time?  If so is there supposed to be a Z at the end of the timestamp?

LiteCore/Support/LogEncoder.cc
```diff
@@ -99,7 +99,7 @@ namespace litecore {
             if ( i == objectMap.end() ) {
                 _writer.write({"?\0", 2});
             } else {
-                _writer.write(slice(i->second.c_str()));
+                _writer.write(slice(i->second.first.c_str()));
```
borrrden, line 102: `second.first` is a hard to read code smell

borrrden (approved):
The effect is nice but the whole map inside a map thing decreases readability

pasin (approved):
`/Repl#167/inserter#169` : Prefer to have naming format consistency (case wise). This can be done as a separate task and not high priority.

PR #1962, opened 2024-02-27
CBL-5448: Attachments flag dropped when applying delta to incoming rev

callumbirks:
`hasBlobReferences` is copied in from master.
You can show the fix works locally, if you comment out the 3 lines added in `IncomingRev.cc`, the test fails

### DIFF
--- a/Replicator/IncomingRev.cc
+++ b/Replicator/IncomingRev.cc
@@ -221,7 +221,7 @@ namespace litecore { namespace repl {
         // First create a Fleece document:
         Doc fleeceDoc;
         C4Error err = {};
-        bool didApplyDelta;
+        bool didApplyDelta = false;
         if (_rev->deltaSrcRevID == nullslice) {
             // It's not a delta. Convert body to Fleece and process:
             FLError encodeErr;
--- a/Replicator/IncomingRev.cc
+++ b/Replicator/IncomingRev.cc
@@ -181,6 +181,9 @@ namespace litecore { namespace repl {
         _mayContainEncryptedProperties = !_options->disablePropertyDecryption()
                                          && MayContainPropertiesToDecrypt(jsonBody);
 
+        logVerbose("_mayContainBlobChanges=%d", _mayContainBlobChanges);
+        logVerbose("_mayContainEncryptedProperties=%d", _mayContainEncryptedProperties);
+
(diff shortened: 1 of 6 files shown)


Replicator/IncomingRev.cc
```diff
@@ -359,6 +359,10 @@ namespace litecore { namespace repl {
             }
         }
 
+        if (!(_rev->flags & kRevHasAttachments) && _db->hasBlobReferences(root)) {
+            _rev->flags |= kRevHasAttachments;
+        }
```
pasin, line 364: As commented in the slack, I think we should reuse the logic in line 330 - 343, as we don't need to find blob for the case that delta sync is not used AND no blobs changes.

The change could be : 

```
if (_mayContainBlobChanges || <didApplyDelta>) {
            _db->findBlobReferences(root, true, [=](FLDeepIterator i, Dict blob, const C4BlobKey &key) {
                _rev->flags |= kRevHasAttachments;
                if (_mayContainBlobChanges) {
                   _pendingBlobs.push_back({_rev->docID,
                                         alloc_slice(FLDeepIterator_GetPathString(i)),
                                         key,
                                         blob["length"_sl].asUnsigned(),
                                         C4Blob::isLikelyCompressible(blob)});
                   _blob = _pendingBlobs.begin();
                }
            });
}
```

pasin (changes requested):

Replicator/IncomingRev.cc
```diff
@@ -218,6 +221,7 @@ namespace litecore { namespace repl {
         // First create a Fleece document:
         Doc fleeceDoc;
         C4Error err = {};
+        bool didApplyDelta;
```
jianminzhao, line 224: initializae

callumbirks: Booleans are already initialized (so are all integral types).
Just structs, classes, etc are uninitialized

jianminzhao: I doubt it. At least GPT does not think so.

callumbirks: Oh I see, Xcode doesn't like it regardless of that.

jianminzhao (approved):

pasin (approved):

pasin (approved):

PR #1963, opened 2024-02-28
CBL-5449: Attachments flag is dropped when applying delta to incoming rev

callumbirks:
Port from release/3.1, CBL-5448.
With the addition of @pasin's optimisation; `else if ( didApplyDelta )`


callumbirks:
Force-pushed to rebase on master and fix EE builds failing from logPrefix errors

PR #1966, opened 2024-02-29
Rewritten query translator

snej:
Total rewrite of QueryParser. The new version is called QueryTranslator. It is much, much cleaner and more extensible.

It operates in three passes:
1. Traverse the JSON/Fleece tree creating a tree of Node objects
2. Do a bit of postprocessing on the Node tree
3. Traverse the Nodes, writing SQL to an output stream.

There's (a bit) more documentation in docs/QueryTranslator.md .


borrrden:
Hmmm I find this unreviewably large so I think I'll schedule another live review.

borrrden (approved):

PR #1967, opened 2024-03-04
CBL-5391: Consistent log format

jianminzhao:
* Changed the name of the default log domain to "Default" from empty string.
* Alwasy show log levels. (Current we don't show Verbase and Info)
* File log and log callback share the same format of obj path.
* Add the end mark for "Replaying binary logs ...". It only apply to unit tests.


LiteCore/Support/DefaultLogger.cc
```diff
@@ -13,5 +13,5 @@
 #include "Logging.hh"
 
 extern "C" {
-CBL_CORE_API litecore::LogDomain kC4Cpp_DefaultLog("", litecore::LogLevel::Info);
+CBL_CORE_API litecore::LogDomain kC4Cpp_DefaultLog("Default", litecore::LogLevel::Info);
```
pasin, line 16: Ideally, I would prefer no or very minimum logs shown in "Default" domain.

pasin (approved):

borrrden (approved):

callumbirks (approved):

PR #1970, opened 2024-03-06
Add debug intruments for CBSE-16470

jianminzhao:
- Add logs to the critical path to c4query_run
- Add instruments to detect when column values are freed as column is encoded into JSON.

### DIFF
--- a/LiteCore/Query/SQLiteQuery.cc
+++ b/LiteCore/Query/SQLiteQuery.cc
@@ -263,6 +263,15 @@ namespace litecore {
         {
             logInfo("Created on {Query#%u} with %llu rows (%zu bytes) in %.3fms",
                 query->objectRef(), rowCount, recording->data().size, elapsedTime*1000);
+
+            if (_iter[0u]) {
+                Array::iterator iter = columns();
+                for (; iter; ++iter) {
+                    if (iter.value()) {
+                        FleeceLogCB::log("column %p", fleece::eColumnValue, iter.value());
+                    }
+                }
+            }
         }
 
         ~SQLiteQueryEnumerator() {
@@ -271,6 +280,10 @@ namespace litecore {
 
         virtual int64_t getRowCount() const override {
             return _recording->asArray()->count() / 2;  // (every other row is a column bitmap)
+        }
+
+        virtual slice getScope() const override {
+            return _recording->data();
         }
 
         virtual void seek(int64_t rowIndex) override {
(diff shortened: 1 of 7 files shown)


LiteCore/Query/SQLiteQuery.cc
```diff
@@ -263,6 +263,15 @@ namespace litecore {
         {
             logInfo("Created on {Query#%u} with %llu rows (%zu bytes) in %.3fms",
                 query->objectRef(), rowCount, recording->data().size, elapsedTime*1000);
+
+            if (_iter[0u]) {
```
pasin, line 267: Is this OK if there is nothing in the array?

LiteCore/Query/SQLiteQuery.cc
```diff
@@ -480,10 +499,12 @@ namespace litecore {
                         throw;
                 }
             }
+            LogTo(SQL, "bindParameters ends");
```
pasin, line 502: What is the default log level of the SQL domain? 

@bmeike I don't remember if enabling verbose log from the platform will also enable SQL domain.

LiteCore/Support/Logging.cc
```diff
@@ -57,6 +57,83 @@ struct ScopedSetter {
     }
 };
 
+
+litecore::LogDomain FleeceLogCB::FLC("Fleece");
+std::unordered_map<const void*, fleece::slice> FleeceLogCB::sQueryEnumNew;
+std::unordered_set<const void*> FleeceLogCB::sColumns;
+std::mutex FleeceLogCB::mutex;
+
+void FleeceLogCB::log(const char* msg, int type, const void* pointer) {
```
pasin, line 66: What is the purpose of this log function? What is your hypothesis?

LiteCore/Query/SQLiteQuery.cc
jianminzhao: It's okay, just there is no columns i

LiteCore/Query/SQLiteQuery.cc
jianminzhao: I did not check. but it definitely work for when I set the callback level to Info

borrrden:
This looks like it will provide some solid information, assuming that the hypothesis of the problem is in the ballpark.

LiteCore/Query/SQLiteQuery.cc
pasin: Oh, this _iter is for the columns. I see.

LiteCore/Support/Logging.cc
pasin: Now I understand.

PR #1971, opened 2024-03-06
Feature/connected client rebased

callumbirks:
This is far from building, many errors. Figured putting up as a PR gives easier way to review the differences between connected-client, and this branch which has been updated to the latest master.
There were many merge conflicts merging master into this branch.


callumbirks:
If anyone stumbles on this PR, the consensus is the Connected Client features should be redone on top of current master, rather than attempting to bring such an out-of-date branch up-to-date, which would be heavily error-prone.

PR #1973, opened 2024-03-07
CBL-5336: Over the bound of FLDicIterator should be banned

jianminzhao:
Revert the change per CBL 5044 in Fleece.


jianminzhao:
The latest fleece that has been pulled in includes the one this PR intends to pull in.

PR #1974, opened 2024-03-08
CBL-5466: Throw NoSuchIndex if querying a missing vector index

snej:


pasin (approved):

pasin:
I have tested the fix with the test in CBL tests and the test was passed. I'm merging this PR now.

PR #1977, opened 2024-03-11
CBL-5438: DateTime standard format parser

callumbirks:
Description copied from Fleece PR:
A proper parser for DateTime formats, rather than just parsing format string as a DateTime.
This has benefits such as additional flexibility in future, but also allows us to keep consistent with the strange "almost-ISO" date format we shipped with [CBL-3](https://issues.couchbase.com/browse/CBL-3).1, because keeping format as a DateTime does not allow us to optionally use colon or no-colon for Timezone offset (i.e. +0500 / +05:00).
I added in parsing for a "subset" of https://howardhinnant.github.io/date/date.html#to_stream_formatting, but it's not necessary and if anyone thinks we don't need it, happy to take it out. Think it's a lot better than the weird "1111-11-11T11:11......` format we copied from Server though.


callumbirks:
Right now this points to the CBL-5438 Fleece branch, for sanity checks.
See Fleece PR https://github.com/couchbase/fleece/pull/213

callumbirks:
Wanted to change to point to 3.2 branch, but it altered the commits. So I'll create another PR later to cherry-pick this one to 3.2

PR #1978, opened 2024-03-12
CBL-5498: Util changes for ConnectedClient

callumbirks:
@snej I tried to make it as simple as possible. So it's just `Error` changes, `Result`, `StringUtil::matchGlobPattern`, `QuietReporter` changes, and `ResultTest`. My goal is to be able to just add in ConnectedClient (and whatever Replicator etc. changes go with that) in one PR. There were some additional stuff in the original PR, like separating `_attachments` metadata functions to `legacy_attachments`, and more, but I've left out as much as possible so it's just the bare essentials. Anything you think is missing from the original branch that is necessary for ConnectedClient?


snej (approved):

borrrden:
You've broken the UWP build but that doesn't matter since we don't support it anymore....someone should really review and merge my PR that removes UWP support.

callumbirks:
@borrrden Thanks, I've approved the UWP PR and re-ran the Jenkins on it. Will try to get it merged today.

PR #1979, opened 2024-03-13
CBL-5450: Remote rev KeepBody flag could be cleared accidentally

callumbirks:
This PR will probably have to sit here until Jianmin is around again, but wanted to just get it up.
I encountered this bug while trying to repro a customer issue before.
It's very difficult to reason about the logic around these flags, and revs bodies being dropped etc. But I can't find any code which ensures the remote rev body is kept. I'm pretty sure there are cases where the body can be dropped.
It needs to be kept to make sure delta sync is working.
I think this is a good way to ensure it will be kept.

### DIFF
--- a/LiteCore/RevTrees/RevTree.cc
+++ b/LiteCore/RevTrees/RevTree.cc
@@ -412,6 +412,7 @@ namespace litecore {
         bool conflict = rev->isConflict();
         for ( auto ancestor = rev->parent; ancestor; ancestor = ancestor->parent ) {
             if ( conflict && !ancestor->isConflict() ) break;  // stop at end of a conflict branch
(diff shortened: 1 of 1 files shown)


callumbirks:
@jianminzhao While attempting to repro the issue where the attachments flag was dropped, I found that in some scenarios the remote revs were not kept, which made it impossible to perform delta sync.
I think its best to stop that from happening at this level, so we can guarantee the remote rev body is not deleted. There are a lot of code paths which affect the flags, bodies, remote revs, etc. But they all lead here, so fixing it here guarantees its fixed everywhere.

callumbirks:
In particular, I saw a case where when creating a new local rev, this method is called to keep the body of the new rev, and the `keepBody` was dropped from the latest remote rev.

jianminzhao (approved):

PR #1980, opened 2024-03-14
Support latest vectorsearch (dev branch) and hybrid queries

snej:
1. Added the official SQLite `carray` extension, because the latest `vectorsearch` library requires it.
2. Support for hybrid vector queries, where the vector search is combined with other criteria on the collection:
  - When `vector_match()` is the only criterion in the WHERE clause, OR if an explicit `max_results` arg is given, it's a "plain" query like already existed.
  - Otherwise it's a "hybrid" query, which invokes the vectorsearch extension differently (with a JOIN constraint on its rowid column.) This is less efficient, but computes distances for all the rows selected by the other WHERE tests, instead of just finding the closest docs in the whole collection, so it gives more accurate results.
3. In a plain query where there is no `max_results` given, but the query itself has a LIMIT, use the LIMIT as the max_results for the vector query. This is intuitive, and makes it so you only need to use max_results if you want to force a plain vector query in combination with other conditions.


pasin:
@snej Can you fix the windows build issue so the PR could get reviewed?

jianminzhao (approved):

pasin (commented):
In beta 1 and 2, the default limit of the vector_match() is 3. 

For the case of hybrid search without specifying the limit, from the test, it seems like there is no default limit = 3 applied anymore? If this is correct, this will be a behavior change that needs to be documented and probably need to see if it will cause any confusion when the default limit will be applied.

pasin:
I have chatted with @jianminzhao to confirm my understanding. As the default limit will not be applied in the hybrid queries, we will need to explain this in the documentation (maybe with some examples). It's intuitive to understand so I hope it will not be hard for users to understand this.

pasin (approved):

PR #1981, opened 2024-03-15
CBL-4661: ROUND_EVEN SQL++ function returning incorrect results

callumbirks:
Fix ROUND_EVEN.
Add test for ROUND_EVEN edge case.

### DIFF
--- a/LiteCore/Query/SQLiteN1QLFunctions.cc
+++ b/LiteCore/Query/SQLiteN1QLFunctions.cc
@@ -929,11 +929,10 @@ namespace litecore {
         roundTo(ctx, argc, argv, [](double d) -> double {
             double fl = floor(d);
             double fr = d - fl;
-            if ( fr < 0.5 || int64_t(fl) % 2 == 0 ) {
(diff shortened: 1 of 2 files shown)


borrrden:
I don't think this algorithm is correct, I just think it happens to get the correct answer this time.  

```c++
// result = 8.8343534, argv[1] = -1
double scale = pow(10, sqlite3_value_int64(argv[1]));
result       = fn(result * scale) / scale;
```

This will give `result * scale` = $8.8343534 * {10^{-1}}$ -> `0.88343534`
`fn` will round to the nearest integer for the result of `result * scale` -> `1`
Then $1 / {10^{-1}}$ -> `10`

However, replace `-1` with `-2` and now you have:

`result * scale` = $8.8343534 * {10^{-2}}$ -> `0.088343534`
`fn` rounds to `0`
Then $0 / {10^{-2}}$ -> `0` when it should be `100`

borrrden:
The PR validation failure is completely unrelated.

PR #1983, opened 2024-03-15
CBL-5521: N1QL Parser has exponential slowdown for redundant parentheses

callumbirks:
Description from ticket:
A query like `SELECT * FROM _ WHERE (((((type == "6")))))` should be the same as `SELECT * FROM _ WHERE (type == "6")`. But the N1QL parser creates a new expression for every parentheses. Because the parser acts recursively, this can cause exponential slowdown. We should strip the redundant parentheses at the top-level, to avoid creating unnecessary expressions.
We must be careful not to remove all parentheses.

### DIFF
--- a/LiteCore/Query/N1QL_Parser/n1ql.cc
+++ b/LiteCore/Query/N1QL_Parser/n1ql.cc
@@ -45,10 +45,10 @@ namespace litecore { namespace n1ql {
 #define YYRELEASE	yyrelease
 #endif
 #ifndef YY_BEGIN
-#define YY_BEGIN	( yy->__begin= yy->__pos, 1)
+#define YY_BEGIN	( yy->_begin= yy->_pos, 1)
 #endif
 #ifndef YY_END
-#define YY_END		( yy->__end= yy->__pos, 1)
+#define YY_END		( yy->_end= yy->_pos, 1)
 #endif
 #ifdef YY_DEBUG
 # define yyprintf(args)	fprintf args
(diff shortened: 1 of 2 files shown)


callumbirks:
This reduces any extra parentheses at the top-level to only one set of parentheses.
i.e. `((((((expr))))))` will become `(expr)`.
In my testing, this reduces parsing time of a test expression from 5 secs to 64ms.

callumbirks:
`__` and `__maxpos` updated to reflect previous changes in the `leg` parser.

snej (approved):

pasin (commented):
As chat in slack, this may not solve the actual customer issues so we will need to wait to see their actual SQLs that they are using. 

The CBSE ticket only provides a simplified sample expression `(((((type == "6")))))` but I'm sure that that is not their actual SQLs. 

For this optimization, can it be put in the _baseExpr?

callumbirks:
Per Slack discussion, `_baseExpr` already contains an expression to strip parentheses (which then calls back into `expression`.
The original cause of multiple levels of parentheses causing exponential slowdown was _because_ the parentheses don't get stripped until they reach `_baseExpr`, which only strips 1 level then goes back to the top with `expression`. Now, as many parentheses as possible are stripped at the top level, avoiding all unnecessary recursive paths through the `expr` chain.

If there are any inner expressions containing extra parentheses, they will also be stripped when `_baseExpr` calls back up to `expression`.

PR #1985, opened 2024-03-21
Handle concurrency between the removing of observers and notifying th…

jianminzhao:
…e observers.

We use a mutex when adding and removing observers from member, _observers, of C4Query. However, to avoid the mutex encompassing the notification callbacks, we make a copy of the observers and leave the mutex before calling notification callback. This renders the copy of observers liable for being deleted. We handle them like weak references.

### DIFF
--- a/C/Cpp_include/c4Query.hh
+++ b/C/Cpp_include/c4Query.hh
@@ -113,7 +113,8 @@ protected:
 
     C4Query(C4Collection*, C4QueryLanguage language, slice queryExpression);
     ~C4Query();
-    void enableObserver(litecore::C4QueryObserverImpl *obs, bool enable);
+    // toDelete can be true only if enable is false.
+    void enableObserver(litecore::C4QueryObserverImpl *obs, bool enable, bool toDelete);
 
 private:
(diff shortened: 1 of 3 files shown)


borrrden (commented):
This has gone so far down a rabbit hole that I cannot follow....

Is there something that the platform should be doing here that it is not?  How are we ending up in this situation now after it having been like this for years?  Is our architecture faulty?  Should there be something keeping these things alive longer?

jianminzhao:
My assumption is that notify() cannot remove itself from the observer list. Reading the note of  [CBL-2336](https://issues.couchbase.com/browse/CBL-2336) this might not be correct

jianminzhao:
Relaxed the assumption. notify() cannot trigger deletion of the observer that it's called upuon.

pasin (commented):
I do not understand the new `toDelete` mode (e.g. when need to use)? 

Is it be better (and possible) to make C4QueryObserverImpl RefCounted so that it will not be deleted if it's still in use?

jianminzhao:
I guess we used plain pointer to avoid circular references.

PR #1988, opened 2024-03-26
Disney CBSE 16662: FOTIFY: thread_mutex crash issue

jianminzhao:
Make C4QueryObserver ref-counted. We notify the observers outside mutex to avoid callbacks recurring back to the locked mutex. This makes observers liable being destroyed as whilst callback. We bump the ref-counts before leaving the mutex.

This change causes reference cycle when the observer is enabled. Client must disable (setEnabled(false)) to break the cycle and free it when done with it.

### DIFF
--- a/C/Cpp_include/c4Query.hh
+++ b/C/Cpp_include/c4Query.hh
@@ -106,7 +106,7 @@ public:
 
     using ObserverCallback = std::function<void(C4QueryObserver*)>;
 
-    std::unique_ptr<C4QueryObserver> observe(ObserverCallback);
+    Retained<C4QueryObserver> observe(ObserverCallback);
 
 protected:
     friend class litecore::C4QueryObserverImpl;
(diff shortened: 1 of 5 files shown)


C/Cpp_include/c4Query.hh
```diff
@@ -117,21 +117,37 @@ protected:
 
 private:
     class LiveQuerierDelegate;
+
+    struct KeyCmp {
+        using is_transparent = void;
```
borrrden, line 122: What does this line do?

C/c4CAPI.cc
```diff
@@ -1504,18 +1504,21 @@ void c4queryenum_release(C4QueryEnumerator *e) noexcept {
...
+    });
 }
 
 void c4queryobs_setEnabled(C4QueryObserver *obs, bool enabled) noexcept {
     obs->setEnabled(enabled);
 }
 
 void c4queryobs_free(C4QueryObserver* obs) noexcept {
-    delete obs;
+    c4base_release(obs);
```
borrrden, line 1521: We'd want to alias c4queryobs_release for public use in the future.  For a VF it is ok.

borrrden (approved):

C/Cpp_include/c4Query.hh
jianminzhao: This is to allow comparison to be done between Key type and another type. Otherwise, plain pointer will be converted to RefCounted before comparison.

C/c4CAPI.cc
jianminzhao: Yes. For released version,  such as 3.1 and 3.2, we keep the name. For Boron/master branch, when I create a port ticket, I will mention it to use stand naming for RefCounted, _new and _release.

C/Cpp_include/c4Query.hh
borrrden: I mean specifically is_transparent = void.  I know what the class is for.

pasin (approved):

C/Cpp_include/c4Query.hh
jianminzhao: Does matter what type is_transparent is aliased to, as long as the alias exists. c++ template quirky thing. std::set::find requires it. Without "is_transparent", "find" can only find by value of KeyType.

PR #1989, opened 2024-03-27
Part of CBL 5579 in order to facilitate VS on .NET Android

borrrden:
The extension library won't auto extract onto the Android device unless it is prefixed with "lib" but the extension loading process needs to accommodate that.

### DIFF
--- a/LiteCore/Storage/SQLiteDataFile.cc
+++ b/LiteCore/Storage/SQLiteDataFile.cc
@@ -199,7 +199,12 @@ namespace litecore {
     // for more extensions.
     static void LoadVectorSearchExtension(sqlite3* sqlite) {
 #ifdef COUCHBASE_ENTERPRISE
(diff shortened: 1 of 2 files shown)


jianminzhao (approved):

bmeike (commented):
I kinda don't love this.  It is a quick fix to solve one problem on one platform.

These things should be called "lib...<suffix>" on every platform that expects them to be named that way (I think that is all *nix variants, including Android, Linux, iOS and MacOS)

Can we have a look at this in the larger scope, before we do a little bitty hack?

borrrden:
None of the platforms _expect_ it to be in this format from SQLite's point of view, they just _allow_ it to be that way on non-Windows.  I plan to take this back out again once we can bump to minSdkVersion API 23.

pasin (approved):

PR #1997, opened 2024-04-02
CBL-5592: Binary Encoder to encode the (Logging) object path

jianminzhao:
CBL-5426: Binary Decoder to account for the new Logging object path

Also, the decoder generates outputs matching the format of plain text logs.


jianminzhao:
Ported from the master branch. All are auto-merged.

pasin (approved):

PR #2002, opened 2024-04-03
CBL-5509: Update Min macOS Support Version to 12.0

callumbirks:


velicuvlad:
I've managed to successfully build iOS Obj-C, still some changes to be done for Swift, but this can be merged.

PR #2008, opened 2024-04-10
CBL-5605: TLS and ZIP logs are excessive

callumbirks:


pasin (approved):

Networking/WebSockets/WebSocketImpl.cc
```diff
@@ -135,7 +135,7 @@ namespace litecore::websocket {
     }
 
     bool WebSocketImpl::send(fleece::slice message, bool binary) {
-        logVerbose("Sending %zu-byte message", message.size);
+        logDebug("Sending %zu-byte message", message.size);
```
pasin, line 138: I'm the fence whether we should log these WebSocket log as Verbose or Debug but we could try this and see. If it's really required, we could change it back in the later release.

pasin (approved):

pasin (approved):

PR #2009, opened 2024-04-11
CBL-5599: LiteCore logging extraneous backtraces

jianminzhao:
We don't want to see the backtrace for error "UnsupportedOperation." Meanwhile, I made the error message little bit more specific, from
  Unable to open extension to check version
to
  Unable to open extension at path/CouchbaseLiteVectorSearch to check version


pasin (approved):
Add `UnsupportedOperation` to the unremakable list seems to be OK as long as those we don't use it for anything internal (non-user operations). I added @borrrden to review this as well when he is back.

borrrden (approved):

PR #2011, opened 2024-04-12
CBL-5569: Text Log Files do not have date included in the log timestamp

jianminzhao:
- Use UTC time both in file logs (cherry picked from (9a09543593 and 17d342b962)
- Include Date in the timestamp.


jianminzhao:
Have many manual picks

PR #2014, opened 2024-04-17
CBL-5642: Null dereference crash in gotHTTPResponse

jianminzhao:
As BLIPIO receives HTTPResponse, it passes it upstream to Connection. When doing it, it calls gotHTTResponse on _connection without checking it against null. This is okay if it's called before it is closed, for we reset it to null when onClose is called. To guard agaisnt this situation, we check it before calling it. To avoid racing with regard to _connection, we dispatch the method to the queue.

### DIFF
--- a/Networking/BLIP/BLIPConnection.cc
+++ b/Networking/BLIP/BLIPConnection.cc
@@ -18,6 +18,7 @@
 #include "Batcher.hh"
 #include "Codec.hh"
 #include "Error.hh"
+#include "Headers.hh"
 #include "Logging.hh"
 #include "StringUtil.hh"
 #include "varint.hh"
@@ -29,7 +30,6 @@
 #include <map>
 #include <unordered_map>
 #include <vector>
-#include "../HTTP/Headers.hh"
 
 using namespace std;
 using namespace fleece;
--- a/Networking/BLIP/BLIPConnection.cc
+++ b/Networking/BLIP/BLIPConnection.cc
@@ -29,6 +29,7 @@
 #include <map>
 #include <unordered_map>
 #include <vector>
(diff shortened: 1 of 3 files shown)


Networking/BLIP/BLIPConnection.cc
```diff
@@ -29,6 +29,7 @@
 #include <map>
 #include <unordered_map>
 #include <vector>
+#include "../HTTP/Headers.hh"
```
borrrden, line 32: This is inappropriate.  I don't know if it is even needed anymore?  What function does this add?  If it is needed, please correct the CMake project to not need `..` paths like this.

borrrden (changes requested):

Networking/BLIP/BLIPConnection.cc
jianminzhao: For websocket::Headers
In Xcode, "#include Headers.hh" works, but I have to use ",," for Linux and Windows

borrrden: I don't see anywhere that is used in this PR.  If you must have it, then edit the search paths in the CMake project to avoid using parent directory paths.

borrrden (changes requested):

Networking/BLIP/BLIPConnection.cc
borrrden: I see where it's used now but it's not something new in this PR.  How was it working before?

jianminzhao: Before, it used reference &headers, Since I send it to the queue, I changed it by value.

pasin (approved):

borrrden (approved):

PR #2016, opened 2024-04-20
CBL-5346: Logging Replicator reasons of state change

jianminzhao:


callumbirks (approved):
LGTM. Was this ticket the point of contention in Slack about logging using run-time strings? I guess without a lot of work re-working the whole logging system its unavoidable.

Replicator/IncomingRev.cc
```diff
@@ -452,12 +452,27 @@ namespace litecore::repl {
         _bodySize       = 0;
     }
 
-    Worker::ActivityLevel IncomingRev::computeActivityLevel() const {
-        if ( Worker::computeActivityLevel() == kC4Busy || _pendingCallbacks > 0 || (_blob != _pendingBlobs.end()) ) {
-            return kC4Busy;
+    Worker::ActivityLevel IncomingRev::computeActivityLevel(std::string* reason) const {
```
borrrden, line 455: Is this ever null?  Why not use `std::string&` ?

Replicator/IncomingRev.cc
```diff
@@ -452,12 +452,27 @@ namespace litecore::repl {
...
+        std::string           reason0;
+        Worker::ActivityLevel workerLevel = Worker::computeActivityLevel(reason ? &reason0 : nullptr);
```
borrrden, line 457: Using a reference instead of a pointer will avoid ugly weird syntax like this...

Replicator/Puller.cc
```diff
@@ -356,21 +356,59 @@ namespace litecore::repl {
         }
     }
 
-    Worker::ActivityLevel Puller::computeActivityLevel() const {
-        ActivityLevel level;
+    Worker::ActivityLevel Puller::computeActivityLevel(std::string* reason) const {
+        ActivityLevel         level;
+        int                   k = -1;
```
borrrden, line 361: ....a variable named `k` ?  Name it something useful please....

Replicator/Puller.cc
```diff
@@ -356,21 +356,59 @@ namespace litecore::repl {
...
+        std::string           reason0;
+        Worker::ActivityLevel workerLevel{kC4Stopped};
         if ( _unfinishedIncomingRevs + _unfinishedIncomingRevoked > 0 ) {
             // CBL-221: Crash when scheduling document ended events
             level = kC4Busy;
+            k     = 0;
```
borrrden, line 367: Instead of this why not just set the reason here if applicable?

Replicator/Pusher.cc
```diff
@@ -512,26 +512,76 @@ namespace litecore::repl {
         Worker::_connectionClosed();
     }
 
-    bool Pusher::isBusy() const {
-        return Worker::computeActivityLevel() == kC4Busy || (_started && (!_caughtUp || !_continuousCaughtUp))
-               || _changeListsInFlight > 0 || _revisionsInFlight > 0 || _blobsInFlight > 0 || !_revQueue.empty()
-               || !_pushingDocs.empty() || _revisionBytesAwaitingReply > 0;
+    bool Pusher::isBusy(std::string* reason) const {
+        std::string reason0;
```
borrrden, line 516: Better than `k` but still not a great variable name.

borrrden (changes requested):

Replicator/IncomingRev.cc
jianminzhao: Right now, it's not null, but I am thinking it can serve as a (one-place) switch to turn it off. If it's too verbose, we can simply turn it off and the only overheads are some wasted assignments to local variable k.

Replicator/IncomingRev.cc
jianminzhao: c.f. above reply. I am open to fix it (the feature) up by changing it to reference.

Replicator/Puller.cc
jianminzhao: I could not think a meaning name. Perhaps name it, decisionPlace, or decisionPoint?

Replicator/Puller.cc
jianminzhao: I designed it in two stages. The first stage record the decision place. The goal is to not alter the current logical flow. The second stage, merely assigns reason to the final surviving k. (k can be overwritten later).

Replicator/Pusher.cc
jianminzhao: It's the reason as we delegate the decision to other functions. It may not be used depending on whether it is overwritten later in the decision path. Maybe potentialReason?

borrrden: Given that it comes from the Worker parent class something like `innerReason` or `parentReason` seems more clear.

Replicator/Puller.cc
borrrden: That seems needlessly complex.  What is this requirement of "not altering the current code"?  The side effect is making the code harder to understand.

Replicator/Puller.cc
borrrden: Probably because it is meaningless, as I pointed out in another comment about not separating this into two steps.

jianminzhao:
In the second commit, I took advantage of the two-stage solution: in Worker::afterEvent, I passed NULL to computeActivityLevel if it's not going to be used, namely when log level is greater than Info, so that to avoid the extra cost of getting the reason. Of course, we may also decide only to print the reason at Verbose level, because there is extra cost for every event whether there is a change or not.

borrrden (commented):
Still opposed to this two stage system whereby checks from the first stage need to be repeated simply for the existence of the second stage.  The "turning on and off" via null pointer could just as easily be done inline and with less confusion.  For example, how can I know from reading the code as a new person that "levelStarts at = 3" else branch means "conflictIMightRetry"?  There is no way, and I have to examine the first stage in order to know that.

jianminzhao:
That's true. One has to compare "case 3: " with "levelSetAt = 3" , but it's only when reading the log and try to locate in the source code. We have if's like, 
   if (x || y || z) level = kC4Busy,
for example. Currently, for the main, important part of the code, it is changed to,
   if (x || y || z) {level = kC4Busy; levelSetAt = 3; }
And we break up the condition chain at the time we assign the reason. Otherwise, to combine the two stages, it would be
  if (x) { level = ; *reason = ; }
  else if (y) ...
  else if (z) ...
makes the main code segment bulky, and runtime branching. We will have very lengthy if-else whereas currently we group if-else in some logical sense.
Yes, it's at the cost of reading the assignment of reasons. 
And to top it off, we have
        } else if ( _options->pull(collectionIndex()) == kC4Continuous || isOpenServer() ) {
            _spareIncomingRevs.clear();
            level      = kC4Idle;
            levelSetAt = 3;
        } 
we have to action for two conditions both true and track individual true for reason.

borrrden:
You can still group them inside the main branch in the same way so that the extra branches are ignored if result is null:

```
if(x) {
    if(reason) {

    }
}

**Comment by jianminzhao** [comment]
It will be
if (x) {
  level = 
  if (reason) *reason = 
} else if (y) {
...
} else if (z) {
}

The worse case is
if (x || y) {
  doSomething();
  level = 
}
It would have to be
if (x || y) {
  doSomething();
  level = 
  if (reason) {
     if (x) *reason = ;
    else *reason = 
 }
}
It's not undoable, just awkward. Originally, (x || y || z) is logical group and we may want to do something  in this logical condition (in the future). For "reason," we broke it up, kind of lost logical group. 
Just my opinion. To understand the behavior, we won't need to read the code in the second part. For the second part, we mostly read the logs. and if critical, we come back to the source to find the place where runtime hits.

**Comment by borrrden** [comment]
@pasin any thoughts ?

**Comment by pasin** [comment]
@jianminzhao when do you think you would like to turn the reason off? I agree with @borrrden that setting the reason right away is more understandable and cleaner - now when I read the code, I need to go up to see where the `levelSetAt` is. Moreover, if I need to add more if/else, the `levelSetAt` will need to be all updated and we may have bug if the update is not proper.

I saw your reply that you are open to change to use reference so if that is the case, the code to set reason will be even more cleanner. Otherwise, you may think about using MARCO instead of checking if (reason) everywhere.

File: Replicator/Pusher.cc [hunk 6]
```diff
@@ -512,26 +512,76 @@ namespace litecore::repl {
...
+            else if ( _revisionsInFlight > 0 )
+                *reason = "revisionsInFlight";
+            else if ( _blobsInFlight > 0 )
+                *reason = "blobsInFlight";
+            else if ( !_revQueue.empty() )
+                *reason = "revQueue";
+            else if ( !_pushingDocs.empty() )
+                *reason = "pushingDocs";
+            else
+                *reason = "revisionBytesAwaitingReply";
```
pasin, line 536: I'm wondering if it's better to have reason = "Busy as revisionsInFlight = x, changeListsInFlight = y, ..." so that we also know their values as well.

Replicator/Replicator.cc
```diff
@@ -368,50 +368,110 @@ namespace litecore::repl {
...
+                case 8:
+                    *reason = "closed";
+                    break;
+                case 9:
+                    *reason = "closed:busyWhenConnecting";
+                    break;
+                default:
+                    DebugAssert(false);
+                    break;
+            }
```
pasin, line 472: It seems like the `reason` return here is just for indicating where the line in the code was executed (It's like I put the breakpoint), and the code added really makes it difficult to understand the original logic. We can do something like "busy as (propery=value, property value)".

jianminzhao:
"when do you think you would like to turn the reason off? " ---
Most obvious reason to turn "reason" off is if it's not going to be used at all in the code, specifically, if willLog(Info) == false. This condition is also adopted in the PR. 

The extra code to assign reasons in the logs is not trivial. I find it an advantage to have separately controllable switch. (this is reason to have "reason" being a pointer instead of reference.) Here is scenario in my mind. Often when dealing with a tough CBSE, first thing we ask SE is to get logs at Verbose. If we alway want Verbose logs when dealing with bug in this detail, probably we may choose to turn this "reason" on only at Verbose. At Info, we have log on transition of state, new and old state. This is not because "reason" add too text in printout of the logs, but the extra cost of getting the reason.

The second issue is separation of logging logic from the main logic (separation of concerns). The main logic determines the behavior of the program and once in the binary, we cannot change it. The logging part only affects the printouts. Even if the reason represented in string is incorrect/mistaken, we can refer back to the source and find out what it ought to be. The separation allows we maintain the main logic without much concern of the logging part, all what is required is to assign unique levelSetAt and come back later to finish the logging part in separate section of the code.

For example, I currently have reason record what variable in a compound condition (e.g. x || y ) that determines the final value of level. Then, you think better to record the value of that variable, or even value of the inactive variable (whose value does not affect the value of compound condition), or even later we want to some critical value in some context. They can be addressed without touching the main part of the code.

For example, in this snippet of main logic,
if ( x || y ) {
   action();
   level = kC4Busy;
}
(Note, in the above, level is not final.) what do we assign the reason? The main logic does not care whether x or y is true. But the logger, we want to know which variable decided the state of level. We have to break it up, but we only need to break it up if reason is turned on. So, the instrumented code would be
if ( x || y ) {
  action();
  level = kC4Busy;
  if (reason) {
    if (x) {
       *reason = "x";
   } else {
       *reason = "y";
  }
}
This more than double the lines of the main code. What the separation does is to substitute "levelSetAt = <num>" for the if-reason block and move it to a separate section. Another benefit is we can decide the content of reason later.

As for readability, I think, if you read the code just to understand the behavior of the program, you won't care about the reason part. (it's just noise). If you do read the reason part, you probably are reading the logs debugging a tough issue.

The above is merely my humble thoughts and rational. Admittedly, it adds a complexity in different dimension. If you all prefer inlining the reason, I can try @borrrden 's suggestion of using macro, something like,
ASSIGN_REASON(COND1, REASON1, ELSE, REASON2)

Replicator/Replicator.cc
jianminzhao: "It seems like the reason return here is just for indicating where the line in the code was executed (It's like I put the breakpoint)" -- exactly. It's like a breakpoint plus reading off the values in the conditions noting down the actual variable that determines the level.
As for the format of reason, refer to my previous reply.

pasin:
What concerns me the most are 
* When making the change to the code, the `levelSetAt` could be easily out-of-sync and now we need to be carefully fix the logging logic on each level.

* The info in the reason seems be like a code or tag identifying the location in the code such as `unfinishedIncomingRevs`, `pushOrPull` and `UnsavedCheckpointer`. Give them a bit more detail would be better such as  `unfinishedIncomingRevs > 0` or `unfinishedIncomingRevs = 10`.

borrrden (changes requested):
Just some readability nits

borrrden (approved):

PR #2021, opened 2024-04-30
CBL-5680 + CBL-5681: ReplacementRev enhancement

callumbirks:


callumbirks:
Seems like this has broken other tests. I will have to run them locally and fix them

callumbirks:
The test "Continuous Super-Fast Push" is failing consistently. The issue seems to be this;
One side receives a replacementRev, which apparently has no history:
```
Obj=/Repl#21/puller#27/incomingrev#29/ {Coll#0} Received revision 'dock' #27-ffff (seq '17') (replaced rev #17-ffff) 
Obj=/Repl#21/puller#27/incomingrev#29/ Server sent no history with 'dock' #27-ffff 
```
Then later, when its receiving the next rev, it has the wrong parent:
```
Obj=/Repl#21/revfinder#26/ {Coll#0} Rejecting proposed change 'dock' #40-ffff with parent 17-ffff (status 409; current rev is 27-ffff) 
```
I guess this is because the other side thinks this side only has rev#17, but it has rev#27

callumbirks:
Fixed the history problem - the Pusher sends history now for the replacement rev.
Still getting rejected proposed change.

callumbirks:
I fixed the rejected rev issues by making sure the remote ancestor on the pusher is set to the new rev, rather than the one which has been replaced.
Although now I see another issue, when the checkpoints are validated in the test:
```
/Users/callumbirks/cblite/litecore/couchbase-lite-core/LiteCore/tests/../../Replicator/tests/ReplicatorLoopbackTest.hh:652: Failure:
  CHECK(doc->body == c4str(body))
with expansion:
  slice["{"local":23,"localCompleted":[0,24,33,8,52,8,70,9,89,9,108,9,128,8,147,8,166,8,184,9,201,1]}"]
==
slice["{"local":201}"]
with message:
  Checking local checkpoint 'cp-i1W0vKdCzXmflOx92CDuWLF3ybleltHuCPbcR9sCCaM=

/Users/callumbirks/cblite/litecore/couchbase-lite-core/LiteCore/tests/../../Replicator/tests/ReplicatorLoopbackTest.hh:652: Failure:
  CHECK(doc->body == c4str(body))
with expansion:
  slice["{"local":23,"localCompleted":[0,24,33,8,52,8,70,9,89,9,108,9,128,8,147,8,166,8,184,9,201,1]}"]
==
slice["{"local":201}"]
with message:
  Checking remote checkpoint 'cp-i1W0vKdCzXmflOx92CDuWLF3ybleltHuCPbcR9sCCaM=
```
ie the `{"local":201}` was expected, but the checkpoint doc was actually the above one with the large array of "localCompleted".

Replicator/Pusher.cc
```diff
@@ -301,6 +305,14 @@ namespace litecore::repl {
              && !_options->properties[kC4ReplicatorOptionDisableDeltas].asBool() )
             _deltasOK = true;
 
+            // re; CBL-5680, CBL-5681. Currently replacementRevs only concerns the passive pusher (active puller).
+            // We can enable this code in future to enable the feature on the active pusher.
+#ifdef ACTIVE_PUSH_REPLACEMENTREVS
+        _sendReplacementRevs = reply->boolProperty("sendReplacementRevs"_sl);
+        logInfo("Peer changes response %s request the pusher to send replacementRevs",
+                _sendReplacementRevs ? "did" : "did not");
+#endif
```
pasin, line 314: I feel like this shouldn't be included as when enabling it some tests will fail. We would need a bit more thought about how the checkpoint, remote rev, and checkpoint will work together for active push.

Replicator/Pusher+Revs.cc
```diff
@@ -114,8 +129,11 @@ namespace litecore::repl {
 
             // Include the document history, but skip the current revision 'cause it's redundant
             alloc_slice history = request->historyString(doc);
-            if ( history.hasPrefix(fullRevID) && history.size > fullRevID.size )
+            if ( !fullReplacementRevID && history.hasPrefix(fullRevID) && history.size > fullRevID.size )
                 msg["history"_sl] = history.from(fullRevID.size + 1);
+            else if ( fullReplacementRevID && history.hasPrefix(fullReplacementRevID)
+                      && history.size > fullReplacementRevID.size )
+                msg["history"_sl] = history.from(fullReplacementRevID.size + 1);
```
pasin, line 136: Code for fullRevID and fullReplacementRevID here are pretty much the same. It can be simplified a bit with something like :

```
alloc_slice effectiveRevID = fullReplacementRev ! = null ? fullReplacementRev : fullRevID;
// then the logic here will just use effectiveRevID
```

pasin (changes requested):

Replicator/Pusher.cc
callumbirks: I just #ifdef'd it out so we don't lose the code, and when we want the pusher side its an easy change, and we can spend more time fixing the issues rather than trying to remember where the code is supposed to go 😆 . Not intended for it to be actually used until we're ready to go ahead with that.

Replicator/tests/ReplicatorLoopbackTest.cc
```diff
@@ -2060,3 +2064,127 @@ TEST_CASE_METHOD(ReplicatorLoopbackTest, "Conflict Includes Rev", "[Push][Sync]"
...
+            break;
+        case TestMode::ServerPush:
+            mutateDoc(_collDB2, "doc"_sl, [](MutableDict props) { props["birthday"_sl] = "1964-11-28"_sl; });
+            break;
+        case TestMode::ServerPull:
+            mutateDoc(_collDB1, "doc"_sl, [](MutableDict props) { props["birthday"_sl] = "1964-11-28"_sl; });
+            break;
+    }
+    waitForReplicators(*clientOpts, *serverOpts);
+    switch ( testMode ) {
```
jianminzhao, line 2175: I would also check the current rev. For example, w.r.t. client pull, remote seq = 1, but current rev = 2-x. Still better, do a new pull, and verify that puller receive 2-x, because remote seq is 1, and puller already has 2-x, and replicator is done without actual BLIP rev msg.

Replicator/tests/ReplicatorLoopbackTest.hh
```diff
@@ -181,6 +181,103 @@ class ReplicatorLoopbackTest
             CHECK(_statusReceived.progress.documentCount == uint64_t(_expectedDocumentCount));
     }
 
+    bool runReplicatorsAsync(const Replicator::Options& opts1, const Replicator::Options& opts2) {
```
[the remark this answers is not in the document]
callumbirks: All it does is run the replicators without waiting for them to stop, expecting a later call to `waitForReplicators`. This is the same model we have in ReplicatorAPITest, ReplicatorCollectionTest, etc.

Replicator/tests/ReplicatorLoopbackTest.cc
```diff
@@ -2060,3 +2064,127 @@ TEST_CASE_METHOD(ReplicatorLoopbackTest, "Conflict Includes Rev", "[Push][Sync]"
     CHECK(revInDb1 == revInDb2);
     REQUIRE(revID_2 == string(docInDb1->revID));
 }
+
+#ifdef DEBUG
```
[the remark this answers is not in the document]
callumbirks: Ah, thank you. For some reason I couldn't find this before. Will update the DEBUG usages now.

Replicator/tests/ReplicatorLoopbackTest.cc
callumbirks: Thanks, I've done so in the latest commit

jianminzhao (approved):

Replicator/Pusher.cc
jianminzhao: Perhaps a brief comment?

pasin (approved):

Replicator/Pusher.cc
callumbirks: I did put a comment above this ifdef already

PR #2022, opened 2024-04-30
CBL-5634 (Fix): ReplacementRev puller enhancement

callumbirks:


callumbirks:
Turns out I had misread the spec before. "replacedRev" contains the revID of the rev we originally requested. "rev" still contains the revID of the rev we are being sent, whether or not it is a replacementRev.
See the testing I've done for this in this PR: #2021

callumbirks:
Whoops I caught my keyboard and it assigned CBL-bot somehow 😆

pasin (approved):

pasin (approved):
As chatted before, we should add some log showing the replacement rev and replaced rev.

callumbirks:
@pasin I've added a log.
It will be for example
> Received revision 'doc1' #2-ff (seq '1') (replaced rev #1-ff)

PR #2025, opened 2024-05-03
CBL-5540: pthread_mutex_lock called on a destroyed mutex

jianminzhao:
Currently, we notify the observers outside mutex to avoid callbacks recurring back to the locked mutex. This makes observers liable of being destroyed after entering the callback. We solve the problem by making C4QueryObserver ref-counted. We increment the ref-count within the mutex and decrement it after callback is done.

This change causes reference cycle when the observer is enabled. We break the cycle in c4queryobs_free by calling setEnabled(false).

C4 clients: c4queryobs_free(obs) will break the cycle.

The C++ client must call SetEnabled(false) to break the reference cycle to avoid memory leaks.

### DIFF
--- a/C/Cpp_include/c4Query.hh
+++ b/C/Cpp_include/c4Query.hh
@@ -98,7 +98,7 @@ struct C4Query final
 
     using ObserverCallback = std::function<void(C4QueryObserver*)>;
 
-    std::unique_ptr<C4QueryObserver> observe(const ObserverCallback&);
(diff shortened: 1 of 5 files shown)


jianminzhao:
Almost auto merged by git

pasin (approved):
> The C++ client must call SetEnabled(false) to break the reference cycle to avoid memory leaks.

This is not accurate as free is calling this.

PR #2026, opened 2024-05-07
CBL-5723: Replicator syncs from beginning when using prebuilt dbs synced from SG

jianminzhao:
Preparation of a VF

### DIFF
--- a/Replicator/tests/ReplicatorCollectionSGTest.cc
+++ b/Replicator/tests/ReplicatorCollectionSGTest.cc
@@ -2544,6 +2544,120 @@ TEST_CASE_METHOD(ReplicatorCollectionSGTest, "Revoked docs queue behind revs", "
     _repl = nullptr;
 }
 
+
+TEST_CASE_METHOD(ReplicatorCollectionSGTest, "Push&Pull Replication with the Copy of Fully Synced Database SG",
+                 "[.SyncServerCollection]") {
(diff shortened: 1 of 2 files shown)


jianminzhao:
This is a draft. It's for review purpose.

Replicator/tests/ReplicatorCollectionSGTest.cc
```diff
@@ -2544,6 +2544,120 @@ TEST_CASE_METHOD(ReplicatorCollectionSGTest, "Revoked docs queue behind revs", "
...
+                    CHECK(total == 2 * docCount);
+                    CHECK(local == docCount);
+                    CHECK(remote == docCount);
+                    break;
+                default:
+                    break;
+            }
+        }
+    };
+    check();
```
pasin, line 2634: This is minor but the question that comes to me is that why do we need to do this check in this test. Can we assume that the replication is correct as there are some other tests that validate this?

Also LiteCore has a way to check the number of docs replicated so the number of docs could be used to verify if all docs are replicated correctly?

jianminzhao: Yes, it's not the object of this test. It's more a sanity check that sets the basis for the final check that no documents are pushed or pulled. I will change it to REQUIRE

pasin (approved):

PR #2030, opened 2024-05-10
CBL-5723: Replicator syncs from beginning when using prebuilt dbs syned from SG

jianminzhao:
…ced from SG


pasin (approved):
Can you update the PR description with the info about the PR before merging?

PR #2031, opened 2024-05-13
De-duplicate tests from Jenkins.

callumbirks:
Currently Jenkins fails the first run on every PR because it takes longer than an hour and gets cut off. We could just change the time limit, but a better solution is not running tests twice that we don't need to.
All of the non-EE tests get run first on GitHub Actions, then again on Jenkins. We only need to run EE tests on Jenkins, so this PR makes that the behaviour.

There is one discussion point, which is the "Code Coverage Results" comment from cbl-bot on every PR relies on running every test, so we lose that.
We could set up another script which runs on a different schedule and gives us coverage results for `master` instead of every PR?


callumbirks:
re; all these formatting fixes. There seem to be some small discrepancies between what my local `clang-format` produces and what the version in the Clang-Format check expects. I will probably need to bump the version in the GitHub Action to 18.

borrrden:
Running the tests twice is somewhat on purpose.  It is to make sure that nothing in the build process for CE vs EE has affected any of the tests.

callumbirks:
@borrrden Yeah, I tend to agree. I'm trying to find other ways of shortening the runtime for the Jenkins pipeline. I'd rather avoid extending the time limit, if there's something we can do to shorten it instead.

There are some specific tests which take a really long time, which can maybe be run on a different schedule than every PR. 

Also I'm still sure the iOS build process shouldn't take 30 minutes. All it does is build the LiteCore.framework for iOS and iOS Simulator.

callumbirks:
Going to close this and investigate other ways to reduce Jenkins' run-time without breaking our code coverage.

PR #2035, opened 2024-05-14
CBL-5539: Add an API to check if a vector index is trained or not

borrrden:


borrrden:
macOS CE failure is concerning but also not related

PR #2037, opened 2024-05-14
CBL-5481:  Lazy vector index updating

jianminzhao:


borrrden (commented):
I'll have to defer to the author of these changes as to whether or not they are correct.

snej (approved):

PR #2046, opened 2024-05-24
CBL-5745: Add an API to check if a vector index is trained or not

borrrden:


bmeike:
Tests failing?  Does that need to be addressed?

borrrden:
That's all fixed up now

bmeike (commented):
Many thanks for adding this.  It's a big help.

Do you think it would be helpful to have a comment that explained what the test *does* do, as well as what it doesn't do?  I worry that when it breaks, the next person won't understand how to fix it.

bmeike:
Thanks a bunch for sorting those tests.  Love me a green checkmark.

jianminzhao (approved):

borrrden:
Oh right the comments about what tests do!  I forgot it even though I was the one that suggested doing it.  It's always a good idea so I will add it.

PR #2048, opened 2024-05-25
Fix SQLiteCpp pointing to an orphaned commit

callumbirks:
release/3.1 is suddenly pointing to an orphaned commit, breaking all Helium builds.
With this PR I'm pointing it back to the correct commit on `2.0.0-couchbase`.
@jianminzhao I think this might be related to your recent breaking and subsequent fixing of the SQLiteCpp pointing to the wrong branch? Could you confirm.

### DIFF
--- a/build_cmake/scripts/check_bd_versions.py
+++ b/build_cmake/scripts/check_bd_versions.py
@@ -3,6 +3,7 @@
 from pathlib import Path
 
 import os, sys, yaml, subprocess, argparse
+from termcolor import colored
 
 def is_changed(git: str) -> bool:
     if len(git) == 0:
@@ -19,7 +20,7 @@ def is_changed(git: str) -> bool:
 def get_submodule_base(git: str):
     if len(git) == 0:
         return None
-        
+
     difflines = git.splitlines()[-2:]
     before = difflines[0].split(" ")[2]
     after = difflines[1].split(" ")[2].replace("-dirty", "")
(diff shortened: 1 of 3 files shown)


callumbirks:
Ignoring blackduck failure. It's failing because of the orphaned commit

pasin (commented):
Is the check_bd_versions.py ported from the other branch?

callumbirks:
> Is the check_bd_versions.py ported from the other branch?

@pasin the script was originally changed with the change which broke the SqLiteCpp branching rules. It seemed to be failing, so I reverted the script back to what it used to be.
I think the changes in the script supported Jianmin's changing away from the `2.0.0-couchbase`, but maybe @jianminzhao can also comment on that.

jianminzhao:
Yes, the deletion of that branch is the cause of the first failure of submodule check.
Now that you pointed it to 90a4aae0238, which is headed by 2.0.0-couchbase, the problem is rectified.

"might be related to your recent breaking and subsequent fixing of the SQLiteCpp pointing to the wrong branch? Could you confirm" - the "recent breaking" is not breaking of the build. It was the the commit that SQLiteCpp was not the one supposed to be. After making 3.2 & 3.3 pointing to 2.0.0-couchbase, we over did it with deletion of that branch.

The current failure of BD check, I think, is because the version is changed between 3.43.0 and 3.45.0. 
I don't know why you had to change the script, check_bd_versions

callumbirks:
@jianminzhao the version is not changed. It is 3.43.0, which it was before. 3.1 still points to the same "commit" with this PR, it's just on a different branch so has different commit hash.

I think the BD failure seems to be it cannot compare with the orphaned commit

pasin (approved):
Assume that the duck script is correct.

callumbirks:
@jianminzhao thanks, that script looks nicely improved! Is this ready to be merged now?

jianminzhao:
Agree with @callumbirks the failure of bd check is due to orphaned commit that SQLiteCpp references has been garbage collected. In theory, there is no way to restore it updated the script to recognize the issue and request manual verification.

callumbirks:
@jianminzhao reviewed the changed script. LGTM and seems to be reflected in the output of the "Check Blackduck Manifest" check.

PR #2049, opened 2024-05-30
CBL-5791: Socket not called to close after WebSocket PING Timeout

callumbirks:
Simply a missed case (SOCKET_OPENED status) in the switch statement. The functions which follow (either `callCloseSocket()` or `callRequestClose()`) both expect the state to be OPENING or OPENED, so this has no side-effect.


pasin:
@callumbirks seem like removing that bracket is causing build issue.

callumbirks:
@pasin Oh okay I'll revert the change

callumbirks:
This is blocked by #2048

pasin (approved):

jianminzhao (approved):

PR #2052, opened 2024-06-04
CBL-5818 + CBL-5796: Lazy index bugs

callumbirks:
CBL-5818: Disallow finishing lazy update without updating/skipping at least one doc
CBL-5796: Fix c4coll_createIndex does not work for lazy index (c4opt not carried over)

PR #2054, opened 2024-06-06
CBL-5805: Add tests following the latest vectorsearch and hybrid queries

jianminzhao:
Added 2 tests to contrast with non-hybrid, or pure, vectorsearches.
    1. Default max_results applies only to the pure vectorsearch.
    2. Giving explicit max_results to vector_match turns it to a pure vectorsearch and causes failure to find the result.

### DIFF
--- a/LiteCore/tests/LazyVectorAPITest.cc
+++ b/LiteCore/tests/LazyVectorAPITest.cc
@@ -145,19 +145,19 @@ class LazyVectorAPITest : public C4Test {
         } catch ( [[maybe_unused]] std::exception& e ) { return nullptr; }
     }
 
-    void checkQueryReturnsWords(C4Query* query, const std::vector<string>& expectedWords) const {
-        auto e = REQUIRED(c4query_run(query, _encodedTarget, ERROR_INFO()));
-        REQUIRE(c4queryenum_getRowCount(e, ERROR_INFO()) == expectedWords.size());
-        for ( const auto& expectedWord : expectedWords ) {
-            REQUIRE(c4queryenum_next(e, ERROR_INFO()));
-            FLArrayIterator columns  = e->columns;
-            slice           word     = Value(FLArrayIterator_GetValueAt(&columns, 0)).asString();
-            float           distance = Value(FLArrayIterator_GetValueAt(&columns, 1)).asFloat();
-            CHECK(word == slice(expectedWord));
-        }
-        CHECK(!c4queryenum_next(e, ERROR_INFO()));
-        c4queryenum_release(e);
-    }
+    //    void checkQueryReturnsWords(C4Query* query, const std::vector<string>& expectedWords) const {
+    //        auto e = REQUIRED(c4query_run(query, _encodedTarget, ERROR_INFO()));
+    //        REQUIRE(c4queryenum_getRowCount(e, ERROR_INFO()) == expectedWords.size());
+    //        for ( const auto& expectedWord : expectedWords ) {
+    //            REQUIRE(c4queryenum_next(e, ERROR_INFO()));
+    //            FLArrayIterator columns  = e->columns;
+    //            slice           word     = Value(FLArrayIterator_GetValueAt(&columns, 0)).asString();
+    //            float           distance = Value(FLArrayIterator_GetValueAt(&columns, 1)).asFloat();
+    //            CHECK(word == slice(expectedWord));
+    //        }
+    //        CHECK(!c4queryenum_next(e, ERROR_INFO()));
+    //        c4queryenum_release(e);
+    //    }
 
     void checkQueryReturnsVectors(C4Query* query, int64_t expectedRowCount,
                                   const std::vector<float>& expectedVectors) const {
@@ -168,7 +168,7 @@ class LazyVectorAPITest : public C4Test {
             FLArrayIterator columns     = e->columns;
             auto            vectorArray = Value(FLArrayIterator_GetValueAt(&columns, 0)).asArray();
             for ( size_t j = 0; j < expectedVectors.size(); j++ ) {
-                float vector = vectorArray.get(j).asFloat();
+                float vector = vectorArray.get((uint32_t)j).asFloat();
                 CHECK(vector == expectedVectors[j]);
             }
         }
--- a/LiteCore/tests/QueryParserTest.cc
+++ b/LiteCore/tests/QueryParserTest.cc
@@ -677,11 +677,17 @@ TEST_CASE_METHOD(QueryParserTest, "QueryParser Vector Search", "[Query][QueryPar
     // Pure vector search (no other WHERE criteria):
     CHECK(parse("['SELECT', {WHERE: ['VECTOR_MATCH()', 'vecIndex', ['[]', 12, 34]],"
                 "ORDER_BY: [ ['VECTOR_DISTANCE()', 'vecIndex'] ],"
-                "LIMIT: 3}]")
+                "LIMIT: 5}]")
+          == "SELECT key, sequence FROM kv_default AS _doc JOIN (SELECT rowid, distance FROM "
+             "\"kv_default:vector:vecIndex\" WHERE vector LIKE encode_vector(array_of(12, 34)) LIMIT 5) AS vector1 ON "
+             "vector1.rowid = _doc.rowid WHERE (true) AND (_doc.flags & 1 = 0) ORDER BY vector1.distance LIMIT MAX(0, "
+             "5)");
+    // Pure vector search with default max_results (3)
+    CHECK(parse("['SELECT', {WHERE: ['VECTOR_MATCH()', 'vecIndex', ['[]', 12, 34]],"
+                "ORDER_BY: [ ['VECTOR_DISTANCE()', 'vecIndex'] ]}]")
           == "SELECT key, sequence FROM kv_default AS _doc JOIN (SELECT rowid, distance FROM "
              "\"kv_default:vector:vecIndex\" WHERE vector LIKE encode_vector(array_of(12, 34)) LIMIT 3) AS vector1 ON "
-             "vector1.rowid = _doc.rowid WHERE (true) AND (_doc.flags & 1 = 0) ORDER BY vector1.distance LIMIT MAX(0, "
-             "3)");
+             "vector1.rowid = _doc.rowid WHERE (true) AND (_doc.flags & 1 = 0) ORDER BY vector1.distance");
     // Pure vector search (explicit limit given):
     CHECK(parse("['SELECT', {WHERE: ['AND', ['VECTOR_MATCH()', 'vecIndex', ['[]', 12, 34], 10],"
                 "['>', ['._id'], 'x'] ],"
--- a/LiteCore/tests/VectorQueryTest.cc
+++ b/LiteCore/tests/VectorQueryTest.cc
@@ -244,6 +244,22 @@ N_WAY_TEST_CASE_METHOD(SIFTVectorQueryTest, "Hybrid Vector Query", "[Query][.Vec
                          {"rec-5300", "rec-4900", "rec-7100", "rec-3600", "rec-8700", "rec-8500", "rec-2400",
                           "rec-4700", "rec-4300", "rec-2600"},
                          {85776, 90431, 92142, 92629, 94598, 94989, 104787, 106750, 113260, 116129});
+
+    queryStr = R"(
+     ['SELECT', {
+        WHERE:    ['AND', ['VECTOR_MATCH()', 'vecIndex', ['$target'], 10],
+                          ['=', 0, ['%', ['._sequence'], 100]] ],
+        WHAT:     [ ['._id'], ['AS', ['VECTOR_DISTANCE()', 'vecIndex'], 'distance'] ],
+        ORDER_BY: [ ['.distance'] ],
+        LIMIT:    10
+     }] )";
+    query    = store->compileQuery(json5(queryStr), QueryLanguage::kJSON);
+
+    Log("---- Querying with $target = data and max_results = 10");
+    // With explicit max_results, it becomes a pure vector query.
+    options                     = optionsWithTargetVector(kTargetVector, kData);
+    Retained<QueryEnumerator> e = query->createEnumerator(&options);
+    CHECK(e->getRowCount() == 0);
 }
 
 // Test joining the result of VECTOR_MATCH with a property of another collection. In particular, it joins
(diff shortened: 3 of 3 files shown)


LiteCore/tests/VectorQueryTest.cc
```diff
@@ -227,23 +227,160 @@ N_WAY_TEST_CASE_METHOD(SIFTVectorQueryTest, "Hybrid Vector Query", "[Query][.Vec
...
+               + " WHERE VECTOR_MATCH(vecIndex, $target) "
+                 " ORDER BY distance";
+    query  = store->compileQuery(queryStr, QueryLanguage::kN1QL);
+    result = collectResults(query->createEnumerator(&options));
+
+    // Argument max_results to VECTOR_MATCH is set by defaut to 3.
+    CHECK(result.size() == 3);
+
+
+    // Case 4: plain query fails to select docs because of the additional WHERE term.
```
pasin, line 320: Why is this not a hybrid query? Is it because of the specified limit = 10?

LiteCore/tests/VectorQueryTest.cc
```diff
@@ -227,23 +227,160 @@ N_WAY_TEST_CASE_METHOD(SIFTVectorQueryTest, "Hybrid Vector Query", "[Query][.Vec
...
+             ['SELECT', {
+                WHERE:    ['AND', ['VECTOR_MATCH()', 'vecIndex', ['$target']],
+                                  ['=', 0, ['%', ['._sequence'], 100]] ],
+                WHAT:     [ ['._id'], ['AS', ['VECTOR_DISTANCE()', 'vecIndex'], 'distance'] ],
+                ORDER_BY: [ ['.distance'] ],
+                LIMIT:    10
+             }] )";
+            queryStr = json5(queryStr);
+        } else if ( queryLan == QueryLanguage::kN1QL ) {
+            queryStr = "SELECT META().id, VECTOR_DISTANCE(vecIndex) AS distance FROM "s + collectionName
```
snej, line 247: I don't think we need to test everything with both query languages. If you want, you could have a separate test that both of those compile to the same SQL, without running anything.

LiteCore/tests/VectorQueryTest.cc
```diff
@@ -227,23 +227,160 @@ N_WAY_TEST_CASE_METHOD(SIFTVectorQueryTest, "Hybrid Vector Query", "[Query][.Vec
...
+        ExclusiveTransaction t(db);
+        writeMultipleTypeDocs(t);
+        t.commit();
+    }
+    createVectorIndex();
+
+    // Case 1: plain query because there is no additional term in the WHERE clause
+    // ===================
+
+    string queryStr = "SELECT META().id, VECTOR_DISTANCE(vecIndex) AS distance FROM "s + collectionName
```
snej, line 282: Don't we already have a test that does this?

LiteCore/tests/VectorQueryTest.cc
```diff
@@ -227,23 +227,160 @@ N_WAY_TEST_CASE_METHOD(SIFTVectorQueryTest, "Hybrid Vector Query", "[Query][.Vec
...
+    std::vector<std::pair<string, float>> result  = collectResults(query->createEnumerator(&options));
+
+    // The outer limit 10 is provided to VECTOR_MATCH as argument max_results
+    CHECK(result.size() == 10);
+
+
+    // Case 2: plain query because of explicitly assigned max_results to VECTOR_MATCH
+    // ===================
+
+    queryStr = "SELECT META().id, VECTOR_DISTANCE(vecIndex) AS distance FROM "s + collectionName
```
snej, line 297: Again, don't we test this already?

LiteCore/tests/VectorQueryTest.cc
```diff
@@ -227,23 +227,160 @@ N_WAY_TEST_CASE_METHOD(SIFTVectorQueryTest, "Hybrid Vector Query", "[Query][.Vec
...
+    // Case3: plain query. If not specified, VECTOR_MATCH will use 3 as max_results.
+    // ==================
+
+    queryStr = "SELECT META().id, VECTOR_DISTANCE(vecIndex) AS distance FROM "s + collectionName
+               + " WHERE VECTOR_MATCH(vecIndex, $target) "
+                 " ORDER BY distance";
+    query  = store->compileQuery(queryStr, QueryLanguage::kN1QL);
+    result = collectResults(query->createEnumerator(&options));
+
+    // Argument max_results to VECTOR_MATCH is set by defaut to 3.
```
snej, line 316: Do we need to test that we get exactly 3 results? It's not an important number, just a placeholder default.

LiteCore/tests/VectorQueryTest.cc
```diff
@@ -227,23 +227,160 @@ N_WAY_TEST_CASE_METHOD(SIFTVectorQueryTest, "Hybrid Vector Query", "[Query][.Vec
...
+               + " WHERE VECTOR_MATCH(vecIndex, $target) AND META().sequence % 100 = 0"
+                 " ORDER BY distance";
+    query  = store->compileQuery(queryStr, QueryLanguage::kN1QL);
+    result = collectResults(query->createEnumerator(&options));
+
+    // Although hybrid query searches the solutions from the docs that
+    // must pass the secondary WHERE term, it may not select all of them.
+    // In this case, 100 docs passed the predicate META().sequence % 100 = 0, but only 34 are
+    // picked up by the vector query.
+    CHECK(result.size() == 34);
```
snej, line 357: Don't test for an exact number! That's an implementation detail in vectorsearch. There's no guarantee it might not change if I alter internal algorithms, or even if SQLite's query engine changes slightly.

LiteCore/tests/VectorQueryTest.cc
```diff
@@ -227,23 +227,160 @@ N_WAY_TEST_CASE_METHOD(SIFTVectorQueryTest, "Hybrid Vector Query", "[Query][.Vec
...
+
+
+    // Case 6: plain query. It's not straightforward to select a number for max_results to also satisfy additional WHERE term.
+    // ===================
+
+    queryStr = "SELECT META().id, VECTOR_DISTANCE(vecIndex) AS distance FROM "s + collectionName
+               + " WHERE VECTOR_MATCH(vecIndex, $target, 50) AND META().sequence % 100 = 0"
+                 " ORDER BY distance";
```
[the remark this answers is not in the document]
snej: @pasin https://github.com/couchbaselabs/mobile-vector-search/blob/main/docs/Implementation.md#hybrid-or-filtered-query-querying-a-subset-of-docids

snej (changes requested):
I don't think it's a good idea to write brittle tests that depend on precise details of the heuristics in the vectorsearch extension.

LiteCore/tests/VectorQueryTest.cc
jianminzhao: Okay, I will take it out to n1ql test.

LiteCore/tests/VectorQueryTest.cc
jianminzhao: In this test, I want to assert that the outer limit will be passed to match_vector as max_results.

LiteCore/tests/VectorQueryTest.cc
jianminzhao: Assert the equivalence for the plain query, whether use LIMIT or max_results.

LiteCore/tests/VectorQueryTest.cc
jianminzhao: Otherwise, how do we that default 3 is used? Or, I can put the test out to some where else we can examine the final sql for sqlite. Putting it here helps to understand the matter better. By the way, it's also testing that hybrid logic is used here.

LiteCore/tests/VectorQueryTest.cc
```diff
@@ -227,23 +227,160 @@ N_WAY_TEST_CASE_METHOD(SIFTVectorQueryTest, "Hybrid Vector Query", "[Query][.Vec
...
+
+    // Asserting that 100 documents are selected by "META().sequence % 100 = 0"
+    queryStr                    = "SELECT count(*) FROM "s + collectionName + " WHERE META().sequence % 100 = 0";
+    query                       = store->compileQuery(queryStr, QueryLanguage::kN1QL);
+    Retained<QueryEnumerator> e = query->createEnumerator();
+    REQUIRE(e->next());
+    auto countBySecondaryWhere = e->columns()[0]->asInt();
+    REQUIRE(countBySecondaryWhere == 100);
+
+    // Case 5: hybrid search may not select all the docs that pass the secondary WHERE term.
```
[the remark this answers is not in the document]
jianminzhao: This probably belongs to documentation. The hybrid logic kicks in if
1. max_results is not provided to VECTOR_MATCH
2. the WHERE clause does not have any predicate other than VETOR_MATCH

LiteCore/tests/VectorQueryTest.cc
jianminzhao: Got it. Will remove it. Maybe it belongs to tutorial or so.

LiteCore/tests/VectorQueryTest.cc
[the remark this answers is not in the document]
jianminzhao: yes, this test is more a case motivating the hybrid method.

pasin (approved):

LiteCore/tests/QueryParserTest.cc
```diff
@@ -697,6 +697,20 @@ TEST_CASE_METHOD(QueryParserTest, "QueryParser Vector Search", "[Query][QueryPar
...
              "vector1.rowid = _doc.rowid WHERE (vector1.vector LIKE encode_vector((array_of(12, 34))) AND _doc.key > "
              "'x') AND (_doc.flags & 1 = 0) ORDER BY vector1.distance");
+    // Pure vector search with default max_results:
+    CHECK(parse("['SELECT', {WHERE: ['VECTOR_MATCH()', 'vecIndex', ['[]', 12, 34]],"
+                "ORDER_BY: [ ['VECTOR_DISTANCE()', 'vecIndex'] ]}]")
+          == "SELECT key, sequence FROM kv_default AS _doc JOIN (SELECT rowid, distance FROM "
+             "\"kv_default:vector:vecIndex\" WHERE vector LIKE encode_vector(array_of(12, 34)) LIMIT 3) AS vector1 ON "
+             "vector1.rowid = _doc.rowid WHERE (true) AND (_doc.flags & 1 = 0) ORDER BY vector1.distance");
+    // Pure vector search, value of LIMIT is passed to VECTOR_MATCH as max_results.
+    CHECK(parse("['SELECT', {WHERE: ['VECTOR_MATCH()', 'vecIndex', ['[]', 12, 34]],"
```
snej, line 707: Isn't this identical to the check at line 678, only with limit 5 instead of 3?

LiteCore/tests/VectorQueryTest.cc
```diff
@@ -246,6 +246,66 @@ N_WAY_TEST_CASE_METHOD(SIFTVectorQueryTest, "Hybrid Vector Query", "[Query][.Vec
...
+        queryStr = "SELECT META().id, VECTOR_DISTANCE(vecIndex) AS distance FROM "s + collectionName
+                   + " WHERE VECTOR_MATCH(vecIndex, $target) AND META().sequence % 100 = 0"
+                     " ORDER BY distance LIMIT "
+                   + limitStrs[i];
+        query                                         = store->compileQuery(queryStr, QueryLanguage::kN1QL);
+        std::vector<std::pair<string, float>> result0 = collectResults(query->createEnumerator(&options));
+        REQUIRE(result0.size() == limits[i]);
+        // As we increase the limit from 1 to 10, the result is consistent.
+        CHECK(std::equal(result0.begin(), result0.begin() + limits[i], result.begin()));
+    }
```
snej, line 306: All this is doing is testing SQLite's `LIMIT` processing; once you've established it returns ≥ 10 results without a limit, setting the limit to 1 or 10 will obviously return only that many rows. This has nothing to do with vectorsearch.

LiteCore/tests/VectorQueryTest.cc
```diff
@@ -246,6 +246,66 @@ N_WAY_TEST_CASE_METHOD(SIFTVectorQueryTest, "Hybrid Vector Query", "[Query][.Vec
...
+    // Hybrid vectorsearch consistentence: as we increase the LIMIT, the results are consistent.
+    // =====================
+
+    queryStr = "SELECT META().id, VECTOR_DISTANCE(vecIndex) AS distance FROM "s + collectionName
+               + " WHERE VECTOR_MATCH(vecIndex, $target) AND META().sequence % 100 = 0"
+                 " ORDER BY distance";
+    query  = store->compileQuery(queryStr, QueryLanguage::kN1QL);
+    result = collectResults(query->createEnumerator(&options));
+
+    REQUIRE(result.size() > 10);
```
snej, line 292: This looks identical to the test "Hybrid Vector Query", except that this one doesn't verify the results.

snej (changes requested):
Sorry, I just don't see how this is testing anything new that isn't already covered by other tests.

LiteCore/tests/QueryParserTest.cc
jianminzhao: Right, but 3 is the magic default. That is, max_results will be 3 regardless of the LIMIT clause

LiteCore/tests/VectorQueryTest.cc
jianminzhao: I basically take it as a truth. The purpose is to check the LIMIT in the hybrid query is consistent, exactly because of the test "Hybrid Vector Query."

LiteCore/tests/VectorQueryTest.cc
jianminzhao: True. I just tried to verify it, that the LIMIT is applied at the place it stands in the SQL. In vectorsearch, I see we pass LIMIT to vector_match for pure query (as an optimization). I tried to assert that the limit in the hybrid query is only used afterwards.

snej (approved):

LiteCore/tests/LazyVectorAPITest.cc
```diff
@@ -145,19 +145,19 @@ class LazyVectorAPITest : public C4Test {
...
-            REQUIRE(c4queryenum_next(e, ERROR_INFO()));
-            FLArrayIterator columns  = e->columns;
-            slice           word     = Value(FLArrayIterator_GetValueAt(&columns, 0)).asString();
-            float           distance = Value(FLArrayIterator_GetValueAt(&columns, 1)).asFloat();
-            CHECK(word == slice(expectedWord));
-        }
-        CHECK(!c4queryenum_next(e, ERROR_INFO()));
-        c4queryenum_release(e);
-    }
+    //    void checkQueryReturnsWords(C4Query* query, const std::vector<string>& expectedWords) const {
```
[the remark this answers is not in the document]
jianminzhao: This piece comes from @callumbirks 's PR, (#2052/6b3045578111c). 
Already communicated with him and he will use the function in a separate PR.
Because my xcode won't let me pass for unused variable, "distance", and I found the entire function is not used so, just commented it out.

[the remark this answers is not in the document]
jianminzhao: I saw his new PR is coming. It will be taken care of after both been merged.

PR #2056, opened 2024-06-07
CBL-5793: Add missing c4 APIs for c4Index.

callumbirks:
`getName` and `getCollection` exist in the C++ API, but not the C API.
This PR rectifies that.

### DIFF
--- a/C/c4CAPI.cc
+++ b/C/c4CAPI.cc
@@ -1060,13 +1060,9 @@ C4IndexUpdater* C4NULLABLE c4index_beginUpdate(C4Index* index, size_t limit, C4E
     return tryCatch<C4IndexUpdater*>(outError, [&] { return index->beginUpdate(limit).detach(); });
 }
 
-C4SliceResult c4index_getName(C4Index* index) noexcept {
-    return tryCatch<C4SliceResult>(nullptr, [&]{ return C4SliceResult(index->getName()); });
-}
+C4Slice c4index_getName(C4Index* index) noexcept { return C4Slice(index->getName()); }
 
-C4Collection* c4index_getCollection(C4Index* index) noexcept {
-    return tryCatch<C4Collection*>(nullptr, [&]{ return index->getCollection(); });
-}
+C4Collection* c4index_getCollection(C4Index* index) noexcept { return index->getCollection(); }
 
 size_t c4indexupdater_count(C4IndexUpdater* update) noexcept { return update->count(); }
 
--- a/C/c4CAPI.cc
+++ b/C/c4CAPI.cc
@@ -1061,11 +1061,11 @@ C4IndexUpdater* C4NULLABLE c4index_beginUpdate(C4Index* index, size_t limit, C4E
 }
 
 C4SliceResult c4index_getName(C4Index* index) noexcept {
-    return C4SliceResult(index->getName());
+    return tryCatch<C4SliceResult>(nullptr, [&]{ return C4SliceResult(index->getName()); });
 }
 
 C4Collection* c4index_getCollection(C4Index* index) noexcept {
-    return index->getCollection();
+    return tryCatch<C4Collection*>(nullptr, [&]{ return index->getCollection(); });
 }
 
 size_t c4indexupdater_count(C4IndexUpdater* update) noexcept { return update->count(); }
--- a/C/c4CAPI.cc
+++ b/C/c4CAPI.cc
@@ -10,7 +10,6 @@
 // the file licenses/APL2.txt.
 //
 
-#include "c4Base.h"
 #include "c4BlobStore.hh"
 #include "c4BlobStore.h"
 #include "c4Certificate.hh"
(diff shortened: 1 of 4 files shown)


C/c4CAPI.cc
```diff
@@ -1059,6 +1060,14 @@ C4IndexUpdater* C4NULLABLE c4index_beginUpdate(C4Index* index, size_t limit, C4E
     return tryCatch<C4IndexUpdater*>(outError, [&] { return index->beginUpdate(limit).detach(); });
 }
 
+C4SliceResult c4index_getName(C4Index* index) noexcept {
```
jianminzhao, line 1063: should handle the exception

C/c4CAPI.cc
```diff
@@ -1059,6 +1060,14 @@ C4IndexUpdater* C4NULLABLE c4index_beginUpdate(C4Index* index, size_t limit, C4E
...
+    return C4SliceResult(index->getName());
+}
+
+C4Collection* c4index_getCollection(C4Index* index) noexcept {
```
jianminzhao, line 1067: handle exception

jianminzhao (approved):

C/c4CAPI.cc
callumbirks: What exception?

C/c4CAPI.cc
callumbirks: What exception?

C/c4CAPI.cc
jianminzhao: noexcept means the app will be terminated should exception popped to here. We do something like, tryCatch<C4 ...>

callumbirks: Nevermind 😅

C/include/c4Index.h
```diff
@@ -136,6 +138,16 @@ CBL_CORE_API bool c4coll_isIndexTrained(C4Collection* collection, C4String name,
     @return  A new `C4IndexUpdater` reference, or NULL if there's nothing to update. */
 CBL_CORE_API C4IndexUpdater* C4NULLABLE c4index_beginUpdate(C4Index* index, size_t limit, C4Error* outError) C4API;
 
+/**
+ * Return the name of this index.
+ **/
+CBL_CORE_API C4SliceResult c4index_getName(C4Index* index) C4API;
```
snej, line 144: The underlying c4Index::getName returns a `slice`, so this result type should be `C4String`, not `C4SliceResult`. (I.e. there's no need to copy the name and force the caller to release it.)

C/c4CAPI.cc
```diff
@@ -1059,6 +1060,14 @@ C4IndexUpdater* C4NULLABLE c4index_beginUpdate(C4Index* index, size_t limit, C4E
     return tryCatch<C4IndexUpdater*>(outError, [&] { return index->beginUpdate(limit).detach(); });
 }
 
+C4SliceResult c4index_getName(C4Index* index) noexcept {
+    return tryCatch<C4SliceResult>(nullptr, [&]{ return C4SliceResult(index->getName()); });
+}
+
+C4Collection* c4index_getCollection(C4Index* index) noexcept {
+    return tryCatch<C4Collection*>(nullptr, [&]{ return index->getCollection(); });
```
snej, line 1068: c4Index::getCollection is a trivial getter that just returns a member, so there's no need to wrap a try/catch around this. (If you like, you could add `noexcept` to the C++ method declaration.)

snej (changes requested):

C/include/c4Index.h
callumbirks: Okay thanks :)

C/c4CAPI.cc
callumbirks: Oh okay I thought we might need it incase of null pointer access exception. But if not, I'll remove it

snej: Null pointer access is a signal/crash, not an exception, in C/C++.

C/include/c4Index.h
```diff
@@ -11,6 +11,8 @@
 //
 
 #pragma once
+#include "c4Base.h"
+#include "c4Compat.h"
```
snej, line 15: That's not necessary. Those headers are transitively included by c4IndexTypes.h, and none of the other headers (to my knowledge) redundantly include them.

C/c4CAPI.cc
```diff
@@ -10,6 +10,7 @@
 // the file licenses/APL2.txt.
 //
 
+#include "c4Base.h"
```
snej, line 13: That's not necessary. That header is transitively included by the other API headers (e.g. c4BlobStore.hh), and none of the other headers (to my knowledge) redundantly include them.

snej (changes requested):

C/c4CAPI.cc
jianminzhao: removed.

C/include/c4Index.h
jianminzhao: removed

snej (approved):

PR #2057, opened 2024-06-07
CBL-5825: Allow inspecting contents of vector index with c4db_getInde…

jianminzhao:
…xRows()

C4Database::getIndexRows() now supports vector indexes

This means you can now inspect the vector data in an index. Remember that, if any encoding is used, the vectors will not be exact.


jianminzhao:
This is completely auto merged by git from @snej 's PR to the master branch

callumbirks (approved):

PR #2058, opened 2024-06-10
Use same vector-index options struct as vectorsearch repo

snej:
- Copied mobile-vector-search's `VectorIndexSpec` cc/hh files into LiteCore -- these define the struct `vectorsearch::IndexSpec`.
- Replaced IndexSpec::VectorOptions with `vectorsearch::IndexSpec`.
- This made it easy to improve `SQLiteKeyStore::getIndex()` to populate _all_ the fields of the spec's vectorOptions.

### DIFF
--- a/C/include/c4Index.h
+++ b/C/include/c4Index.h
@@ -16,110 +16,119 @@
 C4_ASSUME_NONNULL_BEGIN
 C4API_BEGIN_DECLS
 
-/** \defgroup Indexing  Database Indexes
+//======== C4Collection Methods:
+
+/** \defgroup Indexing  Indexes
      @{ */
 
+/** Creates a collection index, of the values of specific expressions across all documents.
+    The name is used to identify the index for later updating or deletion; if an index with the
+    same name already exists, it will be replaced unless it has the exact same expressions.
 
-/** Creates a database index, of the values of specific expressions across all documents.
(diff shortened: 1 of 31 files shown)


snej:
The macOS job failed due to the test "Multiple Collections Incremental Revisions" timing out. Restarted it.

C/include/c4Index.h
```diff
@@ -183,6 +194,21 @@ CBL_CORE_API bool c4indexupdater_finish(C4IndexUpdater* updater, C4Error* outErr
 
 #endif
 
+
+#ifndef C4_STRICT_COLLECTION_API
+//======== SEMI-DEPRECATED DATABASE METHODS:
+NODISCARD CBL_CORE_API bool c4db_createIndex(C4Database* database, C4String name, C4String indexSpecJSON,
```
borrrden, line 200: .NET, at least, doesn't use these anymore so it might be time to actually deprecate, or remove these functions (c4coll_isIndexTrained can wait a bit maybe but since it was JUST introduced it should only be present in a few places).

pasin (approved):

C/include/c4Index.h
snej: Sounds good -- could you file an issue?

PR #2065, opened 2024-06-12
CBL-5843: Crash when call finish() on the IndexUpdater that is alread…

jianminzhao:
…y finished

We throw NotOpen/"The updater may have already been finished," in this case.


jianminzhao:
Redo it because the test case is already in another PR, #2064

PR #2068, opened 2024-06-14
CBL-5843: Crash when call finish() on the IndexUpdater that is alread…

jianminzhao:
…y finished

We throw error at the second call of finish().


jianminzhao:
The test case is already written in PR#2064 for (CBL 5754)

pasin (approved):
Look good to me.

PR #2069, opened 2024-06-14
Merged blaugold's WASM branch, plus fixes

snej:
Cherry-picked @blaugold 's branch that adds WASM support via emscripten.
Made some changes to get it to build again.
This depends on the emscripten support in Fleece, currently https://github.com/couchbase/fleece/pull/232

- C4Tests passes, although the threading test is painfully slow (takes at least 10 minutes to run!)
- CppTests has 3 failures.

I chose to disable all of the socket and listener code -- BuiltInWebSocket, etc. It doesn't make sense to have code that uses custom TCP sockets in a library targeted at web browsers. Instead the hypothetical JS platform code would use the C4Socket API to bridge to the JS WebSocket class, the same way the other (non-C) platforms do.


snej:
FYI, in a `Release` build the size of C4Tests.wasm is 4MB, and CppTests.wasm is 6.2MB.

(In a `RelWithDebSym` build they're 10x larger (!), which I assume is all the debug symbols in the binary.)

For comparison, in a Mac build C4Tests is 4.7MB, or about 2.3MB per architecture; CppTests is 21MB, or about 10.5MB per architecture [which is surprisingly big...]

snej:
To make the WASM smaller I
- Built with `Release` instead of `RelWithDebInfo`
- Enabled LTO
- Forced `-Os` since the `MinSizeRel` build type doesn't seem to do anything with Emscripten

This reduced C4Tests to 3.1MB (1.1MB gzipped), and CppTests to 5.2MB (1.7MB gzipped). Nice!

PR #2071, opened 2024-06-17
CBL-5865: c4indexupdater all calls, except finish(), crash after fini…

jianminzhao:
…sh()

### DIFF
--- a/C/c4Index.cc
+++ b/C/c4Index.cc
@@ -67,7 +67,7 @@ C4IndexUpdater::~C4IndexUpdater() = default;
 
 size_t C4IndexUpdater::count() const {
     if ( hasFinished() ) {
-        WarnError("C4IndexUpdater::count() called on finished updater.");
+        Warn("C4IndexUpdater::count() called on finished updater.");
         return 0;
     }
     return _update->count();
@@ -75,21 +75,24 @@ size_t C4IndexUpdater::count() const {
 
 FLValue C4IndexUpdater::valueAt(size_t i) const {
     if ( hasFinished() ) {
-        litecore::error::_throw(litecore::error::NotOpen, "C4IndexUpdater::valueAt() called on finished updater.");
+        Warn("C4IndexUpdater::valueAt() called on finished updater.");
+        return FLValue{};
     }
     return _update->valueAt(i);
 }
 
 void C4IndexUpdater::setVectorAt(size_t i, const float* vector, size_t dimension) {
     if ( hasFinished() ) {
-        litecore::error::_throw(litecore::error::NotOpen, "C4IndexUpdater::setVectorAt() called on finished updater.");
+        Warn("C4IndexUpdater::setVectorAt() called on finished updater.");
+        return;
     }
     _update->setVectorAt(i, vector, dimension);
 }
 
 void C4IndexUpdater::skipVectorAt(size_t i) {
     if ( hasFinished() ) {
-        litecore::error::_throw(litecore::error::NotOpen, "C4IndexUpdater::SkipVectorAt() called on finished updater.");
+        Warn("C4IndexUpdater::SkipVectorAt() called on finished updater.");
+        return;
     }
     return _update->skipVectorAt(i);
 }
--- a/C/c4Index.cc
+++ b/C/c4Index.cc
@@ -65,20 +65,39 @@ C4IndexUpdater::C4IndexUpdater(Retained<litecore::LazyIndexUpdate> u, C4Collecti
 
 C4IndexUpdater::~C4IndexUpdater() = default;
 
-size_t C4IndexUpdater::count() const { return _update->count(); }
+size_t C4IndexUpdater::count() const {
+    if ( hasFinished() ) {
+        WarnError("C4IndexUpdater::count() called on finished updater.");
+        return 0;
+    }
+    return _update->count();
+}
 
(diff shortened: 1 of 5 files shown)


C/c4Index.cc
```diff
@@ -65,20 +65,39 @@ C4IndexUpdater::C4IndexUpdater(Retained<litecore::LazyIndexUpdate> u, C4Collecti
...
+FLValue C4IndexUpdater::valueAt(size_t i) const {
+    if ( hasFinished() ) {
+        litecore::error::_throw(litecore::error::NotOpen, "C4IndexUpdater::valueAt() called on finished updater.");
+    }
+    return _update->valueAt(i);
+}
 
 void C4IndexUpdater::setVectorAt(size_t i, const float* vector, size_t dimension) {
+    if ( hasFinished() ) {
+        litecore::error::_throw(litecore::error::NotOpen, "C4IndexUpdater::setVectorAt() called on finished updater.");
```
borrrden, line 85: There is not much need to throw an exception here.  Just log and return.  An exception provides no value and wastes cycles with try / catch.

C/c4Index.cc
```diff
@@ -65,20 +65,39 @@ C4IndexUpdater::C4IndexUpdater(Retained<litecore::LazyIndexUpdate> u, C4Collecti
...
+    }
     _update->setVectorAt(i, vector, dimension);
 }
 
-void C4IndexUpdater::skipVectorAt(size_t i) { return _update->skipVectorAt(i); }
+void C4IndexUpdater::skipVectorAt(size_t i) {
+    if ( hasFinished() ) {
+        litecore::error::_throw(litecore::error::NotOpen, "C4IndexUpdater::SkipVectorAt() called on finished updater.");
```
borrrden, line 92: There is not much need to throw an exception here. Just log and return. An exception provides no value and wastes cycles with try / catch.

borrrden (changes requested):

C/c4Index.cc
jianminzhao: This is actually sneaky. 
```
bool c4indexupdater_skipVectorAt(C4IndexUpdater* update, size_t i) noexcept {
    return tryCatch(nullptr, [&] { update->skipVectorAt(i); });
}
```
It relies on tryCatch to return boolean.

C/c4Index.cc
jianminzhao: same answer as above

C/c4Index.cc
borrrden: This makes this API inconsistent with skipVector....as this will return an error while skipVector will not.  I don't mind these calls succeeding as no-ops in this case since the eventual call to finish will alert the user to their mistake.

C/c4Index.cc
borrrden: I didn't notice this returned a boolean but it is now a sort of useless one since the problem is not actually known anymore

pasin: I have discussed with @borrrden before but we have thought that the behavior will be as follows from CBL API if the updater is finished:

count() : Return zero
valueAt() : Return null
setVector : No ops
skipVector : No ops
finish() : Throws CouchbaseLiteException with NotOpen error to indicate the mistake.

pasin: As we do treat this an error (error also means an error needs to be returned), can we return true for no-ops?

borrrden: Yes, I don't think there is any issue with just returning and letting the method return true for this case.  So back to my original recommendation.

borrrden (changes requested):
So the comment thread above about doing no-ops.

C/c4Index.cc
```diff
@@ -67,29 +67,32 @@ C4IndexUpdater::~C4IndexUpdater() = default;
...
         return 0;
     }
     return _update->count();
 }
 
 FLValue C4IndexUpdater::valueAt(size_t i) const {
     if ( hasFinished() ) {
-        litecore::error::_throw(litecore::error::NotOpen, "C4IndexUpdater::valueAt() called on finished updater.");
+        Warn("C4IndexUpdater::valueAt() called on finished updater.");
+        return FLValue{};
```
pasin, line 154: I don't know if this the same as null or not but there `kFLNullValue`.

pasin (approved):

C/c4Index.cc
jianminzhao: Here, FLValue is a pointer type. FLValue{} == nullptr

pasin (approved):

LiteCore/tests/VectorQueryTest.cc
```diff
@@ -659,7 +659,9 @@ TEST_CASE_METHOD(SIFTVectorQueryTest, "Index isTrained API", "[Query][.VectorSea
     enc.writeKey("target");
     enc.writeData(slice(kTargetVector, sizeof(kTargetVector)));
     enc.endDictionary();
-    Query::Options            options(enc.finish());
+    Query::Options options(enc.finish());
+    if ( !expectedTrained )
+        expectedWarningsLogged = 1;  //DB WARNING SQLite warning: vectorsearch: Untrained index; queries may be slow
```
borrrden, line 664: I have a PR in which I remove this because it is no longer useful

borrrden (approved):

PR #2075, opened 2024-06-19
CBL-5879: Use same vector-index options struct as vectorsearch repo

callumbirks:
Port of #2058 (cherry-pick 03ca4bd)
> * Use same vector-index options struct as vectorsearch repo
> Replaced IndexSpec::VectorOptions with vectorsearch::IndexSpec.
> 
> * Reimplemented SQLiteKeyStore::isIndexTrained()
> using a query on the vector index that checks whether any vectors have a bucket not equal to -1.
> As part of this I had to add the `const` attribute to `KeyStore::getIndex()`.
> 
> * c4Collection::getIndexesInfo includes vector options
> c4Collection::getIndexesInfo now includes the vector index options, under the key `vector_options`, encoded in a string in > the same format as the virtual-table args in the CREATE INDEX statement.
> 
> * C4Index API update (added getOptions!)
> - Added c4Index::getOptions, which returns the complete vector or FTS options used to create the index.
> - Also added c4Index methods getType, getQueryLanguage, getExpression, isTrained.
> - Added the equivalent C functions.
> - Moved index related functions from c4Collection.h to c4Index.h.
> - In c4Index.h, moved the older database-based functions like c4db_createIndex to the bottom, bracketed by `#ifndef C4_STRICT_COLLECTION_API`.

### DIFF
--- a/C/Cpp_include/c4Index.hh
+++ b/C/Cpp_include/c4Index.hh
@@ -33,19 +33,30 @@ struct C4Index
 
     slice getName() const noexcept { return _name; }
 
+    C4IndexType     getType() const noexcept;
+    C4QueryLanguage getQueryLanguage() const noexcept;
+    slice           getExpression() const noexcept;
+
+    /// Writes the index options to `opts` and returns true. If there are none, returns false.
+    [[nodiscard]] bool getOptions(C4IndexOptions& opts) const noexcept;
+
 #ifdef COUCHBASE_ENTERPRISE
+    bool isTrained() const;
+
     /// Finds new or updated documents for which vectors need to be recomputed by the application.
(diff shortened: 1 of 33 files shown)


callumbirks:
It seems like VectorOptions.lazy was changed to "lazyEncoding" and "lazyEmbedding". I think I need to set `lazyEmbedding` to the value of C4VectorOptions.lazy? Then what is `lazyEncoding` for, when is it set?

snej:
> Then what is `lazyEncoding` for, when is it set?

It's an unused option in the vectorsearch extension that allows vectors to be added without indexing or encoding them.

snej (approved):

PR #2076, opened 2024-06-19
CBL-5885: Lazy Index cannot be created using c4coll_createIndex

callumbirks:


callumbirks:
I'll echo the same message here from #2075:
It seems like VectorOptions.lazy was changed to "lazyEncoding" and "lazyEmbedding". I think I need to set lazyEmbedding to the value of C4VectorOptions.lazy? Then what is lazyEncoding for, when is it set?

snej (approved):
Oops! 😬 Thanks for fixing this.

PR #2078, opened 2024-06-22
Fix logic error in c4coll_isIndexTrained

borrrden:
It was universally returning true.  Also remove the warnings check from the VS test since it makes no sense anymore and is really hard to in the isTrained test now that "untrained" is a warning.  Add in a test for the C API to make sure it works correctly.

### DIFF
--- a/C/c4CAPI.cc
+++ b/C/c4CAPI.cc
@@ -558,7 +558,7 @@ bool c4db_createIndex2(C4Database* database, C4Slice name, C4Slice indexSpec, C4
 
 bool c4coll_isIndexTrained(C4Collection* collection, C4Slice name, C4Error* outError) noexcept {
     if ( outError ) *outError = kC4NoError;
-    return tryCatch(outError, [=] { return collection->isIndexTrained(name); });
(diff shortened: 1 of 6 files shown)


pasin (approved):

pasin:
Build failed in PR validation.

jianminzhao (approved):

jianminzhao:
[CBL-5906](https://issues.couchbase.com/browse/CBL-5906)
I think this PR does the above. Please confirm.

borrrden:
Yes, I didn't know there was a ticket for that.  I added those checks in.

snej:
I disagree about removing the warnings check; I think it's useful. And I have a separate PR where I fixed the tests by correctly setting the number of expected warnings. Could you take that out of this PR, and we can decide whether or not to do it?

borrrden:
Needing to know how many warnings are logged in order for a test to succeed is really quite obnoxious though.  If I take it out of here I need a system of incrementing the expected warnings count by the number of sections since the check only resets at the end of the test....

borrrden:
Or maybe I'm hallucinating?  I seem to have put the increment in the correct place this time and the test is ok with it.

snej (approved):

PR #2079, opened 2024-06-24
CBL-5907: Adjusting tests due to CBL-5536 in the VectorSearch extension

jianminzhao:
CBL 5536 introduced a new warning when index is not trained. It makes some tests of [.VectorSearch] to fail because we check the number of warning logs.


borrrden:
Can we just remove it instead?  Trying to keep track of the number of warnings is kind of insane and was only there because we had no other way to check certain things.

jianminzhao:
Agreed. It's hard to manage it moving forward.

jianminzhao:
@snej & @pasin , what do you think?

borrrden:
See https://github.com/couchbase/couchbase-lite-core/pull/2078 where I removed it

jianminzhao:
We won't do it.

PR #2081, opened 2024-06-25
CBL-5924: Cannot set C4VectorIndexOptions.numProbes = 0

jianminzhao:
Also fixed a bug in CollectionImpl::createIndex, a missing "break" in switch statement.

### DIFF
--- a/LiteCore/Database/CollectionImpl.hh
+++ b/LiteCore/Database/CollectionImpl.hh
@@ -422,6 +422,7 @@ namespace litecore {
                         switch ( c4Opt.metric ) {
                             case kC4VectorMetricEuclidean:
                                 vecOpt.metric = vectorsearch::Metric::Euclidean2;
+                                break;
                             case kC4VectorMetricCosine:
                                 vecOpt.metric = vectorsearch::Metric::Cosine;
                                 break;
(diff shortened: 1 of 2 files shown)


LiteCore/Database/CollectionImpl.hh
```diff
@@ -422,6 +422,7 @@ namespace litecore {
                         switch ( c4Opt.metric ) {
                             case kC4VectorMetricEuclidean:
                                 vecOpt.metric = vectorsearch::Metric::Euclidean2;
+                                break;
```
borrrden, line 425: If no test caught this....that is an issue.

jianminzhao: It only affects the result, but we don't have tests that test the result with different metric, I think.

borrrden: I think we might be able to assume that FAISS does, and that if the correct results were passed to FAISS then the result should be correct.

LiteCore/tests/LazyVectorAPITest.cc
```diff
@@ -407,6 +407,7 @@ TEST_CASE_METHOD(LazyVectorAPITest, "BeginUpdate on Non-Vector", "[API][.VectorS
 
     C4Error err{};
     auto    _ = c4index_beginUpdate(index, 10, &err);
+    (void)_;
```
snej, line 410:
```suggestion
    CHECK(c4index_beginUpdate(index, 10, &err) == nullptr);
```
Same for the instance below.
This is just basic error checking, guys... 😩

snej (changes requested):

LiteCore/tests/LazyVectorAPITest.cc
jianminzhao: This is because our build checks are less stringent than Xcode. This is the first thing I hit after pulled from the remote and quickly got it over. 
Will do that check

snej (approved):

PR #2084, opened 2024-06-26
CBL-5931: Clear SQL statement bindings after use

jianminzhao:
Using SQLite::Statement::bindNoCopy() with a non-temporary Statement leaves dangling pointers in its bindings. This isn't usually a problem because the bindings get updated before the next time the statement is run ... but if the statement is still around when the db closes, the code that logs open statements will try to deref the bindings and may crash.

Since there were numerous places this needed to be fixed, I improved the `UsingStatement` helper class to clear bindings.


jianminzhao:
It is all auto-merged by Git.

borrrden (approved):

PR #2085, opened 2024-06-27
CBL-5933: Replicator log contains auth info

jianminzhao:
Made a white list based on the current constants of keys of repl::Options except for kC4ReplicatorAuthPassword. We only log properties in the white list.

Platform added options won't be logged because of the white-list.

### DIFF
--- a/Replicator/Worker.cc
+++ b/Replicator/Worker.cc
@@ -46,10 +46,9 @@ namespace litecore::repl {
         for ( Dict::iterator i(dict); i; ++i ) {
             if ( n++ > 0 ) s << ", ";
             slice key = i.keyString();
+            if ( Options::kWhiteListOfKeysToLog.find(key) == Options::kWhiteListOfKeysToLog.end() ) continue;
             s << key << ":";
-            if ( key == slice(C4STR(kC4ReplicatorAuthPassword)) ) {
-                s << "\"********\"";
-            } else if ( i.value().asDict() ) {
+            if ( i.value().asDict() ) {
                 writeRedacted(i.value().asDict(), s);
             } else {
                 alloc_slice json(i.value().toJSON5());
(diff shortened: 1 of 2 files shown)


Replicator/Worker.cc
```diff
@@ -336,4 +335,71 @@ namespace litecore::repl {
...
+            // [1]: Auth dictionary keys:
+            kC4ReplicatorAuthType,
+            kC4ReplicatorAuthUserName,
+            // kC4ReplicatorAuthPassword,
+            kC4ReplicatorAuthEnableChallengeAuth,
+            kC4ReplicatorAuthClientCert,
+            kC4ReplicatorAuthClientCertKey,
+            kC4ReplicatorAuthToken,
+
+            // [2]: auth.type values:
```
borrrden, line 386: These are values, not keys, and thus don't need to be considered

Replicator/Worker.cc
```diff
@@ -336,4 +335,71 @@ namespace litecore::repl {
...
+            kC4AuthTypeClientCert,
+
+            // [3]: Proxy dictionary keys:
+            kC4ReplicatorProxyType,
+            kC4ReplicatorProxyHost,
+            kC4ReplicatorProxyPort,
+            kC4ReplicatorProxyAuth,
+
+            // [4]: proxy.type values:
+            kC4ProxyTypeNone,
```
borrrden, line 400: Same comment

borrrden (changes requested):

Replicator/Worker.cc
jianminzhao: good point

borrrden (approved):

Replicator/Worker.cc
```diff
@@ -336,4 +335,58 @@ namespace litecore::repl {
...
+            kC4ReplicatorCompressionLevel,
+
+            // [1]: Auth dictionary keys:
+            kC4ReplicatorAuthType,
+            kC4ReplicatorAuthUserName,
+            // kC4ReplicatorAuthPassword,
+            kC4ReplicatorAuthEnableChallengeAuth,
+            kC4ReplicatorAuthClientCert,
+            kC4ReplicatorAuthClientCertKey,
+            kC4ReplicatorAuthToken,
```
pasin, line 384: Shall we log this?

jianminzhao: When find it out it's not okay to log, we can take it out from the white list easily.

pasin: This info shouldn't be logged.

jianminzhao: okay, will remove it.

PR #2086, opened 2024-06-28
CBL-5936: IndexUpdater's finish() returns an error when there are any null vector set

callumbirks:
I added in the `modified` property to the IndexUpdater's `Item`, this allows us to check that all vectors have been modified in some way without checking `vector != nullptr`, because any vector set to null would break that condition.

### DIFF
--- a/LiteCore/Query/LazyIndex.hh
+++ b/LiteCore/Query/LazyIndex.hh
@@ -92,12 +92,13 @@ namespace litecore {
 
         using VectorPtr = std::unique_ptr<float[]>;
 
-        bool anyVectorNotUpdatedOrSkipped() const;
+        bool anyVectorNotModified() const;
 
         struct Item {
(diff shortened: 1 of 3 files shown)


LiteCore/Query/LazyIndex.hh
```diff
@@ -92,12 +92,18 @@ namespace litecore {
 
         using VectorPtr = std::unique_ptr<float[]>;
 
-        bool anyVectorNotUpdatedOrSkipped() const;
+        bool anyVectorNotModified() const;
+
+        enum class ItemStatus {
```
jianminzhao, line 97: it can be ItemStatus :  uint8_t

jianminzhao (changes requested):

LiteCore/Query/LazyIndex.hh
callumbirks: Why?

jianminzhao (approved):

PR #2093, opened 2024-07-03
Last-minute changes to vector search SQL++ syntax

snej:
As per recent meetings & email threads, for Server SQL++ compatibility.

- There is no more `VECTOR_MATCH` function.
- `VECTOR_DISTANCE` renamed to `APPROX_VECTOR_DIST`.
  - Parameters are (_vector expression_, _target vector_, [_metric_], [_nprobes_]); last two optional.
  - _metric_, if given, is a case-insensitive string that must match the index's metric. Current supported values are `"EUCLIDEAN_SQUARED"`, `"L2_SQUARED"`, `"COSINE"`.
  - _nprobes_, if given, overrides the value specified when the index was created.
- A non-hybrid vector query must have a `LIMIT` clause. There is no more default limit.
- A hybrid query is one with a `WHERE` clause (and a `APPROX_VECTOR_DIST` call, of course.)
- Two new metrics in the API: `kC4VectorMetricEuclidean1` and `kC4VectorMetricDot`. 
  - I renamed `kC4VectorMetricEuclidean` to `kC4VectorMetricEuclidean2` to avoid ambiguity.
  - Using the new metrics at runtime requires an updated vector-search extension; see https://github.com/couchbaselabs/mobile-vector-search/pull/69
  - Their SQL++ names are `"EUCLIDEAN"` (AKA `"L2"`) and `"DOT"`.

[CBL-5980](https://issues.couchbase.com/browse/CBL-5980)

### DIFF
--- a/C/tests/c4IndexUpdaterTest.cc
+++ b/C/tests/c4IndexUpdaterTest.cc
@@ -423,7 +423,7 @@ TEST_CASE_METHOD(LazyVectorAPITest, "IndexUpdater Set Float Array", "[API][.Vect
     REQUIRE(c4indexupdater_finish(updater, ERROR_INFO()));
 
     auto query = REQUIRED(c4query_new2(db, kC4JSONQuery, alloc_slice(json5(R"({
-            ORDER_BY: ['VECTOR_DISTANCE()', ['.vector'], ['$target']],
+            ORDER_BY: ['APPROX_VECTOR_DIST()', ['.vector'], ['$target']],
             WHAT:  [ ['.word'] ],
             FROM:  [{'COLLECTION':'words'}],
             LIMIT: 300
@@ -518,7 +518,7 @@ TEST_CASE_METHOD(LazyVectorAPITest, "IndexUpdater not update when released witho
     c4indexupdater_release(updater);
 
     const auto query = REQUIRED(c4query_new2(db, kC4JSONQuery, alloc_slice(json5(R"({
-            ORDER_BY: ['VECTOR_DISTANCE()', ['.word'], ['$target']],
+            ORDER_BY: ['APPROX_VECTOR_DIST()', ['.word'], ['$target']],
             WHAT:  [ ['.word'] ],
             FROM:  [{'COLLECTION':'words'}],
             LIMIT: 300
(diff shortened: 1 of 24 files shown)


LiteCore/tests/VectorQueryTest.cc
```diff
@@ -573,8 +568,7 @@ TEST_CASE_METHOD(SIFTVectorQueryTest, "Index isTrained API", "[Query][.VectorSea
             store          = &db->getKeyStore(string(".") + collectionName);
         }
 
-        expectedTrained    = false;
-        expectedPretrained = false;
+        expectedTrained = false;
```
borrrden, line 571: What happened to the preTrained test?

snej: I removed `expectedPretrained` because Clang warned that its value is never read. I assumed it was left in the code unintentionally.

jianminzhao (approved):

C/tests/c4IndexUpdaterTest.cc
```diff
@@ -423,7 +423,7 @@ TEST_CASE_METHOD(LazyVectorAPITest, "IndexUpdater Set Float Array", "[API][.Vect
     REQUIRE(c4indexupdater_finish(updater, ERROR_INFO()));
 
     auto query = REQUIRED(c4query_new2(db, kC4JSONQuery, alloc_slice(json5(R"({
-            WHERE: ['VECTOR_MATCH()', 'words_index', ['$target']],
+            ORDER_BY: ['APPROX_VECTOR_DIST()', ['.vector'], ['$target']],
             WHAT:  [ ['.word'] ],
```
jianminzhao, line 427: Is ['.word'] intentional here, as opposed to ['.vector'] ?

LiteCore/tests/VectorQueryTest.cc
jianminzhao: It seems like an merge error. 7b21a6436b73ed.

C/tests/c4IndexUpdaterTest.cc
snej: Yes. That's the way the query was before, it returns the matching words.

snej:
I added a small commit so that a vector search without an index now returns NoSuchIndex, not InvalidQuery, as per the meeting this afternoon.

LiteCore/tests/VectorQueryTest.cc
jianminzhao: Filed a ticket for it [CBL-5983](https://issues.couchbase.com/browse/CBL-5983) f

pasin:
If this PR is ready to merge, we could merge this first and get the LiteCore build for the platform to use. Any bugs found can be fixed separately.

PR #2094, opened 2024-07-05
CBL-5612: TLS and ZIP logs are excessive

callumbirks:
Cherry-picked 665a506 from CBL-5605.


borrrden:
If we are going to work on a 3.1 release then this should have a cloned ticket for its version.  Reusing the 3.2 one is not a great solution because of how easy it is to miss things.

callumbirks:
@borrrden CBL-5612 is a 3.1.x ticket

bmeike (commented):
:white_check_mark:

PR #2096, opened 2024-07-08
CBL-5979: LiteCore should hold the names of its log domains

jianminzhao:
Most log domains are generated from inside the library with literal string names. The are held by the language. For domains created via c4log_getDomain, we will hold the name in class LogDomain.

Also fixed a bug in test "LogEncoder levels/domains."

### DIFF
--- a/C/c4Base.cc
+++ b/C/c4Base.cc
@@ -184,7 +184,7 @@ CBL_CORE_API C4LogDomain const kC4WebSocketLog = (C4LogDomain)&websocket::WSLogD
 C4LogDomain c4log_getDomain(const char* name, bool create) noexcept {
     if ( !name ) return kC4DefaultLog;
     auto domain = LogDomain::named(name);
-    if ( !domain && create ) domain = new LogDomain(name);
+    if ( !domain && create ) domain = new LogDomain(name, LogLevel::Info, true);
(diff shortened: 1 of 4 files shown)


C/c4Base.cc
```diff
@@ -184,7 +184,7 @@ CBL_CORE_API C4LogDomain const kC4WebSocketLog = (C4LogDomain)&websocket::WSLogD
 C4LogDomain c4log_getDomain(const char* name, bool create) noexcept {
     if ( !name ) return kC4DefaultLog;
     auto domain = LogDomain::named(name);
-    if ( !domain && create ) domain = new LogDomain(name);
+    if ( !domain && create ) domain = new LogDomain(name, LogLevel::Info, true);
```
pasin, line 187: I don't know if the default level should be Info or something else like warning.

pasin (approved):

C/c4Base.cc
borrrden: Probably leave it as the default?  Why explicitly change it here I wonder?

jianminzhao: This is because of 
` explicit LogDomain(const char* name, LogLevel level = LogLevel::Info)`
before this commit, and the new argument, internName, has default false.

PR #2097, opened 2024-07-09
Fixed hybrid vector search checks [CBL-5985]

snej:
- If the WHERE clause consists only of APPROX_VECTOR_DIST() being less than (or equal to) something, it's not a hybrid query.
- A WHERE clause containing APPROX_VECTOR_DIST() inside an OR clause is illegal.
- A change I made to support this (marking the FROM clause in the `_context` stack) resulted in some redundant parens being removed from generated WHERE clauses. I updated the QueryParserTests to reflect this.


pasin (approved):
I have used this in CBL-C tests, and all tests are passed.

jianminzhao (approved):

PR #2099, opened 2024-07-11
CBL-5964: Test APPROX_VECTOR_DIST error cases

jianminzhao:
1. Use APPROX_VECTOR_DIST without a vector index for the specified vector expression.
2. Use APPROX_VECTOR_DIST with a vector index with unmatched distance metric.
3. Use APPROX_VECTOR_DIST on a non-hybrid query without LIMIT:
 - Use APPROX_VECTOR_DIST in ORDER BY
 - Use APPROX_VECTOR_DIST in WHERE
 - Use APPROX_VECTOR_DIST in SELECT
4. Use APPROX_VECTOR_DIST with accurate = true
6. Use APPROX_VECTOR_DIST in a hybrid query with OR expression in WHERE clause.
7. Use APPROX_VECTOR_DIST in a non-hybrid query with OR expression in WHERE clause.
8. Use invalid num probes in the query such as 0, or negative number.

* Also fixed an error message in SQLiteDataFile.cc

### DIFF
--- a/LiteCore/tests/VectorQueryTest.cc
+++ b/LiteCore/tests/VectorQueryTest.cc
@@ -688,50 +688,98 @@ TEST_CASE_METHOD(SIFTVectorQueryTest, "APPROX_VECTOR_DISTANCE Errors (Bad Metric
     string          post = R"() LIMIT 5)";
     Retained<Query> query;
 
+    // c.f. kMetricNames in VectorIndexSpec.cc
+    const char*        kMetrics[]   = {"euclidean",
+                                       "L2",
+                                       "euclidean2",
+                                       "L2_squared",
+                                       "euclidean_squared",
+                                       "cosine",  //5
+                                       "dot",
+                                       "cosine_distance",
+                                       "cosine_similarity",
+                                       "dot_product_distance",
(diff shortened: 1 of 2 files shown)


LiteCore/tests/VectorQueryTest.cc
```diff
@@ -680,4 +680,257 @@ N_WAY_TEST_CASE_METHOD(SIFTVectorQueryTest, "Inspect Vector Index", "[Query][.Ve
...
+                CHECK(err.domain == error::LiteCore);
+                CHECK(err.code == error::InvalidQuery);
+                CHECK("in 3rd argument to APPROX_VECTOR_DISTANCE, cosine does not match the index's metric, euclidean2"s
+                      == err.what());
+            }
+            CHECK(query == nullptr);
+        }
+    }
+
+    SECTION("Non-default Metric") {
```
pasin, line 717: Do you think it it's worth to test all metrics we supported?

jianminzhao: It's good idea. At least there will be place all metrics are touched.
Will do it.

pasin (approved):

PR #2101, opened 2024-07-12
CBL-5479: Test "Rapid Restarts" failing frequently on Linux

jianminzhao:
After the replicator is stopped, it may still take some time to wind down objects Our test currently allows 2 seconds for it. This turns out not enough in situation when there are following errors,

Sync ERROR Obj=/Repl#21/revfinder#26/ Got LiteCore error: LiteCore NotOpen, "database not open"

This error is artificial. It is because we close the database as soon as the replicator goes to stopped, as opposed to when the replicator is deleted. The error follows the exception to the top frame and takes substantial time when there are many of them.

I increased 2 seconds allowance to 20 seconds. We won't always wait for 20 seconds, because the waiter polls the condition every 50ms. A situation of actual memory leak will wait for whole 20 seconds before failure.


Replicator/tests/ReplicatorAPITest.cc
```diff
@@ -696,6 +696,8 @@ TEST_CASE_METHOD(ReplicatorAPITest, "Rapid Restarts", "[C][Push][Pull]") {
     C4Error err;
     REQUIRE(startReplicator(kC4Continuous, kC4Continuous, WITH_ERROR(&err)));
     waitForStatus(kC4Busy, 5s);
+    // give little time in busy
+    std::this_thread::sleep_for(50ms);
```
jianminzhao, line 700: This 50ms is to trigger the situation when 2 seconds are not enough to wait to check memory leak.

callumbirks (approved):
Nice! This test has been bugging me for a while.

PR #2102, opened 2024-07-16
CBL-5978: LiteCore should hold the names of its log domains

jianminzhao:
Most log domains are generated from inside the library with literal string names. They are held by the language. For domains created via c4log_getDomain, we will hold the name in class LogDomain.

Also fixed a bug in test "LogEncoder levels/domains."


bmeike (commented):
:white_check_mark:

callumbirks (approved):

PR #2104, opened 2024-07-19
CBL-6000: Lazy Vector Index Updater doesn't recongize N1QL dictionary…

jianminzhao:
… and array literal expression

* We currently don't store the query language of the index expression in the database. When retrieved from the db, we used heruistics to determine whether it's JSON or N1QL. This heuristic broke down when the vector index is lazy and expression is N1QL, somehow like, "{'key': property}". The heuristic leads to N1QL expression.

To unqmbiguously distinguish between JSON and N1QL, we use illegal prefixes, "=j" for JSON and "=n" for N1QL. We prefix the expression when we register it with database table, and remove them when we retriieve it from db table, if they are present. If the prefix is present, we use the prefix to determine its QueryLanguage. If not present, we assume it is registered prior to this commit and we use the current heuristic.

The prefixed expressions are not parsable, and so they should not exist in the existing index registry, and they should not exist in objects of IndexSpec.

### DIFF
--- a/LiteCore/Query/QueryParser+VectorSearch.cc
+++ b/LiteCore/Query/QueryParser+VectorSearch.cc
@@ -37,10 +37,14 @@ namespace litecore {
         slice metricName;
         if ( auto metricVal = params[2] )
             metricName = requiredString(metricVal, "3rd argument (metric) to APPROX_VECTOR_DISTANCE");
+            // The following requirement does not apply to the lazy index. At this place, we don't know
+            // if uses lazy index.
+#    if 0
         require(expr->type() == kArray,
                 "first argument to APPROX_VECTOR_DISTANCE must evaluate to a vector; did you pass the index name %s "
                 "instead?",
                 exprJSON.c_str());
+#    endif
         return _delegate.vectorTableName(_defaultTableName, exprJSON, metricName);
     }
 
--- a/LiteCore/Query/QueryParser+VectorSearch.cc
+++ b/LiteCore/Query/QueryParser+VectorSearch.cc
@@ -37,14 +37,7 @@ namespace litecore {
         slice metricName;
         if ( auto metricVal = params[2] )
             metricName = requiredString(metricVal, "3rd argument (metric) to APPROX_VECTOR_DISTANCE");
-            // The following requirement does not apply to the lazy index. At this place, we don't know
-            // if uses lazy index.
-#    if 0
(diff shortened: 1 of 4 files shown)


LiteCore/Query/QueryParser+VectorSearch.cc
```diff
@@ -37,10 +37,14 @@ namespace litecore {
...
         if ( auto metricVal = params[2] )
             metricName = requiredString(metricVal, "3rd argument (metric) to APPROX_VECTOR_DISTANCE");
+            // The following requirement does not apply to the lazy index. At this place, we don't know
+            // if uses lazy index.
+#    if 0
         require(expr->type() == kArray,
                 "first argument to APPROX_VECTOR_DISTANCE must evaluate to a vector; did you pass the index name %s "
                 "instead?",
                 exprJSON.c_str());
+#    endif
```
pasin, line 47: I think we just remove this. There is no need to do `if 0`. We could get the code back from git.

LiteCore/Query/SQLiteDataFile+Indexes.cc
```diff
@@ -259,7 +275,10 @@ namespace litecore {
                 auto what = spec.what();
                 // `what()` is defined as an array of 1+ exprs to index; for a vector index there can be only one.
                 // In some cases just that term is passed in, not wrapped in an array.
-                if ( what->count() > 1 || what->get(0)->type() == kArray ) what = (const Array*)what->get(0);
+                if ( what->count() > 1
+                     || (spec.queryLanguage == QueryLanguage::kN1QL || what->get(0)->type() == kArray) ) {
```
pasin, line 279: Is this change needed because what->get(0) may not be an array when using lazy index?

LiteCore/Query/QueryParser+VectorSearch.cc
jianminzhao: I can do it

LiteCore/Query/SQLiteDataFile+Indexes.cc
jianminzhao: yes

bmeike (commented):
:white_check_mark:

pasin (approved):
This approch works fine but I think we should plan to add a language column to the indexes table so that we don't need to do this heuristic check.

PR #2105, opened 2024-07-19
DRAFT: CBL-5997 (cherry picked)

callumbirks:
I don't like this branch very much. The cherry pick caused a lot of conflicts, took more than an hour to resolve, makes way too many changes for a patch release, and many replicator tests are failing.

I'm putting this up in a draft PR to use for comparison to my next attempt (manually adding as little as possible to log the correlation ID).


callumbirks:
Closing this - it was only used as an example why we don't want the cherry-pick

PR #2106, opened 2024-07-19
CBL-5997: Logging correlation ID

callumbirks:
This is an attempt at adding the logging of correlation ID, with as minimal changes as possible.
See #2105 for an example of the opposite.

### DIFF
--- a/Networking/BLIP/BLIPConnection.cc
+++ b/Networking/BLIP/BLIPConnection.cc
@@ -187,8 +187,7 @@ namespace litecore { namespace blip {
         // websocket::Delegate interface:
         virtual void onWebSocketConnect() override {
             _timeOpen.reset();
-            _connection->connected();
-            onWebSocketWriteable();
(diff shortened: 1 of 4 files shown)


Networking/BLIP/BLIPConnection.cc
```diff
@@ -187,8 +187,7 @@ namespace litecore { namespace blip {
         // websocket::Delegate interface:
         virtual void onWebSocketConnect() override {
             _timeOpen.reset();
-            _connection->connected();
-            onWebSocketWriteable();
+            enqueue(FUNCTION_TO_QUEUE(BLIPIO::_onWebSocketConnect));
```
jianminzhao, line 190: This is not related to the title of the CBL ticket. Perhaps add a note in the commit comment.

jianminzhao (approved):

Networking/BLIP/BLIPConnection.cc
callumbirks: Thanks, will do

PR #2109, opened 2024-07-26
CBL-6009: Provide an option to enable full sync in the database

jianminzhao:
Added a flag to C4DatabaseFlags, kC4DB_DiskSyncFull. This flag is passed to DataFile::Options, which is used as we request a connection to the SQLite database.

### DIFF
--- a/C/include/c4DatabaseTypes.h
+++ b/C/include/c4DatabaseTypes.h
@@ -37,7 +37,7 @@ typedef C4_OPTIONS(uint32_t, C4DatabaseFlags) {
     kC4DB_VersionVectors= 0x08, ///< Upgrade DB to version vectors instead of rev trees [EXPERIMENTAL]
     kC4DB_NoUpgrade     = 0x20, ///< Disable upgrading an older-version database
     kC4DB_NonObservable = 0x40, ///< Disable database/collection observers, for slightly faster writes
-    kC4DB_DiskSyncFull  = 0x80, ///< Write to disk after each transaction
+    kC4DB_DiskSyncFull  = 0x80, ///< Flush to disk after each transaction
 };
 
 
--- a/C/include/c4DatabaseTypes.h
+++ b/C/include/c4DatabaseTypes.h
@@ -37,6 +37,7 @@ typedef C4_OPTIONS(uint32_t, C4DatabaseFlags) {
     kC4DB_VersionVectors= 0x08, ///< Upgrade DB to version vectors instead of rev trees [EXPERIMENTAL]
     kC4DB_NoUpgrade     = 0x20, ///< Disable upgrading an older-version database
     kC4DB_NonObservable = 0x40, ///< Disable database/collection observers, for slightly faster writes
+    kC4DB_DiskSyncFull  = 0x80, ///< Write to disk after each transaction
 };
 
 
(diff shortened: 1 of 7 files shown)


bmeike (commented):
:white_check_mark:

C/include/c4DatabaseTypes.h
```diff
@@ -37,6 +37,7 @@ typedef C4_OPTIONS(uint32_t, C4DatabaseFlags) {
     kC4DB_VersionVectors= 0x08, ///< Upgrade DB to version vectors instead of rev trees [EXPERIMENTAL]
     kC4DB_NoUpgrade     = 0x20, ///< Disable upgrading an older-version database
     kC4DB_NonObservable = 0x40, ///< Disable database/collection observers, for slightly faster writes
+    kC4DB_DiskSyncFull  = 0x80, ///< Write to disk after each transaction
```
pasin, line 40: Maybe only SyncFull or FullSync?

pasin (approved):

C/include/c4DatabaseTypes.h
borrrden: The comment is also probably misleading.  All transactions write to the disk after they finish successfully, but the full sync also introduces a manual flush.

LiteCore/tests/c4BaseTest.cc
```diff
@@ -144,6 +146,30 @@ TEST_CASE("C4Error Reporting Macros", "[Errors][C]") {
 }
 
 
+TEST_CASE_METHOD(C4Test, "Database Flag FullSync", "[Database][C]") {
```
borrrden, line 149: There should probably at least be a manual inspection of one of these databases to confirm that the full sync is taking effect at the SQLite setting.

C/include/c4DatabaseTypes.h
jianminzhao: That was my choice in the first attempt, but afterwards, I feel "disk" makes it clearer. It is about the sync between the disk and memory.

LiteCore/tests/c4BaseTest.cc
jianminzhao: I did manual inspection here at 
`            _exec(format("PRAGMA cache_size=%d; "            // Memory cache
                         "PRAGMA mmap_size=%d; "             // Memory-mapped reads
                         "PRAGMA synchronous=%s; "           // Speeds up commits
                         "PRAGMA journal_size_limit=%lld; "  // Limit WAL disk usage
                         "PRAGMA case_sensitive_like=true; "   // Case sensitive LIKE, for N1QL compat
                         "PRAGMA fullfsync=ON",              // Attempt to mitigate damage due to sudden loss of power (iOS / macOS)
                         -(int)kCacheSize/1024, kMMapSize, getOptions().diskSyncFull?"full":"normal",
                         (long long)kJournalSize));
`
by breaking there and inspected the string in either case constructed by "format", particularly, "full" vs "normal".

jianminzhao: Is it sufficient?

C/include/c4DatabaseTypes.h
jianminzhao: "Flush to disk after each transaction" ?
@borrrden

LiteCore/tests/c4BaseTest.cc
borrrden: That seems reasonable

borrrden (approved):

PR #2111, opened 2024-07-29
CBL-6099: Test "Rapid Restarts" failing frequently on Linux

jianminzhao:
After the replicator is stopped, it may still take some time to wind down objects Our test currently allows 2 seconds for it. This turns out not enough in situation when there are following errors,

Sync ERROR Obj=/Repl#21/revfinder#26/ Got LiteCore error: LiteCore NotOpen, "database not open"

This error is artificial. It is because we close the database as soon as the replicator goes to stopped, as opposed to when the replicator is deleted. The error follows the exception to the top frame and takes substantial time when there are many of them.

I increased 2 seconds allowance to 20 seconds. We won't always wait for 20 seconds, because the waiter polls the condition every 50ms. A situation of actual memory leak will wait for whole 20 seconds before failure.

### DIFF
--- a/C/tests/c4Test.cc
+++ b/C/tests/c4Test.cc
@@ -223,7 +223,7 @@ C4Test::~C4Test() {
 
     if ( !current_exception() ) {
         // Check for leaks:
-        if ( !WaitUntil(2000ms, [&] { return c4_getObjectCount() - objectCount == 0; }) ) {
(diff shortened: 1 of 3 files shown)


jianminzhao:
completely auto merged.

jianminzhao:
Code changes are limited to the test folder.

pasin (approved):

PR #2112, opened 2024-07-29
CBL-6104: Flaky test, "Multiple Collections Incremental Revisions"

jianminzhao:
The reason the test occasionally fails because we assumeed that successive revisions all get replicated to the destinations, whereas replicator may skip obsolete revision, say rev-2, when rev-3 exists when the pusher turns to find the revisions to push. In our test case, we create successive revisions in intervals of 500 milliseconds. Most time, 500 ms is enough to set apart when the pusher picks the revisions. On Jenkins machine, the log shows that it found obsolete revisions when the test failed.

We modified the test success criteria: we only check that the latest revisions are replicated to the destination. This is the designed behavior.


jianminzhao:
All auto-merged

jianminzhao:
Code changes are limited to the test folder.

borrrden (approved):

PR #2115, opened 2024-08-15
Build with C++20

snej:
- Changed compiler flags from C++17 to C++20. The code unfortunately no longer builds as C++17 thanks to the lambda-capture changes...
- Capturing `this` in a lambda without mentioning it explicitly in the capture list is deprecated. Unfortunately, adding the explicit `this` is a syntax error in C++17, so we can't easily build both ways, not without specifying every single captured variable.
- The `format` function in StringUtils conflicts with `std::format` so I renamed it `stringprintf`.
- Symbols in our "date.h" header conflict with ones in std::chrono so I added namespace prefixes to disambiguate.
- Updated Fleece for C++20.


snej:
MSVC has a single error: `CookieStore.cc(45,24): error C2672: 'date::parse': no matching overloaded function found`. This has to do with some nasty template stuff in the 3rd-party `date.h` in Fleece; I can't figure out what could be going on.

snej:
The Jenkins builds of both iOS and Android fail due to the use of `std::ranges`. iOS doesn't seem to have `ranges::any_of` or `ranges::sort`, and Android doesn't have that namespace at all. I assume both of these builders are running older toolchains that don't fully support C++20.

borrrden (approved):

PR #2116, opened 2024-08-21
CBL-6136: Annotate all public API methods with required locks

jianminzhao:
Two styles of annotations.
- For methods that are thread-safe, we put following doxygen notes in the header, such as

         \note This function is thread-safe.

- For methods that require locks, the doxygen notes are like the following. ex.1
        \note The caller must use a lock for Database when this function is called.
        ....
        C4Replicator* c4repl_newWithSocket(C4Database* db, C4Socket* openSocket,

"Database" refers to the first argument of function. Prefix "C4" is always omitted.

ex2.
    \note The caller must use a lock for Database when this function is called. */
    C4SequenceNumber c4coll_getLastSequence(C4Collection*) C4API;

"Database" refers to the Database that the first argument, collection, belongs to.

ex3.
    \note The caller must use a lock for Database when this function is called. */
    bool c4queryenum_restart(C4QueryEnumerator* e, C4Error* C4NULLABLE outError)

"Database" refers to the Database this enumerator is created on.

ex4.
    \note The caller must use a lock for Database and Document when this function is called. */
    bool c4doc_save(C4Document* doc, uint32_t maxRevTreeDepth, C4Error* C4NULLABLE outError)

In this case, locks on the Document (first argument) and on the associated Database must be used.


borrrden (approved):
Personally I wouldn't bother annotating functions that are thread-safe.  Functions that lack a note can be considered thread-safe by default and it will save a big note for every API in the HTML docs, but it's not a huge deal.

bmeike (approved):
This is just sharks with fricken lasers AWESOME
Really nice work, @jianminzhao !!!

PR #2120, opened 2024-08-27
CBL-5872: Remove semi-deprecated APIs

callumbirks:
semi-deprecated c4db and c4doc APIs which existed only to delegate to c4coll APIs have been removed.


borrrden (approved):
When you merge this PR, can you edit the PR to contain a line with the warning icon :warning: and an indication that the APIs were removed so it's not a surprise when this gets released?  It needs to remain present in the notification in litecore-release-notify as well

callumbirks:
@borrrden Yes! Will do

PR #2123, opened 2024-08-27
Fixed UB when a CollectionChangeObserver deletes itself in its callback

snej:
The code that calls observers attempted to protect itself from the list being mutated while it was iterating, but apparently it didn't work well enough.

I changed the code so it now iterates the list collecting the observer pointers, and then calls them afterwards. While I was debugging this I noticed SequenceTracker doesn't have a logging identifier, so I added it.

### DIFF
--- a/LiteCore/Database/SequenceTracker.cc
+++ b/LiteCore/Database/SequenceTracker.cc
@@ -14,6 +14,7 @@
 #include "Logging.hh"
 #include "StringUtil.hh"
 #include "Error.hh"
+#include "SmallVector.hh"
 #include <algorithm>
 #include <sstream>
 #include <utility>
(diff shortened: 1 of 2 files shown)


snej:
Jenkins Mac build failed due to a SIGTERM in test "Pull large database no-conflicts".

callumbirks:
Sorry I took so long, I didn't see this until now!
The Jenkins failure is misleading, it is actually because our suite of Jenkins CI takes > 1 hour now, and at that point is terminated. A large part of that time is the iOS compilation, which takes way too long. I need to update the Jenkins file to allow longer than 1 hour.

callumbirks (approved):

jianminzhao (approved):

PR #2124, opened 2024-08-27
Allow a Timer to destruct itself during its callback

snej:
Fixes a longstanding limitation that IIRC we knew about and just avoided. I couldn't avoid it in the code I'm working on, so I finally fixed it.

Manager::unschedule(), if the timer's being deleted, now checks to see if it's being called on the Manager's thread. If so, it no longer waits for the timer's `_triggered` flag to clear (which caused a deadlock.)

Instead it checks if the timer is the same as the new `_triggeredTimer` variable; if so, it clears the variable.

When control returns to the run() method, if `_triggeredTimer` has been cleared it won't clear the `_triggered` flag or check the `_autoDelete` flag.

### DIFF
--- a/LiteCore/Support/Timer.cc
+++ b/LiteCore/Support/Timer.cc
@@ -11,6 +11,7 @@
 //
 
 #include "Timer.hh"
+#include "Error.hh"
 #include "ThreadUtil.hh"
 #include <vector>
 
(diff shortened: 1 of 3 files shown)


snej:
Jenkins Linux build failed with a SIGTERM in test "Multiple Collections Incremental Revisions".

borrrden (approved):

LiteCore/Support/Timer.cc
```diff
@@ -37,18 +37,22 @@ namespace litecore::actor {
...
                 try {
-                    timer->_callback();
+                    _triggeredTimer->_callback();
                 } catch ( ... ) {}
-                timer->_triggered = false;  // note: not holding any lock
-                if ( timer->_autoDelete ) delete timer;
+
+                if ( _triggeredTimer ) {
+                    _triggeredTimer->_triggered = false;  // note: not holding any lock
+                    if ( _triggeredTimer->_autoDelete ) delete _triggeredTimer;
```
jianminzhao, line 53: It appears to me that after the previous time, _triggeredTime may get deleted. Get the value of _triggeredTimer->_autoDelete before the previous line?

snej: Good point. I've pushed another commit to fix the race condition, and also to enforce that an autoDelete Timer is never manually deleted.

jianminzhao (approved):

PR #2128, opened 2024-08-28
Fix DateFormat.cc some more (for GCC)

snej:
- Use `std::nullopt` instead of `{}`. This alone fixed the errors.
- Simplified declarations by using the type instead of `auto`.


borrrden (approved):

callumbirks (approved):
Nice!

PR #2129, opened 2024-08-29
CBL-6189: Annotation of thread-safety, filling in the missing ones af…

jianminzhao:
…ter the first pass.


bmeike (commented):
Brilliant.  Thanks @jianminzhao

C/include/c4Database.h
```diff
@@ -54,6 +54,7 @@ CBL_CORE_API void c4_setExtensionPath(C4String path) C4API;
  * of the extension and the path in which it is supposed to reside.  It makes an attempt
  * to only check things that have the possibility of being corrected by the user (i.e.
  * if there is a bug in the extension and it cannot load functionally that won't be caught)
+ * \note This function is not thread-safe.
```
borrrden, line 57: What do we need to do about it?

borrrden (changes requested):
There appear to be some important changes here that are not just "filling in the missing".  Some have gone from unannotated to "needs a lock".  Could you list the ones that have done so, so that the platforms can update their thread safety?

C/include/c4Database.h
jianminzhao: This is the one, along with the other one, c4_setTempDir, that I had discussion with @bmeike . They are technically not thread safe, but they don't share the memory with any other functions. You just cannot call the very function concurrently. If you must do it, you need use a lock designated for the very function

jianminzhao: How to note it? "Caller must use a lock for this function when it is called?"

borrrden: Ok I understand, but I am still looking for a list of things that have changed from unannotated (which I took as thread safe) to needing a lock (like c4readstream_getLength)

jianminzhao: With regard to c4stream_getLength, I based my note on the following overall instruction in the original file,
/* NOTE: These functions are thread-safe in the same manner as described in the previous
       section, with the additional restriction that a stream cannot be called concurrently on
       multiple threads. */
Reading the actual code, it is thread-safe because it only perform read from the disk file. Annotating it as "thread-safe," I think, may not be too helpful, because there are only two other methods, c4stream_read and c4stream_seek, are in the thread concern and it's hard to image they would be called concurrently with getLength. But I can change it thread-safe, making it an exception with the overall threading note.

jianminzhao: Per conversation with JIm, the thread-safety note in the sections of c4BlobStore.h is not very helpful. I removed them and, by inspecting the code, noted c4stream_getLength as thread-safe. (The section note characterize it as requiring lock for the stream object.)

borrrden (approved):

PR #2134, opened 2024-09-10
CBL-6235: Statically link to libstdc++

borrrden:
so that we don't have to ship libstdc++ with Core


pasin (approved):
We will need a ticket to remove C++ from the support library when we release with this fix.

borrrden:
The JNI lib I doubt will need it.

jianminzhao (approved):

PR #2137, opened 2024-09-13
CBL-6207: Test multi-level unnest w/o index.

jianminzhao:
The tests come from the API Specification.

Note that "group-by" does not work with unnest without index.

### DIFF
--- a/LiteCore/tests/QueryTest.cc
+++ b/LiteCore/tests/QueryTest.cc
@@ -2991,3 +2991,283 @@ TEST_CASE_METHOD(QueryTest, "Invalid collection names", "[Query]") {
         }
     }
 }
+
+namespace {
+    void AddUnnestDoc(KeyStore* store, slice docID, slice fname, slice lname) {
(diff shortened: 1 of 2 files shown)


borrrden (approved):

LiteCore/tests/QueryTest.cc
```diff
@@ -2991,3 +2991,283 @@ TEST_CASE_METHOD(QueryTest, "Invalid collection names", "[Query]") {
...
+            enc.endDictionary();  // contacts[1]
+
+            enc.endArray();  // contacts
+
+            enc.writeKey("likes");
+            enc.beginArray();
+            enc.writeString("Soccer");
+            enc.writeString("Cat");
+            enc.endArray();
+        });
```
pasin, line 3098: Can we just json string instead of the code above for readability?

pasin (changes requested):

pasin (approved):

PR #2138, opened 2024-09-13
More compiler warnings, and sanitizers, in CMake builds

snej:
Added new CMake options: 

- `LITECORE_WARNINGS_HARDCORE` enables _all_ compiler warnings except for a few dozen that are too onerous, plus `-Werror`.
- `LITECORE_SANITIZE` enables the Clang Address and UB sanitizers.

Currently these are only supported in Clang builds, and ignored with other compilers. They are propagated to Fleece (see https://github.com/couchbase/fleece/pull/239)

Also made source changes to fix some new warnings produced by this.

>**NOTE to reviewers:** It's probably clearest if you review the commits individually rather than looking at the "Files changed" tab. Most of the diffs come from the first commit [343bfac](https://github.com/couchbase/couchbase-lite-core/pull/2138/commits/343bfac6af151b21b86573acc1042d977045024d) which is simply replacing `static inline` with `inline` in standalone functions in header files, and from [2356d46](https://github.com/couchbase/couchbase-lite-core/pull/2138/commits/2356d46577f7e6e4965159324569311a82dbefe0) which is a lot of independent little changes fixing compiler warnings.


jianminzhao:
With this PR, will it compile with the master branch of fleece?

borrrden (changes requested):
Ok code wise but fleece needs to be merged first and then this PR updated to point to the master branch.

jianminzhao:
@borrrden the fleece reference actually points to a commit that has same content as master (via git diff ..master). I changed it to be actual master branch.

borrrden (approved):
Blackduck check is a false positive because of source edits (though at what point does it become our own forked product and not a third party one in terms of security scan...)

PR #2140, opened 2024-09-17
CBL-6148: Update Replication Protocol for 4.0

callumbirks:
So far I've been able to do some manual testing with Sync Gateway.
With this change, with Sync Gateway in the main branch, replication succeeds.
With Sync Gateway 3.1.11, replication fails with a `500 Internal Server Error`.

~~We can't merge this PR until we can get a docker image for Sync Gateway's current main branch, because none of the Sync Gateway Replication tests will pass.~~
EDIT: This statement was not quite accurate, we just need a branch of Sync Gateway with VV + 4.0 protocol enabled. Then some method to manually build Sync Gateway for our docker build process.

### DIFF
--- a/Replicator/Replicator.hh
+++ b/Replicator/Replicator.hh
@@ -26,8 +26,7 @@ namespace litecore::repl {
     class Puller;
     class ReplicatedRev;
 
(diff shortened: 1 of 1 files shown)


callumbirks:
Per CBL-6288, for now we have disabled SG replication tests in Actions, so we can merge this protocol change for now and fix the tests after.

jianminzhao (approved):

PR #2143, opened 2024-09-18
Make test logs readable in CLion

snej:
CLion's test runner seems to treat stderr and std::cerr differently!
- Output to stderr is written while the test is running.
- Output to cerr is suppressed while the test is running, then dumped at the end. 

LogDomain::defaultCallback() happens to write the timestamp and domain to cerr, but the message itself to stderr. This means the timestamp/domain are missing in the logs, then show up in a big jumble at the end.

I fixed defaultCallback to write everything to cerr. This shouldn't make a difference to anything but CLion.


snej:
Looks like Jenkins timed out after an hour?

jianminzhao (approved):

PR #2144, opened 2024-09-19
CMake: Fixed duplicate compilation of source files

snej:
When the option LITECORE_BUILD_TESTS is on (as it is by default), all the LiteCore source files were compiled twice: once in the regular LiteCoreObjects target and once in LiteCoreUnitTesting. The latter target has the preprocessor symbol `LITECORE_CPPTEST` defined.

I've removed the LiteCoreUnitTesting target. Instead, LITECORE_CPPTEST is now defined whenever LITECORE_BUILD_TESTS is on. This means that release builds should use LITECORE_BUILD_TESTS=OFF. If they don't, CMake will issue a warning.

I also made LITECORE_BUILD_TESTS default to OFF in release builds.

On my MacBook Pro this improved debug build times by ~30%.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,7 +11,6 @@ CppTests - A test runner that runs test based on the static library
 FleeceObjects - The Fleece serialization library for saving data to a binary format
 FleeceBase - A subset of the above for linking into non LiteCore targets
 LiteCoreObjects - The precompiled object files containing the logic of LiteCore
-LiteCoreUnitTesting - The precompiled object files containing the logic of LiteCore for unit testing.
 LiteCoreStatic - The static LiteCore library (LiteCoreObjects linked statically)
 LiteCore - The shared LiteCore library (LiteCoreObjects linked dynamically)
 LiteCoreREST - A simple library used for enabling testing of the shared library (not used in production)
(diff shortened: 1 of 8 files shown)


snej:
The Windows build fails when starting the tests, with `LiteCore/tests/Debug: No such file or directory`.

I confess I don't know why Windows puts the tests in a subdirectory like this, nor why it isn't doing so now. I didn't change anything that should cause that...

borrrden:
Windows has always done this.  It's a side effect of cmake being a multi config generator for Windows by default, as opposed to a single config like on Unix.

CMakeLists.txt
```diff
@@ -88,7 +87,13 @@ option(LITECORE_PERF_TESTING_MODE "Build LiteCore with more things public than i
 option(BUILD_ENTERPRISE "Set whether or not to build enterprise edition" OFF)
 option(LITECORE_DISABLE_ICU "Disables ICU linking" OFF)
 option(DISABLE_LTO_BUILD "Disable build with Link-time optimization" OFF)
-option(LITECORE_BUILD_TESTS "Builds C4Tests and CppTests" ON)
+
+if (CMAKE_BUILD_TYPE MATCHES "Debug")
```
borrrden, line 91: Note that this is NOT debug by default like you'd think so if no build type is specified it will not enter here.

PR #2149, opened 2024-09-27
CBL-6208: IndexSpec of ArrayIndex with multi-level

jianminzhao:
- C4IndexOptions: added a new field, unnestPath
- N1QL Grammar: changed "UNNEST propertyPath" to "UNNEST expr".
- Although UNNEST can be followed by a general expression, such as "[num[0], num[1]]", but the parser will throw error if the expression is not a propertyPath.
- UNNEST tables are created when createIndex is called with proper propertyPath.
- With multiple levels of UNNEST, the name of unnest table can grow very long. We use SHA1 hash to put a cap on the size of the name.

Others:
- Fixed some warning errors produced by Xcode 16
- Removed two small test json files I put in last commit and extended importJSONFile to allow JSON string inlined in test.

### DIFF
--- a/C/include/c4IndexTypes.h
+++ b/C/include/c4IndexTypes.h
@@ -118,6 +118,11 @@ typedef struct C4IndexOptions {
         separated by spaces. */
     const char* C4NULLABLE stopWords;
 
+    /** The property path to the array property to be unnested.  If this property is nested in a
+        parent array property, e.g. interests in students[i].interests[j],  the unnestPath would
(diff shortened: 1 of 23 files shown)


jianminzhao:
Only the SG check failed and this PR should not affect the PR test at all.

borrrden (approved):
The risk assessment has not changed for patch release right?

jianminzhao:
It's a stretch for a patch release, but there is no new surprise with respect to the risk.

PR #2156, opened 2024-10-05
CBL-6082: IndexSpec of ArrayIndex with multi-level

jianminzhao:
- Triggers on the unnest tables: automatically update the unnest tables when the main kv-tables are modified with CRUD operations.
- Deletion of the Array Index also deletes the auxiliary unnest tables that beome used. It becomes involved in the face of multi-level arrays. Considering two indexes on unnestPaths that overlap, such as, a[].b[].c and a[].b[].d. When Index on c is deleted, we cannot delete b or a until index on d is deleted.
- Unit test on deletion of the Array Index.
- Yet to add tests on the triggers of insertion, update, and deletion.

A minor fix of C4IndexOptions::unnestPath by making it NULLABLE.

### DIFF
--- a/LiteCore/Query/SQLiteDataFile+Indexes.cc
+++ b/LiteCore/Query/SQLiteDataFile+Indexes.cc
@@ -15,6 +15,7 @@
 #include "SQLite_Internal.hh"
 #include "Error.hh"
 #include "Logging.hh"
+#include "SecureDigest.hh"
 #include "SQLiteCpp/Database.h"
 #include "SQLUtil.hh"
 #include "StringUtil.hh"
--- a/LiteCore/Query/SQLiteDataFile+Indexes.cc
+++ b/LiteCore/Query/SQLiteDataFile+Indexes.cc
@@ -100,7 +100,7 @@ namespace litecore {
                         same = schemaExistsWithSQL(indexTableName, "table", indexTableName, indexSQL);
                         break;
                     case IndexSpec::kArray:
-                        same = schemaExistsWithSQL(spec.name, "index", KeyStore::hexName(indexTableName), indexSQL);
+                        same = schemaExistsWithSQL(spec.name, "index", hexName(indexTableName), indexSQL);
                         break;
                     default:
                         same = schemaExistsWithSQL(spec.name, "index", indexTableName, indexSQL);
@@ -142,7 +142,7 @@ namespace litecore {
                     unnestTables.push_back(tableName);
                 }
             }
-            // Assertion: unnestTables.back() == tableName
+            DebugAssert(unnestTables.back() == tableName);
         }
 
         size_t unnestLevel = unnestTables.size();
@@ -162,7 +162,7 @@ namespace litecore {
                 // This table has child table. Don't delete it and its parent
                 if ( stmt.executeStep() ) return;
 
-                tableName = KeyStore::hexName(tableName);  // Now, it's the true table name, hashed.
+                tableName = hexName(tableName);  // Now, it's the true table name, hashed.
                 // Move on to delete the unused tables.
             }
 
--- a/LiteCore/Query/SQLiteDataFile+Indexes.cc
+++ b/LiteCore/Query/SQLiteDataFile+Indexes.cc
@@ -99,6 +99,9 @@ namespace litecore {
                     case IndexSpec::kVector:
                         same = schemaExistsWithSQL(indexTableName, "table", indexTableName, indexSQL);
                         break;
+                    case IndexSpec::kArray:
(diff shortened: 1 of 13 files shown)


LiteCore/Storage/KeyStore.hh
```diff
@@ -84,6 +85,10 @@ namespace litecore {
         /// Returns true if this is a valid collection name, or a scope plus a collection name.
         [[nodiscard]] static bool isValidCollectionNameWithScope(slice name);
 
+        [[nodiscard]] static std::string hexName(string_view name) {
```
snej, line 88: This function isn't specific to KeyStore; I would put it in SecureDigest.hh instead.

LiteCore/Query/SQLiteDataFile+Indexes.cc
```diff
@@ -122,26 +125,68 @@ namespace litecore {
...
+            for ( size_t pos = 0; pos != string::npos; ) {
+                pos = tableName.find(KeyStore::kUnnestLevelSeparator, pos);
+                if ( pos != string::npos ) {
+                    unnestTables.push_back(tableName.substr(0, pos));
+                    pos += KeyStore::kUnnestLevelSeparator.size;
+                } else {
+                    unnestTables.push_back(tableName);
+                }
+            }
+            // Assertion: unnestTables.back() == tableName
```
snej, line 145: Use DebugAssert to specify assertions/invariants.

LiteCore/Query/SQLiteDataFile+Indexes.cc
```diff
@@ -122,26 +125,68 @@ namespace litecore {
...
+            static const char* kTriggerSuffixes[]       = {"ins", "del", "upd", "preupdate", "postupdate", nullptr};
+            static const char* kNestedTriggerSuffixes[] = {"ins", "del", nullptr};
+
+            const char** triggerSuffixes = kTriggerSuffixes;
+            if ( unnestLevel > 1 ) triggerSuffixes = kNestedTriggerSuffixes;
+
+            {
+                stringstream sql;
+                for ( int i = 0; triggerSuffixes[i]; ++i ) {
+                    sql << "DROP TRIGGER IF EXISTS \"" << tableName << "::" << triggerSuffixes[i] << "\";";
```
snej, line 181: This is unsafe if tableName can contain quote characters. Use sqlIdentifier() to quote the trigger name.
(I know this is pre-existing code, but I noticed the issue and it's worth fixing...)

LiteCore/Query/SQLiteKeyStore+ArrayIndexes.cc
```diff
@@ -31,21 +31,23 @@ namespace litecore {
...
+        string      plainTableName    = qp.unnestedTableName(expression);
+        string      unnestTableName   = KeyStore::hexName(plainTableName);
+        string      quotedParentTable = CONCAT(sqlIdentifier(parentTable));
 
         // Create the index table, unless an identical one already exists:
         string sql = CONCAT("CREATE TABLE " << sqlIdentifier(unnestTableName)
                                             << " "
                                                "(docid INTEGER NOT NULL REFERENCES "
-                                            << sqlIdentifier(parentTable)
+                                            << quotedParentTable
```
snej, line 50: This is actually better the other way -- sqlIdentifier() doesn't allocate a string.

snej (changes requested):

LiteCore/Storage/KeyStore.hh
jianminzhao: will do

LiteCore/Query/SQLiteDataFile+Indexes.cc
jianminzhao: Okay

LiteCore/Query/SQLiteDataFile+Indexes.cc
jianminzhao: okay

jianminzhao: Cannot do it. Changing it sqlIdentifier(tableName) breaks multiple tests. I think it's already quoted, "EXISTS \"" and "::ins" should be part of the table name.

callumbirks (approved):
I'm certainly missing some understanding of how the ArrayIndex / table un-nesting works - but otherwise LGTM.

LiteCore/Query/SQLiteDataFile+Indexes.cc
jianminzhao: Got it. done

snej (approved):

PR #2157, opened 2024-10-10
CBL-6245: Query parser regression related to brackets

callumbirks:
The previous implementation caused problems with matching certain expressions, because it stripped the parentheses at the top level. Now we strip parentheses at the bottom level, similar to before, but without recursing through the entire stack for every level of parentheses.


jianminzhao:
ya, this should do it.

jianminzhao (approved):

PR #2158, opened 2024-10-10
CBL-6305: LiteCore API for Enabling and Disabling SQLite's mmap

callumbirks:
Not entirely sure what the best choice for the platform-facing API is. 

There was a gap in `C4DatabaseFlags`, so I put it there for now, although I'm suspicious that there is a reason there was a gap there.

### DIFF
--- a/C/include/c4DatabaseTypes.h
+++ b/C/include/c4DatabaseTypes.h
@@ -35,6 +35,7 @@ typedef C4_OPTIONS(uint32_t, C4DatabaseFlags){
         kC4DB_ReadOnly        = 0x02,    ///< Open file read-only
         kC4DB_AutoCompact     = 0x04,    ///< Enable auto-compaction [UNIMPLEMENTED]
         kC4DB_VersionVectors  = 0x08,    ///< Upgrade DB to version vectors instead of rev trees
+        kC4DB_MmapDisabled    = 0x10,    ///< Disable MMAP in SQLite.
         kC4DB_NoUpgrade       = 0x20,    ///< Disable upgrading an older-version database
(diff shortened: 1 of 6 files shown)


pasin (changes requested):
xcode build fails

C/include/c4DatabaseTypes.h
```diff
@@ -35,6 +35,7 @@ typedef C4_OPTIONS(uint32_t, C4DatabaseFlags){
         kC4DB_ReadOnly        = 0x02,    ///< Open file read-only
         kC4DB_AutoCompact     = 0x04,    ///< Enable auto-compaction [UNIMPLEMENTED]
         kC4DB_VersionVectors  = 0x08,    ///< Upgrade DB to version vectors instead of rev trees
+        kC4DB_MmapDisabled    = 0x10,    ///< Disable MMAP in SQLite.
```
pasin, line 38: @jianminzhao is this the right code? Should it be something after DiskSyncFull?

jianminzhao (approved):

pasin (approved):
I have fixed the PR validation failure.

C/include/c4DatabaseTypes.h
borrrden: This is fine to do, we've done it before.  The former keys "bundled" and "shared keys" used to be 0x8 and 0x10.  They were removed years ago and version vectors became the new 0x8.  Mmap can become the new 0x10.

callumbirks: Thanks for the clarification 😄

PR #2160, opened 2024-10-11
CBL-6131: Race creating the `expiration` column in a collection table

callumbirks:
Fixed by surrounding the column check with the same file lock used when adding the column.

One can verify that the test `Create collection concurrently` fails if you revert the lock change.

### DIFF
--- a/LiteCore/Storage/SQLiteDataFile.hh
+++ b/LiteCore/Storage/SQLiteDataFile.hh
@@ -174,12 +174,12 @@ namespace litecore {
 
             WithNewDocs = 400,  // New document/revision storage (CBL 3.0)
 
-            WithExpirationColumn = 490,  // Added 'expiration' column to KeyStore
             WithDeletedTable     = 500,  // Added 'deleted' KeyStore for deleted docs (CBL 3.0?)
             WithIndexesLastSeq   = 501,  // Added 'lastSeq' column to 'indexes' table (CBL 3.2)
+            WithExpirationColumn = 502,  // Added 'expiration' column to KeyStore
             MaxReadable          = 599,  // Cannot open versions newer than this
 
-            Current = WithDeletedTable
+            Current = WithExpirationColumn
         };
 
         void reopenSQLiteHandle();
--- a/LiteCore/Storage/SQLiteDataFile.hh
+++ b/LiteCore/Storage/SQLiteDataFile.hh
@@ -174,9 +174,10 @@ namespace litecore {
 
             WithNewDocs = 400,  // New document/revision storage (CBL 3.0)
 
-            WithDeletedTable   = 500,  // Added 'deleted' KeyStore for deleted docs (CBL 3.0?)
-            WithIndexesLastSeq = 501,  // Added 'lastSeq' column to 'indexes' table (CBL 3.2)
-            MaxReadable        = 599,  // Cannot open versions newer than this
+            WithExpirationColumn = 490,  // Added 'expiration' column to KeyStore
+            WithDeletedTable     = 500,  // Added 'deleted' KeyStore for deleted docs (CBL 3.0?)
+            WithIndexesLastSeq   = 501,  // Added 'lastSeq' column to 'indexes' table (CBL 3.2)
+            MaxReadable          = 599,  // Cannot open versions newer than this
 
             Current = WithDeletedTable
         };
--- a/LiteCore/Storage/SQLiteKeyStore.cc
+++ b/LiteCore/Storage/SQLiteKeyStore.cc
@@ -522,10 +522,12 @@ namespace litecore {
 
     // Adds the 'expiration' column to the table.
     void SQLiteKeyStore::addExpiration() {
-        if ( mayHaveExpiration() ) return;
-        db()._logVerbose("Adding the `expiration` column & index to kv_%s", name().c_str());
-        db().execWithLock(subst("ALTER TABLE kv_@ ADD COLUMN expiration INTEGER; "
-                                "CREATE INDEX \"kv_@_expiration\" ON kv_@ (expiration) WHERE expiration not null"));
+        db().withFileLock([=]() {
+            if ( mayHaveExpiration() ) return;
+            db()._logVerbose("Adding the `expiration` column & index to kv_%s", name().c_str());
+            db().execWithLock(subst("ALTER TABLE kv_@ ADD COLUMN expiration INTEGER; "
+                                    "CREATE INDEX \"kv_@_expiration\" ON kv_@ (expiration) WHERE expiration not null"));
+        });
         _hasExpirationColumn         = true;
         _uncommittedExpirationColumn = true;
     }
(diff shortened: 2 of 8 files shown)


LiteCore/Storage/SQLiteKeyStore.cc
```diff
@@ -522,10 +522,12 @@ namespace litecore {
 
     // Adds the 'expiration' column to the table.
     void SQLiteKeyStore::addExpiration() {
-        if ( mayHaveExpiration() ) return;
-        db()._logVerbose("Adding the `expiration` column & index to kv_%s", name().c_str());
-        db().execWithLock(subst("ALTER TABLE kv_@ ADD COLUMN expiration INTEGER; "
-                                "CREATE INDEX \"kv_@_expiration\" ON kv_@ (expiration) WHERE expiration not null"));
+        db().withFileLock([=]() {
```
jianminzhao, line 525: First of all, by moving the check (mayHaveExpiration) inside the transaction, it should fix the concurrent calling from two db connections. 
Another option is to leave the read-only transaction, mayHaveExpiration, alone, and include db().execWithLock call in a try state, check the error whether it's because duplicate. If so, it's all right. This will save us not having to open a transaction every time addExpiration is called.

LiteCore/Storage/SQLiteKeyStore.cc
```diff
@@ -522,10 +522,12 @@ namespace litecore {
...
+            if ( mayHaveExpiration() ) return;
+            db()._logVerbose("Adding the `expiration` column & index to kv_%s", name().c_str());
+            db().execWithLock(subst("ALTER TABLE kv_@ ADD COLUMN expiration INTEGER; "
```
jianminzhao, line 528: No need to open a transaction inside a transaction

LiteCore/tests/c4BaseTest.cc
```diff
@@ -139,6 +141,33 @@ TEST_CASE("C4Error Reporting Macros", "[Errors][C]") {
...
+        C4Error err{};
+        C4Error err2{};
+
+        snprintf(buf, 6, "coll%i", i);
+
+        {
+            slice                  collName{buf};
+            const C4CollectionSpec spec{collName, "scope"_sl};
+
+            auto a1 = std::async(std::launch::async, c4db_createCollection, db, spec, ERROR_INFO(&err));
```
jianminzhao, line 162: Access the same db concurrently is not legit anyway

jianminzhao (changes requested):

LiteCore/Storage/SQLiteKeyStore.cc
callumbirks: The particular error (duplicate column) does not have a particular code, the code is `SQLITE_ERROR`. So we would have to check the string, which is not good.

LiteCore/Storage/SQLiteKeyStore.cc
callumbirks: Thanks, fixed.

callumbirks:
@jianminzhao 
I think wrapping the check in a lock is preferable to checking the error, because:
a. The check is guarded by the state `_hasExpirationColumn`, so this lock should only be entered once by each collection instance.
b. Checking the error would require the error string being correct and never changing. Because the Error has code = SQLITE_ERROR, we have to rely on checking the message string to use that alternative.

LiteCore/tests/c4BaseTest.cc
snej: Jianmin is right — LiteCore's API is intentionally not thread-safe. You can’t use a C4Database concurrently on two threads.

I don’t think there’s a good test for this, honestly. You’d have to repeat the test many (hundreds?) times in a tight loop, creating a new database each time, and even then you might not hit the race condition.

snej:
The fix looks good, I just disagree with the test.

jianminzhao:
The new version is okay with me: doing an extra mayHaveExpiration inside the transaction, since the positive return is always good to omit the following transaction.

callumbirks:
> The fix looks good, I just disagree with the test.
@snej I guess you mean as in this is not behaviour we support? I agree, but as this is something that users can do (and now have done), I think its important to test that we keep the logical consistency.

Unless you have something to suggest to change / improve the test?

LiteCore/Storage/SQLiteDataFile.hh
```diff
@@ -174,9 +174,10 @@ namespace litecore {
 
             WithNewDocs = 400,  // New document/revision storage (CBL 3.0)
 
-            WithDeletedTable   = 500,  // Added 'deleted' KeyStore for deleted docs (CBL 3.0?)
-            WithIndexesLastSeq = 501,  // Added 'lastSeq' column to 'indexes' table (CBL 3.2)
-            MaxReadable        = 599,  // Cannot open versions newer than this
+            WithExpirationColumn = 490,  // Added 'expiration' column to KeyStore
```
jianminzhao, line 177: Can you insert a new version before existing versions? How would a version at 500 or 501 get upgraded? There should be a test, I think.

jianminzhao (changes requested):

LiteCore/Storage/SQLiteDataFile.hh
callumbirks: Any database at version >=500 already has the expiration column, because `addExpiration` was added to `KeyStore::reopen` in the same commit that `WithDeletedTable (500)` migration was added.

jianminzhao: So you actually try to retrofit into current upgrade path. This relies on very precise understanding of other part of code and the schema. 
Recognizing this fact, we can also treat it as a new version, but because of the above fact, do
ALTER TABLE ..... IF NOT EXISTS 
I am little concerned of retrofitting a new version into the history

callumbirks: > ALTER TABLE ..... IF NOT EXISTS
There is no such syntax. We need to check the existence of the column manually

jianminzhao: If you check the existence, do we still need to insert a new version into past history? This is the new version, after this new version, we will know schema has the expiration table at the time of creation.

jianminzhao: What's the code path that supports your statement? Is it in the same transaction of the upgrade transaction? Is it possible that some failure occurs outside the upgrade transaction that arrives following state:
- version bumped 500
- some keystore not opened successfully and expiration not added successfully.
Since you removed the addExpiration completely, there is no way to have the column.

callumbirks: SQLiteKeyStore::reopen calls `addExpiration` (on the main branch).
So if a database is version 500, every collection which has been used since it upgraded to version 500 has the expiration column.

jianminzhao: "which has been used since ..." -- what if not used?

snej: I don't think it's worth trying to optimize. Seems like you need to make WithExpiration = 502 and always do it (once) for every database, just to make sure.

jianminzhao (approved):

PR #2164, opened 2024-10-15
CBL-6155: Improve LiteCore Log

velicuvlad:
- repl config is now logged only once, at the time replicator is started
- error response from blob update specified is a blob req error
- certificates errors are all on the same line now - see https://github.com/couchbasedeps/mbedtls/pull/3 (once is it merged, I will update this PR)
- add URL endpoint to Repl log

### DIFF
--- a/Replicator/Replicator.cc
+++ b/Replicator/Replicator.cc
@@ -104,7 +104,6 @@ namespace litecore::repl {
                 DatabaseImpl* impl = asInternal(db);
                 return impl->dataFile()->loggingName();
             });
-            logInfo("DB=%s Instantiated %s", dbLogName.c_str(), string(*options).c_str());
 
             _remoteURL = webSocket->url();
             if ( _options->isActive() ) { prepareWorkers(); }
--- a/Replicator/c4ReplicatorImpl.hh
+++ b/Replicator/c4ReplicatorImpl.hh
@@ -290,8 +290,8 @@ namespace litecore {
             }
 
             setStatusFlag(kC4Suspended, false);
-            logInfo("Starting Replicator %s with config: {%s}\n", _replicator->loggingName().c_str(),
-                    std::string(*_options).c_str());
+            logInfo("Starting Replicator %s with config: {%s} and endpoint: %.*s", _replicator->loggingName().c_str(),
+                    std::string(*_options).c_str(), SPLAT(_replicator->remoteURL()));
             _selfRetain = this;  // keep myself alive till Replicator stops
             updateStatusFromReplicator(_replicator->status());
             _responseHeaders = nullptr;
--- a/Networking/BLIP/BLIPConnection.cc
+++ b/Networking/BLIP/BLIPConnection.cc
@@ -162,11 +162,15 @@ namespace litecore::blip {
 
       protected:
         ~BLIPIO() override {
+            double outboxDepth =
+                    (_countOutboxDepth != 0)
+                            ? (static_cast<double>(_totalOutboxDepth) / static_cast<double>(_countOutboxDepth))
+                            : 0.0;
             LogTo(SyncLog,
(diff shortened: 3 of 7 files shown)


callumbirks:
Ignoring Blackduck failure. It complains because we have changed mbedtls without bumping the version

Replicator/Replicator.cc
```diff
@@ -104,7 +104,6 @@ namespace litecore::repl {
                 DatabaseImpl* impl = asInternal(db);
                 return impl->dataFile()->loggingName();
             });
-            logInfo("DB=%s Instantiated %s", dbLogName.c_str(), string(*options).c_str());
```
pasin, line 107: Is this info not important or is it logged the same info somewhere else?

Replicator/c4ReplicatorImpl.hh
```diff
@@ -296,8 +296,8 @@ namespace litecore {
             }
 
             setStatusFlag(kC4Suspended, false);
-            logInfo("Starting Replicator %s with config: {%s}\n", _replicator->loggingName().c_str(),
-                    std::string(*_options).c_str());
+            logInfo("Starting Replicator %s with config: {%s} with endpoing: %.*s", _replicator->loggingName().c_str(),
+                    std::string(*_options).c_str(), SPLAT(_replicator->remoteURL()));
```
pasin, line 300: What will happen if the replicator is for db-to-db or p2p sync?

callumbirks: Whenever Replicator is instantiated it sets `_remoteURL = webSocket->url()`. So I believe this would be something like `ws://loopback`. I will run test and verify.

callumbirks: DB to DB endpoint looks like `file:////var/folders/dq/4702h1qn7wdclczftpkmv1dr0000gp/T/LiteCore_Tests_1729177323.cblite2/cbl_core_test2.cblite2.cblite2/ `

callumbirks: Oh yeah this is ReplicatorImpl so for P2P the implementation is by platform - so this exact log isn't in the P2P replicators.

Replicator/c4ReplicatorImpl.hh
```diff
@@ -296,8 +296,8 @@ namespace litecore {
             }
 
             setStatusFlag(kC4Suspended, false);
-            logInfo("Starting Replicator %s with config: {%s}\n", _replicator->loggingName().c_str(),
-                    std::string(*_options).c_str());
+            logInfo("Starting Replicator %s with config: {%s} with endpoing: %.*s", _replicator->loggingName().c_str(),
```
callumbirks, line 299: I think it should say "endpoint" instead of "endpoing"

Replicator/c4ReplicatorImpl.hh
callumbirks: Not sure what we should do about this - @pasin any ideas? Should LiteCore be logging the Replicator options from P2P replicator like it does for remote replicator?

pasin: Not sure I understand this. The c4ReplicatorImpl's start() should be used by any types of the replicator. I think we can focus on what is being logged here with the remoteURL() with the replicator with message endpoint is used. Maybe the easiest way to find out is to look at the log when running the message endpoint replicator tests from the platform (e.g. iOS) that uses LiteCore in this PR?

velicuvlad: I've also seen this for the passive replicator 
```endpoint: x-msg-conn:///0x104397e20```

pasin: This is good enough.

callumbirks (approved):

PR #2171, opened 2024-10-23
CBL-6369: Recreate Array Index fails

jianminzhao:
Delete the current index with the same name if it does not have the same IndexSpec.


callumbirks (approved):
👏

PR #2179, opened 2024-10-30
CBL-6337: LiteCore Replicator uses the previous logPrefix passed by t…

jianminzhao:
…he platform replicator

It was a mistake to have loggingClassName class static because the logPrefix passed in by the platform is static per object. We now make it object static, namely fixing it in the constructor.


jianminzhao:
Auto merged from 3.2

pasin (approved):

PR #2181, opened 2024-11-05
CBL-6416: Regression - VS LIMIT query should throw for out of range

jianminzhao:


snej:
Sorry to derail this, but I don't think the 3.2 behavior is the right thing to do. There's no other case where a non-positive LIMIT causes an error; a zero or negative value just means zero results. And why arbitrarily prevent limits above 10000?

I'm pretty sure this is not consistent with vector query on Server, and we're supposed to be compatible with that.

pasin (changes requested):
For limit 0 and negative, the server returns zero results. We can follow that behavior. I also don't think this is a breaking change as nobody can use limit 0 or negative before in CBL. I will create tickets for the platforms to remove the tests that check error for limit 0 and negative. 

For limit 10000 to the vector search, it is a guardrail set by PM so that we have put this in 3.2.0 (It's in the API spec). Can we continue to have this guardrails for 4.0 beta, and we can discuss with PM to remove or change that number after the beta?

pasin:
I have created https://jira.issues.couchbase.com/browse/CBL-6426 for the limit 10000.

pasin (approved):

PR #2182, opened 2024-11-06
CBL-6332: Implicit key change in query results

jianminzhao:
In V3.2, when WHAT clause of the JSON query is absent of empty, we will assign meta properties, _id and _sequence, to the what clause, and use then property names verbatim in the corresponding column names (keep the underscore in the names). In this commit, we mean to keep this behavior with new query translator. The column names are used in the query result of the platform API as the dictionary keys.


pasin:
I have created https://jira.issues.couchbase.com/browse/CBL-6428 to work on disallowing the empty select after 4.0 private beta.

pasin (approved):

PR #2185, opened 2024-11-11
CBL-6422: Some tests of VectorSearch are failing in 4.0

jianminzhao:
- "IndexUpdater Set Float Array": ORDER_BY is more strict than 3.2. It now requires an explicit array.
- "IndexUpdater not update when released without finish": ORDER_BY is more strict than 3.2. It now requires an explicit array.
- "APPROX_VECTOR_DISTANCE Errors (Misc)": error messages have changed.
- "APPROX_VECTOR_DISTANCE Errors (Bad Metric)": error messages have changed.


pasin:
I don't see anything related to "IndexUpdater not update when released without finish" as described in the PR description.

pasin (approved):

PR #2188, opened 2024-11-20
CBL-6302: Fix N1QL parser rejects `META(scope.coll)`

callumbirks:
Test "Various Exceptional Conditions" failed before this fix, when using N_WAY_... due to `scope.coll` not being accepted by N1QL parser for META function.
Now it passes

### DIFF
--- a/LiteCore/Query/N1QL_Parser/n1ql.leg
+++ b/LiteCore/Query/N1QL_Parser/n1ql.leg
@@ -356,7 +356,7 @@ indexTable =                            { a = string(""); }
 
 function =
     "meta"i _ '(' _                    { f = op("meta()");}
-    (c:IDENTIFIER _                    { appendAny(f, c.as<string>());}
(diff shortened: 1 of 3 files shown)


LiteCore/Query/N1QL_Parser/n1ql.leg
```diff
@@ -345,6 +345,11 @@ propertyPath =
 propertyName
     = IDENTIFIER
 
+metaPath =
```
jianminzhao, line 348: Can propertyPath be used here?

callumbirks: No. It doesn't work

callumbirks: propertyPath does some weird quoting and stuff that doesn't work with the way META expression path is parsed. It breaks most META expression tests

jianminzhao (approved):

PR #2193, opened 2024-12-12
CBL-6285: Explain for predictive query no longer names an index

jianminzhao:
The main issue is that we need to know the kv table on which the PredictiveNode is created, and the pass that SourceNodes are assigned table names is run after the pass all nodes are created.

Several technical changes.
- SourceNode::collection() is empty string only if it's not given in the query expression. (Currently it will be turned to empty string if explicitly given "_default."
- We assign the table name to the main data source in the pass of node creation, so that it can be used when PredictiveNode is populated. It can be later altered if flag _usesDeleted is set.
- The result of keystore.compileQuery() should be the same, except that empty data source in the FROM clause will be substituted for with collection corresponding to the invoking key store.

### DIFF
--- a/LiteCore/Query/Translator/IndexedNodes.cc
+++ b/LiteCore/Query/Translator/IndexedNodes.cc
@@ -225,7 +225,7 @@ namespace litecore::qt {
         expr.append(args[1]);
         string id = expressionIdentifier(expr);
 
-        if ( ctx.delegate.hasPredictiveIndex(id) ) {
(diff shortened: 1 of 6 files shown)


snej (approved):
I don't understand all the changes, but I think this is okay.

It sounds as though we should add a LiteCore test for PQ that will do what the failed platform test does, namely run "explain" and verify that the query is really using the index.

PR #2199, opened 2024-12-21
CBL-6592: active pusher should always send proposeChanges regardless …

jianminzhao:
…of whether being version-vector or not.

- There is one test failing, "Push After VV Upgrade", c.f. CBL-6593. It is disabled.

### DIFF
--- a/Replicator/ChangesFeed.cc
+++ b/Replicator/ChangesFeed.cc
@@ -102,6 +102,7 @@ namespace litecore::repl {
         try {
             _db.useLocked([&](C4Database* db) {
                 Assert(db == _checkpointer->collection()->getDatabase());
+                options.flags |= kC4RevIDGlobalForm;
                 C4DocEnumerator e(_checkpointer->collection(), _maxSequence, options);
                 changes.revs.reserve(limit);
                 while ( e.next() && limit > 0 ) {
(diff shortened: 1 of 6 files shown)


Replicator/ChangesFeed.cc
```diff
@@ -102,6 +102,7 @@ namespace litecore::repl {
         try {
             _db.useLocked([&](C4Database* db) {
                 Assert(db == _checkpointer->collection()->getDatabase());
+                options.flags |= kC4RevIDGlobalForm;
```
snej, line 105: Instead of changing C4DocEnumerator, you can just call `DBAccess::convertVersionToAbsolute` on the revID.

Replicator/RevFinder.cc
```diff
@@ -93,7 +93,7 @@ namespace litecore::repl {
             if ( !changes && req->body() != "null"_sl ) {
                 warn("Invalid body of 'changes' message");
                 req->respondWithError({"BLIP"_sl, 400, "Invalid JSON body"_sl});
-            } else if ( (!proposed && _mustBeProposed) || (proposed && _db->usingVersionVectors()) ) {
+            } else if ( !proposed && _mustBeProposed ) {
```
snej, line 96: Delete "But with version vectors, always use "changes"." on line 98

snej (changes requested):

snej (approved):

PR #2201, opened 2025-01-08
CBL-6599: SQL++ query throws error unexpectedly

callumbirks:
The columns in the customers query results have type SQLITE_BLOB but are null. Apparently we didn't check for null before when encoding blob column, now we do.
While it wasn't hit in this issue, also added a null check to the very similar SQLITE_TEXT encoding. That seems like the correct course of action?

### DIFF
--- a/LiteCore/Query/SQLiteQuery.cc
+++ b/LiteCore/Query/SQLiteQuery.cc
@@ -446,6 +446,10 @@ namespace litecore {
                     {
                         if ( i >= _query->_1stCustomResultColumn ) {
                             slice        fleeceData{col.getBlob(), (size_t)col.getBytes()};
+                            if (fleeceData == nullslice) {
+                                enc.writeNull();
+                                return false;
+                            }
                             Scope        fleeceScope(fleeceData, _sk);
                             const Value* value = Value::fromTrustedData(fleeceData);
                             if ( !value )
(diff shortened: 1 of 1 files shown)


jianminzhao:
CBL-6599 is for 4.0. Please use CBL-6625

callumbirks:
@jianminzhao  changed it

LiteCore/Query/SQLiteQuery.cc
```diff
@@ -445,20 +445,24 @@ namespace litecore {
...
-                            slice        fleeceData{col.getBlob(), (size_t)col.getBytes()};
-                            Scope        fleeceScope(fleeceData, _sk);
-                            const Value* value = Value::fromTrustedData(fleeceData);
-                            if ( !value )
-                                error::_throw(error::CorruptRevisionData,
-                                              "SQLiteQueryRunner encodeColumn parsing fleece to Value failing");
-                            enc.writeValue(value);
+                            slice fleeceData{col.getBlob(), (size_t)col.getBytes()};
+                            if ( fleeceData.empty() ) {
+                                enc.writeNull();
```
jianminzhao, line 450: why not return false as we do for "case SQLITEE_NULL" ? 
CC: @snej

jianminzhao (approved):

LiteCore/Query/SQLiteQuery.cc
jianminzhao: Returning false will make it appear as MISSING, and now it appears as NULL.

snej: A SQLite NULL means MISSING. An empty blob (returned from a query) means a JSON/Fleece null.

jianminzhao (approved):

PR #2204, opened 2025-01-10
CBL-6629: Invalid binary version ID (looks like a legacy revID) error…

jianminzhao:
… when pushing rev-tree doc before upgrading to version vector

Have to deal with the situation where VectorDocument's selected revision is in the legacy RevTree form.

### DIFF
--- a/LiteCore/Database/VectorDocument.cc
+++ b/LiteCore/Database/VectorDocument.cc
@@ -190,8 +190,11 @@ namespace litecore {
         }
 
         alloc_slice getSelectedRevIDGlobalForm() const override {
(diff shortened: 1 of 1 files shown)


LiteCore/Database/VectorDocument.cc
```diff
@@ -190,8 +190,11 @@ namespace litecore {
         }
 
         alloc_slice getSelectedRevIDGlobalForm() const override {
-            if ( auto rev = _selectedRevision(); rev ) return rev->versionVector().asASCII(mySourceID());
-            else
+            if ( auto rev = _selectedRevision(); rev ) {
```
snej, line 193: FYI, you don't need the `; rev` part :)

snej (approved):

PR #2209, opened 2025-01-13
CBL-6626: SQL++ query not return expected results

callumbirks:
The issue which caused the customer's query to not return the correct results is that litecore thinks the null value is not null. 
The underlying cause of that is an issue with indexes which causes some values in certain indexes to not be correctly typed as null when passing into the query engine.
So this does not fix the underlying problem, but it is a good workaround that @snej agreed with.
Basically the problematic values do not get given the Null subtype when they should, so our usual check for the null subtype fails. Now, we also check if the size is 0 if the subtype has not been set.

### DIFF
--- a/LiteCore/Query/SQLiteN1QLFunctions.cc
+++ b/LiteCore/Query/SQLiteN1QLFunctions.cc
@@ -280,7 +280,9 @@ namespace litecore {
 
     // Test for N1QL NULL value (which is an empty blob tagged with kFleeceNullSubtype)
     static inline bool isNull(sqlite3_value* arg) {
-        return sqlite3_value_type(arg) == SQLITE_BLOB && sqlite3_value_subtype(arg) == kFleeceNullSubtype;
+        if (sqlite3_value_type(arg) != SQLITE_BLOB) return false;
+        auto subtype = sqlite3_value_subtype(arg);
+        return subtype == kFleeceNullSubtype || (subtype == 0 && sqlite3_value_bytes(arg) == 0);
     }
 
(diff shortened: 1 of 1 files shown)


jianminzhao:
@snej I see Blob&&subtype==0 is being used in fl_result. I would defer to you whether to accept it as a patch to isNull() function.

jianminzhao:
Please also fix fl_result where we only check blob's subtype but not the size. Also for isNull(), size of the blob (being 0) is sufficient. Subtype is secondary. (non-zero size with subtype fleecenull reveals a bug)

LiteCore/Query/SQLiteFleeceFunctions.cc
```diff
@@ -235,8 +235,17 @@ namespace litecore {
...
+                                    // If the size is 0, it may actually be a Fleece Null with missing subtype.
+                                    if ( sqlite3_value_bytes(arg) == 0 ) {
+                                        slice encoded = Encoder::kPreEncodedNull;
+                                        sqlite3_result_blob(ctx, encoded.buf, int(encoded.size), SQLITE_STATIC);
+                                        return;
+                                    } else {
+                                        // Untagged blob with size is already Fleece data
+                                        break;
+                                    }
+                                }
```
jianminzhao, line 248: Is it equivalent to:
case 0:
case kFleeceNullSubtype:
    if (size > 0) break; // untagged blob ...
   else {
     // Fleece/JSON null
    return;
  }

callumbirks: I believe the latest commit is more appropriate

jianminzhao (approved):

PR #2211, opened 2025-01-15
CBL-6636: Rev message contains the full history in the revID field

jianminzhao:
For the rev property of the rev message, We use the revID of the current version for Version Vector revisions and the rest of of the versions will be in the history property.

### DIFF
--- a/Replicator/Pusher+Revs.cc
+++ b/Replicator/Pusher+Revs.cc
@@ -17,6 +17,7 @@
 #include "HTTPTypes.hh"
 #include "Increment.hh"
 #include "StringUtil.hh"
+#include "VersionVector.hh"
 #include "fleece/Mutable.hh"
 #include <cinttypes>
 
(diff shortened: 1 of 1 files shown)


Replicator/Pusher+Revs.cc
```diff
@@ -106,21 +107,43 @@ namespace litecore::repl {
...
+        alloc_slice     currentRevID = fullRevID;
+        constexpr slice vvSeparator  = ",;"_sl;
+        if ( currentRevID.findAnyByteOf(vvSeparator) ) {
+            try {
+                VersionVector vv = VersionVector::fromASCII(currentRevID);
+                currentRevID     = vv.current().asASCII();
+            } catch ( ... ) {
+                // it's not a VV even with comma or semi-colon in it.
+                throw;
+            }
```
snej, line 120: This try-catch is pointless; it doesn't do anything.

Replicator/Pusher+Revs.cc
```diff
@@ -106,21 +107,43 @@ namespace litecore::repl {
...
+        alloc_slice currentReplacementRevID = fullReplacementRevID;
+        if ( currentReplacementRevID && currentReplacementRevID.findAnyByteOf(vvSeparator) ) {
+            try {
+                VersionVector vv        = VersionVector::fromASCII(currentReplacementRevID);
+                currentReplacementRevID = vv.current().asASCII();
+            } catch ( ... ) {
+                // it's not a VV even with comma or semi-colon in it.
+                throw;
+            }
+        }
```
snej, line 134: This is the same code as above. Factor it out into a function you can reuse.

snej (changes requested):

Replicator/Pusher+Revs.cc
jianminzhao: Right, I was thinking to add some logs there. Perhaps not that useful; hopefully we will get stacktrace.

snej (approved):

PR #2220, opened 2025-01-30
CBL-6671: Options of FTS partial index is not fully encoded in the SQ…

jianminzhao:
…Lite table when it is registered

1. Bumped Schema version.
2. In the new version, added a column, whereClause, to table indexes.
3. Upgrade the tables in previous version to include the new column
4. IndexSpec::whereClause is registered in table "indexes" by the new column.
5. c4index_getOptions will have "where" passed in with C4IndexOptions.
6. Tests of the above.

### DIFF
--- a/C/tests/c4QueryTest.cc
+++ b/C/tests/c4QueryTest.cc
@@ -286,8 +286,9 @@ N_WAY_TEST_CASE_METHOD(C4QueryTest, "C4Query expression index", "[Query][C]") {
 }
 
 N_WAY_TEST_CASE_METHOD(C4QueryTest, "C4Query partial value index", "[Query][C]") {
-    C4Error err;
-    auto    defaultColl = getCollection(db, kC4DefaultCollectionSpec);
+    C4Error              err;
+    auto                 defaultColl = getCollection(db, kC4DefaultCollectionSpec);
+    const vector<string> result{"0000099"};
     for ( int withIndex = 0; withIndex < 2; ++withIndex ) {
         if ( withIndex ) {
             C4IndexOptions options{};
@@ -295,23 +296,19 @@ N_WAY_TEST_CASE_METHOD(C4QueryTest, "C4Query partial value index", "[Query][C]")
             REQUIRE(c4coll_createIndex(defaultColl, C4STR("length"), c4str("length(name.first)"), kC4N1QLQuery,
                                        kC4ValueIndex, &options, WITH_ERROR(&err)));
         }
-        c4query_release(query);
-        const char* queryStr = "SELECT META().id FROM _ WHERE length(name.first) = 9 AND gender = 'female'";
-        query                = c4query_new2(db, kC4N1QLQuery, c4str(queryStr), nullptr, ERROR_INFO(err));
+        compileSelect("SELECT META().id FROM _ WHERE length(name.first) = 9 AND gender = 'female'", kC4N1QLQuery);
         REQUIRE(query);
         checkExplanation(withIndex);
-        CHECK(run() == (vector<string>{"0000099"}));
+        CHECK(run() == result);
 
         if ( withIndex ) {
-            c4query_release(query);
             // Logically equivalent query, changing gender = 'female' to gender != 'male'.
             // Because the condition is not exact as the condition in partial index, it would not use
             // the index.
-            const char* queryStr2 = "SELECT META().id FROM _ WHERE length(name.first) = 9 AND gender != 'male'";
-            query                 = c4query_new2(db, kC4N1QLQuery, c4str(queryStr2), nullptr, ERROR_INFO(err));
+            compileSelect("SELECT META().id FROM _ WHERE length(name.first) = 9 AND gender != 'male'", kC4N1QLQuery);
             REQUIRE(query);
             checkExplanation(!withIndex);
-            CHECK(run() == (vector<string>{"0000099"}));
+            CHECK(run() == result);
         }
     }
 }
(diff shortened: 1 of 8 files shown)


C/tests/c4QueryTest.cc
```diff
@@ -464,6 +470,41 @@ N_WAY_TEST_CASE_METHOD(C4QueryTest, "C4Query FTS", "[Query][C][FTS]") {
     c4slice_free(matched);
 }
 
+N_WAY_TEST_CASE_METHOD(C4QueryTest, "C4Query FTS (Partial)", "[Query][C][FTS]") {
```
pasin, line 473: I have seen the other test above was changed to test both JSON and SQL++. Would it be worth to do that here too?

LiteCore/Query/SQLiteDataFile+Indexes.cc
```diff
@@ -99,6 +105,20 @@ namespace litecore {
...
+                                else if ( whatA )
+                                    same = whatA->toJSONString() == whatB->toJSONString();
+                            }
+                            if ( same ) {
+                                auto whereA = spec.where(), whereB = existingSpec->where();
+                                if ( !whereA ? whereB != nullptr : !whereB ) same = false;
+                                else if ( whereA )
+                                    same = whereA->toJSONString() == whereB->toJSONString();
+                            }
+                        }
```
pasin, line 121: Maybe a bit clearer if move this to kFullText case itself.

LiteCore/Query/SQLiteDataFile+Indexes.cc
```diff
@@ -99,6 +105,20 @@ namespace litecore {
                     case IndexSpec::kFullText:
                     case IndexSpec::kVector:
                         same = schemaExistsWithSQL(indexTableName, "table", indexTableName, indexSQL);
+                        if ( spec.type == IndexSpec::kFullText ) {
+                            if ( same ) {
+                                auto whatA = spec.what(), whatB = existingSpec->what();
+                                if ( !whatA ? whatB != nullptr : !whatB ) same = false;
```
pasin, line 111: Could what be practically NULL? I understand that it's a required field.

pasin (commented):
I have a couple minor comments. Apart from that, look good to me.

LiteCore/Query/SQLiteDataFile+Indexes.cc
snej: You can compare Values more simply than this:
`same = v1 ? v1->isEqual(v2) : !v2;`

LiteCore/Query/SQLiteDataFile+Indexes.cc
jianminzhao: I think the purpose here is to check the sameness. NULL should be avoided somewhere else.
I tried it, empty expression will cause crash before coming to here.
Do platforms check it?

C/tests/c4QueryTest.cc
jianminzhao: I can do it

LiteCore/Query/SQLiteDataFile+Indexes.cc
jianminzhao: This is intentional. I think the logic applies both to Vector and FTS, only at this time, I don't want to touch Vector

snej (changes requested):

LiteCore/Query/IndexSpec.hh
```diff
@@ -112,15 +112,19 @@ namespace litecore {
         /** The optional WHERE clause: the condition for a partial index */
         const fleece::impl::Array* where() const;
 
+        void setWhereClause(string_view whereClause_) const {
+            if ( !whereClause_.empty() ) whereClause = alloc_slice::nullPaddedString(whereClause_);
```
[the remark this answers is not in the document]
jianminzhao: I see. I intend it to be constant, not to be changed after first assigned.. How about throw if current value is not null? Maybe throw is too rough, but most likely the second assignment is not right. Maybe I can add a warning.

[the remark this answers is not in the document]
jianminzhao: @snej I like it to be almost const, not to be changed after the first assignment. What do you prefer in following?
A. throw if assigned value second time
B. log an warning
C. don't try to limit it

LiteCore/Query/SQLiteDataFile+Indexes.cc
pasin: If LiteCore returns the error for the empty expression, the platform will return or throw an exception to the user for that error.

pasin (approved):

LiteCore/Query/IndexSpec.hh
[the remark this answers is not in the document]
snej: It shouldn't be a `const` method: it changes the visible state of the object.

LiteCore/Query/IndexSpec.hh
```diff
@@ -112,15 +112,21 @@ namespace litecore {
...
-        std::string const name;           ///< Name of index
-        Type const        type;           ///< Type of index
-        alloc_slice const expression;     ///< The query expression
-        alloc_slice const whereClause;    ///< The where clause. If given, expression should be the what clause
-        QueryLanguage     queryLanguage;  ///< Is expression JSON or N1QL?
-        Options const     options;        ///< Options for FTS and vector indexes
+        std::string const   name;           ///< Name of index
+        Type const          type;           ///< Type of index
+        alloc_slice const   expression;     ///< The query expression
+        mutable alloc_slice whereClause;    ///< The where clause. If given, expression should be the what clause
```
snej, line 127: Should not be mutable

snej (changes requested):

snej (approved):

PR #2226, opened 2025-02-10
Actually disable Housekeeper when kC4DB_NoHousekeeping is set

snej:
Bug-fix for 3.1.9-vf1. I accidentally left out the most important line of the fix while cleaning up the code for the PR.

This commit makes it so a collection does not create a Housekeeper if the database's kC4DB_NoHousekeeping flag is set.

### DIFF
--- a/C/c4CAPI.cc
+++ b/C/c4CAPI.cc
@@ -484,10 +484,13 @@ C4Database* c4db_openNamed(C4String name,
 }
 
 
-C4Database* c4db_openAgain(C4Database* db,
+C4Database* c4db_openAgain(C4Database* database,
                            C4Error *outError) noexcept
 {
-    return c4db_openNamed(c4db_getName(db), c4db_getConfig2(db), outError);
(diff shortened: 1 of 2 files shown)


pasin (approved):

C/c4CAPI.cc
```diff
@@ -484,10 +484,13 @@ C4Database* c4db_openNamed(C4String name,
 }
 
 
-C4Database* c4db_openAgain(C4Database* db,
+C4Database* c4db_openAgain(C4Database* database,
                            C4Error *outError) noexcept
 {
-    return c4db_openNamed(c4db_getName(db), c4db_getConfig2(db), outError);
+    if ( database == nullptr ) return nullptr;
```
pasin, line 490: `database` is not nullable.  I checked the other places in the file and when the database is not nullable, I didn't see why have try to specifically handle that case.

jianminzhao: I guess I don't have to, but not-nullable is not runtime checked in the release, I think.

PR #2227, opened 2025-02-10
Actually disable Housekeeper when kC4DB_NoHousekeeping is set

snej:
Bug-fix for the previous fix (#2223). I accidentally left out the most important line of the fix while cleaning up the code for the PR.

This commit makes it so a collection does not create a Housekeeper if the database's kC4DB_NoHousekeeping flag is set.

(cherry-picked from 3.1.9-vf2)


jianminzhao:
Would also need the latest commit on 3.1.9-vf2

pasin (approved):

PR #2228, opened 2025-02-10
Disable "housekeeping" tasks for db opened with openAgain()

snej:
Adds a flag `C4DB_NoHousekeeping` that suppresses database "housekeeping" tasks, and makes C4Database::openAgain() set this flag in the database it opens.

Housekeeping tasks like expiring documents and SQLite vacuuming only need to be done by one database connection (C4Database), but we do them in all connections. This means CBL wastes some time and also opens twice as many SQLite connections as it needs to (that _backgroundDB that DatabaseImpl opens.)

It also enables a nasty deadlock condition when the replicator closes its database(s) -- the close call happens on an actor thread, and ends up calling Housekeeper::stop, which sends an Actor message to the Housekeeper and blocks until it's processed. But if all actor threads are doing this, there are no threads left to actually run the Housekeeper. [CBSE-19205]

_(Cherry-picked from release/3.2. Includes the bonus "Actually disable Housekeeper" commit, without which this doesn't actually fix the bug.)_


pasin (approved):

jianminzhao:
Perhaps also need to cherry pick 3da7dd250ec3707b from 3.1.9-vf2

snej:
> Perhaps also need to cherry pick [3da7dd2](https://github.com/couchbase/couchbase-lite-core/commit/3da7dd250ec3707bb16cc246a617799e71573963) from 3.1.9-vf2

Yes, that is the fix I mentioned above.

PR #2232, opened 2025-02-12
Moved tool_support dir here from the mobile-tools repo

snej:
In order to keep submodule dependencies sane, we're moving the code shared between couchbase-mobile-tools and Edge Server into LiteCore. 

That code is the base classes for CLI tools (`Tool` and `LiteCoreTool`) and some code it uses for reading command lines and parsing them into arguments.

These come from the `edge_server` branch of the tools repo (which is going away.) The Tool and LiteCoreTool classes have a bunch of changes from the master branch there -- accumulated over time on the `dev` branch and then later the `edge_server` branch.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -25,3 +25,6 @@
 [submodule "tools"]
 	path = tools
 	url = https://github.com/couchbaselabs/litecore-download-script
+[submodule "vendor/linenoise-mob"]
+	path = vendor/linenoise-mob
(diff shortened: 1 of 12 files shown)


borrrden (approved):
There is an important thing that we need to now remember.  In a black duck manifest, any tools created using this will need to have linenoise-mob in their manifest (once linenoise-mob gets added to the BD database)

PR #2239, opened 2025-02-17
Adjust build scripts for Windows and Android

borrrden:
To not use the now nonexistent root level CMakeLists.txt, but instead use the one from this repo.


borrrden:
Tested via toy builds

jianminzhao (approved):

PR #2240, opened 2025-02-19
Logging: Use a separate mutex to protect sObjectMap [CBL-6787]

snej:
To avoid a deadlock on Android, we prevent the `Logging` class's destructor from requiring the `sLogMutex`. Instead, the functions that use `sObjectMap` will use their own mutex.

This is straightforward except for the way Logging passes a reference to `sObjectMap` to the LogEncoder, totally breaking encapsulation. Sigh. I fixed this the way I did on the log-observer branch, by having the LogEncoder take the object path directly and giving the logger a way to tell whether it needs to compute the path.

### DIFF
--- a/LiteCore/Support/LogEncoder.cc
+++ b/LiteCore/Support/LogEncoder.cc
@@ -65,20 +65,35 @@ namespace litecore {
         return _out.tellp();
     }
 
+    bool LogEncoder::isNewObject(ObjectRef object) const {
+        lock_guard<mutex> lock(_mutex);
+        return _isNewObject(object);
+    }
+
+    bool LogEncoder::_isNewObject(ObjectRef object) const {
+        return object != ObjectRef::None && _seenObjects.find(object) == _seenObjects.end();
+    }
+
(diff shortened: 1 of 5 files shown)


jianminzhao:
(1) In LogDomain::vlog, although not explicitly, there is a lock on sLoggingObjectsMutex
addObjectPath() -> getObjectPath() -> unique_lock lock(sLoggingObjectsMutex)

(2) whereas in Logging::~Logging(), there is a path
 LogDomain::unregisterObject() -> unique_lock lock(sLoggingObjectsMutex);

snej:
> (1) In LogDomain::vlog, although not explicitly, there is a lock on sLoggingObjectsMutex addObjectPath() -> getObjectPath() -> unique_lock lock(sLoggingObjectsMutex)
> 
> (2) whereas in Logging::~Logging(), there is a path LogDomain::unregisterObject() -> unique_lock lock(sLoggingObjectsMutex);

That is true but irrelevant. It doesn't lead to a deadlock.

borrrden (approved):

PR #2244, opened 2025-03-11
Housekeeping fix

jianminzhao:


jianminzhao:
The failure of Blackduck check is due to a change in stopwordset.cc to fix missing header.

borrrden (approved):

PR #2246, opened 2025-03-25
CBL-6878: Compile 16k Android Page Size

borrrden:


bmeike (approved):
This is very cool.  I'm going to do exactly the same thing.

bmeike (commented):
Why do you include "-Wl" twice and "CACHE INTERNAL" twice?

cmake/platform_android.cmake
```diff
@@ -10,6 +10,9 @@ function(setup_globals)
     # See: https://github.com/android-ndk/ndk/issues/289
     # Work around an NDK issue that links things like exception handlers in the incorrect order
     set(LITECORE_SHARED_LINKER_FLAGS "-Wl,--exclude-libs,libgcc.a" CACHE INTERNAL "")
+
+    # Enable 16k page size
+    set(LITECORE_SHARED_LINKER_FLAGS "${LITECORE_SHARED_LINKER_FLAGS} -Wl,-z,max-page-size=16384" CACHE INTERNAL "")
```
bmeike, line 15: Is that comma after the -z correct?  From the ld docs here: https://sourceware.org/binutils/docs/ld/Options.html, it looks to me like it should be "-z max-page-size=16384". Maybe this is something about how CMAKE works?

bmeike (approved):

cmake/platform_android.cmake
bmeike: It appears to be correct.   It works when I use it.  Must be CMAKEish.

borrrden: Not related to CMake, it's just the way to specify linker arguments when the same program (clang) is used for both compilation and linking (clang will invoke ld / lld, etc on our behalf).

jianminzhao (approved):

PR #2248, opened 2025-03-26
CBL-6810: Array indexes with path containing uppercase letter are always recreated

callumbirks:
The issue is the existing spec unnestPath is escaped (e.g. `foo\\Bar`) whereas the new spec is not escape (e.g. `fooBar`), so we check against an escaped version of the new spec unnestPath.

### DIFF
--- a/LiteCore/Query/SQLiteKeyStore+ArrayIndexes.cc
+++ b/LiteCore/Query/SQLiteKeyStore+ArrayIndexes.cc
@@ -30,12 +30,15 @@ namespace litecore {
 
         auto currSpec = db().getIndex(spec.name);
         if ( currSpec ) {
+            // Escape capitals in the unnest path of the new spec, to match the existing spec.
+            alloc_slice specUnnestPath{
+                    SQLiteKeyStore::transformCollectionName(spec.arrayOptions()->unnestPath.asString(), true)};
             // If there is already index with the index name,
             // eiher delete the current one, or use it (return false)
             while ( true ) {
                 if ( currSpec->type != IndexSpec::kArray ) break;
                 if ( !currSpec->arrayOptions() ) break;
-                if ( currSpec->arrayOptions()->unnestPath != spec.arrayOptions()->unnestPath ) break;
+                if ( currSpec->arrayOptions()->unnestPath != specUnnestPath ) break;
                 if ( !currSpec->what() || !spec.what() ) break;
                 if ( currSpec->what()->toJSONString() != spec.what()->toJSONString() ) break;
                 // Same index spec and unnestPath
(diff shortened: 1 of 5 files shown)


LiteCore/Query/SQLiteKeyStore+ArrayIndexes.cc
```diff
@@ -30,12 +30,15 @@ namespace litecore {
...
+            // Escape capitals in the unnest path of the new spec, to match the existing spec.
+            alloc_slice specUnnestPath{
+                    SQLiteKeyStore::transformCollectionName(spec.arrayOptions()->unnestPath.asString(), true)};
             // If there is already index with the index name,
             // eiher delete the current one, or use it (return false)
             while ( true ) {
                 if ( currSpec->type != IndexSpec::kArray ) break;
                 if ( !currSpec->arrayOptions() ) break;
-                if ( currSpec->arrayOptions()->unnestPath != spec.arrayOptions()->unnestPath ) break;
+                if ( currSpec->arrayOptions()->unnestPath != specUnnestPath ) break;
```
jianminzhao, line 41: Better to change arrayOptions()->unnestPath. It should return the original input (non-escaped)

jianminzhao (changes requested):

LiteCore/Query/SQLiteKeyStore+ArrayIndexes.cc
callumbirks: Why?

jianminzhao: Because options supposed to return back to the C4 users what they passed to the the config. Quoting of upper case is completely internal to the Core.

callumbirks: Resolved in the new commit.

jianminzhao:
But I think options should return the unescaped string, as it is passed in when the index is created. The original  motivation of it is so that platform can test the options against what's in index config. Looks like we don't have the test case with upper cases in the unnest path.

callumbirks:
> But I think options should return the unescaped string, as it is passed in when the index is created. The original  motivation of it is so that platform can test the options against what's in index config. Looks like we don't have the test case with upper cases in the unnest path.

The options do now contain the unescaped string.

LiteCore/Storage/SQLiteKeyStore.cc
```diff
@@ -233,6 +233,18 @@ namespace litecore {
         return ss.str();
     }
 
+    string SQLiteKeyStore::untransformCollectionName(std::string_view name) {
```
jianminzhao, line 236: Why introduce a new method, rather than using the above method? I don't think we allow backslash in the collection names.

jianminzhao (changes requested):

jianminzhao (approved):

PR #2254, opened 2025-04-10
CBL-6902: External Keys to work with client cert

jianminzhao:
- Override ExternalPrivateKey::isPrivateKeyDataAvailable to return false.
- Add a new bool option, kC4ReplicatorAuthClientCertKeyIsExternal, to C4ReplicatorParameters.
- Decode the new parameter in BuiltInWebScoket.

### DIFF
--- a/C/include/c4ReplicatorTypes.h
+++ b/C/include/c4ReplicatorTypes.h
@@ -262,7 +262,6 @@ typedef struct C4ReplicatorParameters {
 #define kC4ReplicatorAuthClientCertKey "clientCertKey"  ///< Client cert's private key (data)
 #define kC4ReplicatorAuthToken         "token"          ///< Session cookie or auth token (string)
 
-
 // [2]: auth.type values:
 #define kC4AuthTypeBasic         "Basic"           ///< HTTP Basic (the default)
 #define kC4AuthTypeSession       "Session"         ///< SG session cookie
--- a/C/include/c4ReplicatorTypes.h
+++ b/C/include/c4ReplicatorTypes.h
@@ -203,7 +203,9 @@ typedef struct C4ReplicatorParameters {
     const C4SocketFactory* C4NULLABLE                 socketFactory;    ///< Custom C4SocketFactory, if not NULL
     C4ReplicationCollection*                          collections;
     size_t                                            collectionCount;
-    C4KeyPair* C4NULLABLE                             externalKey;
+#ifdef COUCHBASE_ENTERPRISE
+    C4KeyPair* C4NULLABLE externalKey;
+#endif
 } C4ReplicatorParameters;
 
 #pragma mark - CONSTANTS:
--- a/C/include/c4ReplicatorTypes.h
+++ b/C/include/c4ReplicatorTypes.h
@@ -203,7 +203,7 @@ typedef struct C4ReplicatorParameters {
     const C4SocketFactory* C4NULLABLE                 socketFactory;    ///< Custom C4SocketFactory, if not NULL
     C4ReplicationCollection*                          collections;
     size_t                                            collectionCount;
-    const C4KeyPair* C4NULLABLE                       externalKey;
+    C4KeyPair* C4NULLABLE                             externalKey;
 } C4ReplicatorParameters;
 
 #pragma mark - CONSTANTS:
@@ -260,9 +260,7 @@ typedef struct C4ReplicatorParameters {
                      ///< Implemented by BuiltInWebSocket.
 #define kC4ReplicatorAuthClientCert    "clientCert"     ///< TLS client certificate (value platform-dependent)
 #define kC4ReplicatorAuthClientCertKey "clientCertKey"  ///< Client cert's private key (data)
-#define kC4ReplicatorAuthClientCertKeyIsExternal                                                                       \
-    "clientCertKeyIsExternal"           ///< if true key (data) is a pointer to External PrivateKey
-#define kC4ReplicatorAuthToken "token"  ///< Session cookie or auth token (string)
+#define kC4ReplicatorAuthToken         "token"          ///< Session cookie or auth token (string)
 
 
 // [2]: auth.type values:
--- a/C/include/c4ReplicatorTypes.h
+++ b/C/include/c4ReplicatorTypes.h
@@ -259,9 +259,9 @@ typedef struct C4ReplicatorParameters {
                      ///< Implemented by BuiltInWebSocket.
 #define kC4ReplicatorAuthClientCert    "clientCert"     ///< TLS client certificate (value platform-dependent)
 #define kC4ReplicatorAuthClientCertKey "clientCertKey"  ///< Client cert's private key (data)
-#define kC4ReplicatorAuthClientCertKeyIsExternal \
-    "clientCertKeyIsExternal"                           ///< if true key (data) is a pointer to External PrivateKey
-#define kC4ReplicatorAuthToken         "token"          ///< Session cookie or auth token (string)
+#define kC4ReplicatorAuthClientCertKeyIsExternal                                                                       \
+    "clientCertKeyIsExternal"           ///< if true key (data) is a pointer to External PrivateKey
+#define kC4ReplicatorAuthToken "token"  ///< Session cookie or auth token (string)
 
 
 // [2]: auth.type values:
--- a/C/include/c4ReplicatorTypes.h
+++ b/C/include/c4ReplicatorTypes.h
@@ -203,6 +203,7 @@ typedef struct C4ReplicatorParameters {
     const C4SocketFactory* C4NULLABLE                 socketFactory;    ///< Custom C4SocketFactory, if not NULL
     C4ReplicationCollection*                          collections;
     size_t                                            collectionCount;
+    const C4KeyPair* C4NULLABLE                       externalKey;
 } C4ReplicatorParameters;
 
 #pragma mark - CONSTANTS:
--- a/C/include/c4ReplicatorTypes.h
+++ b/C/include/c4ReplicatorTypes.h
@@ -259,7 +259,10 @@ typedef struct C4ReplicatorParameters {
                      ///< Implemented by BuiltInWebSocket.
 #define kC4ReplicatorAuthClientCert    "clientCert"     ///< TLS client certificate (value platform-dependent)
 #define kC4ReplicatorAuthClientCertKey "clientCertKey"  ///< Client cert's private key (data)
+#define kC4ReplicatorAuthClientCertKeyIsExternal \
+    "clientCertKeyIsExternal"                           ///< if true key (data) is a pointer to External PrivateKey
 #define kC4ReplicatorAuthToken         "token"          ///< Session cookie or auth token (string)
+
 
 // [2]: auth.type values:
 #define kC4AuthTypeBasic         "Basic"           ///< HTTP Basic (the default)
--- a/Networking/WebSockets/BuiltInWebSocket.cc
+++ b/Networking/WebSockets/BuiltInWebSocket.cc
@@ -15,7 +15,6 @@
 #include "HTTPLogic.hh"
 #include "Certificate.hh"
 #include "CookieStore.hh"
-#include "c4Certificate.hh"
 #include "c4Database.hh"
 #include "c4ReplicatorTypes.h"
 #include "c4Socket+Internal.hh"
--- a/Networking/WebSockets/BuiltInWebSocket.cc
+++ b/Networking/WebSockets/BuiltInWebSocket.cc
@@ -257,8 +257,7 @@ namespace litecore::websocket {
             if ( parameters().externalKey ) {
 #ifdef COUCHBASE_ENTERPRISE
                 Retained<crypto::Cert> cert = make_retained<crypto::Cert>(certData);
-                _tlsContext->setIdentity(
-                        new crypto::Identity(cert, const_cast<C4KeyPair*>(parameters().externalKey)->getPrivateKey()));
+                _tlsContext->setIdentity(new crypto::Identity(cert, parameters().externalKey->getPrivateKey()));
                 return true;
 #else
                 closeWithError(c4error_make(LiteCoreDomain, kC4ErrorUnsupported, "External Key is not supported"_sl));
--- a/Networking/WebSockets/BuiltInWebSocket.cc
+++ b/Networking/WebSockets/BuiltInWebSocket.cc
@@ -31,8 +31,8 @@ using namespace litecore::websocket;
 void C4RegisterBuiltInWebSocket() {
     // NOLINTBEGIN(performance-unnecessary-value-param)
     C4SocketImpl::registerInternalFactory(
-            [](websocket::URL url, fleece::alloc_slice options, std::shared_ptr<DBAccess> database) -> WebSocketImpl* {
-                return new BuiltInWebSocket(url, C4SocketImpl::convertParams(options), database);
+            [](websocket::URL url, fleece::alloc_slice options, std::shared_ptr<DBAccess> database, const C4KeyPair* externalKey) -> WebSocketImpl* {
+                return new BuiltInWebSocket(url, C4SocketImpl::convertParams(options, externalKey), database);
             });
     // NOLINTEND(performance-unnecessary-value-param)
 }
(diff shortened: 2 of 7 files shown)


C/include/c4ReplicatorTypes.h
```diff
@@ -259,7 +259,10 @@ typedef struct C4ReplicatorParameters {
                      ///< Implemented by BuiltInWebSocket.
 #define kC4ReplicatorAuthClientCert    "clientCert"     ///< TLS client certificate (value platform-dependent)
 #define kC4ReplicatorAuthClientCertKey "clientCertKey"  ///< Client cert's private key (data)
-#define kC4ReplicatorAuthToken         "token"          ///< Session cookie or auth token (string)
+#define kC4ReplicatorAuthClientCertKeyIsExternal                                                                       \
+    "clientCertKeyIsExternal"           ///< if true key (data) is a pointer to External PrivateKey
```
snej, line 263: This is a bad idea! You'll have to add a field to C4ReplicatorParameters instead.

snej (changes requested):

C/include/c4ReplicatorTypes.h
jianminzhao: Okay

Networking/WebSockets/BuiltInWebSocket.cc
```diff
@@ -252,7 +254,22 @@ namespace litecore::websocket {
                                             "Missing TLS client cert in C4Replicator config"_sl));
                 return false;
             }
-            if ( slice keyData = auth[kC4ReplicatorAuthClientCertKey].asData(); keyData ) {
+            if ( bool isExternal = auth[kC4ReplicatorAuthClientCertKeyIsExternal].asBool(); isExternal ) {
+#ifdef COUCHBASE_ENTERPRISE
```
pasin, line 258: I think it's not important to check for COUCHBSE_ENTERPRISE here as the external key feature itself is not an enterprise feature even though URLEndpointListener and TLSIdentity public API are enterprise only feature?

Replicator/c4RemoteReplicator.hh
```diff
@@ -218,8 +221,10 @@ namespace litecore {
 
 
       private:
-        alloc_slice const      _url;
-        const C4SocketFactory* _socketFactory{nullptr};
+        alloc_slice const                                            _url;
+        const C4SocketFactory*                                       _socketFactory{nullptr};
+        std::unique_ptr<const C4KeyPair, void (*)(const C4KeyPair*)> _socketExternalKey{
```
pasin, line 226: Wondering why not just using `Retained<C4KeyPair>`?

Networking/WebSockets/BuiltInWebSocket.cc
jianminzhao: It's engaged by the client cert. I can take the #ifdef out.

Replicator/c4RemoteReplicator.hh
jianminzhao: Good question. This took me hours. I used it at first and both Xcode and cmake for me on my local machine. The GH checks are all failing not seeing the complete definition of C4KeyPair even after I include c4certificate.hh. Oh, now I know why. It's because complete definition is inside #ifdef COUCHBASE_ENTERPRISE
This also answer the above question. CE does not see C4KeyPair
Interestingly, c4keypair_xyz are visible to CE.

C/include/c4ReplicatorTypes.h
```diff
@@ -259,7 +260,10 @@ typedef struct C4ReplicatorParameters {
...
```
snej, line 264: No longer necessary

Networking/WebSockets/BuiltInWebSocket.cc
```diff
@@ -22,17 +22,19 @@
 #include "StringUtil.hh"
 #include "ThreadUtil.hh"
 #include <string>
+#include "c4Certificate.hh"
```
snej, line 25: Project includes go above std ones

Networking/WebSockets/BuiltInWebSocket.cc
```diff
@@ -252,7 +254,22 @@ namespace litecore::websocket {
                                             "Missing TLS client cert in C4Replicator config"_sl));
                 return false;
             }
-            if ( slice keyData = auth[kC4ReplicatorAuthClientCertKey].asData(); keyData ) {
+            if ( bool isExternal = auth[kC4ReplicatorAuthClientCertKeyIsExternal].asBool(); isExternal ) {
```
snej, line 257: You don't need this flag, you can just check if `externalKey != nullptr`

Networking/WebSockets/BuiltInWebSocket.cc
```diff
@@ -252,7 +254,22 @@ namespace litecore::websocket {
...
+#ifdef COUCHBASE_ENTERPRISE
+                if ( !parameters().externalKey ) {
+                    closeWithError(c4error_make(LiteCoreDomain, kC4ErrorInvalidParameter,
+                                                "Missing externalKey when 'clientCertKeyIsExternal' is true"_sl));
+                    return false;
+                }
```
snej, line 263: Again, you don't need this.

Networking/WebSockets/BuiltInWebSocket.cc
```diff
@@ -252,7 +254,22 @@ namespace litecore::websocket {
...
+                Retained<crypto::Cert> cert = new crypto::Cert(certData);
+                _tlsContext->setIdentity(
+                        new crypto::Identity(cert, const_cast<C4KeyPair*>(parameters().externalKey)->getPrivateKey()));
```
snej, line 266: It's cleaner to use `make_retained` than `new`. (I know there's old code that uses `new`, but that's before we had `make_retained`.)

Networking/WebSockets/WebSocketImpl.hh
```diff
@@ -45,6 +47,7 @@ namespace litecore::websocket {
             int                 heartbeatSecs;       ///< WebSocket heartbeat interval in seconds (default if 0)
             fleece::alloc_slice networkInterface;    ///< Network interface
             fleece::AllocedDict options;             ///< Other options
+            const C4KeyPair*    externalKey;  ///< Client cert uses external key. Its life is tied to the Replicator.
```
snej, line 50: Please use `Retained` for safety.

Replicator/c4RemoteReplicator.hh
snej: Use `Retained`. If the compiler gives you trouble, it's because the destructor is being inlined, so add an explicit destructor in the .cc file.

C/include/c4ReplicatorTypes.h
```diff
@@ -203,6 +203,7 @@ typedef struct C4ReplicatorParameters {
     const C4SocketFactory* C4NULLABLE                 socketFactory;    ///< Custom C4SocketFactory, if not NULL
     C4ReplicationCollection*                          collections;
     size_t                                            collectionCount;
+    const C4KeyPair* C4NULLABLE                       externalKey;
```
snej, line 206: It doesn't make sense to use `const C4KeyPair*`, since C4KeyPair is immutable. (There are a bunch of other places this appears later on, too.)

snej (changes requested):

Networking/WebSockets/WebSocketImpl.hh
jianminzhao: I already "retained" in c4RemoteReplicator. It lives with the Replicator.

C/include/c4ReplicatorTypes.h
jianminzhao: Will remove the const.
As for Retained, the compiler wants to see the complete definition, instead of forward declaration. The entire class definition of C4KeyPair is under #ifdef COUCHBASE_ENTERPRISE

C/include/c4ReplicatorTypes.h
jianminzhao: will change

Networking/WebSockets/BuiltInWebSocket.cc
jianminzhao: will do

Networking/WebSockets/BuiltInWebSocket.cc
jianminzhao: will do

Networking/WebSockets/BuiltInWebSocket.cc
jianminzhao: yes

Networking/WebSockets/BuiltInWebSocket.cc
jianminzhao: will do

Networking/WebSockets/WebSocketImpl.hh
jianminzhao: No, it's hard to retain it because class definition of C4KeyPair is completely hidden to CE. To retain, it has to be inside EE ifdef.

Replicator/c4RemoteReplicator.hh
```diff
@@ -220,6 +223,12 @@ namespace litecore {
       private:
         alloc_slice const      _url;
         const C4SocketFactory* _socketFactory{nullptr};
+        // _socketExternalKey comes from C4ReplicatorParameters::externalKey. This parameter is
+        // part of kC4ReplicatorOptionAuthentication. When kC4ReplicatorAuthClientCertKeyIsExternal is true, we
+        // don't have the private key data in kC4ReplicatorAuthClientCertKey. Instead, the private key will
+        // be in the form of object, externalKey.
+        std::unique_ptr<C4KeyPair, void (*)(C4KeyPair*)> _socketExternalKey{nullptr,
+                                                                            [](C4KeyPair* k) { c4keypair_release(k); }};
```
pasin, line 231: One thing I have noticed in `C4Cert` is that the base declaration of the `C4Cert` struct, including its inheritance is outside `COUCHBASE_ENTERPRISE`. If doing the same for C4KeyPair, will it solve the problem here?

```
struct C4Cert final
    : public fleece::RefCounted
    , public fleece::InstanceCountedIn<C4Cert>
    , C4Base {
#ifdef COUCHBASE_ENTERPRISE
...
#endif  // COUCHBASE_ENTERPRISE
```

pasin (approved):
Functionality wise, look good to me. I will leave the discussion regarding the code structure with C4KeyPair between you and Jens.

C/include/c4ReplicatorTypes.h
pasin: One thing I have noticed in `C4Cert` is that the base declaration of the `C4Cert` struct, including its inheritance is outside `COUCHBASE_ENTERPRISE`. If doing the same for C4KeyPair, will it solve the problem here?

```
struct C4Cert final
    : public fleece::RefCounted
    , public fleece::InstanceCountedIn<C4Cert>
    , C4Base {
#ifdef COUCHBASE_ENTERPRISE
...
#endif  // COUCHBASE_ENTERPRISE
```

jianminzhao: This will suddenly expose a class definition of C4KeyPair to the CE. Currently, the only thing CE sees about C4KeyPair is a opaque declaration, except for c4keypair_retain ad c4keypair_release. (Don't whether it's intentional.) I think the cleaner way could be putting "#ifdef" around "externalKey" every to please the compiler. Although there is no way C4KeyPair can be created in CE, but the compiler won't compile if any functions of C4KeyPair is used except for _retain and _release.

jianminzhao: Basically, the code we are concerned about, using unique_ptr in lieu of Retained would not be hit in CE. We may put ENTERPRISE in C4ReplicatorParameters, or start the "#ifdef" in Core/c4RemoteReplicator and downward just to get compiled.

Replicator/c4RemoteReplicator.hh
jianminzhao: First, the destructor "C4KeyPair::~C4KeyPair() = default;" is not inlined.
Second, the entire class definition is #if'ed
`#ifdef COUCHBASE_ENTERPRISE

#    pragma mark - KEY PAIRS:

struct C4KeyPair final`

C/include/c4ReplicatorTypes.h
pasin: > This will suddenly expose a class definition of C4KeyPair to the CE

I think this is fine as we already did it for C4Cert?. Also as you mentioned, c4keypair_retain / release is also visible to CE already. If feels weird when having C4KeyPair passing through `C4ReplicatorParameters` when C4KeyPair definition is not actually available. I cannot comment about #ifdef `externalKey`.

> We may put ENTERPRISE in C4ReplicatorParameters, or start the "#ifdef" in Core/c4RemoteReplicator and downward just to get compiled.

This is counterpart alternative.

snej (changes requested):

Networking/WebSockets/WebSocketImpl.hh
```diff
@@ -45,6 +46,9 @@ namespace litecore::websocket {
             int                 heartbeatSecs;       ///< WebSocket heartbeat interval in seconds (default if 0)
             fleece::alloc_slice networkInterface;    ///< Network interface
             fleece::AllocedDict options;             ///< Other options
+#ifdef COUCHBASE_ENTERPRISE
+            Retained<C4KeyPair> externalKey;  ///< Client cert uses external key. Its life is tied to the Replicator.
```
snej, line 50:
```suggestion
            Retained<C4KeyPair> externalKey;  ///< Client cert uses external key.
```

Replicator/c4Socket+Internal.hh
```diff
@@ -22,7 +22,8 @@ namespace litecore::repl {
     // Main factory function to create a WebSocket.
     fleece::Retained<websocket::WebSocket> CreateWebSocket(const websocket::URL&, const fleece::alloc_slice& options,
                                                            std::shared_ptr<DBAccess>, const C4SocketFactory*,
-                                                           void* nativeHandle = nullptr);
+                                                           void*      nativeHandle = nullptr,
+                                                           C4KeyPair* externakKey  = nullptr);
```
snej, line 26:
```suggestion
                                                           C4KeyPair* externalKey  = nullptr);
```

Replicator/c4RemoteReplicator.hh
```diff
@@ -117,8 +120,14 @@ namespace litecore {
             _c4db_setDatabaseTag(dbOpenedAgain, DatabaseTag_C4RemoteReplicator);
             auto dbAccess =
                     make_shared<DBAccess>(dbOpenedAgain, _options->properties["disable_blob_support"_sl].asBool());
-            auto webSocket = CreateWebSocket(_url, socketOptions(), dbAccess, _socketFactory);
-            _replicator    = new Replicator(dbAccess, webSocket, *this, _options);
+            auto webSocket = CreateWebSocket(_url, socketOptions(), dbAccess, _socketFactory, nullptr
+#ifdef COUCHBASE_ENTERPRISE
+                                             ,
+                                             _socketExternalKey.get());
```
snej, line 126:
```suggestion
                                             _socketExternalKey);
```

Replicator/c4RemoteReplicator.hh
```diff
@@ -54,6 +54,9 @@ namespace litecore {
                 _customSocketFactory = *params.socketFactory;
                 _socketFactory       = &_customSocketFactory;
             }
+#if COUCHBASE_ENTERPRISE
+            if ( params.externalKey ) _socketExternalKey = params.externalKey;
```
snej, line 58:
```suggestion
           _socketExternalKey = params.externalKey;
```

Networking/WebSockets/BuiltInWebSocket.cc
```diff
@@ -15,6 +15,7 @@
 #include "HTTPLogic.hh"
 #include "Certificate.hh"
 #include "CookieStore.hh"
+#include "c4Certificate.hh"
```
snej, line 18: I don't think this is necessary since you're just passing a pointer around.

snej (changes requested):
Just a few minor tweaks

Networking/WebSockets/BuiltInWebSocket.cc
jianminzhao: done

Networking/WebSockets/WebSocketImpl.hh
jianminzhao: done

Replicator/c4Socket+Internal.hh
jianminzhao: done

Replicator/c4RemoteReplicator.hh
jianminzhao: done

pasin (approved):

snej (approved):

PR #2255, opened 2025-04-15
Keep pulled revs from being echoed back to the peer by the Pusher

snej:
This fixes a performance issue in bidirectional continuous sync. After the puller saves revs received from the peer, the ChangesFeed observes those revs being added to the db and tells the pusher to send them to the peer, which then rejects them with a 304.

The fix is to briefly track the revisions being added to the local db, then ignore them when they're reported by the C4DatabaseObserver.

I'm fixing this now because this situation will happen in all P2P replications, and since peers will be running multiple replications we want to get rid of any obvious wastes of CPU  and I/O like this one.

### DIFF
--- a/Replicator/ChangesFeed.cc
+++ b/Replicator/ChangesFeed.cc
@@ -132,13 +132,6 @@ namespace litecore::repl {
             uint32_t nChanges = nextObservation.numChanges;
             if ( nChanges == 0 ) break;
 
-            //TODO: Detect & ignore changes made by a concurrent pull replication from the same remote;
-            // i.e don't echo changes back to the peer we got them from in the first place!
-            // It's not harmful, but it's wasteful.
-            // (There used to be old code here that checked the `external` flag to do that,
-            // but it turned out it had been nonfunctional for quite a while. We'd need a better
-            // mechanism for discovering where the changes came from.) --Jens, Feb 2025
-
             logVerbose("Observed %u db changes #%" PRIu64 " ... #%" PRIu64, nChanges, (uint64_t)c4changes[0].sequence,
                        (uint64_t)c4changes[nChanges - 1].sequence);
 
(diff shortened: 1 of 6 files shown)


Replicator/ChangesFeed.cc
```diff
@@ -262,7 +261,14 @@ namespace litecore::repl {
     ReplicatorChangesFeed::ReplicatorChangesFeed(Delegate& delegate, const Options* options, DBAccess& db,
                                                  Checkpointer* cp)
         : ChangesFeed(delegate, options, db, cp)  // DBAccess is a subclass of access_lock<C4Database*>
-        , _usingVersionVectors(db.usingVersionVectors()) {}
+        , _usingVersionVectors(db.usingVersionVectors()) {
+        if ( _options->push(_collectionIndex) == kC4Continuous ) _db.echoCanceler.trackCollection(_collectionIndex);
+    }
+
```
jianminzhao, line 267: "only for collections that have bidirectional continuous replication" - but looks like either above, (continuous push), or below (continuous pull), would trigger trackCollection?

callumbirks: +1

Replicator/EchoCanceler.hh
```diff
@@ -0,0 +1,90 @@
...
+#include <unordered_map>
+#include <vector>
+
+C4_ASSUME_NONNULL_BEGIN
+
+namespace litecore::repl {
+
+    /** A set of {docID, revID} pairs that's used to avoid echoing a peer's own revisions back to it.
+        (There's actually one of these sets per collection, but only for collections that have
+        bidirectional continuous replication.)
```
callumbirks, line 27: Regarding what Jianmin pointed out above, maybe just this comment is slightly inaccurate?

callumbirks (approved):
Other than Jianmin's comment LGTM

PR #2257, opened 2025-04-17
Update self and third parties to modern cmake minimum

borrrden:
This includes a large update to mbedTLS to go from 2.x to 3.x


borrrden:
NOTE: Depends on https://github.com/couchbase/fleece/pull/260

snej (approved):

PR #2287, opened 2025-06-03
CBL-7058: c4log_writeToBinaryFile doesn't set logging information whe…

jianminzhao:
…n level is None

When the log level is None but file path is not empty, we will stage default LotFiles, which is to be activated when the level is set by setBinaryFileLevel.

### DIFF
--- a/C/c4Log.cc
+++ b/C/c4Log.cc
@@ -220,8 +220,10 @@ void c4log_initConsole(C4LogLevel level) noexcept {
 
 #pragma mark - FILE LOGGING:
 
-static C4LogObserver* sDefaultLogFiles;
-static C4LogLevel     sDefaultLogFilesLevel = kC4LogNone;
+static C4LogObserver*                                           sDefaultLogFiles;
+static C4LogLevel                                               sDefaultLogFilesLevel = kC4LogNone;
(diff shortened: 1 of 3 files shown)


C/c4Log.cc
```diff
@@ -220,30 +220,50 @@ void c4log_initConsole(C4LogLevel level) noexcept {
...
+            auto obs = make_retained<LogFiles>(convertFileOptions(options));
+            stagedDefaultLogFiles.reset(toExternal(std::move(obs)));
+        }
+
+        return true;
+    }  // Post-conditions: options.log_level != kC4LogNone
+
+    if ( options.base_path.empty() ) {
+        // treat it as setLevel
+        c4log_setBinaryFileLevel(options.log_level);
```
jianminzhao: Author's notes: I handle it this way to achieve some symmetry between path and level. we stage the path when the level is none and I try here to do with the level when the path is empty. What is still not symmetric is that we don't stage the level.
Please offer your suggestions.

borrrden: I don't know as there is a need to stage at all.  Can't you just have the log files set up in sDefaultLogFiles?  They won't do anything if the level is None right?

borrrden (approved):

snej (approved):

borrrden (approved):

PR #2289, opened 2025-06-11
CBL-7064: Pull filter can cause assertion failure during replication

callumbirks:
The original problem was essentially that the IncomingRev was settings its own status to Stopped and resetting its parent while another function of the IncomingRev (`handleRev`) was executing on the Puller's thread.

The solution is to make sure IncomingRev doesn't report Stopped when the Puller is running `handleRev`, and to avoid status updates on the IncomingRev's thread, and instead update status on the Puller's thread.

Basically the whole lifecycle of the IncomingRev (`handleRev` to begin, `reset` to end) is run on the Puller's thread.
The IncomingRev only uses `Stopped` or `Busy` status. `Busy` status should always be set once the Puller has called `handleRev`, `Stopped` status should always be set once the Puller has called `reset`. 

We now use an atomic boolean to track this so the IncomingRev will report the correct status.

We also now override `afterEvent` on IncomingRev to avoid updating status on the IncomingRev's thread, and we instead manually call `Worker::afterEvent` in `IncomingRev::reset` which is called on the Puller's thread. This means the status is safely synchronously updated BEFORE the IncomingRev re-enters the `_spareIncomingRevs` queue.

### DIFF
--- a/Replicator/IncomingRev.cc
+++ b/Replicator/IncomingRev.cc
@@ -467,7 +467,7 @@ namespace litecore::repl {
         Worker::afterEvent();
         _rev            = nullptr;
         _parent         = nullptr;
-        _handlingRev = false;
+        _handlingRev    = false;
         _remoteSequence = {};
         _bodySize       = 0;
     }
@@ -476,7 +476,7 @@ namespace litecore::repl {
     // We will calculate status on the Puller's thread.
     // This is because the start (`handleRev`) and end (`reset`) of the IncomingRev lifecycle
     // is all already done on the Puller's thread.
-    void IncomingRev::afterEvent() { }
+    void IncomingRev::afterEvent() {}
 
     Worker::ActivityLevel IncomingRev::computeActivityLevel(std::string* reason) const {
         std::string           parentReason;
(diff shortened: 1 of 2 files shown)


jianminzhao:
I have concern on the change on afterEvent. It changes the invariant that afterEvent is always called on the actor's thread. Since "Basically the whole lifecycle of the IncomingRev (handleRev to begin, reset to end) is run on the Puller's thread," why cannot we check _handlingRev at the time when attempt to reuse it. If it's still true, we can just, instead of pop_back, we create a new IncomingRev?

callumbirks:
That was one approach I thought of, and tested, and it "works", but I don't think it makes sense for IncomingRev.

It makes sense for other Actors who control their own lifecycle, but IncomingRev is almost 100% ran synchronously on Puller thread, except if `parseAndInsert` is delayed to run pull filter or delta. And `parseAndInsert` doesn't actually affect the lifecycle or status of the IncomingRev.

That's why I think moving the critical status-check onto the Puller thread makes the most sense. The status changes only occur on the Puller thread.

Yes, this solution technically "changes the invariant that afterEvent is always called on the actor's thread", although really I think it doesn't make much sense for that invariant to apply to IncomingRev - again the only time the Actor queue is ever used in IncomingRev is to offload `parseAndInsert` work to later IF pull filter or delta is used, and that method does not affect the lifecycle or status, it defers to the Puller to manage the status.

pasin:
From my chat with Callum, the master branch also has the same race but the issue doesn't happen because master branch as a different way of getting the collection. The fix is fine with me but I think it's a good idea to ask Jens to review this as well to make sure that it's inline with the architecture that he has designed.

PR #2294, opened 2025-06-26
CBL-7064 (Attempt 2): Manage IncomingRev lifecycle properly

callumbirks:


Replicator/IncomingRev.cc
```diff
@@ -479,8 +482,11 @@ namespace litecore::repl {
     // Puller we are done.
     void IncomingRev::afterEvent() {
         Worker::afterEvent();
-        if ( _shouldNotifyPuller ) { _puller->revWasHandled(this); }
-        _shouldNotifyPuller = false;
+        if ( _shouldNotifyPuller.exchange(false) ) {
```
jianminzhao: Do you know why it needs be so complicated. Isn't _insertWasEnqueued all is needed? Can it be reset when reset() is called? We need to read it only in IncomingRev::finish is called. And afterEvent is called only if _insertWasEnqueued is true. It doesn't even need to be atomic because it's assigned value before it is enqueued, and is read in IncomingRev's queue.

callumbirks: Because `afterEvent` may be fired before, at the same time, or after `finish`. And `finish` is always fired on `Inserter` thread.

Replicator/IncomingRev.cc
```diff
@@ -459,11 +460,11 @@ namespace litecore::repl {
...
         _rev->trim();
 
-        // If IncomingRev is not the active Actor, this was called on the Puller's thread, so we can notify the
-        // Puller straight away.
-        // Otherwise, we will call `revWasHandled` later, in `Worker::afterEvent`.
-        if ( Actor::currentActor() != this ) _puller->revWasHandled(this);
-        else { _shouldNotifyPuller = true; }
+        // If insert was enqueued, the last code to fire will be in `afterEvent`, so delay return to
+        // Puller until `afterEvent`. Otherwise, notify Puller now so this IncomingRev can be recycled.
+        if ( _insertWasEnqueued.exchange(false) ) _shouldNotifyPuller = true;
```
jianminzhao: _shouldNotifyPuller is not atomically exchanged in afterEvent().

callumbirks: Yes it is?

jianminzhao: I meant collectively.  Let's look at this case
1a. _insertWasEnqueued.exchange(false) == true.   // in finish()
1b. _shouldNotifyPuller.exchange(false)   == true or false // afterEvent(), depending 2a

2a. _shouldNotifyPuller = true.  // in finish()

callumbirks: Should be resolved in the latest commit using Atomic Enum to track state.

jianminzhao (approved):

PR #2297, opened 2025-06-27
C4PeerSync shutdown fixes

snej:
- CBL-7101, CBL-7102: Ensure proper shutdown of C4PeerSync / MeshManager by making it safe to release it before the stop method has completed.
- CBL-7134: Ensure Observers are removed from ObserverLists before they destruct.

(most of the changes are in https://github.com/couchbase/couchbase-lite-core-EE/pull/58 .)


pasin:
I tested the fix and all of the issues mentioned above are not reproducible any more.

pasin (approved):

PR #2299, opened 2025-07-01
CBL-7013: Native crash when using LiteCore 3.3.0-80

jianminzhao:
I used LiteCore 3.3.0-80 to reproduce the native crash when the following tests are reinstated,
- ReplicatorPendingDocIdTest
- ReplicatorLocal2LocalTest

The issue is specific to LoopbackWebSocket. We need to guard against the case when Replictor::_onConnect comes after it is stopped.

### DIFF
--- a/Replicator/Replicator.cc
+++ b/Replicator/Replicator.cc
@@ -607,8 +607,6 @@ namespace litecore::repl {
     }
 
     void Replicator::_onConnect() {
-        logInfo("Connected!");
-
         // onConnect may come after the replicator is asked to stop. This situation is
         // guarded against in WebSocketImpl::onConnect. However, LoopbackWebSocket does
         // not go throuth WebSocketImpl. We catch the case here.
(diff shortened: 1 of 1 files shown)


Replicator/Replicator.cc
```diff
@@ -609,6 +609,15 @@ namespace litecore::repl {
...
         logInfo("Connected!");
 
+        // onConnect may come after the replicator is asked to stop. This situation is
+        // guarded against in WebSocketImpl::onConnect. However, LoopbackWebSocket does
+        // not go throuth WebSocketImpl. We catch the case here.
+        if ( _connectionState != Connection::kConnecting ) {
+            logInfo("Method=_onConnect connectionState=%d Replicator is not in Connecting state, ignoring _onConnect",
+                    _connectionState);
+            return;
+        }
```
pasin, line 616: What do you think about only log "Connected"  after this if as it's actually not connected if that case? 

Also, when having `Method=_onConnect`, it feels like a debugging log message. How about something like this?

```
logInfo("Replicator not in connecting state (connectionState=%d); ignoring onConnect.", _connectionState);
```

jianminzhao: super

pasin: Should `logInfo("Connected!");` move under `if (_connectionState != Connection::kConnecting) {... }` as it should be one or the other?

PR #2302, opened 2025-07-02
CBL-7125: c4Log sometimes does not log anything

jianminzhao:
Refactor c4log_writeToCallback and c4log_setCallbackLevel so that they function independently of each other. In particular, calling setCallbackLevel with LevelNone won't remove the default logCallback. The new behavior matches with that of the pair of writeToBinary and setBinaryFileLevel.


jianminzhao:
I have tested in Java as described in the ticket, "entire Java LogTest suite", before and after the fix.

PR #2306, opened 2025-07-15
CBL-7161: Multipeer replication fails when trying to perform merge co…

jianminzhao:
…nflicts

The issue is caused by the function that generates the rev ID of RevTrees. In this function, the document body passed in as an argument is encoded with the SharedKeys. The actual keys are not present in "body." In lieu of them are the indexes in the SharedKeys. Therefore, documents that differ only by keys may yield identical digests which are fed to the function that generates the rev ID.

We fixed it by applying the SharedKeys to the document body before the digest is computed.

Tests that assert rev IDs are adjusted to changes in computing the digest.

### DIFF
--- a/LiteCore/Database/TreeDocument.cc
+++ b/LiteCore/Database/TreeDocument.cc
@@ -612,7 +612,11 @@ namespace litecore {
                 mutable_slice mslice(digest.asSlice());
                 SecureRandomize(mslice);
             } else {
-                SHA1 tmp = (SHA1Builder() << revLen << slice(parentRevID.buf, revLen) << delByte << body).finish();
+                Scope        scope(body, _collection->dbImpl()->dataFile()->documentKeys());
+                const Value* bodyNoSK = Value::fromTrustedData(body);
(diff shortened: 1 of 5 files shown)


LiteCore/Database/TreeDocument.cc
```diff
@@ -612,7 +612,11 @@ namespace litecore {
                 mutable_slice mslice(digest.asSlice());
                 SecureRandomize(mslice);
             } else {
-                SHA1 tmp = (SHA1Builder() << revLen << slice(parentRevID.buf, revLen) << delByte << body).finish();
+                Scope        scope(body, _collection->dbImpl()->dataFile()->documentKeys());
```
snej: The documentKeys were already looked up just above on line 611, and the call to hasEncryptables also creates a Scope, so why not combine them? Create the Scope above the call to hasEncryptables. Then that function doesn't have to take the SharedKeys as a parameter.

jianminzhao: done

PR #2316, opened 2025-07-30
CBL-7196: Predictive Queries test failing when running with CBL 3.3.0

jianminzhao:
The failure is caused by the limit we put on the number of arguments, particularly, of function array_of. We changed the limit to 127, which is the limit that sQLite has on functions.


snej (changes requested):
You should remove the comment `max 9 means "unlimited"` on line 214, too.

snej (approved):

PR #2318, opened 2025-08-04
Fixed Clang errors by disabling some warnings

snej:


borrrden:
Does this depend on an EE PR or something?  Build failures in both Android and iOS.

borrrden (approved):

PR #2332, opened 2025-08-22
CBL-7347: Crash in puller _revsFinished

jianminzhao:
When errors occur (while waiting attachments), the IncomiingRev would be sent to queue _returningRevs multiple times. This can cause BAD ACCESS exception as the second time it is handled in _revsFinished.


snej:
I'd rather we fixed this by only adding the rev to the queue once. Was it impractical to do that?

PR #2335, opened 2025-08-29
Fixed problems syncing partially VV-upgraded database [CBL-6666, CBL-6593]

snej:
The key changes:

- Implemented [the scheme I'd come up with but forgotten about](https://docs.google.com/document/d/1hqeuQ7dlCD1XWsjnxRW6L5E15diLXkfo0Xn5F9vs13c) to represent a legacy revID as a version.
- Improved VectorRecord to be able to save a new revision with a legacy revid, which can happen when pulling from a peer that made further changes to the same doc before upgrading the db.

### DIFF
--- a/LiteCore/RevTrees/Version.cc
+++ b/LiteCore/RevTrees/Version.cc
@@ -12,6 +12,7 @@
 
 #include "Version.hh"
 #include "Base64.hh"
+#include "Endian.hh"
 #include "Error.hh"
 #include "HybridClock.hh"
 #include "RevID.hh"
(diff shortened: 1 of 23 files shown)


borrrden:
Is the failing test of any consequence or is it known to fail?

LiteCore/RevTrees/Version.cc
```diff
@@ -23,9 +24,17 @@ namespace litecore {
...
+    Version Version::legacyVersion(revid oldRev) {
+        // From "Migrating RevIDs to Version Vectors"
+        // <https://docs.google.com/document/d/1hqeuQ7dlCD1XWsjnxRW6L5E15diLXkfo0Xn5F9vs13c>
+        // - The generation becomes the upper 24 bits of the clock
+        // - The first 40 bits of the digest become the lower 40 bits of the clock
+        // - The legacy sourceID is kLegacyRevSourceID, which in ASCII is "Revision+Tree+Encoding".
+        slice    digest = oldRev.digest();
+        uint64_t n      = 0;
+        ::memcpy(&n, digest.buf, std::min(size_t(8), digest.size));
+        n = (fleece::endian::dec64(n) >> 24) | (uint64_t(oldRev.generation()) << 40);
```
torcolvin, line 36: If I am reading this correctly, when generation => 16777216, we will start counting at zero again.

This is excessively unlikely, but I'd want Sync Gateway to match this behavior, so let's define this as such.

snej: Yeah, it's part of the data loss inherent into stuffing the revid into 64 bits.

borrrden (approved):

PR #2341, opened 2025-09-05
CBL-6969: Don't use env vars in CMake project

jianminzhao:


borrrden (approved):
I'm assuming you have tested that this works correctly.  The unguarded "set version" logic makes me nervous but I assume it's going to honor the passed values rather than the defaults that it sets.

PR #2343, opened 2025-09-08
Fixed race condition calling Replicator::httpResponse() [CBL-7419]

snej:
...after the connection has closed.

Solution is to copy the response to the Replicator when it arrives, instead of assuming the WebSocket is still around at the time of the call.

### DIFF
--- a/Replicator/Replicator.cc
+++ b/Replicator/Replicator.cc
@@ -268,7 +268,12 @@ namespace litecore::repl {
             _subRepls[coll].puller->start(_subRepls[coll].checkpointer->remoteMinSequence());
     }
 
-    pair<int, websocket::Headers> Replicator::httpResponse() const { return webSocket()->httpResponse(); }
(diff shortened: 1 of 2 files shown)


snej:
Windows build failed in Jenkins; looks like the tests timed out? Restarting the build.

Replicator/Replicator.cc
```diff
@@ -268,7 +268,12 @@ namespace litecore::repl {
             _subRepls[coll].puller->start(_subRepls[coll].checkpointer->remoteMinSequence());
     }
 
-    pair<int, websocket::Headers> Replicator::httpResponse() const { return webSocket()->httpResponse(); }
+    pair<int, websocket::Headers> Replicator::httpResponse() const {
```
jianminzhao, line 271: _httpStatus/_httpHeaders are captured only when _onConnect is called. Is webSocket()->httpResponse() also captured when connection is made?

snej: Yes.

PR #2344, opened 2025-09-10
Adding quotes to variables (BldNum & Version) inside the powerscript …

jianminzhao:
…function.


borrrden (changes requested):
Change the CMakeLists set(BLD_NUM) to use STRING instead of the non-existent INTEGER

borrrden (approved):

PR #2351, opened 2025-09-16
Fixed Xcode 26 / Clang 17 build errors

snej:
- Disabled new warnings that complain about identifiers or macros beginning with an underscore
- Did same to Fleece; updated submodule
- Removed unused private instance variables in VectorRecord, per new warning
- Added a missing include of <ostream> in SQLUtil.hh

⚠️  Depends on PR https://github.com/couchbase/fleece/pull/271 -- the Check Submodules CI test will fail until that's merged.

### DIFF
--- a/C/tests/c4DatabaseTest.cc
+++ b/C/tests/c4DatabaseTest.cc
@@ -30,6 +30,7 @@
 #include <cerrno>
 #include <future>
 #include <iostream>
+#include <mutex>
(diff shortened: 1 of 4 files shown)


snej:
macOS seems to have hung in test "Multiple Collections Incremental Revisions"; it hadn't progressed in over an hour so I killed it.

PR #2352, opened 2025-09-18
Updated VV conflict resolution when keeping winning rev's body

snej:
Per @torcolvin's new design.

When resolving a conflict, if the merged revision has the same body as the winning rev, generate a "trivial merge" version vector. This doesn't have the form of a merge VV; instead it just adds the versions from the losing VV but keeps the same current version as the winner.

In terms of the API, a trivial merge results when calling `c4doc_resolveConflict` and the `mergedBody` is either nullslice or is an equivalent FLDict to the winning rev's body.

Reference : CBL-7455

### DIFF
--- a/LiteCore/RevTrees/VersionVector.cc
+++ b/LiteCore/RevTrees/VersionVector.cc
@@ -18,6 +18,7 @@
 #include "slice_stream.hh"
 #include "NumConversion.hh"
 #include <algorithm>
+#include <unordered_map>
 
 namespace litecore {
     using namespace std;
@@ -320,7 +321,7 @@ namespace litecore {
                     ++i;
             }
         }
-        // Now add the new version:
+        // Now add the new version and clear the 'isMerge' state:
         _vers.insert(_vers.begin(), v);
         _nCurrent = 1;
 #if DEBUG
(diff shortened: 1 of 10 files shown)


LiteCore/RevTrees/VersionVector.cc
```diff
@@ -422,6 +428,57 @@ namespace litecore {
         return result;
     }
 
+    VersionVector VersionVector::trivialMerge(const VersionVector& winner, const VersionVector& loser) {
+        // Build a map from an author to its newest timestamp in either vector:
+        unordered_map<SourceID, logicalTime> merge;
+        for ( auto [author, time] : winner._vers ) merge.insert({author, time});
```
torcolvin, line 434: I think this code is probably correct, but without the same test cases that are in https://github.com/couchbase/sync_gateway/blob/e67f6dced72f502e44b553e06882daf8de570419/db/hybrid_logical_vector_test.go#L1429 I can not be sure of this, especially around merge vectors.

The reference implementation is in https://github.com/couchbase/sync_gateway/blob/e67f6dced72f502e44b553e06882daf8de570419/db/hybrid_logical_vector.go#L851

I think it is OK to merge this code and follow up in https://jira.issues.couchbase.com/browse/CBL-7402 to write tests for this function.

PR #2353, opened 2025-09-19
Detect redundant merges when pulling revs [CBL-7457]

snej:
When a VV document adds an existing [pulled] revision, it now checks if it's a redundant merge: if it has the same body and the same merge parents (MV) as the current revision, even if the revID (CV) is different.

In this situation it doesn't update the current local revision, though it updates the remote's current rev as usual.

The example from the unit test I added:
- Doc's current VV is `8@*, 7@AliceAliceAliceAliceAA, 6@BobBobBobBobBobBobBobA; 4@CarolCarolCarolCarolCA` with body `{'ok':'go'}`.
- Pull revision `9@ZegpoldZegpoldZegpoldA, 7@AliceAliceAliceAliceAA, 6@BobBobBobBobBobBobBobA; 4@CarolCarolCarolCarolCA`, also with body `{'ok':'go'}`
- This is not treated as a conflict. Instead the current version is unaltered, though the remote's current version is updated.

### DIFF
--- a/LiteCore/Database/VectorDocument.cc
+++ b/LiteCore/Database/VectorDocument.cc
@@ -401,22 +401,38 @@ namespace litecore {
                 return -1;
             }
 
-            // Compare it with the current document revision:
-            versionOrder order = VersionVecWithLegacy::compare(newVers, curVers);
-            logPutExisting(curVers, newVers, order, remote);
+            Doc fldoc = _newProperties(rq, outError);
+            if ( !fldoc ) return -1;
+            Dict newProperties = fldoc.asDict();
 
-            // Check for no-op or conflict:
-            int commonAncestor = 1;
-            if ( order != kNewer ) {
-                if ( remote == RemoteID::Local ) {
-                    if ( order == kConflicting ) {
-                        c4error_return(LiteCoreDomain, kC4ErrorConflict, nullslice, outError);
-                        return -1;
-                    } else {
-                        return 0;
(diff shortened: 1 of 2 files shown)


LiteCore/Database/VectorDocument.cc
```diff
@@ -401,22 +401,38 @@ namespace litecore {
...
+            Dict newProperties = fldoc.asDict();
+
+            versionOrder order;
+            int          commonAncestor;
+            if ( newVers.vector.mergesSameVersions(curVers.vector) && newProperties.isEqual(_doc.properties()) ) {
+                // This a redundant merge: same body, same MV.
+                order = kSame;
+                logPutExisting(curVers, newVers, order, remote);
+                commonAncestor = 0;
+                if ( remote == RemoteID::Local ) return commonAncestor;
```
torcolvin, line 415: I don't understand the complexities of this code, but I don't understand why the scenario where you'd hit `remote != RemoteID::Local` and why you'd want to continue to to add the doc.

From my Sync Gateway based perspective, this is treated the same as if you tried to add 2-def to 2-def, return as not adding the revision.

snej: That test checks whether the revision being added is coming from an active (client) replicator; if it isn't, it bails out as a no-op because there's no need to remember which remote peer is came from. Basically this is the case when acting as a passive/server.

PR #2354, opened 2025-09-19
CBL-7443: Segmentation fault in Android 3.3.0-59 build while running …

jianminzhao:
…volume test

The stack trace points to litecore::websocket::WebSocket::validatePeerCert(). It uses a callback function that captures "this" of PeerReplicateTask. To prevent it from being called after the object is deleted, we keep a weak_ptr in WebSocket. The callback shared_ptr will be nullified in ReplicateTask::stop.

### DIFF
--- a/Networking/WebSockets/WebSocketInterface.hh
+++ b/Networking/WebSockets/WebSocketInterface.hh
@@ -134,9 +134,13 @@ namespace litecore::websocket {
 
         using PeerCertValidator = std::function<bool(slice certData, std::string_view hostname)>;
 
-        void setPeerCertValidator(PeerCertValidator cv) { _peerCertValidator = std::move(cv); }
+        void setPeerCertValidator(std::shared_ptr<PeerCertValidator> cv) {
+            _peerCertValidator = std::move(cv);
+            // If we are not given a valid validator, the validator will always return true.
+            if ( cv ) _hasPeerCertValidator = true;
+        }
 
-        bool hasPeerCertValidator() const { return _peerCertValidator != nullptr; }
+        bool hasPeerCertValidator() const { return _hasPeerCertValidator; }
 
         /** The HTTP response status and headers.
             Available after a client socket connects. Otherwise zero status and empty headers. */
@@ -158,7 +162,11 @@ namespace litecore::websocket {
         ~WebSocket() override;
 
         bool validatePeerCert(slice certData, std::string_view hostname) {
-            return !_peerCertValidator || _peerCertValidator(certData, hostname);
+            if ( !_hasPeerCertValidator ) return true;
+            if ( auto validator = _peerCertValidator.lock() ) return (*validator)(certData, hostname);
+            else
+                // This is the case when the validator is expired by ReplicateTask
+                return false;
         }
 
         /** Called by the public connect(Delegate*) method. This should open the WebSocket. */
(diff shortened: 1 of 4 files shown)


snej:
The WebSocket only connects once, but the C4Replicator can reconnect multiple times so it needs to hang onto it.

It's not a problem having a cycle as long as it gets broken. When the PeerReplicateTask stops it clears the callback, which breaks the cycle.

The two lines you quoted about a race condition are in different classes setting different variables, so those aren't a race condition. Please show me the actual place where you think the race condition would occur.

jianminzhao:
Before this PR, repl->setPeerTLSCertificateValidator() is called only once on the callback (non-null). And this is passed to WebSocket::_peerCertValidator as std::function.  This is good. However, the goal this PR is to avoid calling it if the Task object is freed. Now I got to this situation,
A. in ReplicateTask::stop(): _peerTLSCertificateValidator = nullptr; 
B. in C4RemoteReplicator:  webSocket->setPeerCertValidator(_peerTLSCertificateValidator);
(A) is writer, and (B) is reader. This is the race condition I thought you were concerned about.

Here are reference relationships: 
PeerReplicatorTask -> _replicator -> _peerTLSCertificateValidator -> PeerReplicatorTask
It looks like we cannot keep the validator if the PeerReplicateTask is deleted.

snej:
You're describing this confusingly. It's very important to know what members are being accessed, and you're not saying that clearly.

> A. in ReplicateTask::stop(): _peerTLSCertificateValidator = nullptr;

That's not in ReplicateTask::stop(). That's in a method it calls, C4Replicator::setPeerTLSCertificateValidator.

So I think you're referring to C4RemoteReplicator:: _peerTLSCertificateValidator as the value with the race condition?

jianminzhao:
Yes. Sorry for the ambiguity.

Networking/WebSockets/WebSocketInterface.hh
```diff
@@ -134,9 +134,13 @@ namespace litecore::websocket {
 
         using PeerCertValidator = std::function<bool(slice certData, std::string_view hostname)>;
 
-        void setPeerCertValidator(PeerCertValidator cv) { _peerCertValidator = std::move(cv); }
+        void setPeerCertValidator(std::shared_ptr<PeerCertValidator> cv) {
+            _peerCertValidator = std::move(cv);
+            // If we are not given a valid validator, the validator will always return true.
+            if ( cv ) _hasPeerCertValidator = true;
```
snej: Should set the flag when cv is null too:
```suggestion
           _hasPeerCertValidator = !!cv;
```

Replicator/c4ReplicatorImpl.cc
```diff
@@ -184,10 +184,18 @@ namespace litecore {
 
 
 #ifdef COUCHBASE_ENTERPRISE
-    void C4ReplicatorImpl::setPeerTLSCertificateValidator(PeerTLSCertificateValidator v) {
+    using PeerTLSCertificateValidator = C4Replicator::PeerTLSCertificateValidator;
+
+    void C4ReplicatorImpl::setPeerTLSCertificateValidator(std::shared_ptr<PeerTLSCertificateValidator> v) {
+        std::scoped_lock<std::mutex> lock(_peerValidatorMutex);
```
snej: You don't need a scoped_lock here. That's only for locking multiple mutexes. Just use a regular unique_lock. Same thing on line 195.

jianminzhao: Done

Networking/WebSockets/WebSocketInterface.hh
jianminzhao: done

PR #2357, opened 2025-09-29
CBL-7461: Possible racing in Replicator & WebSocket compound

jianminzhao:
Addressing a possible race condition as
- Replicator::onConnect() and
- BLIPIO::_closed()

BLIPIO::_closed() can be called from WebSocket not going through the state Connection::kClosing, but because of an error condition. It can happen as the Replicator goes from kConnecting to onConnect(). When this race occurs, Replicator::onConnect() can encouter the webSocket being nullified by BLIPIO::_closed.

We protect it by introducing Connection::webSocketSafe(). With it, if Replicator gets non-null webSocket, the the webSocket won't be deleted in BLIPIO::_closed until Replicator::onConnect() is done.

### DIFF
--- a/Networking/BLIP/BLIPConnection.cc
+++ b/Networking/BLIP/BLIPConnection.cc
@@ -114,6 +114,7 @@ namespace litecore::blip {
         Stopwatch                                       _timeOpen;
         atomic_flag                                     _connectedWebSocket = ATOMIC_FLAG_INIT;
         Retained<WeakHolder<Delegate>>                  _weakThis{new WeakHolder<Delegate>(this)};
+        mutable std::mutex                              _webSocketMutex;
 
       public:
         BLIPIO(Connection* connection, WebSocket* webSocket, Deflater::CompressionLevel compressionLevel)
@@ -153,6 +154,16 @@ namespace litecore::blip {
         }
 
         WebSocket* webSocket() const { return _webSocket; }
+
+        Retained<WebSocket> webSocketSafe() const {
+            std::unique_lock lock(_webSocketMutex);
+            return _webSocket;
+        }
+
+        void resetWebSocket() {
+            std::unique_lock lock(_webSocketMutex);
+            _webSocket = nullptr;
+        }
 
         std::string loggingIdentifier() const override {
             return _connection ? _connection->name()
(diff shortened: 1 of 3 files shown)


Networking/BLIP/BLIPConnection.cc
```diff
@@ -154,6 +155,16 @@ namespace litecore::blip {
 
         WebSocket* webSocket() const { return _webSocket; }
 
+        Retained<WebSocket> webSocketSafe() const {
```
snej: Why don't we just name this method `webSocket()`, i.e. replace the existing method with this one. There's no need to have two when there's only one caller of this method.

Networking/BLIP/BLIPConnection.cc
```diff
@@ -696,4 +707,6 @@ namespace litecore::blip {
 
     websocket::WebSocket* Connection::webSocket() const { return _io->webSocket(); }
 
+    Retained<websocket::WebSocket> Connection::webSocketSafe() const { return _io ? _io->webSocketSafe() : nullptr; }
```
snej: Similarly, let's replace the `webSocket()` method with this one.

That will mean also changing `Replicator::webSocket()` to return `Retained<WebSocket>`.

snej (changes requested):

Networking/BLIP/BLIPConnection.cc
jianminzhao: I was meaning to make as little change as possible. This is a very special case. In general, the webSocket is like a private member of connection. In fact except for uses by tests, Replicator::onConnect is the only place where webSocket is used. And it's used only to get role(). If we store the role in Connection at the construction time, we don't even need to expose it.
However, as you pointed, since we are exposing it, why not to expose only safe version.
I will make the change.

Networking/BLIP/BLIPConnection.hh
```diff
@@ -99,6 +99,7 @@ namespace litecore::blip {
         int8_t                                   _compressionLevel;
         std::atomic<State>                       _state{kClosed};
         CloseStatus                              _closeStatus;
+        mutable std::mutex                       _webSocketMutex;
```
snej: I don't think this is used anywhere.

Replicator/Replicator.cc
```diff
@@ -620,9 +620,9 @@ namespace litecore::repl {
             return;
         }
 
-        logInfo("Connected!");
+        if ( Retained<websocket::WebSocket> socket = webSocket(); socket->role() == websocket::Role::Client ) {
+            logInfo("Connected!");
```
pasin: The original crash indicates that the returned websocket is null, can you confirm that this line doesn’t need to check if the retained socket is valid before getting role like :

; socket & socket->role() == …

(From what I read C++ doesn’t check the initialized var?)

jianminzhao: Actually I have fixed it,

PR #2451, opened 2026-03-27
Deal with new warnings in Clang 21 / Xcode 26.4

snej:
Common ones:
- Unnecessary semicolons
- Initializing an enum with 0
- Unnecessary 'virtual' on methods in a final class
- Implicit fallthrough between switch cases

NOTE: Requires https://github.com/couchbase/fleece/pull/290


borrrden (changes requested):
Just need to take care of the fleece submodule issue

pasin (approved):
I tested it with XCode 26.4 and it worked (Need a callback order change in LiteCoreBTSocket.mm but that is in EE repo).

borrrden (approved):

PR #2453, opened 2026-03-27
Added TLSContext::registerPrivateKey() [CBL-8083]

snej:
This is a workaround for the problem of how to use a custom (external) PrivateKey with the outgoing version of TLSCodec::wrapSocketFactoryInTLS, which does not take a TLSContext and has no way to convey such an object to the code that eventually creates a TLSContext.

### DIFF
--- a/Networking/TLSContext.cc
+++ b/Networking/TLSContext.cc
@@ -20,6 +20,7 @@
 #include "Logging.hh"
 #include "fleece/Fleece.hh"
 #include "sockpp/mbedtls_context.h"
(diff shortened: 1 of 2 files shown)


borrrden (approved):
I'd like to see a note here to replace this with persistent key storage support in the future.  I believe that we can set it up in such a way that it fails gracefully if the Java gets mismatched later, but with some helper methods like those in KeyStoreManagerDelegate.java we could have proper support for persistent keys on Android too.

PR #2454, opened 2026-03-30
CBL-8010: Guard object ref logging callback with mutex

borrrden:
Otherwise, if another one happened quickly on another thread the sFormatBuffer could get corrupted


jianminzhao (approved):

borrrden:
Validation failure looks unrelated to the PR

pasin (approved):

PR #2459, opened 2026-04-02
fix: skip live table scan for deleted-only queries

Nao-ris:
## Context

Hello 👋

In our park of users some have large cblite databases (many > 2GB).
To keep their size in check we periodically query for tombstones to purge them after a successful push replication.
We run the query `SELECT _._id FROM _ WHERE _._deleted LIMIT 100` and we noticed it is very slow even with a LIMIT. After investigation, it appears the query generates a UNION scan of both the live docs table (`kv_default`) and the deleted docs table (`kv_del_default`). The live table scan finds zero matches but still reads every row.

We see two potential solutions:

1. skip live table scan for deleted-only queries
2. create a new API endpoint in `couchbase-lite-core` & `couchbase-lite-C` to retrieve the ids of all deleted documents

This PR is my proposal to tackle solution 1.
Feel free to tell me if you see issues with it, or if you would prefer solution 2.

## Summary

This change detects when the WHERE clause guarantees only deleted documents can match, and routes the query to `kv_del_<collection>` directly instead of the `all_<collection>` UNION view. This resolves the `//FIXME: Support kDeletedDocs` at QueryTranslator.cc:144.

## Patterns optimized

- `WHERE _._deleted` (bare boolean)
- `WHERE _._deleted = true` / `IS TRUE`
- `WHERE _._deleted != false` / `IS NOT FALSE`
- All of the above inside AND chains

Patterns that reference `_deleted` without guaranteeing it (OR, SELECT, `= false`, etc.) correctly fall back to the UNION view.

## Safety: default collection migration

The default collection may have deleted docs in `kv_default` if the database was upgraded from a version prior to 3.1 (when `kv_del_` tables were introduced) and the background migration has not yet completed. A new `isDeletedTableComplete()` delegate method checks this: non-default collections always return true; the
default collection checks `kMaxRowidWithDeletedInDefault`. When migration is incomplete, the query falls back to the UNION view.

## Known limitation

Live queries (change tracking) are not registered for deleted-only queries, consistent with the existing behavior for queries using the `all_*` UNION view.

## Test plan

- [x] CppTests: 685,963 assertions, 553 cases — all passed
- [x] C4Tests: 543,111 assertions, 178 cases — all passed
- [x] clang-format applied

### DIFF
--- a/LiteCore/Query/Translator/ExprNodes.cc
+++ b/LiteCore/Query/Translator/ExprNodes.cc
@@ -241,6 +241,13 @@ namespace litecore::qt {
         return nullopt;
     }
 
+    optional<bool> LiteralNode::asBool() const {
+        if ( _literal.index() == 0 ) {
+            if ( Value v = get<Value>(_literal); v.type() == kFLBoolean ) return v.asBool();
+        }
+        return nullopt;
+    }
+
     string_view LiteralNode::asString() const {
         switch ( _literal.index() ) {
             case 0:
--- a/LiteCore/Query/Translator/ExprNodes.hh
+++ b/LiteCore/Query/Translator/ExprNodes.hh
@@ -53,6 +53,7 @@ namespace litecore::qt {
 
         FLValueType            type() const;
         std::optional<int64_t> asInt() const;
+        std::optional<bool>    asBool() const;
         string_view            asString() const;
 
         void setInt(int64_t);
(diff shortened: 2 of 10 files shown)


jianminzhao:
Hi @Nao-ris, have you signed CLA, https://developer.couchbase.com/open-source-projects/ (Steps for Contributing to a Project)?
Thanks!

Nao-ris:
After discussing with Fabrice Leray, this change will be handled internally by a CBSE.

PR #2465, opened 2026-04-09
Use '-r quiet' in lieu of '-r list' for CppTests and C4Tests

jianminzhao:


jianminzhao:
I tried to use "-r quiet" for windows, but retracted because it still causes hanging.

borrrden (approved):

PR #2467, opened 2026-04-10
CBL-8074: Do not retry replicator connections if WebSocket being clos…

jianminzhao:
…ed abnormally

The Abnormal Close occurs after the client is already decided to close the socket but fails to receive response from the server. We don't want to retry replicating on this condition. Hence, we move this status code out of the candidates of retrying; remove the Transient classification from this status code.

### DIFF
--- a/C/c4Error.cc
+++ b/C/c4Error.cc
@@ -335,7 +335,6 @@ __cold bool C4Error::mayBeTransient() const noexcept {
                                            502, /* Bad Gateway */
                                            503, /* Service Unavailable */
                                            504, /* Gateway Timeout */
-                                           websocket::kCodeAbnormal,
                                            websocket::kCloseAppTransient,
                                            0};
     static ErrorSet kTransient          = {// indexed by C4ErrorDomain
--- a/C/c4Error.cc
+++ b/C/c4Error.cc
@@ -335,6 +335,7 @@ __cold bool C4Error::mayBeTransient() const noexcept {
                                            502, /* Bad Gateway */
                                            503, /* Service Unavailable */
                                            504, /* Gateway Timeout */
+                                           websocket::kCodeAbnormal,
                                            websocket::kCloseAppTransient,
                                            0};
     static ErrorSet kTransient          = {// indexed by C4ErrorDomain
--- a/Networking/WebSockets/WebSocketImpl.cc
+++ b/Networking/WebSockets/WebSocketImpl.cc
@@ -556,7 +556,7 @@ namespace litecore::websocket {
                     if ( clean ) {
                         status.reason = kWebSocketClose;
                         // if (!expected) then _closeSent => !_closeReceived
-                        if ( !expected ) status.code = _closeSent ? kCodeStatusCodeExpected : kCodeAbnormal;
+                        if ( !expected ) status.code = _closeSent ? kCodeNormal : kCodeAbnormal;
                         else if ( !_closeMessage )
                             status.code = kCodeNormal;
                         else {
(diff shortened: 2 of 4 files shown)


Replicator/tests/ReplicatorAPITest.cc
```diff
@@ -864,6 +864,37 @@ TEST_CASE_METHOD(ReplicatorAPITest, "Stop after transient connect failure", "[C]
...
+TEST_CASE_METHOD(ReplicatorAPITest, "OneShot WebSocket kCodeAbnormal", "[C][Push][Pull]") {
+    _mayGoOffline           = true;
+    C4SocketFactory factory = {};
+    factory.open            = [](C4Socket* socket, const C4Address* addr, C4Slice options, void* context) {
+        c4socket_opened(socket);
+    };
+
+    factory.close = [](C4Socket* socket) {
+        // Not invoked
+        REQUIRE(false);
```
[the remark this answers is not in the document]
jianminzhao: It's for the same purpose of assertion; close is not supposed to be called in this test.

C/c4Error.cc
```diff
@@ -335,7 +335,6 @@ __cold bool C4Error::mayBeTransient() const noexcept {
                                            502, /* Bad Gateway */
                                            503, /* Service Unavailable */
                                            504, /* Gateway Timeout */
-                                           websocket::kCodeAbnormal,
```
pasin, line 338: From the meeting, I understand that we want to ignore the error meaning that the replicator will be stopped without the abnormal close error (Otherwise, all customers who are using Google Cloud Provider will never get clean stopped). 

Does removing this error along is enough to ignore the error?

jianminzhao: Perhaps I don't get the question, "Does removing this error along is enough to ignore the error?"
We just removed this particular from the list of transient errors. Just like other non-transient errors, the replicator is stopped, and the status contains the error code. 

"Otherwise, all customers who are using Google Cloud Provider will never get clean stopped" 
-- This is actually what happens with GCP as of today. It could be fixed by GCP in the future.

snej (approved):

C/c4Error.cc
pasin: There is a possibility but it could be low that it will get fixed anytime soon. Technically, it's not incorrect to return the error. However, in my opinion, the error is not useful for the application and we might get a CBSE ticket again from another customer about the error.

Replicator/c4ReplicatorImpl.cc
```diff
@@ -344,6 +344,14 @@ namespace litecore {
             if ( _status.level == kC4Stopped ) {
                 _replicator->terminate();
                 _replicator = nullptr;
+                if ( _status.error.code == websocket::kCodeStatusCodeExpected ) {
+                    // c.f. WebSocketImpl.cc
+                    // The replicator stopped with this error when it already sent
+                    // CLOSE to the remote but fail to receive a response. This may happen
+                    // with remote behind a Load Balancer as it tries to close the connection
+                    // as it is done. This error is interested to the client.
```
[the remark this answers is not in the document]
jianminzhao: Will make the change as suggested

Replicator/c4ReplicatorImpl.cc
```diff
@@ -344,6 +344,14 @@ namespace litecore {
...
+                    _status.error.code = 0;
```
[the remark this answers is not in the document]
jianminzhao: will change

snej (approved):

Networking/WebSockets/WebSocketImpl.cc
```diff
@@ -555,7 +555,8 @@ namespace litecore::websocket {
 
                     if ( clean ) {
                         status.reason = kWebSocketClose;
-                        if ( !expected ) status.code = kCodeAbnormal;
+                        // if (!expected) then _closeSent => !_closeReceived
```
pasin, line 558: Left over comment

PR #2470, opened 2026-04-21
CBL-7977: [Enhancement] c4repl_getCorrelationID to return the Correlation ID of the latest Replicator task after it's ended

jianminzhao:
Currently, the API returns null if the underlying Replicator is released. (Each Replicator has unique Correlation ID corresponding to the unique WebSocket). In this PR, we have it keep return the Correlation ID of the late Replicator until the new Replicator is started.

### DIFF
--- a/Replicator/c4ReplicatorImpl.cc
+++ b/Replicator/c4ReplicatorImpl.cc
@@ -59,6 +59,7 @@ namespace litecore {
         }
 
         if ( !_replicator ) {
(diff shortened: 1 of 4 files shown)


jianminzhao:
The companion PR in EE repo, https://github.com/couchbase/couchbase-lite-core-EE/pull/137

borrrden (approved):
PR title is not related at all to the content, it should be changed to something else.

PR #2473, opened 2026-04-28
CBL-8229: Upgrade SQLite to 3.53.0

jianminzhao:
Adjusted test, "SQLite numeric ops", because of the new version increased numerical precision.


pasin (commented):
I'm not entirely sure that this is a breaking change to users or not. Let's discuss about this before committing the change. Otherwise, we need to upgrade to a lower version.

pasin (approved):

PR #2474, opened 2026-04-28
CBL-7386: Crash in puller _revsFinished

jianminzhao:
Manually ported from 73d1b9ee. Two files are affected, IncomingRev.{hh,cc}, and they are already in sync on branches release/3.3, release/4.0, and master.


pasin (approved):
This fix is much simpler than the original fix.

PR #2484, opened 2026-04-30
CBL-8181: Upgrade mbedTLS to 3.6.6

jianminzhao:


pasin (approved):
Copilot suggested some improvement in CMake but I don't think that is critical as long as the build successes. However, there is a typo at the license header. Could you please fix that?

PR #2489, opened 2026-05-15
CBL-7841: Handle conflict when revision history appears broken

jianminzhao:
This is a fallback for cases where the general comparison algorithm yields a Conflict while applying a remote revision to the local document. The target scenario is a deleted document being resurrected in CBS -- when that happens, the resurrected document has no history prior to the deletion.

### DIFF
--- a/LiteCore/Database/VectorDocument.cc
+++ b/LiteCore/Database/VectorDocument.cc
@@ -428,6 +428,62 @@ namespace litecore {
             } else {
                 // Compare it with the current document revision:
                 order = VersionVecWithLegacy::compare(newVers, curVers);
+
+                // CBL-7841: Handle the case where revision history may appear broken.
+                //
+                // Concrete scenario for a single document ID:
+                //   1. Local DB deletes the doc.
+                //   2. Local pushes the tombstone to a remote SGW.
+                //   3. In the target CB database, a new doc is created with the same ID.
+                //      (In CBS, a deleted doc is effectively purged, so the ID is free
+                //      for reuse by an unrelated document.)
+                //   4. Local DB performs a pull.
+                //
+                // At step 4 the BLIP message carries a document that starts at a new
+                // version with no history. In general, this conflicts with whatever
+                // revision Local currently holds, and it is indistinguishable from a
+                // genuinely new document authored elsewhere (i.e. one that never went
+                // through the sequence above). The following analysis shows how we can
+                // recognize the sequence above and treat `newVers` as a newer revision
+                // rather than a conflict.
+                //
+                // Proposition:
+                //   For a given document at a given remote, the set of authors in the
+                //   Version Vector may only grow.
+                //
+                // Applying the proposition here:
+                //   - `remote`   : the remote index that `newVers` came from.
+                //   - `rev`      : the latest revision we know we have shared with
+                //                  `remote`.
+                //   - `newVers`  : the revision arriving from `remote` via BLIP.
+                //   - `curVers`  : the current local revision of the document.
+                //
+                // Because the remote has already seen `rev`, its Version Vector must
+                // contain `rev`'s author. Our general VV comparison relies on this. If
+                // that author is now missing from `newVers`, it must have been dropped.
+                // From this we can draw the following conservative conclusion:
+                //
+                //   `newVers` is "newer" than any version that shares an author with
+                //   `rev` at the same or older age (lower or equal logical clock).
+                //
+                // We further posit that `is_newer_or_same` is transitive. Because `rev`
+                // has been seen by the remote, we have:
+                //
+                //   newVers is_newer_or_same rev
+                //
+                // By transitivity, if rev is_newer_or_same curVers, then
+                // newVers is_newer_or_same curVers. The "same" case is already handled
+                // by the general comparison, so here we are only concerned with the
+                // strict is_newer case.
+
+                if ( order == kConflicting && remote != RemoteID::Local ) {
+                    if ( optional<Revision> rev = _doc.loadRemoteRevision(remote); rev ) {
+                        VersionVecWithLegacy landmark{rev->revID};
+                        versionOrder         orderWithLandmark = VersionVecWithLegacy::compare(landmark, curVers);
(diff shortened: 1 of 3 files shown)


LiteCore/Database/VectorDocument.cc
```diff
@@ -428,6 +428,62 @@ namespace litecore {
             } else {
                 // Compare it with the current document revision:
                 order = VersionVecWithLegacy::compare(newVers, curVers);
+
+                // CBL-7841: Handle the case where revision history may appear broken.
+                //
```
snej: Having a detailed explanation is good, but this is already quite a long method. One thing I've done before is to make the explanation a "footnote" at the end of the file, with a one-line reference to it inline.

LiteCore/Database/VectorDocument.cc
```diff
@@ -428,6 +428,62 @@ namespace litecore {
...
+                // newVers is_newer_or_same curVers. The "same" case is already handled
+                // by the general comparison, so here we are only concerned with the
+                // strict is_newer case.
+
+                if ( order == kConflicting && remote != RemoteID::Local ) {
+                    if ( optional<Revision> rev = _doc.loadRemoteRevision(remote); rev ) {
+                        VersionVecWithLegacy landmark{rev->revID};
+                        versionOrder         orderWithLandmark = VersionVecWithLegacy::compare(landmark, curVers);
+                        if ( orderWithLandmark == kSame || orderWithLandmark == kNewer ) order = kNewer;
+                    }
```
snej: As I understand it, this says "if the last known rev at the remote was the same or newer than the local rev, treat the remote's new rev as also newer than the local rev, instead of as a conflict."

I think this is too general. This situation isn't the same as the rev-tree "branch switch". It's specifically that the remote has an entirely new history, because Couchbase Server recreated the doc from scratch after it was deleted. We can detect that by checking whether `rev` and `newVers` have _no_ Versions in common, and also the local revision is a tombstone. What do you think?

snej (changes requested):

LiteCore/Database/VectorDocument.cc
jianminzhao: I did not find specific check of "the local revision is a tombstone". Did I miss the check somewhere?

jianminzhao: I think whether we accept the newVers only if the curVers is a tombstone is a decision by its own right, provided we already decided that "Couchbase Server recreated the doc from scratch." I tried to create the the doc in Couchbase Server after I deleted in Lite and syned to the SGW. There is no friction in the way. I also tried to create the doc in CB after the doc is syned to SGW before deleted. In this case CB won't let me do it because the doc already exists. I did not try to create the doc in CB first before CB sees it. This would simulate the case where a doc with same docID is created in Lite and CB concurrently, and we will encounter the same situation where the latestSynedInRemote shares no versions at at all. The only difference is whether curVers is a tombstone or not.

LiteCore/Database/VectorDocument.cc
jianminzhao: I will put the comments in the CBL ticket, and leave a short note in the source.

LiteCore/Database/VectorDocument.cc
```diff
@@ -428,6 +428,75 @@ namespace litecore {
...
+                // the general VV comparison to yield kConflict. We focus on the
+                // following conditions:
+                // 1. Conflict with genuine remote
+                // 2. curVers is a tombstone
+                // 3. Remote revision is available at `remote`
+                // 4. curVers == RemoteRevision(remote)
+                // 5. curVers.vector shares no version with newVers.vector
+                // 6. curVers.legacy[0] is not in newVers.legacy (RevTree parent)
+
+                for ( bool goOn = order == kConflicting && remote != RemoteID::Local; goOn; goOn = false ) {
```
snej: This is confusing. It is effectively an `if`, since it never runs more than once, but I'm guessing you made it a `for` so you could use `break` to exit it.
In my experience, when you have a block with multiple exits like this, it's time to turn it into a function. Or if that would add too many parameters to pass, use a `do{...}while(false)` statement along with a comment that this is just to allow`break`.

LiteCore/Database/VectorDocument.cc
```diff
@@ -428,6 +428,75 @@ namespace litecore {
...
+
+                    // 3. Remote revision is available
+                    optional<Revision> rev = _doc.loadRemoteRevision(remote);
+                    if ( !rev ) break;
+
+                    // 4. curVers == RemoteRevision(remote)
+                    VersionVecWithLegacy lastSynced{rev->revID};
+                    if ( curVers.legacy != lastSynced.legacy || curVers.vector != lastSynced.vector ) break;
+
+                    // 5. curVers.vector shares no version with newVers.vector
```
snej, line 461: This whole thing through line 488 should be a method on VersionVector, not inlined here.

LiteCore/Database/VectorDocument.cc
```diff
@@ -428,6 +428,75 @@ namespace litecore {
...
+                    unordered_map<SourceID, logicalTime>       newVersions;
+                    optional<std::pair<SourceID, logicalTime>> extra;
```
snej: This pair is identical to `Version`, so use that instead.

LiteCore/Database/VectorDocument.cc
```diff
@@ -428,6 +428,75 @@ namespace litecore {
...
+                        return false;
+                    };
+                    if ( intersect() ) break;
+
+                    // 6. curVers.legacy[0] is not in newVers.legacy (RevTree parent)
+                    auto isCurVersLegacyParentOfNewVers = [&]() {
+                        if ( curVers.legacy.empty() ) return false;
+                        return std::ranges::find(newVers.legacy, curVers.legacy[0]) != newVers.legacy.end();
+                    };
+                    if ( isCurVersLegacyParentOfNewVers() ) break;
```
snej: Why is this a lambda and not just a plain if statement?

snej (changes requested):

LiteCore/Database/VectorDocument.cc
jianminzhao: done

snej (approved):

PR #2490, opened 2026-05-26
CBL-8057: Better stack trace logging and crash handling

borrrden:
It's almost impossible to have a release build that properly prints out stack traces at runtime with frame names and such.  This adds the relative address of the frame to the stack trace log so that it can be looked up after the fact, and also adds a new crash log that prints a stack trace to a special file in the event of a crash.

As a side effect of updating Fleece this also fixes CBL-8363

### DIFF
--- a/LiteCore/Logging/LogFiles.cc
+++ b/LiteCore/Logging/LogFiles.cc
@@ -11,6 +11,8 @@
 //
 
 #include "LogFiles.hh"
+
+#include "Backtrace.hh"
 #include "c4ExceptionUtils.hh"
 #include "Error.hh"
 #include "FilePath.hh"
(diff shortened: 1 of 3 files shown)


pasin (commented):
Earlier this week in Slack, you mentioned that this likely wouldn't be included in 4.1.0 because it was probably too late. Has that decision changed?

pasin:
From LiteCore channel, seem like it needs to go in because of the change in fleece.

LiteCore/Logging/LogFiles.cc
```diff
@@ -159,6 +161,10 @@ namespace litecore {
         // Initialize LogFile objects before opening them:
         for ( int i = 0; i < kNumLogLevels; i++ ) _files[i] = make_unique<LogFile>(LogLevel(i), _options);
         for ( int i = 0; i < kNumLogLevels; i++ ) _files[i]->open();
+
+        // The expectation here is that you will only create one active LogFiles at once
+        auto crash_path = options.directory + FilePath::kSeparator + "cbl_crash.log";
+        BacktraceSignalHandler::setLogPath(crash_path.c_str());
```
pasin, line 167: Maybe in the future, we could have multiple crash log?

pasin (approved):

LiteCore/Logging/LogFiles.cc
borrrden: Maybe but it requires having all of the files open preemptively so I figured it's not worth it

borrrden:
> Earlier this week in Slack, you mentioned that this likely wouldn't be included in 4.1.0 because it was probably too late. Has that decision changed?

Yes, my hand is kind of forced because it's already on Fleece master so I either have to keep going or back out.  The thing definitely not going into 4.1.0 is the std::filesystem stuff.

PR #2494, opened 2026-06-02
CBL-7976 Implement Parse and Compare Non-numeric checkpoint resolution

nitishgupta25:
Implementation is done to enhance the checkpoint resolution process to be able to process and compare non-numeric checkpoints so that the replication can continue from the lower checkpoint if they are mismatched, instead of starting from the beginning.

Jira ticket : https://jira.issues.couchbase.com/browse/CBL-7975

Design doc for change: https://docs.google.com/document/d/1fIQsdT8Xkhs4HcH94POYV2mBlNA4gSHczMd3x5Qi69E/edit?tab=t.0

### DIFF
--- a/Replicator/Checkpoint.cc
+++ b/Replicator/Checkpoint.cc
@@ -116,8 +116,8 @@ namespace litecore::repl {
                   remoteSequences._remote.toJSONString().c_str());
 
             ParsedSequenceID localParsed, remoteParsed;
-            bool localParseable  = _remote.toParsedSequenceID(localParsed);
-            bool remoteParseable = remoteSequences._remote.toParsedSequenceID(remoteParsed);
+            bool             localParseable  = _remote.toParsedSequenceID(localParsed);
+            bool             remoteParseable = remoteSequences._remote.toParsedSequenceID(remoteParsed);
             if ( localParseable && remoteParseable ) {
                 if ( remoteParsed.before(localParsed) ) {
                     LogTo(SyncLog, "Rolling back to earlier remote sequence from server, some redundant changes may be "
--- a/Replicator/Checkpoint.cc
+++ b/Replicator/Checkpoint.cc
@@ -114,18 +114,20 @@ namespace litecore::repl {
         if ( _remote && _remote != remoteSequences._remote ) {
             LogTo(SyncLog, "Remote sequence mismatch: I had '%s', remote had '%s'", _remote.toJSONString().c_str(),
                   remoteSequences._remote.toJSONString().c_str());
-            if ( _remote.isInt() && remoteSequences._remote.isInt() ) {
-                if ( _remote.intValue() > remoteSequences._remote.intValue() ) {
+
+            ParsedSequenceID localParsed, remoteParsed;
+            if ( _remote.toParsedSequenceID(localParsed) && remoteSequences._remote.toParsedSequenceID(remoteParsed) ) {
+                if ( remoteParsed.before(localParsed) ) {
                     LogTo(SyncLog, "Rolling back to earlier remote sequence from server, some redundant changes may be "
                                    "proposed...");
                     _remote = remoteSequences._remote;
                     match   = false;
                 } else {
-                    LogTo(SyncLog, "Ignoring remote sequence on server since client side is older, some redundant "
-                                   "changes may be proposed...");
+                    LogTo(SyncLog, "Ignoring remote sequence on server since client side is older or equal, some "
+                                   "redundant changes may be proposed...");
                 }
             } else {
-                Warn("Non-numeric remote sequence detected, resetting replication back to start.  Redundant changes "
+                Warn("Unparseable remote sequence detected, resetting replication back to start.  Redundant changes "
                      "will be proposed...");
                 _remote = {};
                 match   = false;
--- a/Replicator/Checkpoint.cc
+++ b/Replicator/Checkpoint.cc
@@ -116,7 +116,9 @@ namespace litecore::repl {
                   remoteSequences._remote.toJSONString().c_str());
 
             ParsedSequenceID localParsed, remoteParsed;
-            if ( _remote.toParsedSequenceID(localParsed) && remoteSequences._remote.toParsedSequenceID(remoteParsed) ) {
+            bool localParseable  = _remote.toParsedSequenceID(localParsed);
+            bool remoteParseable = remoteSequences._remote.toParsedSequenceID(remoteParsed);
+            if ( localParseable && remoteParseable ) {
                 if ( remoteParsed.before(localParsed) ) {
                     LogTo(SyncLog, "Rolling back to earlier remote sequence from server, some redundant changes may be "
                                    "proposed...");
@@ -127,8 +129,10 @@ namespace litecore::repl {
                                    "redundant changes may be proposed...");
                 }
             } else {
-                Warn("Unparseable remote sequence detected, resetting replication back to start.  Redundant changes "
-                     "will be proposed...");
+                Warn("Unparseable remote sequence: locally-saved remote seq '%s' is %s, "
+                     "server-side remote seq '%s' is %s. Resetting replication.",
+                     _remote.toJSONString().c_str(), localParseable ? "parseable" : "UNPARSEABLE",
+                     remoteSequences._remote.toJSONString().c_str(), remoteParseable ? "parseable" : "UNPARSEABLE");
                 _remote = {};
                 match   = false;
             }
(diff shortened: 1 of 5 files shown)


Replicator/ParsedSequenceID.hh
```diff
@@ -0,0 +1,151 @@
...
+        static bool parse(const std::string& str, ParsedSequenceID& out) {
+            if ( str.empty() ) return false;
+
+            // Split into up to 4 tokens by ':' (4th would mean invalid).
+            uint64_t values[3] = {};
+            bool     empty[3]  = {};
+            int      count     = 0;
+            size_t   start     = 0;
+
+            for ( size_t i = 0; i <= str.size(); ++i ) {
```
jianminzhao: you may use litecore::split in StringUtil.hh to split by ':', and you can use slice_istream parse a uint64_t

Replicator/tests/ParsedSequenceIDTest.cc
```diff
@@ -0,0 +1,448 @@
...
+    }
+
+    SECTION("both components equal") {
+        auto s = mustParse("50:50");
+        CHECK(s.triggeredBy() == 50);
+        CHECK(s.seq() == 50);
+        CHECK(s.isBackfillFormat());
+    }
+
+    SECTION("triggeredBy larger than seq") {
```
adamcfraser: This looks like the same case as 'basic backfill' (triggeredBy larger than Seq).  You shouldn't need to handle the reverse case (200:300 is an invalid triggeredBy/Seq combination).

Replicator/ParsedSequenceID.hh
```diff
@@ -0,0 +1,151 @@
...
+        ParsedSequenceID() = default;
+
+        ParsedSequenceID(uint64_t seq, uint64_t triggeredBy, uint64_t lowSeq) : _seq(seq), _triggeredBy(triggeredBy), _lowSeq(lowSeq) {}
+
+        [[nodiscard]] uint64_t seq()         const { return _seq; }
+        [[nodiscard]] uint64_t triggeredBy() const { return _triggeredBy; }
+        [[nodiscard]] uint64_t lowSeq()      const { return _lowSeq; }
+
+        [[nodiscard]] bool isLowSeqFormat()    const { return _lowSeq > 0; }
+        [[nodiscard]] bool isBackfillFormat()  const { return _lowSeq == 0 && _triggeredBy > 0; }
```
adamcfraser: I don't think 'isBackfillFormat' is exactly the right description for this, as lowSeq:triggeredBy:seq **does** represent a backfill - it's just one that's happening while lowSeq is also populated.

Replicator/tests/ParsedSequenceIDTest.cc
```diff
@@ -0,0 +1,448 @@
...
+    }
+
+    SECTION("same triggeredBy, same seq → not before") {
+        CHECK_FALSE(mustParse("50:50").before(mustParse("50:50")));
+    }
+}
+
+TEST_CASE("ParsedSequenceID before - LowSeq vs LowSeq", "[ParsedSequenceID]") {
+    // Both "LowSeq:TB:Seq": compare lowSeq first, then inner pair as tie-break.
+
```
adamcfraser, line 205: This section needs a case confirming "20:100:35" is before "20::100".  (the backfill boundary case, but with matching lowseq).

adamcfraser (dismissed):
Generally looks good, some comments and recommendations on an additional test case.

Replicator/tests/ParsedSequenceIDTest.cc
nitishgupta25: Agreed — "triggeredBy larger than seq" with 200:1 is structurally identical to "basic backfill". Since TriggeredBy < Seq (e.g. 200:300) is an invalid SG combination, Will remove the section.

Replicator/tests/ParsedSequenceIDTest.cc
nitishgupta25: will add a section to testing "20:100:35" before "20::100". The logic already handles it correctly (tie-break recurses into inner backfill-vs-simple comparison, 100 <= 100 → true), but the test was missing. Will also add the reverse check (false, due to strict <) to confirm the asymmetry.

Replicator/ParsedSequenceID.hh
nitishgupta25: yes..got it. renamed to isTriggeredRevision() to describe the format structurally - it identifies the "TriggeredBy:Seq" format where triggeredBy is the primary comparison value and no lowSeq is present. 
This avoids implying that LowSeq:TriggeredBy:Seq is not a backfill. Similarly renamed isLowSeqFormat() to isLowSeqRevision() to keep all three classifiers consistent with the existing isSimpleRevision() naming pattern:
isSimpleRevision() — "Seq"
isTriggeredRevision() — "TriggeredBy:Seq"
isLowSeqRevision() — "LowSeq:TriggeredBy:Seq" / "LowSeq::Seq"

Replicator/ParsedSequenceID.hh
nitishgupta25: I think yes using  litecore::split(str, ":") to split the components and a readUInt lambda backed by slice_istream::readDecimal() + eof() to parse each uint64_t. This removes the manual for loop, the std::stoull exception handling, and the explicit negative-value guard..thanks

adamcfraser (commented):
LGTM from a functional perspective - test cases are comprehensive and sequence comparison algorithm is correct.  Leaving to Jianmin for final approval of the implementation details.

Replicator/ParsedSequenceID.hh
```diff
@@ -0,0 +1,147 @@
...
+                    if ( !readUInt(parts[0], v[0]) ) return false;
+                    if ( !parts[1].empty() && !readUInt(parts[1], v[1]) ) return false;  // empty → 0 for "LowSeq::Seq"
+                    if ( !readUInt(parts[2], v[2]) ) return false;
+                    out = ParsedSequenceID(v[2], v[1], v[0]);  // LowSeq:TriggeredBy:Seq
+                    return true;
+                default:
+                    return false;
+            }
+        }
+
```
jianminzhao, line 107: I am using https://github.com/couchbase/sync_gateway/blob/main/db/sequence_id.go#L187 as reference for "truth."

Replicator/ParsedSequenceID.hh
```diff
@@ -0,0 +1,147 @@
...
+        [[nodiscard]] bool before(const ParsedSequenceID& seqID2) const {
+            const auto& a = *this;
+            const auto& b = seqID2;
+
+            if ( a.isLowSeqRevision() ) {
+                if ( b.isLowSeqRevision() ) {
+                    if ( a.lowSeq() != b.lowSeq() ) return a.lowSeq() < b.lowSeq();
```
jianminzhao, line 114: Per the reference, n1:100:m1 .before. n2:101:m2  for any n and m, but the this rule compares n1 and n2 they are both greater than 0.

jianminzhao (changes requested):

Replicator/ParsedSequenceID.hh
nitishgupta25: We should not use the safesequence comparison...instead you should use this corrected code which is recently done by adam:
https://github.com/couchbase/sync_gateway/commit/0bc83f2ad2428748dd5c295897b78ffc18e52331

Replicator/ParsedSequenceID.hh
nitishgupta25: yes this is expected..

Replicator/Checkpoint.cc
```diff
@@ -114,18 +114,20 @@ namespace litecore::repl {
...
                     match   = false;
                 } else {
-                    LogTo(SyncLog, "Ignoring remote sequence on server since client side is older, some redundant "
-                                   "changes may be proposed...");
+                    LogTo(SyncLog, "Ignoring remote sequence on server since client side is older or equal, some "
+                                   "redundant changes may be proposed...");
                 }
             } else {
-                Warn("Non-numeric remote sequence detected, resetting replication back to start.  Redundant changes "
+                Warn("Unparseable remote sequence detected, resetting replication back to start.  Redundant changes "
```
jianminzhao: whether _remote is unparseable or remoteSeuqnece._remote is unparseable may be useful when debugging. In theory, _remote should be parseable because it is early saved, but explicit noting which is not parseable may avoid "guessing" at debugging CBSE.

jianminzhao (changes requested):

Replicator/Checkpoint.cc
nitishgupta25: yes agree...i will correct this ambiguity

jianminzhao (approved):

jianminzhao (approved):

PR #2497, opened 2026-06-05
CBL-7976 added a fix for quoted compound sequence parsing

nitishgupta25:
### `Replicator/RemoteSequence.hh`
- Strip surrounding JSON string quotes in `toParsedSequenceID()` before parsing.
  Works for both quoted (Value constructor path) and unquoted (slice constructor 
  path) sequences.

### DIFF
--- a/Replicator/RemoteSequence.hh
+++ b/Replicator/RemoteSequence.hh
@@ -75,14 +75,20 @@ namespace litecore::repl {
 
         bool operator!=(const RemoteSequence& other) const noexcept FLPURE { return _value != other._value; }
 
-        /** Convert this RemoteSequence to a ParsedSequenceID. Returns false if unparseable. */
+        /** Convert this RemoteSequence to a ParsedSequenceID. Returns false if unparseable.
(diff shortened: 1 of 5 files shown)


Replicator/RemoteSequence.hh
```diff
@@ -75,14 +75,19 @@ namespace litecore::repl {
...
+            Value constructor path). These are stripped before parsing so that compound Sync
+            Gateway sequences like "20:100:35" are handled correctly whether quoted or not. */
         [[nodiscard]] bool toParsedSequenceID(ParsedSequenceID& out) const {
             if ( !*this ) return false;
             if ( isInt() ) {
                 out = {intValue(), 0, 0};
                 return true;
             }
-            return ParsedSequenceID::parse(std::string(sliceValue()), out);
+            std::string s(sliceValue());
```
jianminzhao, line 88: In general, this is json string. We can parse it to a Value and go on only if its type is kFLString.

jianminzhao (changes requested):

jianminzhao (approved):

PR #2502, opened 2026-06-15
CBL-8473: Fixed dumb bug in RingBuffer causing BT TLS streams to fail

snej:
See Claude's analysis in the Jira ticket.


borrrden:
This needs to go into release/4.1, I'm going to change the base

borrrden (approved):

PR #2503, opened 2026-06-17
CBL-8475: Use AtomicRetained in Replicator::Subreplicators

snej:
Fixes a race condition when a Replicator is torn down.

Some code had to be altered, because AtomicRetained is a bit harder to use: you have to first call `get()` to get a temporary Retained object, then use that.

### DIFF
--- a/Replicator/Replicator.cc
+++ b/Replicator/Replicator.cc
@@ -252,9 +252,13 @@ namespace litecore::repl {
 
     // Called after the checkpoint is established.
     void Replicator::startReplicating(CollectionIndex coll) {
-        if ( _options->push(coll) > kC4Passive ) _subRepls[coll].pusher->start();
-        if ( _options->pull(coll) > kC4Passive )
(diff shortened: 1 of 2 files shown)


borrrden:
The downstream stuff needs to go into fleece release/4.1 before it can go into lite-core release/4.1

snej:
Ubuntu build failed with a GCC error building Fleece tests.
```
couchbase-lite-core/vendor/fleece/Tests/SupportTests.cc:476:15: error: alias template 'AtomicRef' requires template arguments; argument deduction only allowed for class templates
  476 |     AtomicRef r = new AtomicRetainedTest();
```
Why wasn't this caught by Fleece's CI? ... hm, one difference is that Fleece's Ubuntu builds use GCC, while LiteCore's use Clang. But Clang on macOS didn't get this error. Different versions of Clang, perhaps?

borrrden (approved):

PR #2516, opened 2026-06-29
CBL-8578: Fixed dumb bug in RingBuffer causing BT TLS streams to fail.

jianminzhao:


jianminzhao:
Ported from CBL-8473. There is no merge conflict.

borrrden (approved):

PR #2520, opened 2026-07-14
Bump fleece to get fix for CBL 8465

borrrden:


borrrden:
Alright this will need some reworking since the tests will need to be different between Windows and others

jianminzhao (approved):

PR #2524, opened 2026-08-03
CBL-8715: Pusher UAF in ChangesFeed's DB-change observer

jianminzhao:
The observer callback captured a raw `this` into Pusher's object graph and fired on an arbitrary thread. If a DB write landed while Pusher was mid-teardown, the callback could retain an already-destructed Pusher -- confirmed by an ASan tombstone: "RefCounted object <Pusher> retained while it had an invalid refCount of 0".

ChangesFeed now captures a Retained<Replicator> + CollectionIndex instead of a raw pointer, and reaches Pusher only through the new Replicator::getSubReplPusher(), which safely reads the already- AtomicRetained SubReplicator::pusher field. If Pusher has already been torn down, the callback is a no-op instead of a stray retain.

### DIFF
--- a/Replicator/ChangesFeed.cc
+++ b/Replicator/ChangesFeed.cc
@@ -19,6 +19,8 @@
 #include "ChangesFeed.hh"
 #include "Checkpointer.hh"
 #include "ReplicatorOptions.hh"
+#include "Replicator.hh"
+#include "Pusher.hh"
 #include "DBAccess.hh"
 #include "StringUtil.hh"
 #include "c4DocEnumerator.hh"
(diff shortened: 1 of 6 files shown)


snej:
This does look like the source of the problem, but I disagree about the fix. ChangesFeed is a field of Pusher (`_changesFeed`), not a separate allocation. Invoking the Replicator to find the Pusher isn't going to help because if the Pusher is freed the ChangesFeed gets freed along with it, so any call into it is UB.

I think the best fix is to have Pusher tell the ChangesFeed to stop responding to change messages; ChangesFeed can do that by clearing its `_changeObserver`. That can be done whenever the Pusher is told to stop, or in its destructor as a last resort.

jianminzhao:
Yes, agreed. There is even another argument: the very reason of using raw pointer, I think, is to avoid cyclic reference. Now, capture of Replicator in the callback introduces a cycle. The hard part if the synchronization between where the observer is disabled and where the pusher is enqueued. "disable" cannot be in the destructor because at this time refCount is already 0.

snej (commented):
I think Claude's gone into the weeds. This PR is overcomplicated and just doesn't make sense. The only part that looks valid is the `stopObserving()` method.

PR #2525, opened 2026-08-07
CBL-8715: Pusher UAF in ChangesFeed's DB-change observer

jianminzhao:
The root cause is that the changes notifier is not timely removed from the SequenceTracker. We remove it whent the status turns to Stoppped.

The removal and the use of the notifier callback are under Collection's lock on the SequenceTracker; the removal has to wait for the callback to complete.

### DIFF
--- a/LiteCore/Database/SequenceTracker.cc
+++ b/LiteCore/Database/SequenceTracker.cc
@@ -467,8 +467,6 @@ namespace litecore {
         if ( tracker ) { tracker->removePlaceholder(_placeholder); }
     }
 
-    // This method is called under collection's SequenceTrack lock,
-    // in CollectionImpl::documentSaved().
     void CollectionChangeNotifier::notify() noexcept {
         if ( callback ) {
             logInfo("posting notification");
--- a/LiteCore/Database/SequenceTracker.cc
+++ b/LiteCore/Database/SequenceTracker.cc
@@ -467,6 +467,8 @@ namespace litecore {
         if ( tracker ) { tracker->removePlaceholder(_placeholder); }
     }
 
+    // This method is called under collection's SequenceTrack lock,
+    // in CollectionImpl::documentSaved().
     void CollectionChangeNotifier::notify() noexcept {
         if ( callback ) {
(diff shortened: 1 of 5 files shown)


LiteCore/Database/SequenceTracker.cc
```diff
@@ -467,6 +467,8 @@ namespace litecore {
         if ( tracker ) { tracker->removePlaceholder(_placeholder); }
     }
 
+    // This method is called under collection's SequenceTrack lock,
+    // in CollectionImpl::documentSaved().
```
snej: It's actually called from four places; I don't think it's worth having a comment describing it.

Replicator/ChangesFeed.cc
```diff
@@ -252,6 +252,10 @@ namespace litecore::repl {
         return true;
     }
 
+    // This will remove the notifier from the SequenceTracker under the following lock,
+    // _collection->sequenceTracker().useLocked([&](SequenceTracker& st)
```
snej: This comment is really specific to the current bug-fix, and I don't think we need to add it to the source code since it's not generally relevant.

Replicator/ChangesFeed.hh
```diff
@@ -97,6 +98,7 @@ namespace litecore::repl {
         void                _dbChanged();
         Retained<RevToSend> makeRevToSend(C4DocumentInfo&, C4DocEnumerator*);
         bool                shouldPushRev(RevToSend*, C4DocEnumerator*) const;
+        void                stopObserving();
```
snej: This should be part of ChangesFeed's public API. No need to make it private.

Replicator/ChangesFeed.hh
```diff
@@ -106,6 +108,7 @@ namespace litecore::repl {
         CollectionIndex const  _collectionIndex;
         bool                   _getForeignAncestors{false};  // True in propose-changes mode
       private:
+        friend Pusher;
```
snej: ...which makes this declaration unnecessary.

snej (approved):

PR #2526, opened 2026-08-11
CBL-7986: Remove "flags&1=0" in the result of QueryTranslator whereve…

jianminzhao:
…r warranted by C4Database::isDeletedTableComplete()

### DIFF
--- a/LiteCore/Query/Translator/QueryTranslator.cc
+++ b/LiteCore/Query/Translator/QueryTranslator.cc
@@ -51,6 +51,10 @@ namespace litecore {
             mutableThis->assignTableNameToSource(source, ctx);
         };
         root.translatorDefaultCollection = [&]() -> string_view { return _defaultCollectionName; };
+        root.isDeletedDocsFullyTracked   = [&](string_view collection) {
+            if ( collection.empty() ) collection = _defaultCollectionName;
+            return collection != "_default" || _delegate.isDeletedDocsFullyTracked();
+        };
         return root;
     }
 
@@ -140,9 +144,14 @@ namespace litecore {
             if ( name.empty() ) name = _defaultCollectionName;
             if ( !source->scope().empty() ) name = string(source->scope()) + "." + name;
 
-            DeletionStatus delStatus = source->usesDeletedDocs() ? kLiveAndDeletedDocs : kLiveDocs;
-            //FIXME: Support kDeletedDocs
-
+            DeletionStatus delStatus{kLiveDocs};
+            if ( source->onlyDeletedDocs() && ctx.delegate.isDeletedDocsFullyTracked(name) ) {
+                // WHERE clause guarantees only deleted docs match, and the delegate
+                // confirms that all deleted docs are in the dedicated kv_del_ table.
+                delStatus = kDeletedDocs;
+            } else if ( source->usesDeletedDocs() ) {
+                delStatus = kLiveAndDeletedDocs;
+            }
             tableName = _delegate.collectionTableName(name, delStatus);
             if ( name != _defaultCollectionName && !_delegate.tableExists(tableName) )
                 fail("no such collection \"%s\"", name.c_str());
--- a/LiteCore/Query/Translator/SelectNodes.cc
+++ b/LiteCore/Query/Translator/SelectNodes.cc
@@ -351,7 +351,10 @@ namespace litecore::qt {
             for ( WhatNode* what : _what ) what->parseChildExprs(ctx);
 
             // Parse the WHERE clause:
-            if ( Value where = getCaseInsensitive(select, "WHERE") ) { setChild(_where, ExprNode::parse(where, ctx)); }
+            if ( Value where = getCaseInsensitive(select, "WHERE") ) {
+                auto nodeWhere = reduceDeleted(ExprNode::parse(where, ctx), ctx);
+                if ( nodeWhere ) setChild(_where, nodeWhere);
+            }
 
             if ( Value order = getCaseInsensitive(select, "ORDER_BY") ) {
                 for ( Value orderItem : requiredArray(order, "ORDER BY") ) {
@@ -434,7 +437,7 @@ namespace litecore::qt {
         for ( SourceNode* source : _sources ) {
             string_view coll{source->collection()};
             if ( coll.empty() ) coll = ctx.delegate.translatorDefaultCollection();
-            if ( !source->_usesDeleted && coll == "_default" && source->isCollection() ) {
+            if ( !source->_usesDeleted && source->isCollection() && !ctx.delegate.isDeletedDocsFullyTracked(coll) ) {
                 // The default collection may contain deleted documents in its main table,
                 // so if the query didn't ask for deleted docs, add a condition to the WHERE
                 // or ON clause that only passes live docs:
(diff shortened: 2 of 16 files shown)


snej (commented):
There's a lot of code in here to remove the AND clause that tests for the deleted flag. Wouldn't it be better to just not generate that AND clause in the first place?

jianminzhao:
@snej I added a piece of code to initialize kv_info inside SQLDataFile::reopen for the following reason. Currently we initialize the info table for kMaxRowidWithDeletedInDefault in HouseKeeper which is linked to Database. In LiteCore DataFileTestFixture, tests use DataFile without the owning Database.

LiteCore/Query/Translator/NodesToSQL.cc
```diff
@@ -68,7 +68,17 @@ namespace litecore::qt {
...
-        writeMetaSQL(aliasDot, _property, ctx);
+        if ( _source && _property == MetaProperty::deleted && source()->usesDeletedTable() ) {
+            // As we are determined that we will use the deleted table for this source, Metadata test for "deleted"
+            // will always evaluate to true.
+            fleece::Encoder encoder;
+            encoder.writeBool(true);
+            fleece::Doc   doc       = encoder.finishDoc();
+            fleece::Value trueValue = doc.root();
+            LiteralNode   literalNode(trueValue);
+            literalNode.writeSQL(ctx);
```
snej: Why do all this instead of just writing `true`?

LiteCore/Query/Translator/QueryTranslator.cc
```diff
@@ -51,6 +51,14 @@ namespace litecore {
             mutableThis->assignTableNameToSource(source, ctx);
         };
         root.translatorDefaultCollection = [&]() -> string_view { return _defaultCollectionName; };
+        root.isDeletedDocsFullyTracked   = [&](string_view collection) {
+            // collection is either the name of the collection or name of the kv table.
+            return (collection.starts_with("kv_") ? collection != "kv_default" : collection != "_default")
+                   || _delegate.isDeletedDocsFullyTracked();
```
snej: This is ambiguous. A collection name can start with `kv_`. This function needs to accept only collection names or only table names, not both.

In both places this is called, the calling code has a collection name and has to make a call to turn it into a table name. So just make this function take a collection name, and remove the extra code at the call site.

LiteCore/Query/Translator/QueryTranslator.cc
```diff
@@ -51,6 +51,14 @@ namespace litecore {
...
+        };
+        root.collectionKeyStoreName = [this](string_view collection) -> string {
+            return _delegate.collectionTableName((string)collection, kLiveDocs);
+        };
```
snej: The name implies the function returns a KeyStore name, but the implementation returns a table name. (The difference is the table name has a `kv_` prefix.) Should probably be renamed.

LiteCore/Query/Translator/SelectNodes.cc
```diff
@@ -421,20 +421,39 @@ namespace litecore::qt {
...
+                if ( prop == MetaProperty::none || prop == MetaProperty::deleted ) meta->source()->setUsesDeletedDocs();
             } else if ( auto fn = dynamic_cast<FunctionNode*>(&node) ) {
                 // Look for aggregate functions:
                 if ( fn->opFlags() & kOpAggregate ) _isAggregate = true;
             }
         });
 
+        // Check if the WHERE clause filters the query to only deleted docs.
+        if ( _where ) {
+            [&](ExprNode* root) {
```
snej: Why is there a lambda wrapped around this? (lines 433 and 444).

LiteCore/Query/Translator/SelectNodes.hh
```diff
@@ -85,7 +85,17 @@ namespace litecore::qt {
 
         string_view collection() const { return _collection; }  ///< Collection name, or empty if default
 
-        bool usesDeletedDocs() const { return _usesDeleted; }  ///< True if exprs refer to deleted docs
+        bool usesDeletedDocs() const {
+            return _deletionStatus >= kUsesDeleted;
+        }  ///< True if an expr refers to a deleted doc from this source
```
snej: Please put the doc-comment before the method not after it.

LiteCore/Query/Translator/SelectNodes.hh
```diff
@@ -85,7 +85,17 @@ namespace litecore::qt {
...
+
+        bool usesOnlyDeletedDocs() const {
+            return _deletionStatus >= kUsesOnlyDeleted;
+        }  ///< True if WHERE guarantees only deleted docs are selected
```
snej: Please put the doc-comment before the method not after it.

LiteCore/Query/Translator/SelectNodes.hh
```diff
@@ -113,20 +123,31 @@ namespace litecore::qt {
         void         disambiguateColumnName(ParseContext&);
         void         writeASandON(SQLWriter& ctx) const;
 
-        void setUsesDeleted() {
-            _usesDeleted = true;  // It will turn kv_.xyz to all_.xyz
-            _tableName   = string_view{};
+        void setUsesDeletedDocs() {
+            if ( _deletionStatus < kUsesDeleted ) {
+                _deletionStatus = kUsesDeleted;  // It can turn kv_.xyz to all_.xyz
+                _tableName      = string_view{};
```
snej:
```suggestion
                _tableName      = "";
```

LiteCore/Query/Translator/SelectNodes.hh
```diff
@@ -113,20 +123,31 @@ namespace litecore::qt {
...
+            }
+        }
+
+        void setUsesOnlyDeletedDocs() {
+            if ( _deletionStatus < kUsesOnlyDeleted ) {
+                _deletionStatus = kUsesOnlyDeleted;  // WHERE guarantees only deleted docs; can use kv_del_ directly
+                _tableName      = string_view{};
```
snej:
```suggestion
                _tableName      = "";
```

LiteCore/Storage/SQLiteDataFile.cc
```diff
@@ -441,6 +441,22 @@ namespace litecore {
...
+        if ( options().writeable ) {
+            auto& infoStore = getKeyStore(DataFile::kInfoKeyStoreName, KeyStore::noSequences);
+            if ( !infoStore.get(kMaxRowidWithDeletedInDefault).exists() ) {
+                ExclusiveTransaction t(this);
+                if ( !infoStore.get(kMaxRowidWithDeletedInDefault).exists() ) {
+                    Record putRec{DataFile::kMaxRowidWithDeletedInDefault};
+                    putRec.setBodyAsUInt(asSQLiteKeyStore(&defaultKeyStore())->maxRowid());
+                    infoStore.setKV(putRec, t);
+                }
+                t.commit();
```
snej: This duplicates code in Housekeeper. Both of these blocks seem to belong in DataFile. It would be good to have all the code that manages this piece of state in one place.

snej (changes requested):

LiteCore/Query/Translator/NodesToSQL.cc
jianminzhao: I would be possible but I walked safe route to follow the pattern by submitting it to literal node. Could literal bool node turns it to TRUE?

LiteCore/Query/Translator/QueryTranslator.cc
jianminzhao: "the calling code has a collection name" - I made it this way exactly to guard against the ambiguity of "collection name" at the calling site. It's directly from the query text. For the default collection, it could be the db name, or ".default.default". That's why, before I calling this method, I run it through ctx.delegate.collectionKeyStoreName(coll), which turns generous form in the query text to the formal (live) key store name. Thus, I can reliably apply the logic applicable to the default collection.

LiteCore/Query/Translator/QueryTranslator.cc
jianminzhao: Will rename it. Again, this is guard against the generous textual form of the default collection.

LiteCore/Query/Translator/SelectNodes.cc
jianminzhao: To allow recursion.

LiteCore/Query/Translator/SelectNodes.hh
jianminzhao: Will change it.

LiteCore/Query/Translator/SelectNodes.hh
jianminzhao: Will change it

LiteCore/Query/Translator/SelectNodes.hh
jianminzhao: Will change it as suggested.

LiteCore/Query/Translator/SelectNodes.hh
jianminzhao: Will change it.

LiteCore/Storage/SQLiteDataFile.cc
jianminzhao: I can remove the initialization in Housekeeper. HouseKeeper does more than initialization. It may kick off based on the reading of this property.

LiteCore/Query/Translator/NodesToSQL.cc
snej: We know we just want a SQL `true` (really `1`) here. We don't need to know what LiteralNode does.

LiteCore/Query/Translator/QueryTranslator.cc
snej: Then make this function take only a table name. Change the doc-comment and change the first line to just compare against `kv_default`.

You are using "table name" and "keystore name" interchangeably. They are not the same thing, as I said in the comment below.

LiteCore/Query/Translator/SelectNodes.cc
snej: I understand the recursive part in the inner lambda. I'm asking what the outer lambda -- `[&](ExprNode* root) { ... }(_where)` -- is for. It doesn't seem to serve any purpose.

LiteCore/Storage/SQLiteDataFile.cc
snej: I mean the code that saves the value. There are at least three copies of this snippet in the source:
```c++
                    Record putRec{DataFile::kMaxRowidWithDeletedInDefault};
                    putRec.setBodyAsUInt(____);
                    infoStore.setKV(putRec, t);
```

LiteCore/Query/Translator/QueryTranslator.cc
jianminzhao: Will do.

LiteCore/Query/Translator/SelectNodes.cc
jianminzhao: I see. I tried to use the "cool" pattern, [](){}(argument);
Will remove the outer wrapper.

LiteCore/Storage/SQLiteDataFile.cc
jianminzhao: I see. Will refactor to consolidate.

LiteCore/Query/Translator/NodesToSQL.cc
jianminzhao: Done

LiteCore/Query/Translator/QueryTranslator.cc
jianminzhao: Done

LiteCore/Query/Translator/QueryTranslator.cc
jianminzhao: Done

LiteCore/Query/Translator/SelectNodes.cc
jianminzhao: done

LiteCore/Query/Translator/SelectNodes.hh
jianminzhao: Done

LiteCore/Query/Translator/SelectNodes.hh
jianminzhao: Done

LiteCore/Query/Translator/SelectNodes.hh
jianminzhao: Done

LiteCore/Query/Translator/SelectNodes.hh
jianminzhao: Done

snej (approved):

PR #2530, opened 2026-08-20
CBL-8733: Pusher UAF in ChangesFeed's DB-change observer

jianminzhao:
The root cause is that the changes notifier is not timely removed from the SequenceTracker. We remove it whent the status turns to Stoppped.

The removal and the use of the notifier callback are under Collection's lock on the SequenceTracker; the removal has to wait for the callback to complete.


pasin (approved):
Can you select the reason as well?

PR #2531, opened 2026-08-27
CBL-8759: Fix race when two TLS handshakes happen at once

borrrden:
See https://github.com/couchbasedeps/sockpp/pull/20 for details


borrrden:
Disregard blackduck manifest, since this is a fork and the version was not changed

jianminzhao (approved):

PR #2532, opened 2026-09-01
CBL-8812: Bump fleece submodule, add DeDuplicateEncoder regression test

jianminzhao:
Bumps vendor/fleece to pick up the HeapDict::_count fix (see the paired fleece PR for the root-cause writeup): a remove()-then-set() of a key that exists in a MutableDict's immutable _source left count() one too low, corrupting HeapDict::kvArray()'s cache-array sizing -- an out-of-bounds write that only fails cleanly in debug builds (HeapArray::setting()'s bounds check), silently corrupting the heap in shipping builds. Root cause of the crashes reported in CBSE-23608.

Adds a regression test exercising the actual code path that crashed in the field: DeDuplicateEncoder (used only by
VectorRecord::encodeBodyAndExtra) encoding a MutableDict in this remove-then-set state. Before the fix this doesn't fail cleanly -- it crashes the process, since HeapDict::kvArray() is reached through Dict::iterator's public constructor, itself the C-linkage function FLDictIterator_Begin(), and a C++ exception can never safely cross an extern "C" boundary. So the test asserts the correct post-fix behavior (successful encode, correct roundtrip) rather than trying to catch the crash, and is only meaningful -- and only safe to run in CI -- with the fleece fix in place.

Full CppTests suite run clean: 575 cases (572 baseline + 3 new: this one plus the two in the paired fleece PR), 35 pre-existing/unrelated failures (missing test-fixture files), zero new regressions.

### DIFF
--- a/LiteCore/tests/VectorRecordTest.cc
+++ b/LiteCore/tests/VectorRecordTest.cc
@@ -19,6 +19,7 @@
 #include "c4.hh"
 #include "HybridClock.hh"
 #include "VectorRecord.hh"
+#include "DeDuplicateEncoder.hh"
 #include "VersionVector.hh"
 #include "fleece/Mutable.hh"
 #include <iostream>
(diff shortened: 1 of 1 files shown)


jianminzhao:
There is no production code in this PR. The fix is in the linked fleece.

borrrden (approved):

Issue #19, opened 2016-11-17
Need REST API Bindings.


djpongh:

snej:
We have a limited implementation of this already, mostly for QE purposes, but it's far from the complete REST API.

Since this issue was filed, we've been leaning more toward creating a direct JavaScript binding to CBL, as a preferred solution for PhoneGap / ReactNative / etc.

borrrden:
Should I bother migrating this?  It seems less and less likely that this is going to happen.

borrrden:
Closing because we won't be doing this.

PR #2534, opened 2026-09-16
CBL-8862: bump sockpp for mTLS PEM client key fix

nitishgupta25:
## CBL-8862
https://jira.issues.couchbase.com/browse/CBL-8862

## What
Bumps `vendor/sockpp` to pick up the mTLS PEM client-key parse fix.
sockpp PR: https://github.com/couchbasedeps/sockpp/pull/21
sockpp bump: ed77f20 → 03e6c83 (adds only the PEM fix; master already had #19/#20).

## Why
`mbedtls_context::set_identity` parsed the replication client private key with
`mbedtls_pk_parse_key(..., size())` instead of `size() + 1`. Without the
terminating NUL, mbedTLS treats a PEM key as DER and fails with
`-0x3D00 "PK - Invalid key tag or value"`, breaking edge-to-edge / replicator
mTLS with a PEM client cert. The fix adds the `+1` only for PEM (mirrors
`mbedtls_pk_load_file`); DER keys are unaffected. RSA and EC both affected.

## Verification
Built Edge Server against the bumped sockpp and reproduced the customer scenario
(2 edge servers, mTLS replication): PEM client key now loads and the replication
connects/syncs; no `PK - Invalid key tag or value`.


borrrden:
BD manifest can be disregarded, it's still bad at false failures when a CB specific patch is applied to a third party source

jianminzhao (approved):

PR #2535, opened 2026-09-17
CBL-8805: Replicator reads wrong revision key from getCollections che…

jianminzhao:
…ckpoint response

We now allow both "rev" and "_rev" in the response of getCollections.


Replicator/Replicator.cc
```diff
@@ -893,7 +893,10 @@ namespace litecore::repl {
                         sub.remoteCheckpointRevID.reset();
                     } else {
                         remoteCheckpoints[i].readDict(dict);
-                        sub.remoteCheckpointRevID = dict["rev"].asString();
+                        auto checkpointrevid = dict["_rev"];
+                        if ( checkpointrevid.type() == kFLUndefined ) checkpointrevid = dict["rev"];
```
snej:
```suggestion
                        if ( !checkpointrevid ) checkpointrevid = dict["rev"];
```

snej (approved):
One minor simplification you can add if you want

